PR #1, opened 2017-05-01
Add a hook so that an AmrLevel can force a post-timestep regrid

maxpkatz:
The motivation for this change is that sometimes, during the course of a coarse
timestep, the timestep will found to be unstable due to a lack of resolution. In
this case the coarse level may want to signal that more resolution is required
immediately (as opposed to waiting for a future regrid), because even a single
unstable step may cause significant harm to the integrity of the simulation. An
example is a spurious numerical detonation caused by insufficient resolution in
a thermonuclear burning region. If this numerically-seeded detonation is detected
at the end of a timestep, more levels can be added immediately so that the fine
levels can improve the situation and quench the detonation (since they will
overwrite the coarse data at the end of the timestep).

Each AmrLevel now has a member variable post_step_regrid. To trigger a regrid
at the end of a step, but before any fine timesteps would be taken, they only need
to set this variable to one. At the beginning of every timestep, it will be reset to 0.

### DIFF
--- a/Src/Amr/AMReX_Amr.cpp
+++ b/Src/Amr/AMReX_Amr.cpp
@@ -1759,6 +1759,9 @@ Amr::timeStep (int  level,
     //      when regridding is called with possible lbase > level.
     which_level_being_advanced = level;
 
+    // Update so that by default, we don't force a post-step regrid.
+    amr_level[level]->setPostStepRegrid(0);
+
     //
     // Allow regridding of level 0 calculation on restart.
     //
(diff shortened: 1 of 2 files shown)


WeiqunZhang:
Should `level_count` be reset for the new levels?

What happens to the coarse fluxes stored the flux register that is a member of the fine level?

maxpkatz:
>Should level_count be reset for the new levels?

I think not. This should be considered a 'special' regrid that does not interfere with the normal regrid process. This will lead sometimes lead to back-to-back regrids (one at the end of a step and then another at the beginning of the next) but this is fine with me; those regrids are not necessarily redundant, and in any case it's a price the user is choosing to pay by opting in to this flexibility.

WeiqunZhang:
I might have misunderstood this.   Each AmrLevel has a post_step_regrid flag.  It looks like it indicates whether or not the fine levels should have new grids, because regrid doesn't change the base level.  But if the fine levels change, shouldn't the current level be run again so that FluxRegister can be set properly.  Or do you simply save all the currently fluxes?

maxpkatz:
I originally started with the idea of running the current level again, but I realized that it is a big win to avoid re-doing the step if it is possible. So for this implementation, I think that saving the fluxes is a requirement. At the beginning of an advance, the fine level should check whether the level below it has triggered post_step_regrid. If so, it should zero out the coarse fluxes and re-fill them.

PR #3, opened 2017-05-01
Dtg branch

dtgraves:
Marc and Ray will need the new tool (EBNormalizeByVolumeFraction) that I just ported and tested.


Src/Base/AMReX_FabArrayBase.cpp
```diff
@@ -1244,7 +1244,10 @@ FabArrayBase::getTileArray (const IntVect& tilesize) const
 #pragma omp critical(gettilearray)
 #endif
     {
-	BL_ASSERT(getBDKey() == m_bdkey);
+      if(getBDKey() != m_bdkey)
```
WeiqunZhang: Could we put BL_ASSERT back?  If one compile with `USE_ASSERTION=TRUE`, assertion will be turned on without having to run in debug mode.

Tests/GeometryShop/regression/Backtrace.rg_0_rl_0
```diff
@@ -7,47 +7,47 @@
             readelf -wl my_exefile | grep my_line_address'
     to find out the offset for that line.
 
- 0: /home/spencer/users/graves/_amrex/amrex/Tests/GeometryShop/regression/dataArith2d.gnu.DEBUG.COMM_PROF.ex() [0x506975]
```
WeiqunZhang: This file should be removed.

Tests/GeometryShop/regression/bl_prof/bl_prof_H
```diff
@@ -1,8 +1,11 @@
 BLProfVersion 1
 NProcs  1
 NOutFiles  1
+phFName "AggSten.constructor"
```
WeiqunZhang: This whole bl_prof directory probably should be removed.

PR #13, opened 2017-05-17
EB bug fixes for Marc

dtgraves:


WeiqunZhang:
Dan, can you resolve the conflicts?

I have a suggestion for `#define BL_TO_FORTRAN_FAB(x) (x).dataPtr(), ARLIM_3D((x).loVect()), ARLIM_3D((x).hiVect()), x.nCompPtr()`.  Can we use (x).nComp() to puss by value so that we don't have to have the `nCompPtr` function?  On the Fortan side, the variable becomes `integer, value :: n`.

dtgraves:
I am not sure I trust sending anything to fortran that is not a real* or an
int*.    It is not more risky than sending dataPtr.
  As a compromise, we can keep only the const version of nCompPtr,
maybe?

I resolved the conflict.

dtg

WeiqunZhang:
The `value` attribute is in Fortran 2003 standard.  So I think we can trust it.  Anyway, the compromise is fine.   So I am going to merge it, and you can remove the non-const version after the merge.

PR #23, opened 2017-05-31
python script added to convert ascii amrex particle files to vtp

hsitaram:
I have added python scripts to convert amrex particle file to vtp format, read by paraview.

Thanks,
Hari NREL


zingale:
I don't think it's a good idea to store plotfiles in the git repo itself.  github doesn't like this.  Perhaps the github large file storage could be an alternative if they are needed as an example?

hsitaram:
No problem, I can remove them. They are not very big anyway. All of them contribute to about 1 MB.

hsitaram:
Is it ok to leave may 3 files instead of the 20 I have in there?

WeiqunZhang:
OK.  Let's leave 3 files.  But you may have to generate a new commit that is NOT derived from this commit.  Otherwise they will still be in the repo.   Sorry for the trouble.

WeiqunZhang:
I guess I can squash the two commits into one.

PR #25, opened 2017-06-05
Update CVODE docs

bcfriesen:
These commits move the documentation for using CVODE in AMReX from a Readme file to the user guide.


bcfriesen:
@WeiqunZhang Do you prefer a different organization for CVODE in the user doc than that provided here? I made it a new chapter but don't know if it is significant enough to merit a standalone chapter.

WeiqunZhang:
This is fine.

PR #40, opened 2017-06-17
An approach to managed memory for particle data

atmyers:


Tests/Particles/ManagedCUDADeposition/make.out
```diff
@@ -0,0 +1,14 @@
+Loading ../../..//Tools/GNUMake/comps/pgi.mak...
```
maxpkatz: This text file looks like it got added in by accident.

maxpkatz: Another approach which would make the particle code a little simpler would be to use your custom allocator always, but switch between malloc/new and cudaMalloc/cudaMallocManaged (inside AMReX_CudaManagedAllocator.H) depending on whether CUDA is defined.

atmyers: I think I have a slight preference for the current approach, which makes `AMReX_Particles.H` uglier but easier to understand IMO, since you'd have to go to a different file to see that CudaManagedAllocator may or may not do something with related to Cuda. But, I can change it if you prefer.

PR #51, opened 2017-06-30
Have the test suite also send an email if it has to abort for…

atmyers:
… whatever reason (if emails are turned on...)


zingale:
can this replace the `email_developers` function embedded in `regtest.py`?

PR #56, opened 2017-07-11
add a print_state (and print_state_n) method

zingale:
Useful for dumping out the state for a single zone, for debugging


zingale:
well, it only outputs a single zone, so it should not be an issue

bcfriesen:
Oh whoops, right you are. I can't read good.

zingale:
this is a prettier version, all in C++ now

Src/Base/AMReX_MultiFabUtil.cpp
```diff
@@ -377,4 +379,34 @@ namespace amrex
...
+
+
+#ifdef _OPENMP
+#pragma omp parallel
+#endif
+      for (MFIter mfi(mf); mfi.isValid(); ++mfi)
+	{
+	  if (mfi.validbox().contains(cell)) {
+	    if (n >= 0) {
+	      amrex::AllPrint() << mf[mfi](cell, n) << std::endl;
```
WeiqunZhang: Let's set precision to maximum.  `amrex::AllPrint().SetPrecision(17)`.

PR #76, opened 2017-08-02
clrdDiffuSolve

akreienbuehl:
Added `amrex.clr_map` flag to `AMReX_ParallelDescriptor.*` to allow for a convenient switching between different processor-to-color mappings.


WeiqunZhang:
In `ParallelDescriptor::Translate`, `ParmParse` is used.  It would be better if we don't have to query the ParmParse database repeatedly.  Maybe that information can be initialized in the beginning when `ParallelDescriptor` is initialized and is stored somewhere `Translate` can access.

PR #79, opened 2017-08-03
Do not use hardcoded numbers for real kinds

certik:
Such numbers are compiler specific. In particular, the NAG compiler fails to
compile this line. One must use some platform independent way to get the real
kind. In this case, the number 8 means double precision in gfortran, so I have
changed it to kind(0d0) where 0d0 is a double precision number, and kind()
returns the proper kind on all platforms. Now this line compiles with NAG.

### DIFF
--- a/Src/Base/AMReX_mempool_f.f90
+++ b/Src/Base/AMReX_mempool_f.f90
@@ -7,7 +7,7 @@ module mempool_module
 
   implicit none
 
-  integer, parameter, private :: c_real = 8
+  integer, parameter, private :: c_real = kind(0d0)
 
   ! We could/should use Fortran 2008 c_sizeof here.
   integer (kind=c_size_t), parameter, private :: szr = c_real


WeiqunZhang:
If you print out the value of `c_real = kind(0.d0)`, is it 8 for NAG?  If not, we have a problem.  `c_real` is later used to define the size of a real (i.e., `szr` in that module).  That needs to be 8 for double, or we have to use `c_sizeof`, which NAG does not support.

nncarlson:
`kind(0d0)` is fine if the intent is for `double precision`.  Perhaps a better alternative is to use `real64` from the standard `iso_fortran_env` module.

certik:
Yes, `real64` works too. `kind(0.d0)` is equal to 2 in NAG, but that is irrelevant, as one should never use such numbers directly.

WeiqunZhang:
In principle, we should do the following because we may run in single precision mode.

    use amrex_fort_module, only : c_real=>amrex_real
    integer (kind=c_size_t), parameter, private :: szr = c_sizeof(c_real)

WeiqunZhang:
I think we are going to use macros to solve this problem because NAG doesn't support `c_sizeof`.

WeiqunZhang:
Your commit will not have work because of the size of real is wrong.  It should be 8, but for NAG it's 2.  I made some changes in the development branch to fix this problem.  Could you test it and let us know?

certik:
In NAG, the size of double precision real is 8 bytes, as it should be. But the kind, which is an integer, happens to be 2 in NAG, but that's not right or wrong. It's just an internal compiler number that one should never depend on, nor use directly.

I am going to test your changes.

certik:
So the changes done in the commit 9a2888d84f96ad88eacc8961c66e7ad684287630 in the development branch fix the problem. Thank you @WeiqunZhang. I am closing this issue.

PR #80, opened 2017-08-03
Compare real kinds directly

certik:
In order to determine if the real kinds are equal, one can just compare them
directly, they are just integers. This is simpler, and also allows these lines
to be compiled with NAG. For some reason NAG does not seem to have `c_sizeof()`
as part of the `iso_c_binding` module and so it fails on these lines. This
commit fixes it.

### DIFF
--- a/Src/F_Interfaces/Base/AMReX_parallel_mod.F90
+++ b/Src/F_Interfaces/Base/AMReX_parallel_mod.F90
@@ -89,9 +89,9 @@ contains
     end if
     call MPI_Comm_Size(m_comm, m_nprocs, ierr)
     call MPI_Comm_Rank(m_comm, m_myproc, ierr)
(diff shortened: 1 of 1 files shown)


nncarlson:
The original code using `c_size()` is simply wrong (not just more complicated). `amrex_real` is an integer variable (or parameter), and `c_size(amrex_real)` is the C size of a Fortran integer, not the size of a `real(kind=amrex_real)` variable which was intended.  The pull request code does exactly the right thing.

If there is any similar use of `c_size()` in the code, it's wrong and needs to be fixed.

WeiqunZhang:
You are absolutely right.

WeiqunZhang:
Fortunately, that's the only place `c_sizeof` is used.

certik:
Thanks for merging @WeiqunZhang!

PR #81, opened 2017-08-03
NAG crash workaround

certik:
For some reason, NAG 6.1 gives the following error:

	/tmp/AMReX_tagbox_mod.024836.f90:4:3: error: expected specifier-qualifier-list before ‘__NAGf90_ChDope4’

	   ^
	/tmp/AMReX_tagbox_mod.024836.f90:1:8: error: unknown type name ‘__NAGf90_ChDope4’
	 # 1 ".../amrex/Src/F_Interfaces/AmrCore/AMReX_tagbox_mod.F90"
			^~~~~~~~~~~~~~~~
	.../amrex/Src/F_Interfaces/AmrCore/AMReX_tagbox_mod.F90:43:18: error: conflicting types for ‘amrex_tagbox_module_MP_amrex_tagboxarray_dataptr’
	   function amrex_tagboxarray_dataPtr (this, mfi) result(dp)
					  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
	/tmp/AMReX_tagbox_mod.024836.f90:1:25: note: previous declaration of ‘amrex_tagbox_module_MP_amrex_tagboxarray_dataptr’ was here
	 # 1 ".../amrex/Src/F_Interfaces/AmrCore/AMReX_tagbox_mod.F90"
							 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
	.../amrex/Tools/GNUMake/Make.rules:147: recipe for target 'tmp_build_dir/o/2d.gnu.MPI.EXE/AMReX_tagbox_mod.o' failed

And it looks like it crashes due to the combination of returning the `dp` array
from the function and adding the function as a method of the class. One simple
workaround is just:

	-     procedure :: dataPtr       => amrex_tagboxarray_dataptr
	+!     procedure :: dataPtr       => amrex_tagboxarray_dataptr

But a better workaround implemented in this commit is to return the `dp` array as an `intent(out)` argument. Then everything seems to work.

### DIFF
--- a/Src/F_Interfaces/AmrCore/AMReX_tagbox_mod.F90
+++ b/Src/F_Interfaces/AmrCore/AMReX_tagbox_mod.F90
@@ -40,10 +40,10 @@ contains
     this%p = p
   end subroutine amrex_tagboxarray_install
 
-  function amrex_tagboxarray_dataPtr (this, mfi) result(dp)
+  subroutine amrex_tagboxarray_dataPtr (this, mfi, dp)
     class(amrex_tagboxarray) :: this
     type(amrex_mfiter), intent(in) :: mfi
-    character(c_char), contiguous, pointer, dimension(:,:,:,:) :: dp
+    character(c_char), contiguous, pointer, dimension(:,:,:,:), intent(out) :: dp
     type(c_ptr) :: cp
     character(c_char), contiguous, pointer :: fp(:,:,:,:)
     integer(c_int) :: n(4)
@@ -53,6 +53,6 @@ contains
     n(4)   = 1
     call c_f_pointer(cp, fp, shape=n)
     dp(bx%lo(1):,bx%lo(2):,bx%lo(3):,1:) => fp
-  end function amrex_tagboxarray_dataPtr
+  end subroutine amrex_tagboxarray_dataPtr
   
 end module amrex_tagbox_module


certik:
This change likely broke some code that uses it. I have installed AMReX as a library and our code works, but that's because we do not use this functionality. 

How can I run tests? So that I can fix code that needs to be fixed.

WeiqunZhang:
Could you resubmit this as a pull-request merging to development branch?

WeiqunZhang:
Several derived types have `dataPtr` functions that return a pointer.  For consistence, if we change it to subroutine for tagboxarray, we should change others too.   I am not sure that's a good idea.  We have to think about this.

nncarlson:
I think we also need to understand what the NAG compiler is complaining about before doing something.

certik:
Yes, I don't know what the problem is. This is just one way to fix it without losing functionality.

I rebased on top of the development branch in #83.

nncarlson:
@certik I've updated the NAG compiler on the petaca machines.  Please test the development branch code again with the new compiler (same nag/6.1 module file) and see if this compiler error doesn't go away.

certik:
Previously I used the `NAG Fortran Compiler Release 6.1(Tozai) Build 6113` compiler. Now I tried with the new `NAG Fortran Compiler Release 6.1(Tozai) Build 6136` compiler and it gives exactly the same error.

WeiqunZhang:
This looks like a compiler bug.  What if you made `dataPtr` `generic` and add `amrex_tagboxarray_dataptr` to the list of `procedure,private`?  Maybe then the compiler is happy.

nncarlson:
Okay.  That is pretty clearly a NAG compiler bug.   I should be able to work up a small reproducer and report it.  For our purposes, if we don't use that function, we can comment it out as you did as our own personal patch.  But that change, or the #83 change, really has no business being merged.

certik:
@WeiqunZhang is this patch what you had in mind?
```diff
--- a/Src/F_Interfaces/AmrCore/AMReX_tagbox_mod.F90
+++ b/Src/F_Interfaces/AmrCore/AMReX_tagbox_mod.F90
@@ -12,8 +12,9 @@ module amrex_tagbox_module
      type(c_ptr) :: p = c_null_ptr
    contains
      generic   :: assignment(=) => amrex_tagboxarray_assign, amrex_tagboxarray_install
-     procedure :: dataPtr       => amrex_tagboxarray_dataptr
+     generic :: dataPtr       => amrex_tagboxarray_dataptr
      procedure, private :: amrex_tagboxarray_assign, amrex_tagboxarray_install
+     procedure, private :: amrex_tagboxarray_dataPtr
   end type amrex_tagboxarray
 
   interface
```
But unfortunately NAG gives the same error.

WeiqunZhang:
Yes, it is.

WeiqunZhang:
I just pushed a commit changing the tagboxarray module to get around a PGI comiler issue.  

https://github.com/AMReX-Codes/amrex/commit/c50e086bd9616a716bcbaefad27ab810aa452d7a

I wonder if this will have an effect on NAG.

nncarlson:
Your change for PGI had no effect on NAG.  I've reported this bug to NAG (with a 8 line example) and expect a fix within a couple of weeks.  I think this PR and #83 can be closed.

PR #87, opened 2017-08-06
[WIP] Add GNUmakefile support for the NAG Fortran compiler

nncarlson:
This PR adds "nag" as one of the compiler options for the `configure` script and adds the corresponding make includes to define compiler options, etc.  This uses gcc for the C/C++ compilers and nagfor for the Fortran compiler.


nncarlson:
This may likely be incomplete, but does at least work for building/installing the library (with several additional source tweaks to work around NAG compiler issues).  What tests should I run to ensure that it is complete?  Is this documented somewhere?

WeiqunZhang:
We don't have an acceptance test for pull request.  We run daily regression tests on the development branch.   It's unlikely your commit is going to break anything.  So I am happy to merge it now.

We do have a script for testing compilation of a number of tests.  Maybe you want to try to see if those tests can be compiled with NAG.  To do that, you need the latest development branch that includes some minor updates to the script.  You need to set `AMREX_HOME` to appropriate value, and then run the following in `$AMREX_HOME` directory,

    Tools/CompileTesting/compiletesting.py --full --make_flags COMP=nag

PR #89, opened 2017-08-09
Fix character kind declarations

nncarlson:
This replaces `character(c_char)` declarations with `character(kind=c_char)`, which is surely what was intended.  Unlike other intrinsic types, a single (unqualified) type parameter is taken as the `len` parameter for character, not `kind`.  This is a really easy error to make.  We lucked out here because with most compilers `c_char` seems to have the value 1 and the default kind is also 1.

### DIFF
--- a/Src/Base/AMReX_error_mod.F90
+++ b/Src/Base/AMReX_error_mod.F90
@@ -13,12 +13,12 @@ module amrex_error_module
   interface
      subroutine amrex_fi_error (message) bind(c)
        import
-       character(c_char), intent(in) :: message(*)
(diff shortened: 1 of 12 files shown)


WeiqunZhang:
aha.  This is a Fortran gotcha.  Is this why NAG fails to build `amrex_tagbox_module`?

nncarlson:
No, that's a NAG compiler bug.  I've heard back from NAG and should get a fixed compiler within a couple of days.  I also have a real workaround for it, but I don't think it's worth bothering you with at this point.

PR #93, opened 2017-08-22
Add specific build routines for constructing boxarrays and distromaps from user-provided lists in Fortran

kweide:
I add these changes to gain access to some constructors from Fortran. These are for generating boxarrays and distibution maps directly from user-provided data. We need something like this for our work transitioning FLASH to AMReX, so we can generate AMReX objects that represent the same grids as an existing PARAMESH grid.

The interface isn't pretty and can certainly be improved...


WeiqunZhang:
I have some comments.

In `subroutine amrex_boxarray_build_bxs`, local arrays `loa` and `hia` are created.  Why is that necessary?

In `amrex_fi_new_boxarray_from_bxfarr`, there is `Box bxarr[bxslen];`.  I suggest, instead of built-in array, we use `BoxList` and `push_back`.  Then we make a new `BoxArray` from `BoxList`.  The built-in array is constructed on the stack.  We may run into troubles if the size is large.

kweide:
Local arrays `loa`, `hia` may not be necessary at all. I think I wanted to ensure that Fortran does   not somehow pass `lo` and `hi` as assumed shape arrays to `amrex_fi_new_boxarray_from_bxfarr`, but now I think I needn't have worried.

I also appreciate your other comments re `amrex_fi_new_boxarray_from_bxfarr`. Would appreciate it even more if you could implement your suggestion. My C++ foo is slight and I would have to do a lot of trial-and-erroring to implement your suggestion.

Also, on the Fortran side maybe you could come up with a nicer (more natural) interface to `amrex_boxarray_build_bxs`. For reference, currently I call this as

``        call amrex_boxarray_build(ba,dimLimits(LOW,:NDIM,1:ibLoc),dimLimits(HIGH,:NDIM,1:ibLoc) )``

but it doesn't have to look like this - any reasonable way to explicitly pass a list of box index limits would be good.

WeiqunZhang:
In Fortran, I would pass `dimLimits` directly into `amrex_boxarray_build` instead of slicing to avoid copying the data.  We can then pass that directly to C++ as well.  What are the bounds of `dimLimits`?

kweide:
`  integer, dimension(LOW:HIGH,NDIM,MAXBLOCKS) :: dimLimits`

which is concretely

`  integer, dimension(1:2,2,1000) :: dimLimits`

kweide:
It's common in the FLASH code to store block limits in arrays shaped like `blkLimits(1:2,MDIM)`. I was just trying to avoid doing a copy to separate `lo` and `hi` arrays explicitly before a call to `amrex_boxarray_build`. But again, that's not important; I will use whichever convention you suggest.

Also I think efficiency shouldn't be much of a concern here. I assume the these variants of building a boxarray (and multifab) won't be used much in production code.

WeiqunZhang:
I am going to merge this first, and then make some changes.

WeiqunZhang:
I made some changes.  But I didn't test it.  Let me know if this doesn't work for you.

kweide:
Thank you @WeiqunZhang, I was able to compile and link with the changed code. Did not seriously test running it yet, I don't expect problems but will let you know if I find any.

Issue #50, opened 2017-06-28
Clarifications on ownership of nodal data (via fortran)


pbrady:
I apologize if these are already answered in the manual by I couldn't find it.  As noted in the manual, on nodal data, the boxes associated with an `mfiter` are not disjoint.  

1)  Are the boxes returned by `tilebox()` considered to be valid non-ghost cells even though they overlap?  If so, are these overlapping non-ghost cells which are "shared" by different multifabs ever updated via `fillboundary` calls or is that job of the application developer?

2)  Is there an idiomatic way of looping over them such that they appear disjoint?  (My main interest here is in interfacing with external solvers which represent the grid as a disjoint union among the ranks)

asalmgren:
Peter -- have you looked at Figures 4.3 through 4.8 in the User's Guide?  Note that the boxes returned by tilebox do not overlap even in the case of nodal data.

WeiqunZhang:
I think Peter is talking about the boxes returned by `tilebox()` from different Fabs.  Yes, they could overlap in index space, but the memory location in arrays do not.

`fillboundary` does not modify those overlapping non-ghost cells.  If you can give more details about what you want to achieve, maybe there is a way to do this.

As for question 2, `MultiFab` in C++ has a function called `OverlapMask`.  This or this type of functions might be useful for this problem.  If you give more details, that will help us understand.

PS.  There are many `tilebox` types of functions.  The Fortran interface is incomplete on this.  We will finish the porting as soon as possible.

pbrady:
Thanks for the responses.  For a little context, I'm exploring amrex by implementing a simple incompressible flow solver on a staggered mesh.  At the moment I'm restricting the investigation to a single level of refinement.  Pressure is cell centered and each velocity component is nodal in its direction.  With this setup, I have some initialization code (with added serialization to make sense of the io):
```fortran
    do i=1,amrex_parallel_nprocs()
       if (i-1.eq.amrex_parallel_myproc()) then
          print *, amrex_parallel_myproc(), "init_pressure"
          call amrex_mfiter_build(mfi, P)
          do while(mfi%next())
             bx = mfi%tilebox()
             call amrex_print(bx)

             dp => P%dataPtr(mfi)
             dp = 0.0d0
          end do
          call amrex_mfiter_destroy(mfi)
       end if
       call MPI_BARRIER(MPI_COMM_WORLD, ierr)
    end do
```
I have a similar loop for the `u` velocity and the output for a particular set of parameters is:
[...]

PR #114, opened 2017-09-05
Additions to the Fortran imultifab interface

nncarlson:
This extends the Fortran imultifab interface to include the setval method, and adds the optional nodal argument to amrex_imultifab_build, as is defined for amrex_multifab_build.


nncarlson:
@WeiqunZhang, I'm testing your `amrex_multifab_build_owner_mask` code that you added for us in #50 and found I needed the capabilities provided by this PR to use it.  Kind of my fault, since I asked you to make the mask integer-valued instead of real-valued.  I've tested this code in combination with your code.

WeiqunZhang:
Looks good to me.  Thanks for the contribution!

PR #118, opened 2017-09-06
Print build target on Cray

bcfriesen:
AMReX does not have separate compiler flags for different architectures on Cray
systems, e.g., separate compile flags for AVX2 vs. AVX-512 instructions. The
user is responsible for choosing the correct build target, via the
"craype-<arch>" modules. To make clear which target the user is compiling for,
print the target at link time.

### DIFF
--- a/Tools/GNUMake/Make.rules
+++ b/Tools/GNUMake/Make.rules
@@ -32,6 +32,10 @@ else
 endif
 #	@echo SUCCESS
 
+ifdef CRAY_CPU_TARGET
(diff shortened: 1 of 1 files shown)


zingale:
would it make sense to reflect this in the executable name?

WeiqunZhang:
fine with me.  Only need to add the following to `Make.defs`, I think.

    ifdef CRAY_CPU_TARGET
        machineSuffix += .$(strip $(CRAY_CPU_TARGET))
    endif

PR #143, opened 2017-10-06
Migrate convergence util

rgrout:
I think Marc already did this, but couldn't find it in the repo - changes to comparison utilities to work with current AMReX.


drummerdoc:
Sorry, notification of this musta got sent to my JUNK folder.

PR #144, opened 2017-10-11
Base: add support for querying Process Management Interface (PMI)

bcfriesen:
PMI is available on Cray systems and provides information about the topology of
the job on the network. One can use this information to apply optimizations at
run time which exploit higher bandwidth, lower-latency connections between
particular groups of processes on the network.

PMI support is disabled by default, and can be activated by setting
USE_PMI=TRUE in the makefile.


WeiqunZhang:
Could we have a longer name instead of just `PMI`?  `#ifdef AMREX_PMI`?

bcfriesen:
Sure. Changed "PMI" to "AMREX_PMI". And printed a simple header so it's clearer what the stuff printed to STDOUT means.

PR #152, opened 2017-10-25
Create a dimension agnostic, multi-component filcc

maxpkatz:
This version of filcc is both dimension agnostic, and works over an arbitrary number of components. It gives the same results as the dimensionful versions of filcc. It will be necessary for performance reasons to use a version like this on the GPU.

This also applies the new filcc to amrex_fab_filcc.

### DIFF
--- a/Src/Base/AMReX_filcc_mod.F90
+++ b/Src/Base/AMReX_filcc_mod.F90
@@ -1,7 +1,10 @@
 
+#include "AMReX_BC_TYPES.H"
+#include "AMReX_CONSTANTS.H"
+
 module amrex_filcc_module
 
-  use amrex_fort_module, only : amrex_real, amrex_spacedim
+  use amrex_fort_module, only : amrex_real, amrex_spacedim, get_loop_bounds
 
   implicit none
 
@@ -11,7 +14,7 @@ module amrex_filcc_module
   end interface amrex_filcc
 
   private
-  public :: amrex_filcc, amrex_fab_filcc
+  public :: amrex_filcc, amrex_fab_filcc, filccn
 
 contains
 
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
Do you mind if I make some changes to the code?  For example

    do k = lo(3), hi(3)
             do j = lo(2), hi(2)
                do i = lo(1), hi(1)

                   if (i < ilo) then
                      q(i,j,k,n) = q(ilo,j,k,n)
                   end if

                end do
             end do
          end do

We can have a test outside the loop so that interior boxes do not have to go through k, j and i loops just to find out there are no works to do.

maxpkatz:
I don't mind, but I've found that GPU performance is pretty sensitive to how conditionals are done in this routine. So I'd like to test your changes on the GPU before we merge.

WeiqunZhang:
I want to change it to

          if (lo(1) < ilo) then
             do k = lo(3), hi(3)
                do j = lo(2), hi(2)
                   do i = lo(1), ilo-1
                      q(i,j,k,n) = q(ilo,j,k,n)
                   end do
                end do
             end do
          end if

maxpkatz:
On the GPU, we're doing this on a zone-by-zone basis; lo == hi. So that's why I wrote it the way I did.

WeiqunZhang:
My version should be the same for gpu.  Correct?  But it will make a big difference for cpu.

WeiqunZhang:
In fact the test for `if (lo(1) < ilo` can be outisde the whole `if (bc(1,1,n) ... else ...`.  I will make a new fillcc2 branch for you to test.

maxpkatz:
On the GPU, we assign one thread to every possible zone. So, assume ilo = 0 and q_lo = -4. Then there's an individual thread for i = -4, i = -3, i = -2, and i = -1. With your version, we violate this assumption because now some threads are doing work on other thread's zones.

maxpkatz:
It would work if we did, say, `do i = lo(1), min(ilo-1, hi(1))`.

WeiqunZhang:
Got it. But we can do test outside the nested loops.

maxpkatz:
I think so, yes.

WeiqunZhang:
I think this should work

       if (lo(1) < ilo) then
          imin = lo(1)
          imax = min(hi(1),ilo-1)

          if (bc(1,1,n) .eq. EXT_DIR) then

             ! Do nothing.

          else if (bc(1,1,n) .eq. FOEXTRAP) then
             
             do k = lo(3), hi(3)
                do j = lo(2), hi(2)
                   do i = imin, imax
                      q(i,j,k,n) = q(ilo,j,k,n)
                   end do
                end do
             end do
             
             else if (bc(1,1,n) .eq. HOEXTRAP) then
                ......

maxpkatz:
Makes sense to me.

PR #165, opened 2017-11-07
EB updates

bcfriesen:
These commits attempt to optimize some of the EB neighbor cell computations. Switching the Combustor code from calling `get_neighbor_cells()` to `get_neighbor_cells_int_single()` yielded a 22% speedup in overall time step execution time on KNL, using 64 MPI processes on 1 node, with a 256^3 grid decomposed into 64^3 boxes. This was due entirely to vectorization of the (i,j,k) loop in the `eb_diff_mol_3d()` function, which was not possible with the original `get_neighbor_cells()` calls.

### DIFF
--- a/Src/EB/AMReX_ebcellflag_mod.F90
+++ b/Src/EB/AMReX_ebcellflag_mod.F90
@@ -5,7 +5,7 @@ module amrex_ebcellflag_module
   private
   public :: is_regular_cell, is_single_valued_cell, is_multi_valued_cell, &
        is_covered_cell, get_neighbor_cells, num_neighbor_cells, &
-       set_regular_cell, amrex_ebcellflag_count
+       set_regular_cell, amrex_ebcellflag_count, get_neighbor_cells_int_single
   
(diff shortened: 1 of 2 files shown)


WeiqunZhang:
Could `get_neighbor_cells_int_single` be elemental?

bcfriesen:
Ah yes it could. In fact, an `ELEMENTAL` subroutine is also `PURE` unless it is explicitly declared to be `IMPURE`. I will change that.

bcfriesen:
Note also that, with AMReX's current default optimization flags for the Intel compiler, it still does not vectorize the loop in `eb_diff_mol_3d()`, even with the call to `get_neighbor_cells_int_single()`. It claims that the vectorization is too inefficient. One must upgrade the `-ip` flag to `-ipo` for the loop to vectorize. (`-ipo` considers inter-procedural optimization for all functions throughout the code at link time. It increases the link time considerably, but also increases the chance of functions being inlined and vectorized.)

PR #172, opened 2017-11-24
Add target attribute to c_loc'ed fab

maxpkatz:
I picked this up while accidentally compiling with gcc 4.3. Newer versions of gfortran don't complain.

### DIFF
--- a/Src/F_BaseLib/fab.f90
+++ b/Src/F_BaseLib/fab.f90
@@ -820,7 +820,7 @@ contains
          integer(kind=c_size_t), intent(in), value :: n
        end subroutine amrex_real_array_init
     end interface
-    type(fab), intent(out) :: fb
+    type(fab), target, intent(out) :: fb
     type(box), intent(in)  :: bx
     integer, intent(in), optional :: ng, nc
     logical, intent(in), optional :: nodal(:)


maxpkatz:
I am not sure if this is a bug in gcc 4.3 since we're doing c_loc on a member of the fab, not the fab itself. But I don't think it hurts to add the target attribute.

WeiqunZhang:
I am quite sure this is a bug in gcc 4.3.

WeiqunZhang:
When I run `gcc -dumpversion` with my gcc 4.8.4, I only get `4.8` without the patchlevel version.  Not sure how to get the patchlevel version reliably.  When I was searching for a way of doing this, I was surprised to find out that some systems only print the major version.  https://stackoverflow.com/questions/45168516/gcc-7-1-1-on-fedora-26-dumpversion-now-only-includes-major-version-by-default

Anyway, I am assuming dumpversion will give me x.y and throw an error if it is less than 4.8.

WeiqunZhang:
Since gcc 4.3 cannot compile amrex anyway, I don't think we should add `target` just for a gcc 4.3 bug.  Adding `target` means maybe somewhere there is a pointer pointing to this object.  Maybe it won't hurt performance, but maybe it could.

PR #175, opened 2017-12-05
fix macros leading to incorrec endianness in fab io

JBlaschke:
In: `/Src/F_BaseLib/fabio_c.c` a macro checked for the Darwin kernel and switched to little-endian. No doubt to check if the code is running on a Mac: assuming that Macs used the powerpc architecture, requiring little-endian encoding. 

This would have been "correct" until about 13 years ago, when Macs switches to intel cpus. The patch fixes this oversight. 

This error interfered with tools like fextract running on a modern Mac.


atmyers:
As @JBlaschke and I discussed, the real fix would be to infer the encoding of the data from the plt file itself, rather than what machine fextract is running on. But I'm merging this for now, as the current code is clearly inconsistent with the rest of AMReX.

PR #178, opened 2017-12-13
CVODE, Tools: use Cray version of SUNDIALS if module is loaded

bcfriesen:
Cray includes SUNDIALS in the "cray-tpsl" module. Use that if the module is
loaded. If the module is not loaded, then look for the home-cooked version.


bcfriesen:
Forgot to update README to reflect this option. Closing this until I fix that.

PR #184, opened 2017-12-21
Some fixes to the dot, sum, and norm routines of BaseFab.

atmyers:


maxpkatz:
Please wrap CUDA API calls in the CudaAPICheck() macro.

PR #187, opened 2018-01-05
Tools: remove "-h std" flags from Cray C/C++ compiler

bcfriesen:
As of CCE 8.6.2, the compiler defaults to "-h std=c11" for the C compiler, and
"-h std=c++14" for the C++ compiler, so these flags are not necessary.


WeiqunZhang:
We might want to wait.  The default Cray compiler on some machines, for example titan, still needs the flag.

PR #210, opened 2018-02-16
GNUMake respects Make.local's compile commands

JBlaschke:
The user's specification of compile commands in the Make.local no longer is overwritten by the settings in `Tools/GNUMake/comps/*.mak`. This way the user can specify non-standard compiler commands by defining the variables `CXX`, `CC`, `FC`, and `F90` in the `Make.local`. Note that the version check in `Tools/GNUMake/comps/gnu.mak` has been removed: most users will have a sufficiently contemporary gcc.

This has also been documented. While I was at it, I cleaned up some of the doc's formatting ...


Tools/GNUMake/comps/gnu.mak
```diff
@@ -13,25 +13,25 @@ F90FLAGS =
...
-ifeq ($(gcc_major_le_4),1)
-  ifeq ($(gcc_minor_lt_8),1)
-    $(error GCC >= 4.8 required! Your version is $(gcc_version))
-  endif
-endif
+#gcc_major_le_4 = $(shell expr $(gcc_major_version) \<= 4)
+#gcc_minor_lt_8 = $(shell expr $(gcc_minor_version) \< 8)
+#ifeq ($(gcc_major_le_4),1)
+#  ifeq ($(gcc_minor_lt_8),1)
+#    $(warning GCC >= 4.8 required! Your version is $(gcc_version))
```
maxpkatz: This error was added because I ran on a system relatively recently where the default compiler was gcc 4.3, and in this case the code will not compile due to some things that depend on relatively recent gcc versions, so I think the error is still appropriate.

JBlaschke: The problem with the old way was that `Tools/GNUMake/comps/gnu.mak` sets "default" compilers and then tests against them:
```
CXX = g++
CC  = gcc
FC  = gfortran
F90 = gfortran

...

gcc_version       = $(shell $(CXX) -dumpfullversion -dumpversion | head -1 | sed -e 's;.*  *;;')
gcc_major_version = $(shell $(CXX) -dumpfullversion -dumpversion | head -1 | sed -e 's;.*  *;;' | sed -e 's;\..*;;')
gcc_minor_version = $(shell $(CXX) -dumpfullversion -dumpversion | head -1 | sed -e 's;.*  *;;' | sed -e 's;[^.]*\.;;' | sed -e 's;\..*;;')

...

gcc_major_le_4 = $(shell expr $(gcc_major_version) \<= 4)
gcc_minor_lt_8 = $(shell expr $(gcc_minor_version) \< 8)
ifeq ($(gcc_major_le_4),1)
  ifeq ($(gcc_minor_lt_8),1)
    $(error GCC >= 4.8 required! Your version is $(gcc_version))
  endif
endif
```

_before_ loading Make.local. The version test unfortunately forces GNUMakes lazy evaluation to evaluate `CXX` before the user's own compilers (in Make.local) have been specified. This isn't too much of a problem, if the system default gcc is > 4.8. For a Mac running homebrew this breaks because homebrew installs gcc to gcc-version_number (e.g. gcc-7). So then even if the user specifies the "correct" gcc in the Make.local, the code above breaks.

As an alternative, I suggest replacing a warning instead of an error. Something like:
"Your default gcc version is $(gcc_version). This might break during build. We therefore highly recommend that you specify GCC >= 4.8 in your Make.local. See the docs for an example"

maxpkatz:
I can't seem to find the comment chain to reply, but at any rate, the error to warning thing is fine with me. If we really cared to fix it we could perhaps put the check in at the end of Make.defs so that it gets evaluated after the user updates, but I'm not worried about it.

Tools/GNUMake/comps/gnu.mak
```diff
@@ -13,23 +13,23 @@ F90FLAGS =
 
 ########################################################################
 
-gcc_version       := $(shell $(CXX) -dumpfullversion -dumpversion | head -1 | sed -e 's;.*  *;;')
-gcc_major_version := $(shell $(CXX) -dumpfullversion -dumpversion | head -1 | sed -e 's;.*  *;;' | sed -e 's;\..*;;')
-gcc_minor_version := $(shell $(CXX) -dumpfullversion -dumpversion | head -1 | sed -e 's;.*  *;;' | sed -e 's;[^.]*\.;;' | sed -e 's;\..*;;')
+gcc_version       = $(shell $(CXX) -dumpfullversion -dumpversion | head -1 | sed -e 's;.*  *;;')
```
maxpkatz: This change breaks what I'm doing on the GPU branch -- was the immediate expansion of these variables removed for a particular reason?

WeiqunZhang: This was intended to fix a flaw in the make system.  If `CXX` is redefined in `Make.local`, we would like it to behave as if it is defined with the overriding value in say `gnu.mak`.  With immediate expansion, unfortunately `gcc_version` etc. still have the old value.   So we changed it.

How does this break the GPU branch?  Could we find a way around this?

maxpkatz: nvcc has not historically had good support for PGI and XL as host C++ compilers, but we do need them to compile CUDA Fortran. So I opted to go with COMP=PGI or COMP=IBM for compiling GPU code with USE_CUDA=TRUE, but when this is done, we use nvcc + g++ as the C++ compiler. To do this, in pgi.mak, I do an include of gnu.mak so that I get the correct compiler flags for g++ in CXXFLAGS, and then set CXX to nvcc and pass the CXXFLAGS as an option to nvcc. 

The problem this creates without immediate expansion is that DEFINES += -DBL_GCC_VERSION='$(gcc_version)' is evaluated on each compile statement, but CXX no longer is gcc by that point, so it doesn't work, because -dumpversion and -dumpfullversion aren't recognized by non-gcc compilers. (Weirdly though, it seems to be pgc++ that is being invoked with -dumpversion, which should not be CXX at this point. I don't understand that.)

maxpkatz: I think I can fix this by doing DEFINES := $(DEFINES) after I do the include of gnu.mak, to force immediate expansion. That would break any later updates in Make.local though, I guess?

maxpkatz: To be clear, it would only break Make.local updates when using CUDA.

PR #224, opened 2018-03-07
Minimal mods required to avoid BL_ASSERT error and to write a clean c…

marient:
I might need to add the time variable to the checkpoint too.
The long term solution is to create a working plt file for the particles.


drummerdoc:
You need to adjust this to request a merge into development instead of master...sorry.

PR #225, opened 2018-03-07
Development

marient:
I might need to add the time variable to the checkpoint too.
The long term solution is to create a working plt file for the particles.


drummerdoc:
Sorry again :)  This has to be a merge from your new branch on your fork to the development branch on the amrex repo.  Although I could change the target branch (on our end), I cannot change the source branch on yours.  Gotta delete this again....

PR #226, opened 2018-03-07
Avoid BL_ASSERT error message.

marient:
Please let it be the right combination of branches.

### DIFF
--- a/Src/Particle/AMReX_ParticleContainerI.H
+++ b/Src/Particle/AMReX_ParticleContainerI.H
@@ -1987,8 +1987,10 @@ ParticleContainer<NStructReal, NStructInt, NArrayReal, NArrayInt>
                 HdrFile << "real_comp" << i << '\n';
             }
         } else {
-            BL_ASSERT(real_comp_names.size() == NStructReal + NArrayReal);
-            for (int i = 0; i < NStructReal + NArrayReal; ++i ) {
+            std::cout << "PC " << NStructReal << " " << NArrayReal << " " << real_comp_names.size() 
+            //BL_ASSERT(real_comp_names.size() == NStructReal + NArrayReal);
+            for (int i = 0; i < real_comp_names.size(); ++i ) {
+            //for (int i = 0; i < NStructReal + NArrayReal; ++i ) {
                 HdrFile << real_comp_names[i] << '\n';
             }
         }
@@ -2767,11 +2769,11 @@ ParticleContainer<NStructReal, NStructInt, NArrayReal, NArrayInt>::WriteAsciiFil
         if (!File.good())
             amrex::FileOpenFailed(filename);
 
-        File << nparticles  << '\n';
-        File << NStructReal << '\n';
-        File << NStructInt  << '\n';
-        File << NArrayReal  << '\n';
-        File << NArrayInt   << '\n';
+//      File << nparticles  << '\n';
+//      File << NStructReal << '\n';
+//      File << NStructInt  << '\n';
+//      File << NArrayReal  << '\n';
+//      File << NArrayInt   << '\n';
             
         File.flush();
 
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
std::cout << "PC " << NStructReal << " " << NArrayReal << " " << real_comp_names.size()

This line should be removed.

Travis says it doesn't build in 3d.

Src/Particle/AMReX_ParticleContainerI.H
```diff
@@ -1987,8 +1987,10 @@ ParticleContainer<NStructReal, NStructInt, NArrayReal, NArrayInt>
                 HdrFile << "real_comp" << i << '\n';
             }
         } else {
-            BL_ASSERT(real_comp_names.size() == NStructReal + NArrayReal);
-            for (int i = 0; i < NStructReal + NArrayReal; ++i ) {
+            std::cout << "PC " << NStructReal << " " << NArrayReal << " " << real_comp_names.size() 
+            //BL_ASSERT(real_comp_names.size() == NStructReal + NArrayReal);
```
atmyers: Your application code can call the ParticleContainer::Checkpoint() routine with or without specifying the names of the various particle components (mass, velocity_x, etc...) in a Vector of strings. If you don't specify them, then in the checkpoint file they will just be called `particle_real_comp0`, and so on. If you do, they will have reasonable names in the checkpoint (useful for post processing), but the number of names has to match the number of components. My guess is that, in your application code, Checkpoint is being called with a string of component names with the wrong number of entries.  I'd prefer that be fixed in your application, either by passing in the right number of names or by omitting the names altogether.

marient:
Sure, you can remove it. The printout is just a reminder to me that NStructReal is not set to the correct number right now and that I need to look into it.
Marco

Src/Particle/AMReX_ParticleContainerI.H
marient: Does it mean that this is an input problem? Why does the checkpoint file needs a specification of the file components? Is there an example I can follow?

atmyers: It doesn't *need* it, but for post-processing it is sometimes useful to have the variable names be stored in the checkpoint file so you can refer to them in an intuitive way. 

For example, the place in Nyx where the ParticleContainer::Checkpoint method is called is here: 
    https://github.com/AMReX-Astro/Nyx/blob/master/Source/NyxParticleContainer.H#L418

You should have something similar in your application. The `real_comp_names` Vector is set up in Nyx for dark matter particles here:
    https://github.com/AMReX-Astro/Nyx/blob/master/Source/DarkMatterParticleContainer.H#L16

You may need to do something more complicated if the number of components depends on the number of spatial dimensions or other parameters. If you don't need the components to be labelled, the easiest thing to do is just remove the `real_comp_names` argument from the call to ParticleContainer::Checkpoint in your application.

marient: I understand, I'll see what I can do. So far I have been actually using the checkpoint for post-processing, but I thought that was a hack from my part while I was looking for the real output file. But it sounds like the boundary between plotting and check-pointing is thinner than for the amr boxes.

marient: I think I have it fixed from the application side. However, I also need to add "time" in the output line. Is it better to try and add it though the checkpoint or is there an example of plot file for particles that I could use? Thanks.

atmyers: The simulation time is already stored in the regular (non-particle) plt and chk files - would it be possible for you to get it from there? It's in the Header file in the top of the directory structure.

asalmgren:
One thing we do in Nyx is to write a separate ascii one-line file into the
checkpoint and plotfiles that holds the redshift -- this is something you
can trivially add at the application level then you could cat those files
together into a single ascii file that would hold the times.

atmyers:
I was about to make the same suggestion - I'd prefer that be done with a sidecar file in the application checkpoint / restart routines, so as not to disrupt things for other codes.

atmyers:
I'm going to close this pull request for now - let me know if you'd like help implementing this in your application code.

PR #227, opened 2018-03-16
Multi component mlmg

bsrunnels:
Modification of MLMG, MLCGSolver, MLLinOp, MLCellLinOp to work for multicomponent MultiFabs.


WeiqunZhang:
@bsrunnels  I made a bunch of small changes to MultiComponentMLMG branch.  I think there are two remaining issues.

(1) Solvability.  For singular problems (e.g., Poisson with all Neumann boundaries), we modify the RHS to make the problem solvable.  We do this in three places: the beginning of a solve, the beginning of each V-cycle on the coarsest AMR level, and before the bottom solve.  In the first two places, the fix is currently only for component 0.   For the bottom solve, the fix appears incorrect for multi-components.  In cell-centered case, the sum is done on component 0 and then the offset is applied to all components.  Do your problems have solvability issues?  If not, we could simply put an assertion saying singular fix is not supported for multi components and we will fix it when needed.

(2) I think the modification to applybc to fill corners have some issues.  First, for corner cells covered by other grids, the corner cells could be filled by valid cells of other grids, instead of by extraploating from neighboring cells.  Second, for each box there is a loop over faces.  Currently the corner fix is called more times than it is needed, because it doesn't need to be called for every loop over faces.  I am going to submit a pull-request to the MultiComponentMLMG branch to fix these.   There is another issue that is not easy to fix.  (But it seems possible it doesn't really matter.)  The corner of a Box may be a non-corner ghost cell of another Box.  If we know how to fill that non-corner ghost cell, that value should be preferred for filling the corner cell rather than exptrapolation.

bsrunnels:
@WeiqunZhang Thank you for working on this. To answer your comments:

(1) I did not consider singular problems. I don't have any solvability issues, but all of my applications involve self-adjoint operators, so there may be problems I'm not catching. So I'm perfectly fine with putting in the assertion as you suggested.
(2) I did not realize how involved the corner-filling would be and thank you for your work on improving it. I agree that the corner fix is likely using sub-optimal data and is being called more times than needed. But from a cache-awareness standpoint, it seems like it might actually be more efficient to simply recompute the value than to do a possibly expensive memory access?

PR #228, opened 2018-03-16
Add isCrossStencil virtual function to MLCellLinOp.  For non-cross st…

WeiqunZhang:
…encil, extend mask to corners and extrapolation is done only for non-covered corner cells.  Corners covered by valid cells from other grids are filled by FillBoundary.


WeiqunZhang:
Note that I haven't test it.

Issue #241, opened 2018-04-06
Commit be25c21d broke for the NAG Fortran compiler


nncarlson:
Commit be25c21 reworked some Fortran code (conversion from f77 to f90?) and in the process relied on some C processor features not recognized by the NAG compiler (and generally not regarded as Fortran preprocessing directives). Culprit files are `AMReX_COORDSYS_3D.F90` and the `1D` and `2D` versions; there may be others.  They include the new header file `AMRex_COORDSYS_F.H` which defines a bunch of macros like `DIM` which use the concatenation operator ## or an apparent alternative /**/ (empty C comment?).  It seems like this issue has cropped up before (and been fixed).

maxpkatz:
That commit did not involve any change to code, and that is not a new header file. Something about the change to running them through the F90 compiler instead of the F77 compiler must have broken this for NAG.

nncarlson:
Okay, let me investigate further.  Github seemed to indicate these were new files, but perhaps they were just renamed.

maxpkatz:
Thanks -- I still think we should fix this if we can, just want to better understand what the actual issue is.

WeiqunZhang:
In AMReX build system, `.F` files go through cpp preprocessor invoked by us.  But we don't do that for F90 files.  That's why.

WeiqunZhang:
We do not want to call `cpp` ourselves because it makes debug messages clearer if compiler does the preprocessing.

nncarlson:
Ah, that's it.
There's a way to define the concatenation operator from more primitive operations. Let me see if I can't fix up the include file for NAG.  Hopefully that is the only issue.

WeiqunZhang:
Could you give the latest development branch a try? @nncarlson

nncarlson:
@WeiqunZhang I didn't intend for this to distract you from your main work (just yet).  But thanks!  I tried the latest development branch and there are still some macro issues in `AMReX_REAL.H`

Yesterday I used a different approach of fixing the header files for NAG.  I made a PR just now to show you what that looks like.  I should have done the PR earlier, but got distracted by some issues on my end while testing it.  You *might* want to consider it, as it's much less invasive than your changes to source (although the new source that doesn't rely as much on macros is cleaner IMHO).  Note that I don't believe any of my code uses this part of AMReX, so I'm not testing whether the changes are good, just that it compiles.

Also note that the PR changes to `AMReX_REAL.H` likely will fix the remaining errors from your commit.

[...]

PR #243, opened 2018-04-07
Nag fix for #241

nncarlson:
This is a fix for #241 that only involves some additional special cases for the NAG compiler in several header files.  This is a possible alternative to 2f4d3a8 (which didn't fully fix the problem).

*I'm necessarily looking for this to be merged.*  A PR seemed to be the easiest way of  showing a possible fix.  The definition of the `CONCAT` macro for NAG is done in two separate places where it is needed, and there may be a better place to put it.

### DIFF
--- a/Src/Base/AMReX_ArrayLim.H
+++ b/Src/Base/AMReX_ArrayLim.H
@@ -111,6 +111,55 @@
 #define DIMARG(a) a##(1),a##(2),a##(3)
 #endif
 
+#elif NAGFOR
+
+#ifndef CONCAT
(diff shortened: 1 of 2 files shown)


nncarlson:
Note that this "low tech" version of the concatenation operator should work with any Fortran preprocessor, not just NAG.  So all the special cases could be eliminated in favor of this single approach (especially the one using the /**/ empty C comment).  Not as convenient as ##, but works for all.

nncarlson:
This is now moot.

PR #246, opened 2018-04-12
Sy/hypre - HYPRE IJ matrix interface

shashankNREL:
Added the HYPRE IJ Matrix in AMREX. With this functionality we can provide a generalized matrix to HYPRE to solve using Krylov solvers with BoomerAMG as preconditioner or BoomerAMG as the solver. Removes the following 7 point stencil constraints we had with Struct and SStruct interfaces in HYPRE. Will be useful for higher order discretization. 
Have tested it and compared against the SStruct interface already implemented in HYPRE. Used the Tutorials/HYPRE/ABecLaplacian test case to test this functionality. @asalmgren and @WeiqunZhang  please review this and let me know what you think.

### DIFF
--- a/Src/Extern/HYPRE/AMReX_HABEC_3D.F90
+++ b/Src/Extern/HYPRE/AMReX_HABEC_3D.F90
@@ -121,6 +121,7 @@ subroutine hbvec(vec, &
   else
      print *, "hbvec: impossible face orientation"
   endif
+
 end subroutine hbvec
 
 subroutine hbvec3(vec, &
@@ -300,6 +301,7 @@ subroutine hbvec3(vec, &
   else
      print *, "hbvec3: impossible face orientation"
   endif
+
 end subroutine hbvec3
 
 subroutine hmac(mat, a, &
@@ -332,6 +334,7 @@ subroutine hmac(mat, a, &
         enddo
      enddo
   endif
+  
 end subroutine hmac
 
 
@@ -459,7 +462,7 @@ end subroutine hmbc
 
 subroutine hmbc_ij(b, DIMS(bbox), DIMS(reg), &
                    beta, dx, n, &
-                   Amat, Index, DIMS(Gbox), ndim, bcFace) bind(C, name="amrex_hmbc_ij")
+                   Amat, Index, DIMS(Gbox)) bind(C, name="amrex_hmbc_ij")
 
   use amrex_fort_module, only : rt => amrex_real
   integer :: DIMDEC(bbox)
@@ -469,10 +472,9 @@ subroutine hmbc_ij(b, DIMS(bbox), DIMS(reg), &
   real(rt)  :: beta, dx(3)
   integer :: DIMDEC(Gbox)
   integer :: Index(DIMV(Gbox))
-  integer, intent(in) :: ndim
-  integer,dimension(ndim) :: bcFace
   ! HYPRE objects
   integer(kind=8) :: Amat
+
   integer :: nrows, ncols, rows
   integer, dimension(3) :: cols
   real(rt), dimension(3) :: values
@@ -485,9 +487,10 @@ subroutine hmbc_ij(b, DIMS(bbox), DIMS(reg), &
   if (n == 0) then
 
      fac = beta / (dx(1)**2)
+     
      do k = reg_l3, reg_h3
         do j = reg_l2, reg_h2
-           do i = reg_l1+bcFace(1), reg_h1-bcFace(4)
+           do i = reg_l1+1, reg_h1-1
               
               rows      = Index(i,j,k)
               
@@ -496,21 +499,21 @@ subroutine hmbc_ij(b, DIMS(bbox), DIMS(reg), &
               
               cols(2)   = Index(i-1,j,k)
               values(2) = -fac * b(i,j,k)
-
+              
               cols(3)   = Index(i+1,j,k)
               values(3) = -fac * b(i+1,j,k)
               
               call HYPRE_IJMatrixAddToValues(Amat, nrows, ncols, rows, cols, values, ierr)
               
            end do
-        end do 
+        end do
      end do
-
+     
   elseif (n == 1) then
 
      fac = beta / (dx(2)**2)
      do k = reg_l3, reg_h3
-        do j = reg_l2+bcFace(2), reg_h2-bcFace(5)
+        do j = reg_l2+1, reg_h2-1
            do i = reg_l1, reg_h1
               
               rows = Index(i,j,k)
@@ -533,7 +536,8 @@ subroutine hmbc_ij(b, DIMS(bbox), DIMS(reg), &
   else
 
      fac = beta / (dx(3)**2)
-     do k = reg_l3+bcFace(3), reg_h3-bcFace(6)
+
+     do k = reg_l3+1, reg_h3-1
         do j = reg_l2, reg_h2
            do i = reg_l1, reg_h1
 
@@ -577,6 +581,7 @@ subroutine hmmat(mat, &
   real(rt)         :: h, fac, bfm, bfv
   real(rt)         :: bfm2, h2, th2
   integer :: i, j, k
+
   if (cdir == 0 .OR. cdir == 3) then
      h = dx(1)
   elseif (cdir == 1 .OR. cdir == 4) then
@@ -584,7 +589,9 @@ subroutine hmmat(mat, &
   else
      h = dx(3)
   endif
+
   fac = beta / (h**2)
+
   if (bct == LO_DIRICHLET) then
      if (bho >= 1) then
         h2 = 0.5e0_rt * h
@@ -602,6 +609,7 @@ subroutine hmmat(mat, &
      print *, "hmmat: unsupported boundary type"
      stop
   endif
+  
   if (cdir == 0) then
      i = reg_l1
      do k = reg_l3, reg_h3
@@ -707,9 +715,10 @@ subroutine hmmat_ij(DIMS(reg), &
   real(rt)         :: h, fac, bfm, bfv
   real(rt)         :: bfm2, h2, th2
   integer :: i, j, k, ierr
+
   integer, dimension(3)  :: cols
   real(rt), dimension(3) :: values
-  integer, dimension(1) :: ncols
+  integer :: ncols
   integer :: rows, nrows
 
   if (cdir == 0 .OR. cdir == 3) then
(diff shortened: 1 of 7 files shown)


WeiqunZhang:
I am going to make some comments.

Src/Extern/HYPRE/AMReX_HABEC_3D.F90
```diff
@@ -11,6 +11,7 @@ module amrex_habec_module
 
   use amrex_fort_module, only : rt => amrex_real
   implicit none
+  integer(kind=8), parameter :: HYPRE_PARCSR = 5555
```
WeiqunZhang: This integer parameter is not used anywhere. Do we really need it?

Src/Extern/HYPRE/AMReX_HABEC_3D.F90
```diff
@@ -299,13 +301,13 @@ subroutine hbvec3(vec, &
...
 
+end subroutine hbvec3
 
 subroutine hmac(mat, a, &
-                DIMS(abox), &
-                DIMS(reg), &
-                alpha) bind(C, name="amrex_hmac")
+     DIMS(abox), &
+     DIMS(reg), &
+     alpha) bind(C, name="amrex_hmac")
```
WeiqunZhang: If possible, could we get rid of those white space only changes?  There are a lot of white space changes in several files, not just here.

Tests/LinearSolvers/MLMG/GNUmakefile
```diff
@@ -11,12 +11,17 @@ USE_MPI   = TRUE
 #USE_OMP   = TRUE
 USE_OMP   = FALSE
 
+USE_HYPRE = TRUE
```
WeiqunZhang: Could we make the default of this be FALSE?

Tests/LinearSolvers/MLMG/GNUmakefile
```diff
@@ -11,12 +11,17 @@ USE_MPI   = TRUE
...
+TINY_PROFILE = TRUE
+
 include $(AMREX_HOME)/Tools/GNUMake/Make.defs
 
 include ./Make.package
 include $(AMREX_HOME)/Src/Base/Make.package
 include $(AMREX_HOME)/Src/Boundary/Make.package
 include $(AMREX_HOME)/Src/LinearSolvers/MLMG/Make.package
 include $(AMREX_HOME)/Src/LinearSolvers/C_CellMG/Make.package
+include $(AMREX_HOME)/Src/Extern/HYPRE/Make.package
```
WeiqunZhang: How about

    ifeq ($(USE_HYPRE),TRUE)
        include $(AMREX_HOME)/Src/Extern/HYPRE/Make.package
    endif

Src/Extern/HYPRE/AMReX_HypreABecLap3.cpp
```diff
@@ -0,0 +1,503 @@
...
+
+  // Arrays needed to build the global indices for all procs
+  CellsGIndex.resize(grids.size());
+  numCellsProc.resize(num_procs+1);
+
+  // These arrays store the starting global indices of all the boxes involved
+  StartIndex.resize(grids.size());
+
+  // Initialize all the arrays to 0
+  std::fill(CellsGIndex.begin(), CellsGIndex.end(), 0);
```
WeiqunZhang: When the vector is resized, we could do `CellsGIndex.resize(grids.size(), 0);`.  So `std::fill` is unnecessary.

Src/Extern/HYPRE/AMReX_HypreABecLap3.cpp
```diff
@@ -0,0 +1,503 @@
...
+  for (int i = 0; i < grids.size(); i++) {
+    const Box& bx = grids[i];
+    CellsGIndex[i] = numCellsProc[ dmap[i]+1 ];
+    numCellsProc[ dmap[i]+1 ] = numCellsProc[ dmap[i]+1 ] + bx.numPts();
+  }
+
+  // making sure that counting starts from 0 for procId = 0
+  for (int i = 1; i< (num_procs+1); i++) {
+    if (numCellsProc[i] == 0) {
+      amrex::Abort("No rows in this core - Check domain decomposition");
```
WeiqunZhang: Do we have to abort?  Does Hypre work if process 0 has no cells?

Src/Extern/HYPRE/AMReX_HypreABecLap3.cpp
```diff
@@ -0,0 +1,503 @@
...
+  // Box starting index starting from the proc offset
+  for (int i = 0; i < grids.size(); i++) {
+    CellsGIndex[i] = CellsGIndex[i] + numCellsProc[dmap[i]];
+  }
+
+  // Starting and ending global index for each box
+  for (int i = 0; i < (grids.size()-1); i++) {
+    StartIndex[i] = CellsGIndex[i];
+  }
+  StartIndex[grids.size()-1] = CellsGIndex[grids.size()-1];
```
WeiqunZhang: Could this be merged into the loop above?

Src/Extern/HYPRE/AMReX_HypreABecLap3.cpp
```diff
@@ -0,0 +1,503 @@
...
+                           ARLIM(reg.hiVect()), CellsGIndex[i]);
+  }
+
+  const int nghost = 0;
+  bool local = true;
+  int ilower = GbInd.min(0, nghost, local);
+  int iupper = GbInd.max(0, nghost, local);
+
+  // Fill the global indices in the ghost cells along the grid edges
+  GbInd.FillBoundary();
```
WeiqunZhang: Let's do `GbInd.FillBoundary(geom.periodicity())` to fill periodic boundary too.

Src/Extern/HYPRE/AMReX_HypreABecLap3.cpp
```diff
@@ -0,0 +1,503 @@
...
+
+    // build the global index for all cells on this level
+    amrex_BuildGlobalIndex(BL_TO_FORTRAN(GbInd[mfi]), ARLIM(reg.loVect()),
+                           ARLIM(reg.hiVect()), CellsGIndex[i]);
+  }
+
+  const int nghost = 0;
+  bool local = true;
+  int ilower = GbInd.min(0, nghost, local);
+  int iupper = GbInd.max(0, nghost, local);
```
WeiqunZhang: Seems that min and max are already in `numCellsProc`.

WeiqunZhang: If we do that, we need to modify the upper bound of the loop at line 67.

Src/Extern/HYPRE/AMReX_HypreABecLap3.cpp
```diff
@@ -0,0 +1,503 @@
...
+
+  // Fill the numpoints in the box and in the proc where this box resides
+  for (int i = 0; i < grids.size(); i++) {
+    const Box& bx = grids[i];
+    CellsGIndex[i] = numCellsProc[ dmap[i]+1 ];
+    numCellsProc[ dmap[i]+1 ] = numCellsProc[ dmap[i]+1 ] + bx.numPts();
+  }
+
+  // making sure that counting starts from 0 for procId = 0
+  for (int i = 1; i< (num_procs+1); i++) {
```
WeiqunZhang: Seem that this can be done with http://en.cppreference.com/w/cpp/algorithm/partial_sum

Src/Extern/HYPRE/AMReX_HypreABecLap3.cpp
```diff
@@ -0,0 +1,503 @@
...
+          bd.setBoundCond(Orientation(n, Orientation::high),
+                          i, comp, ibnd);
+        }
+      }
+    }
+  }
+}
+
+void HypreABecLap3::loadMatrix() {
+  static_assert(BL_SPACEDIM > 1, "HypreABecLap2: 1D not supported");
```
WeiqunZhang: HypreABecLap2 --> 3

Src/Extern/HYPRE/AMReX_HypreABecLap3.cpp
```diff
@@ -0,0 +1,503 @@
...
+
+  for (MFIter mfi(acoefs); mfi.isValid(); ++mfi) {
+    int i = mfi.index();
+    const Box &reg = mfi.validbox();
+
+    int volume = reg.numPts();
+
+    // build matrix interior
+
+    amrex_hmac_ij(BL_TO_FORTRAN(acoefs[mfi]), ARLIM(reg.loVect()),
```
WeiqunZhang: amrex_hmac_ij and other Fortran routines call functions like `HYPRE_HYPRE_IJMatrixAddToValues`.  These are C functions, right?  I am worried about portability issues of calling them from Fortran.  Also `integer(kind=8)` does not necessarily mean an 8-bytes integer.

I think we do the following,

    IArrayBox cols, rows, ncols;
    FArrayBox values;
    for (MFIter mfi(acoefs); ....) {
         // .....
        cols.resize(reg);  rows.resize(reg); ....; values.resize(reg);
        amrex_hmac_ij(...., cols.dataPtr(), rows.dataPtr(), ncols.dataPtr(), values.dataPtr(),...);
        HYPRE_IJMatrixAddtoValuse(A, nrows.dataPtr(), ncols.dataPtr(), ....);
    }

    subroutine hmac_ij(....., cols, ...)
        integer :: cols(*)
    end subroutine hmac_ij

WeiqunZhang: I see later you call `HYPRE_IJMatrixAddToValues` inside loops in Fortran.  So my approach above won't work in that case.  An alternative approach is pass `&A` as type(c_ptr) by value to Fortran.  In Fortran, we call our own C function that in turn call Hypre.

    amrex_hmmat_ij(..., &A, ....);

    void amrex_hmmat_ij(..., void*, ...);  // in _F.H 

    interface
      void our_own_c_function_that_wraps_HYPRE_IJMatrixAddToValues(A,...) bind(c)
          type(c_ptr), value :: A
    end interface

     subroutine hmac_ij(..., Amat, ...) bind(c)
        type(c_ptr), value :: Amat
        ! ...
        call our_own_c_function_that_wraps_HYPRE_IJMatrixAddToValues(Amax,...)

     void our_own_c_function_that_wraps_HYPRE_IJMatrixAddToValues(double* A, ...)  // Is there Hypre_real that's typedefed double?

Src/Extern/HYPRE/AMReX_HypreABecLap3.cpp
shashankNREL: I tried but HYPRE seems to crash if a process has no cells. We create a sub communicator and make sure all the cells are distributed among all the ranks in the sub communicator and so we should be ok.

Src/Extern/HYPRE/AMReX_HABEC_3D.F90
shashankNREL: Good catch. I'll remove it.

Tests/LinearSolvers/MLMG/GNUmakefile
shashankNREL: Sure. Made the change.

Src/Extern/HYPRE/AMReX_HypreABecLap3.cpp
shashankNREL: All very good suggestions. Made the changes and will push the code now.

PR #266, opened 2018-05-31
Adding some particle support to the F_Interfaces

atmyers:
This adds support for particles to the Fortran Interfaces and adds tracer particles to the Advection_F tutorial. Right now this is somewhat hard-wired for tracers, but this will be relaxed in the future.


WeiqunZhang:
Will review this after tomorrow's monthly merge.

Src/F_Interfaces/Particle/AMReX_particlecontainer_mod.F90
```diff
@@ -0,0 +1,135 @@
...
+       type(c_ptr), value :: pc
+     end subroutine amrex_fi_init_particles_one_per_cell
+
+     subroutine amrex_fi_write_particles (pc, dirname, pname, is_checkpoint) bind(c)
+       import
+       implicit none
+       type(c_ptr), value :: pc
+       character(kind=c_char), intent(in) :: dirname(*)
+       character(kind=c_char), intent(in) :: pname(*)
+       logical, intent(in)                :: is_checkpoint
```
WeiqunZhang: (1) This needs to be value.  (2) logical is not necessary the same type as C++ bool.  For portability, we need to use `integer(c_int)` and `int`.  (Note that we still want to use `logical` for Fortran function argument, but we do a conversion to `integer(c_int)` before the call to `amrex_fi_write_particles`.

Src/F_Interfaces/Particle/AMReX_particlecontainer_fi.cpp
```diff
@@ -0,0 +1,82 @@
...
+            np = particle_tile.numParticles();
+            if (np > 0) {
+                auto& aos = particle_tile.GetArrayOfStructs();
+                dp = aos.data();
+            } else {
+                dp = NULL;
+            }            
+        } else {
+            np = 0;
+            dp = NULL;
```
WeiqunZhang: Why not using `nullptr`?

Src/F_Interfaces/Particle/AMReX_particlecontainer_mod.F90
```diff
@@ -0,0 +1,135 @@
...
+    character(len=*), intent(in) :: dirname
+    character(len=*), intent(in) :: pname
+    logical, intent(in)          :: is_checkpoint    
+    call amrex_fi_write_particles(particlecontainer, &
+         amrex_string_f_to_c(dirname), amrex_string_f_to_c(pname), is_checkpoint)
+  end subroutine amrex_write_particles
+
+  subroutine amrex_particle_redistribute (lev_min,lev_max,nghost)
+    integer, optional, intent(in) :: lev_min, lev_max, nghost
+    integer(c_int) :: min, max, ng
```
WeiqunZhang: We might want to avoid using `min` and `max` since they are Fortran keywords.

Src/F_Interfaces/Particle/AMReX_particlecontainer_mod.F90
```diff
@@ -0,0 +1,135 @@
...
+    if(present(lev_max)) max = lev_max
+    if(present(nghost )) ng = nghost
+    call amrex_fi_particle_redistribute(particlecontainer, min, max, ng)
+  end subroutine amrex_particle_redistribute
+
+  function amrex_get_particles(lev, mfi) result(particles)
+    integer(c_int),     intent(in) :: lev
+    type(amrex_mfiter), intent(in) :: mfi
+    type(amrex_particle),  pointer :: particles(:)
+    type(c_ptr), target            :: data
```
WeiqunZhang: Probably no need to be `target`.

Tutorials/Amr/Advection_F/Source/Src_2d/advect_2d_mod.F90
```diff
@@ -124,4 +124,91 @@ subroutine advect(time, lo, hi, &
...
+             e_hi = w_hi
+             e_lo = w_lo
+             e_hi(d) = length(d) - cc_cell(d)
+             do j = 1, 2
+                if (e_hi(j) > 1.d0) then
+                   e_hi(j) = 1.d0
+                end if
+                if (e_hi(j) < 0.d0) then
+                   e_hi(j) = 0.d0
+                end if
```
WeiqunZhang: Since this is Fortran, I believe we can write `e_hi = max(0.d0,min(1.d0,e_hi))`.

PR #267, opened 2018-06-01
Fix --ghost option.  Do not advance farg in case(--ghost) section bec…

cgilet:
…ause --ghost does not take any arguments.


maxpkatz:
Can you please change the target to the development branch?

PR #270, opened 2018-06-05
Forkjoin

cychan-lbnl:
some changes to support forkjoin


WeiqunZhang:
It does not compile when USE_MPI=FALSE.

cychan-lbnl:
Should be fixed now.

WeiqunZhang:
There is a problem with

```
enum class ReduceOp : MPI_Op {
        max  = MPI_MAX,
        min  = MPI_MIN,
        sum  = MPI_SUM,
        lor  = MPI_LOR,
        land = MPI_LAND
    };
```
`MPI_Op` is not necessarily an integral type.  So this does not compile with OpenMPI.

cychan-lbnl:
Thanks!  I changed it to use an unordered_map to convert the enum to MPI_Op type.  Hopefully that fixed it.

PR #283, opened 2018-07-10
Amr derive spectrum

johnpwakefield:


zingale:
fyi... we've been hacking on AmrDeriveSpectrum here:

https://github.com/AMReX-Astro/AmrDeriveSpectrum

PR #290, opened 2018-07-21
Merge particle support into GPU branch

maxpkatz:


Tools/GNUMake/comps/pgi.mak
```diff
@@ -54,7 +54,8 @@ ifeq ($(USE_OMP),TRUE)
 endif
 
 ifeq ($(USE_ACC),TRUE)
-  GENERIC_PGI_FLAGS += -acc -Minfo=acc -ta=nvidia -mcmodel=medium
+  # GENERIC_PGI_FLAGS += -acc -Minfo=acc -ta=nvidia -mcmodel=medium
```
maxpkatz: Can you please put back -mcmodel=medium, and remove the commented line?

Tools/GNUMake/comps/pgi.mak
```diff
@@ -54,7 +54,8 @@ ifeq ($(USE_OMP),TRUE)
...
+  GENERIC_PGI_FLAGS += -acc -ta=tesla:$(CUDA_VERSION),cuda9.0,lineinfo,ptxinfo -Minfo=accel
```
maxpkatz: I think the cuda9.0 should not be here?

maxpkatz: Same for the USE_CUDA block below.

maxpkatz:
Is it OK to do this as a squash merge?

asalmgren:
It would be so cool if I knew what a squash merge is!

maxpkatz:
@asalmgren it means that the commit history will not be preserved; instead, all changes in the gpu_particles branch will be condensed into a single new commit that is added to the gpu branch.

atmyers:
That's fine with me.

PR #296, opened 2018-08-01
Add OpenMP GPU offload version of HeatEquation_EX1_C tutorial

bcfriesen:


bcfriesen:
Don't know why the CI failed:

```
In file included from tmp_build_dir/s/3d.gnu.MPI.EXE/AMReX_RedistributeStrategyCPU.cpp:1:0:
/home/travis/build/AMReX-Codes/amrex/Src/Particle/AMReX_RedistributeStrategy.H:13:23: fatal error: Particles.H: No such file or directory
 #include <Particles.H>
                       ^
compilation terminated.
```

I did not change any of those files ...

bcfriesen:
Oops, didn't mean to close this.

bcfriesen:
I'm confused, that file is nowhere in the `gpu` branch or in my `omp45` branch.

bcfriesen:
Oh I see, @atmyers just pushed 2221e56fec1b6175946e45be50dad07a208e8c28 which adds that file. Blame him!

atmyers:
That was my fault and happens on the vanilla gpu branch - it has nothing to do with your PR. I'll fix momentarily.

maxpkatz:
I want to think about the incompatibility with my UVM allocations, please don't merge this yet.

bcfriesen:
OK. I will rebase this on top of `gpu` with the fixed stuff so that the CI passes (I don't like those merge commits).

Tools/GNUMake/Make.defs
```diff
@@ -111,6 +111,12 @@ ifeq ($(USE_PMI), TRUE)
   DEFINES += -DAMREX_PMI
 endif
 
+ifeq ($(USE_OMP_OFFLOAD), TRUE)
+  DEFINES += -DOMP_OFFLOAD
+  INCLUDE_LOCATIONS += $(CUDA_DIR)/include
```
maxpkatz: There is some logic down below that should cover this if MAKE_CUDA_PATH is set.

Src/Base/AMReX_BArena.cpp
```diff
@@ -31,6 +34,13 @@ amrex::BArena::alloc (std::size_t _sz)
 	gpu_malloc(&pt, &_sz);
 
     }
+// If combining the OpenMP offloading code with the existing CUDA_UM and
+// AMREX_USE_CUDA code, the tutorials generate the wrong answer.  I do not know
+// why. So until that gets sorted out, the OpenMP offloading code can't re-use
+// any of the existing CUDA code here. This will lead to some duplication (such
+// as these calls to cudaMallocManaged() and cudaFree()).
+#elif (defined(OMP_OFFLOAD))
+    cudaMallocManaged(&pt, _sz);
```
maxpkatz: Can you test whether this issue is due specifically to the use of CArena by commenting out DEFINES += -DBL_COALESCE_FABS in Make.defs?

Tools/GNUMake/Make.defs
bcfriesen: Ah OK, I deleted these and it works, thanks.

Src/Base/AMReX_BArena.cpp
bcfriesen: I tested with `USE_OMP_OFFLOAD=TRUE USE_CUDA=TRUE` and removed the line `DEFINES += -DBL_COALESCE_FABS`. Removing that line makes no difference - it gives the same result as if the line were present (which is good), but the answer is still wrong (which is bad).

maxpkatz: OK. One thing I realized is that we're still using CUDA Fortran to do cudaMallocManaged which I would independently like to change and do it in C++. Let me try that and see if that makes any difference.

PR #298, opened 2018-08-12
Do not initialize dt_level on levels that do not yet exist

maxpkatz:
The way we set dt_level and dt_min in Amr::FinalizeInit is problematic if the level 0 timestep when the simulation starts is significantly different from the timestep later on in the simulation when we add refinement. In particular, this will happen if the simulation code shrinks the initial timestep by some factor. If the fine levels are only added later on once the timestep has caught back up to its normal limited level, then the first timestep on the fine level will be effectively limited by the initial level 0 timestep, which can sharply decrease the timestep for no reason.

### DIFF
--- a/Src/Amr/AMReX_Amr.cpp
+++ b/Src/Amr/AMReX_Amr.cpp
@@ -1285,7 +1285,7 @@ Amr::FinalizeInit (Real              strt_time,
     dt_min[0]  = dt_level[0];
     n_cycle[0] = 1;
 
-    for (int lev = 1; lev <= max_level; lev++)
+    for (int lev = 1; lev <= finest_level; lev++)
     {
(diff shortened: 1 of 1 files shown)


asalmgren:
Does this change require any change in the application code to make sure
that dt_level is set correctly on a level newly formed mid-simulation?

maxpkatz:
I do not think it requires any change. When a new level is created after a regrid, we already divide the timestep on those new levels by dt[lev-1] / n_cycle regardless of whether the application code does anything special. With that being said, I am not completely certain this will have zero downstream effect so it would be nice to see if the test suites for Castro and at least one other code show any diffs. @zingale can you do this for Castro on this branch?

zingale:
Castro passed

PR #300, opened 2018-08-16
Tutorial AdvectionScalar_F with checkpoint restart implemented

nataraj2:
Have used the restart routines from the Advection_AmrCore tutorial and implemented in this Fortran tutorial. Have verified the results by comparing against the results from the Advection_AmrCore tutorial (cases where the number of levels is changed during restart have also been checked for).

This new tutorial was created since the existing Advection_F tutorial has particle advection by default, and checkpoint restart does not handle particle advection. This tutorial contains no particle advection. Just a single scalar advection.


WeiqunZhang:
Sorry for closing this.  I don't think we will have time to review this.

PR #308, opened 2018-09-07
Add per-fab version of FineAdd to fluxregister's Fortran interface

kweide:
Adding a per-fab version of FineAdd, of which various versions already exist in the C++ code, to fluxregister's Fortran interface.

This is a step towards having Fortran code (FLASH in particular) using flux arrays that do not always have to be allocated for all boxes in a whole level.

Suggestions for changes are welcome.

Tested within a modified version of FLASH-subset with 2D Cartesian Sedov.


WeiqunZhang:
I have a suggestion.  `FArrayBox` has a constructor that takes a pointer and builds a non-owning `FArrayBox`.  By non-owning, I mean the pointer will not be deleted by the destructor of `FArrayBox`.

    FArrayBox (const Box& b, int ncomp, Real* p);

If we use it, we don't have to add a new `FineAdd` function to `FluxRegister` that is almost identical to the existing one.

kweide:
Sure, I agree with your suggestion! I might have done it that way if it had been obvious to me how...

Would you please do the necessary changes? Or alternatively, let me know the elegant way to use the `FArrayBox` constructor in `amrex_fi_fluxregister_fineadd_1fab_1dir`?

WeiqunZhang:
OK.  I will do it.

PR #309, opened 2018-09-10
use __BYTE_ORDER__ to detect endianess

WeiqunZhang:
It seems that using `__BYTE_ORDER__` is more portable way of detecting endianess.


atmyers:
I tested this on Summit-dev with pgi and ibm compilers, and also on my apple laptop with clang. Those all seem to work.

PR #310, opened 2018-09-11
Fortran build routine for non-owning fab

kweide:
@WeiqunZhang Your improvement to #308 works for me, thanks!

I wanted something like this on the Fortran side, too, i.e., a routine 
"that takes a pointer and builds a non-owning" FAB. Thus this new PR.

Again please let me know if you would prefer to do this in a different way - or just make appropriate changes.

### DIFF
--- a/Src/F_Interfaces/AmrCore/AMReX_fluxregister_mod.F90
+++ b/Src/F_Interfaces/AmrCore/AMReX_fluxregister_mod.F90
@@ -145,10 +145,10 @@ contains
     call amrex_fi_fluxregister_fineadd(this%p, mf, scale)
   end subroutine amrex_fluxregister_fineadd
 
-  subroutine amrex_fluxregister_fineadd_1fab (this, fluxfabs, boxno, scale)
+  subroutine amrex_fluxregister_fineadd_1fab (this, fluxfabs, gridIdx, scale)
     class(amrex_fluxregister), intent(inout) :: this
     type(amrex_fab),  intent(in) :: fluxfabs(amrex_spacedim)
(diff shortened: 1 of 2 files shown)


kweide:
(1) It's fine to have the generic name be the same as one of the specific names in Fortran, at least I am not aware of any problems with it. But to avoid confusion, sure we can change one of the names.

(2) Yes I did not pay much attention to error checking. My idea of the interface was that bx has to be consistent with the Fortran pointer passed in, that is the sizes in all three spatial indices have to be the same, and nc `.le.` the last size of dp.

kweide:
(Is there an ASSERT macro for use in Fortran code?)

WeiqunZhang:
Unfortunately, the word macro does not appear in the Fortran standard and it's known some compilers don't like it.   But `ifdef` seems to have wide support and we have relied on it anyway.  So you can put the checks inside `#ifdef AMREX_DEBUG`.

kweide:
I have pushed an additional commit to address your points (1) and (2).

I am using ERROR STOP, do not know about compatibility but I assume that Fortran compilers that are modern enough to compile AMReX understand it. My old GNU Fortran (Ubuntu 4.8.4-2ubuntu1~14.04.3) 4.8.4 does. Is there another preferred way - for AMReX Fortran code in particular?

WeiqunZhang:
We like to use `amrex_error` in `amrex_error_module`.   For example, `call amrex_error("error message")`.   This could provide a backtrace.

kweide:
Now using `amrex_error` as suggested.

PR #331, opened 2018-10-05
Mc nodelinop

bsrunnels:
This update makes minor modifications to MLNodeLinOp and MLMG to support multicomponent solves.


Src/LinearSolvers/MLMG/AMReX_MLLinOp.cpp
```diff
@@ -342,6 +342,7 @@ MLLinOp::make (Vector<Vector<MultiFab> >& mf, int nc, int ng) const
         {
             const auto& ba = amrex::convert(m_grids[alev][mlev], m_ixtype);
             mf[alev][mlev].define(ba, m_dmap[alev][mlev], nc, ng, MFInfo(), *m_factory[alev][mlev]);
+	    mf[alev][mlev].setVal(0.0);
```
WeiqunZhang: Do we have to have this setVal here?  Could we do it where make is called?

Src/LinearSolvers/MLMG/AMReX_MLNodeLap_nd.F90
```diff
@@ -32,26 +32,29 @@ subroutine amrex_mlndlap_copy_fine_node (lo, hi, d, dlo, dhi, s, slo, shi, m, ml
...
-                res(i,j,k) = 0.0
-             else
-                res(i,j,k) = b(i,j,k) - res(i,j,k)
-             end if
+    integer :: i,j,k,n
+
+    do          k = lo(3), hi(3)
+       do       j = lo(2), hi(2)
+          do    i = lo(1), hi(1)
+             do n = 1, nc
```
WeiqunZhang: Let's make `do n` the outermost loop.

Src/LinearSolvers/MLMG/AMReX_MLLinOp.cpp
bsrunnels: My thought was that it is good practice to instantiate right after definition. However we can certainly instantiate elsewhere if that is the amrex convention or if it would improve performance.

Src/LinearSolvers/MLMG/AMReX_MLNodeLap_nd.F90
bsrunnels: Sure. I will update accordingly.

Src/LinearSolvers/MLMG/AMReX_MLLinOp.cpp
WeiqunZhang: Besides performance, it will help debugging.

bsrunnels: Ok sounds good. I will make the change.

WeiqunZhang:
My question is about why we need to insert `res.setVal` there.  If the code breaks without that setVal, something might be wrong and we need to understand.

bsrunnels:
I see. In our implementation we assumed that all levels were pre-initialized and errored out if they were not. In that case I will remove `setVal` from here and we will just modify our code accordingly.

PR #333, opened 2018-10-09
gpu-mm branch into development.

kngott:
This is the first attempt to merge the gpu-mm branch into the current development+gpu branch. The overall intention: ensure CPU functionality doesn't degrade and Castro functionality doesn't break, but add the additional tools and functionality that has been developed in the branch.

This branch doesn't work with Cuda version 9.2, as the --expt-relaxed-constexpr --expt-extended-lambda flags are explicitly turned off by the compiler in Cuda 9.2.

Comparison of Castro GPU Regression test, Sedov:
--------------
After converting Castro to the same lambda method found in gpu-merge-test (found in Castro gpu-merge-test2), the branches have nearly identical results as the development branches for large box sizes (gpu-merge-test ~1.5% faster at 128). For smaller box sizes, gpu-merge-test branches show substantial improvement (~30% speedup).

Castro development + amrex gpu-merge-test shows a slow down overall (60% slow down for large box sizes, 6% for small boxes) however the overall conversion is simple, and the difference is negligible at smaller box sizes. (The current theory: this is due to the differences in managed and non-managed object metadata between the two branches.)

Numbers:
............
CUDA initialized with 1 GPU per MPI rank
MPI initialized with 4 MPI processes

amr.n_cell           =  256 256 256
amr.max_grid_size   = 128

............

Castro development + AMReX development
Run time = 25.26208159
Run time without initialization = 12.80058071
Average number of zones advanced per microsecond: 13.107

Castro development + AMReX gpu-merge-test
Run time = 33.51515886
Run time without initialization = 20.95590505
Average number of zones advanced per microsecond: 8.006

Castro gpu-merge-test + AMReX gpu-merge-test
Run time = 25.08297292
Run time without initialization = 12.59683173
Average number of zones advanced per microsecond: 13.319

............
amr.n_cell           =  256 256 256
amr.max_grid_size   = 32

Castro development + AMReX development
Run time = 57.04528955
Run time without initialization = 43.15197161
Average number of zones advanced per microsecond: 3.888

Castro development + AMReX gpu-merge-test
Run time = 59.24962566
Run time without initialization = 45.94612525
Average number of zones advanced per microsecond: 3.651

Castro gpu-merge-test + AMReX gpu-merge-test
Run time = 48.55417838
Run time without initialization = 35.57122984
Average number of zones advanced per microsecond: 4.717

Change Details:
--------------
`AMReX_CudaQualifiers.H` contains some macros for marking functions \_\_host__, \_\_device__ and \_\_global__.

`AMReX_CudaLaunch.H` contains some function templates for kernel launching and some helper functions.

A new class template `CudaArray` is introduced.  For non-GPU build, this is an alias to `std::array`.  For GPU build (i.e., `USE_CUDA=TRUE`), `std::array` is not supported by CUDA.  Our `CudaArray` is a very simple array class that can be used for GPU code.  Macros in `AMReX_ArrayLim.H` using `std::array` have been changed to use `CudaArray`.  This allows us to use these dimension agnostic macros in GPU code. Additional std::array functions can be also added as needed.

In `AMReX_CudaAllocators.H`, some aliases have been added so that it can be included and used when not using CUDA.

A new class `Managed` is introduced providing `operator new` and `delete` that allocate/deallocate managed memory.

`amrex::Abort`, `Assert`, `Error`, and `Warning` have been modified so that they can be called from \_\_host__ and \_\_device__.

Many functions in `IndexType`, `IntVect`, `Orientation`, `OrientationIter`, and `Box` have been marked host and device.  If we need to pass these objects to kernels, we pass them by value.

`Geometry` class with static data is not CUDA friendly.  We add a new member function `data` that returns a POD `GeometryData` by value. This will allow us pass the data we need from `Geometry` to GPU kernels by value.

`BaseFab` now inherits the `Managed` class so that objects created with `new` have their meta-data in managed memory.  The data in `BaseFab` is allocated with `The_Arena()` (unless on device already) in managed memory when `USE_CUDA=TRUE`.  A lot of `BaseFab` functions are marked \_\_host__ and \_\_device___.  Many functions have specialization for the `Real` type if both `USE_CUDA=TRUE` and `USE_GPU_PRAGMA=TRUE` and `AMREX_USE_BASEFAB_CPP` macro is not defined.   The functions containing `pragma gpu` are processed by `write_cuda_headers.py`.  By default `USE_GPU_PRAGMA` is false and `AMREX_USE_BASEFAB_CPP` is undefined.  Thus by default the C++ template versions without pragmas are used.  

The `write_cuda_headers.py` script is modified so that these BaseFab functions can be used as device functions. 

We have modified the way `the_FA_arena` is initialized to avoid file scope static variable initialization issue.  More specifically, this guarantees the `FabArrayBase::use_cuda_aware_mpi` parameter is initialized before the arena is created.

`MFIter.H` has a new function `TilingIfNotGPU` that return true if not in launching region and false otherwise.  Currently, when `USE_CUDA=TRUE` and `USE_GPU_PRAGMA=TRUE`, tiling is effectively turned off because of the default tile size set in `AMReX_FabArrayBase.cpp`.   But by default (i.e., `USE_GPU_PRAGMA=FALSE`), the tile size is not changed.  Tiling can be turned off by using `TilingIfNotGPU`.

Currently in `development` branch, by default, if GPU codes are available, they will be used.  This is controlled by a boolean flag in `Device` class.  One can use functions in `Device` class to toggle the flag.  Here we have added a `ParmParse` parameter `device.in_launch_region` to change the default.

In this branch, the new kernel launch approach has been applied wherever needed for Castro's GPU regression test. However, the strategy is essentially the same as `#pragma gpu`.  In `AMReX_FabArrayCommI.H`, it is used to do copying data from Fab to Fab, and between Fab and buffer.  It is also used in some functions in `AMReX_MultiFab.cpp`.

Made a slightly modified version of particle_threads_and_blocks, n_particle_threads_and_blocks, that returns dim3 and slightly improves the robustness of the calculation. More can be done in the future.

Numerous Tutorials have been added to Tutorials/GPU to show how different varieties of these changes could be implemented.


WeiqunZhang:
For max_grid_size=32, the difference is not due to # of blocks and threads.  I tested AMReX gpu-merge-test and Castro-merge-test with a change in AMReX that made all kernels use the same threading strategy as development (Device::grid_stride_threads_and_blocks plus tri-nested loops).  The results are
```
Run time = 50.16782409
Run time without initialization = 37.15213844

  Average number of zones advanced per microsecond: 4.516
```

maxpkatz:
For whatever reason one of the things I found when I made the change to use grid-strided loops rather than one thread for every zone is that it resulted in a significant slowdown. This is something I had been intending to fix eventually by generalizing #pragma gpu to accept a box as an argument so that it could launch the thread-per-zone model in addition to the strided loop.

atmyers:
I have tested this with WarpX and everything is working smoothly. I think it is ready to merge. 

By my count, there are now four (!) versions of the Electromagnetic PIC tutorial that replicate a lot of code. Given that these aren't working at the moment, could you remove them from this PR and stash them on another branch? I'd like to fix the existing Tutorial and merge your lambda function version into it in a way that minimizes code duplication, but now that WarpX itself is (partially) running on GPUs, fixing the Tutorial hasn't been a high priority task. This way, it wouldn't hold up merging this PR, which I think is ready to go.

WeiqunZhang:
This past weekend, I ran regression tests on AMReX, Castro, IAMR and Nyx using this branch with USE_CUDA=FALSE.  All tests passed.

WeiqunZhang:
For consistence, I made Castro gpu-merge-test branch for working together with AMReX gpu-merge-test branch.

asalmgren:
Shall I try the mfix regression tests with the gpu-merge-test branch?

WeiqunZhang:
Yes, that would be great.

PR #339, opened 2018-10-13
Sensei insitu development

burlen:
@asalmgren This branch integrates the SENSEI in situ framework into AMReX. I tested it using the AmrLevel and AmrCore Advection SingleVortex tutorials and with IAMR. There is draft of documentation in the Visualization section of the sphinx docs.


drummerdoc:
Burlen, the GNUmakefile.in in the root folder is missing an endif before the `ifeq ($(USE_SENSEI_INSITU),TRUE)` line, and this is breaking the tests.  If you fix that, the pull request should automatically update and we can try again.

burlen:
Ah, I know what happened , recently I've been testing using the tutorials and iamr,  I hadn't tested building amrex as a library since rebasing from the development branch. I think that missing endif came about after rebase. it should be fixed now.

drummerdoc:
Approving this PR.  Note that there is/will be similar capabilities through Ascent.  For now it seems that any duplication here is fine.

PR #344, opened 2018-10-16
Intersection in fortran

jared321:
Hi, @WeiqunZhang.  I was playing around with performing intersections of boxes today.  I added the following to the Fortran interface to do this.  If this works for you, should we add it officially to the Fortran interface?  Please feel free to clean it up as you see fit.

### DIFF
--- a/Src/F_Interfaces/Base/AMReX_box_fi.cpp
+++ b/Src/F_Interfaces/Base/AMReX_box_fi.cpp
@@ -11,4 +11,21 @@ extern "C"
 	Box box {IntVect(lo), IntVect(hi), IntVect(nodal)};
 	AllPrint() << box << "\n";
     }
+ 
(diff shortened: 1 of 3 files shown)


WeiqunZhang:
The code is fine, but I have two comments.

(1) I think you can do it in Fortran without having the overhead of calling C++ functions.

(2) If this is used inside a loop over Boxes in a BoxArray, it will not scale.  You might want to consider using BoxArray::intersections functions.

PS.  For the fillpatch thing you mentioned, the data on intersecting Boxes might be owned by other MPI processes.

jared321:
For (1), are you suggesting that the Fortran interface reimplement directly the intersection code already in C++?

Thanks for the information in (2).  I am toying around with the idea of intersecting the region that is given to fillPhysicalBC with the domain.  No looping over a boxarray in this case.

WeiqunZhang:
For (1), yes.  It will be something similar to `amrex_box_intersects_box`.

PR #351, opened 2018-10-22
sensei data adaptor fixes

burlen:
* set data set origin, fixes issue with simulations where the lower left
  corner in physical space is not at 0,0,0
* fix a memory leak in AmrMeshDataAdaptor
* let simulations set control parameters explicilty
* validate control parameters during initialization


drummerdoc:
Is frequency in units of coarseTimeStep calls?

PR #369, opened 2018-11-16
Sensei init ghost zones

burlen:
I need to fill ghost cells as we convert from face and edge centering for visualization. this change adds functions that do that.

I added an option to force mesh origin to 0,0,0 to help deal with moving meshes in visit.

both of these came up when working with warpx


burlen:
Maybe I don't need to pass ngrow. can you advise on what is the best way to accomplish what I need?

on the analysis side we need cell centered data to have a layer of ghost cells. For ex. correctly computing a gradient, or iso surface on cell centered data requires a layer of ghost cells.

I wanted to initialize the ghost cells during conversion from edge and face data to cell centering rather than trying to do a bunch of exchanges after the fact.

It's coming up with warpx, which has face and edge centered fields. here is the code that uses these functions. https://github.com/burlen/WarpX/blob/sensei_init_ghost_cells/Source/WarpXIO.cpp#L491

Is this the best way to do it? Maybe this code belongs in Src/Extern/SENSEI?

drummerdoc:
Sorry for the delay, Burlen.  I am approving this pull request, but it would be good to separate out the demonstration of Sensei into its own Tutorial.  It should be very simple to copy everything over to a new folder there and edit out controls for the background problem and just make the demo about Sensei.  Doing this now so that the files remain compatible with impending AMReX edits.  Let me know if you need this comment translated :)

burlen:
@drummerdoc Thank you. I can do that. Would like to learn more about the impending changes and how I can keep our parts compatible

PR #374, opened 2018-11-27
add conduit blueprint support for particles

cyrush:
This PR adds support for wrapping AMReX ParticleContainers into a Conduit Mesh Blueprint tree, for consumption by Ascent. 

One question I have -- is there a recipe to get a unique id (across mpi tasks) for each particle tile?


atmyers:
If you use the particle iterator to access the tiles on each level, then you can get a unique id **per level** for each tile:

`
for (int lev = 0; lev < num_levels; ++lev) {
    for (MyParIter pti(pc, lev); pti.isValid(); ++pti) {
        int tile_id = pti.uniqueIndex();        
    }
}
`

This will already be unique across mpi tasks. If you also want something that is also unique across levels, you could do `int unique_id = tile_id + lev * num_levels` or something.

cyrush:
@atmyers thanks, i'll follow that strategy!

cyrush:
@atmyers 

Working on testing using `pti.uniqueIndex()` & it seems to only be unique within an mpi task, not across mpi tasks. Here is output echoed from a single level test (my variant of CellSortedParticles)

There are 8 total particle tiles in this example. 

`[Z]` is the MPI rank in the ouput. 

Run with 1 MPI Task:

```
[0] level_id 0
[0] tile_id 0
[0] tile_id 1
[0] tile_id 2
[0] tile_id 3
[0] tile_id 4
[0] tile_id 5
[0] tile_id 6
[0] tile_id 7
```

Run with 4 MPI Tasks:

```
[3] level_id 0
[3] tile_id 0
[3] tile_id 1
[0] level_id 0
[0] tile_id 0
[0] tile_id 1
[1] level_id 0
[1] tile_id 0
[1] tile_id 1
[2] level_id 0
[2] tile_id 0
[2] tile_id 1

Here is the basic loop:

```
    for (int lev = 0; lev < num_levels; ++lev) 
    {
        std::cout << "[" << rank  <<"] level_id " << lev << std::endl;
        
        for (MyParConstIter pti(pc, lev); pti.isValid(); ++pti)
        {
            int tile_id = pti.uniqueIndex();        
            std::cout << "[" << rank  <<"] tile_id " << tile_id << std::endl;
         // ...
        }
    }
```


I can generate my own unique ids, but it seems maybe the particle iterator might not reflect the intended semantics.

atmyers:
Apologies, I misread the code yesterday.

 `pti.index()` will give you a unique id for each *grid* across mpi domains, but it won't take tiling into account. If you want each tile to have its own id, you could do:

`int tile_id = pti.LocalTileIndex() + pti.numLocalTiles() * pti.index()` 

and then add in the offset for levels as well. If that works, we could wrap that into a function - `uniqueGlobalIndex` maybe?

WeiqunZhang:
What if `pti.numLocalTiles()` is different for each grid?

cyrush:
hmm -- I am not sure I understand the nested relationships for iterating over particle tiles. 

Maybe CellSortedParticles is too simple to tease out the index logic. Is there a better multi-level particle example that I should work with?

I know I can get right by counting + reducing -- I may go with that strategy.

atmyers:
@WeiqunZhang then I don't think the `uniqueIndex` function currently in there is right either...

The most general particle example in terms of the multi-level structure is probably the Advection Tutorial at `Tutorials/Amr/Advection_AmrLevel/`

Do you need a unique int for each tile or is a pair of ints sufficient? Does it have to be an int or can it be a long?

cyrush:
@atmyers thanks, I'll work with that example . 

I think some of the confusion on my part is that ParticleTiles aren't quite like Grid Tiles, at least as far as how the iterators work.

atmyers:
The difference is that the grid tiles are logical, whereas the particle tiles actually affect how the data is laid out in memory. When tiling is enabled, there is still only 1 multi-dimensional array per box for mesh data, but there are separate vectors of particle data for each tile. 

The iterators are consistent with each other, in that for a given BoxArray, DistributionMapping, and tile_size, the `mfi` and `pti` will execute the same number of times, once per tile, but the if the goal is to have a single integer id for each ParticleTile, then they will need to be numbered differently. We usually use a pair of ints to label the ParticleTiles, one unique one corresponding to the global box id, and a non-unique one specifying the tile number on that box.

cyrush:
@atmyers @WeiqunZhang, for the blueprint particle support : I added logic to generate unique ids for each particle tile. That was the last thing I wanted to work out for this PR.  Let me know if you have any questions or issues. Thanks!

atmyers:
Thanks!

cyrush:
@atmyers thanks for merging!

PR #375, opened 2018-11-29
Spline Implicit function for EB2

shashankNREL:
@asalmgren 

Added this additional functionality of constructing an implicit function for EB, based on splines and line segments. User provides input in terms of a few control points on a spline which is then used to construct a surface separating the fluid and solid zones. First version of this code developed by Ray Grout for EB1 formulation, that code was adapted to EB2 in this current version.

### DIFF
--- a/Src/EB/AMReX_distFcnElement.H
+++ b/Src/EB/AMReX_distFcnElement.H
@@ -1,0 +1,110 @@
+#ifndef AMREX_DISTFCNELEMENT_H_
+#define AMREX_DISTFCNELEMENT_H_
+
+#include <vector>
+#include "AMReX_RealVect.H"
+//#include "AMReX_BaseIF.H"
+#include "AMReX.H"
+#include "AMReX_Vector.H"
+
+class distFcnElement2d {
+ public:
+  // Constructor
+  distFcnElement2d() {}
+  ~distFcnElement2d() {}
+
+  virtual distFcnElement2d* newDistFcnElement2d() const;
+
+  virtual  amrex::Real cpdist(amrex::RealVect pt, amrex::RealVect& cp) const;
+  virtual  amrex::Real cpside(amrex::RealVect pt, amrex::RealVect& cp) const;
+  int solve_thomas(std::vector<amrex::Real> diagminus,
+                   std::vector<amrex::Real> diag,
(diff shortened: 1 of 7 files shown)


Src/EB/AMReX_distFcnElement.H
```diff
@@ -0,0 +1,109 @@
+#ifndef AMREX_DISTFCNELEMENT_H_
+#define AMREX_DISTFCNELEMENT_H_
+
+#include <vector>
+#include "AMReX_RealVect.H"
+#include "AMReX.H"
+#include "AMReX_Vector.H"
+
+class distFcnElement2d {
```
WeiqunZhang: Could you put these classes inside `namespace amrex`?

Src/EB/Make.package
```diff
@@ -42,16 +42,18 @@ CEXE_sources += AMReX_EB_levelset.cpp
 F90EXE_sources += AMReX_compute_normals.F90 AMReX_EB_geometry.F90
 F90EXE_sources += AMReX_EB_levelset_F.F90
 
+CEXE_headers += AMReX_distFcnElement.H
+CEXE_sources += AMReX_distFcnElement.cpp
```
WeiqunZhang: Please update CMakeLists.txt too.

Tutorials/EB/GeometryGeneration/main.cpp
```diff
@@ -8,59 +8,111 @@ using namespace amrex;
...
 
-        Geometry geom;
-        {
-            RealBox rb({-0.5,-0.5,-0.5}, {0.5,0.5,0.5}); // physical domain
-            Array<int,AMREX_SPACEDIM> is_periodic{false, false, false};
-            Geometry::Setup(&rb, 0, is_periodic.data());
-            Box domain(IntVect(0), IntVect(n_cell-1));
-            geom.define(domain);            
-        }
+    {
```
WeiqunZhang: Could you keep the old setup and add the new one as an additional example?  Maybe you can add an inputs ParmParse parameter for the geometry type.

rgrout:
I'm going to close this and replace it with a new pull request from an amrex branch where I can resolve a few conflicts by hand.

PR #378, opened 2018-12-02
make the extrema use a mask to find the true extrema in the domain

### LINKED ISSUE #291: fextrema should mask out coarse levels

zingale:
closes #291


maxpkatz:
This also fixes a bug in development where -v wasn't actually restricting to the requested variable -- thanks!

PR #388, opened 2018-12-19
Bunch of changes to support MLMG performance analysis and tuning

cychan-lbnl:


WeiqunZhang:
Thanks for the pull-request!  It's a lot of changes.  So I will probably review this after the new year.

WeiqunZhang:
I made a pull-request to the cy-perf branch.  I also have some additional comments.

`perf::Counters`: Could we have a function that can reset all the counters to zero?  We might also want to put that in `The_Finalize_function_Stack`.

 `CommCache comm_cache` is a variable in an anonymous namespace in `AMReX_MLLinOp.cpp`.  The constructor of `CommCache` calls `amrex::ExecOnFinalize(finalize_comm_cache)` defined in `AMReX.cpp` that uses the `The_Finalize_Function_Stack` variable in another anonymous namespace.  C++ rules on the initialization of variables with static storage are too complicated for me to be confident on this.  But I think the initialization order of these two variables is unspecified.   What do you think?

`Machine the_machine` is an anonymous namespace variable.  I think we had better clear the cache in `amrex::Finalize()`.  This is because a code is supposed to be able to repeatedly call `amrex::Initialize` and `amrex::Finalize` with different communicators passed to amrex.

WeiqunZhang:
Another comment.  Could we add an optional argument to `DistributionMapping::makeSFC (const BoxArray& ba)` so that its default behavior doesn't change?

cychan-lbnl:
Thanks for the comments!

Good point about the static initialization ordering problem.  I can add
Finalize functions to reset the perf counters and clear the neighborhood
cache and add a function (e.g. MLMG::Initiialize) called from
amrex::Initialize to register the finalize functions.

For the makeSFC behavior, I did have a version that read an input parameter
(e.g. mg.makesfc_use_box_vol) that switched the function's behavior.  I had
taken it out for the pull request because I thought it would be more
desirable under most circumstances versus the default where every box is
weighted equally.  What do you think?

WeiqunZhang:
Regarding `makeSFC`, it makes sense and I agree with you.  So maybe we can add an optional argument that defaults to the new behavior.

WeiqunZhang:
We cannot put MLMG and machine::initialize in AMReX.cpp because MLMG source codes are not always included.

cychan-lbnl:
Okay, what if we have MLMG initialize the first time an MLMG object is
created instead?  This can be done with a static variable.

Since machine::initialize does not depend on MLMG (only
Src/Base/AMReX_Machine.cpp/.H), is it okay to have it in AMReX.cpp?

WeiqunZhang:
Yes, machine is fine because it's in `Base/`.   Yes, you can make `MLLinOp`'s (not `MLMG`'s) constructor check a static boolean flag for initialization.

WeiqunZhang:
Cy, if you don't have any changes, I will merge it.

cychan-lbnl:
Okay, thanks!

PR #404, opened 2019-02-02
New heat tutorials

ylunalin:
Added OpenACC and OpenMP versions. CUDA version still works, but I noticed its answer seems to be different from CPU or OpenACC/OpenMP versions.


atmyers:
Hi Luna,

Do you have any idea why the CUDA version is giving different answers? One possible difference is that the CUDA version uses different streams, while the OpenACC / OpenMP versions are synchronous by default.

ylunalin:
There is a typo in the cuda source code, in calling update_phi, the fabs phiNew, phiOld are both from phi_old. After correcting that, there's still a large difference. We are not talking about round off level of difference, the relative difference is 18.7%.

ylunalin:
Hey Andrew, I spent some time tracking the error. It seems like writing the update_phi device on multiple lines makes a different. I wonder if it has something to do with the function being  inlined? I fixed it, and now the answers of all three agree.

WeiqunZhang:
Thanks for finding my bug.  It's not because of multiple lines.  It's
because in C++

    a = 1, + 2, +3;

is legitimate but is not 1+2+3.  Somehow I typed some extra commas.

PR #406, opened 2019-02-08
Add floating point awareness to the plot_per/check_per logic

maxpkatz:
The current logic can be slightly erroneous due to floating point issues. For example, if the timestep is 0.01 and the plot_per is 0.1, you would expect plotfiles at steps that are multiples of 10, but due to roundoff errors the current logic will sometimes plot in the 11th step instead of the 10th. By adding a small number to both sides of the comparison, we can guarantee plotting on exactly the timestep we intended.


WeiqunZhang:
There might be a flaw because `dt_buffer` changes due to the change of `dt_level`.  Suppose at step `n-1`, `num_per_old` and `num_per_new` are equal, but the pre-conversion double value of `num_per_new` is very very close to `num_per_new+1`.  Suppose `dt_level` at next step becomes bigger.  Then `num_per_old` and `num_per_new` at step `n` might be same again.  The problem is in the new way `num_per_new` at step `n-1` can be different from `num_per_old` at step `n`.  When that happens, we will miss a plotfile.

maxpkatz:
I agree that this could happen. I haven't been able to actually reproduce it in Castro with a targeted test, but I see how it could happen if I got the numbers just right. I will come up with something else that is immune to this.

maxpkatz:
I am able to make a related issue happen though: the plotfile can happen a step early, for essentially the same reason. I can reproduce this in the Castro KH problem by running with:
```
mpirun -np 1 ./Castro2d.gnu.MPI.OMP.ex inputs.2d amr.checkpoint_files_output=0 max_step=12 stop_time=1.2e3 castro.max_dt=100.0 amr.plot_per=1.e3 castro.init_shrink=0.99999999 castro.change_max=1.000000001 castro.output_at_completion=0 geometry.prob_hi=1.e6 1.e6
```
This should plot at step 11 only, but actually prints at step 10, when the current time is 999.9999945.

maxpkatz:
I have fixed the issue. What we do now is, first do the normal calculation of num_per_old and num_per_new. Then, if we detect that either the current time, or the last time (`cumtime - dt_level[0]`), is within machine epsilon of a plot_interval, then we increment num_per_new or num_per_old, respectively.

WeiqunZhang:
I like you current approach.  And I have two suggestions.

(1) `epsilon` in `numeric_limits` is epsilon to 1, and it can be quite different for other numbers.   For example, for `1.e-5`, its epsilon is about `1.7e-21`, and for `1.e5`, `1.5e-11`.  So I suggest we use

    std::numeric_limits<Real>::epsilon()*10.0*std::abs(cumtime)

instead.

(2) Maybe we can make `num_per_old` static so that we don't have to repeat the calculation.

maxpkatz:
(1) I agree and I have made that change.

(2) This is a pretty cheap calculation, so I'd rather just pay the small cost of repeating it and avoid the potential code errors involved from saving it (e.g. from dealing with restarts).

WeiqunZhang:
I think I was clear in my second comment.  I was simply thinking of

    if (check_per > 0.0)
    {
        static bool first = true;
        static int num_per_old = (cumtime-dt_level[0]) / check_per;
        ...
        if (first) {
            first = false;
            // may need to do num_per_old += 1;
        }
        ...
        num_per_old = num_per_new;
    }

I was worried about (t+dt)-dt may not be equal to t.  But even if it could happen, the chance is probably smaller we will be invaded by an alien species tomorrow.  So I am OK your current changes.

WeiqunZhang:
I meant I was not clear.

PR #415, opened 2019-02-28
Nodeghostcells

bsrunnels:
This pull request extends MLMG and MLCG to be able to treat operators that have arbitrary numbers of ghost cells/nodes. It adds a virtual `getNGrow` function to `MLLinOp` that can be overridden; by default it returns 0 for `MLNodeLinOp` and 1 for `MLCellLinOp`. 

The 2d and 3d versions of `amrex_mlmg_lin_nd_interp` were rewritten to account for interpolating between coarse/fine nodal fabs in which the nodal fab has only one layer of ghost nodes. 

Finally, the `final` keyword was removed for `solutionResidual` and `correctionResidual` allowing them to be overridden. This was necessary when writing the nodal solver with the 2-ghost-node reflux method, because the original `FillBoundary` does not work in this case.


WeiqunZhang:
I think the goal should be all existing solvers in the repo should not change behaviors.  This doesn't appear to be the case.

WeiqunZhang:
In `Tutorials/LinearSolvers/ABecLaplacian_C` (nodeghostcells branch), I do `make DEBUG=TRUE`, and then run with `mpiexec -n 4 ./main3d.gnu.DEBUG.MPI.ex inputs`, there are assertion failures.

You added virtual function `getNGrow`.  I think you should make it return 0 for all solvers in the official repo.  For all changes of ghost cell argument in function calls, you can change from 0 to the virtual `getNGrow()`, but not to `MultiFab::nGrow()`.

bsrunnels:
Thanks for checking that. I have updated both `MLNodeLinOp` and `MLCellLinOp` so that they return `getNGrow()=0`. I have also changed `MLMG` so that if `getNGrow() <= 1` then there is no change; the change only happens if `getNGrow() > 1`. 

As a side question: is there a systematic set of unit/regression tests that I can run before submitting future pull requests? The LinearSolver tutorials now pass, but it seems like there are some out-of-date tests and tutorials, and I'm not sure which ones are supposed to pass and which ones are not.

WeiqunZhang:
You can use the tests at https://ccse.lbl.gov/pub/RegressionTesting/AMReX/.

It seems to me whenever the virtual funciton `getNumGrow()` is called, it is always used as followed.

    const int nghost = linop.getNGrow() > 1 ? linop.getNGrow() : 0;

And all the LinOp classes in AMReX returns 0.  So why not hide the logic in your derived class?

drummerdoc:
Is it written somewhere what the point of the new function is?  Is this to make sure the user creates data a certain way, or?  Sorry if it's obvious...I'm being lazy here.

bsrunnels:
@WeiqunZhang : The problem with putting the logic in the derived class is that there are exceptions. E.g. in `MLMG::interpCorrection` the variable `ng=1` for cell-centered `LinOp`. So (as you know) in most cases the logic is based on whether the fab is node or cell centered - and there are instances where `ng=1` for cells and `ng=0` for nodes *and* vice versa. I like the idea of putting all the logic in the `getNGrow` of the derived class, but would it be possible to put the same logic in for node and cell based operators?

@drummerdoc : for my application I have a nodal operator that needs two layers of ghost nodes, and `getNGrow` communicates the number of ghost nodes/cells to `MLMG`.

WeiqunZhang:
I don't know how you can do it.   Maybe you have to add more virtual functions.  Maybe macro is actually the best way.  May we can make MLMG vritual, so you can derive.  I don't know.  But my goal is to make what's in the repo as clean as possible. 

Currently, the code does not compile in 2D.  https://ccse.lbl.gov/pub/RegressionTesting/IAMR/

bsrunnels:
I understand and completely agree re keeping the code clean :-) 

I'd rather not derive from `MLMG` since I'd need to re-write almost every function. What about adding a `getOperatorType()` to `MLLinOp` that returns an `OperatorType` enum with possible values, e.g. `CellBased,NodeBased,NodeWithGhostCells`, and then `nghost` is adjusted accordingly for each case?

Latest commit should fix the 2D compile error. I don't know about the comparison failures though. My changes should not have changed the existing behavior. Is there a systematic way to trigger regression tests (or to run them on my platform) so I can make sure they pass next time before submitting the pull request?

WeiqunZhang:
https://github.com/AMReX-Codes/regression_testing

We use this to do regression testing.

Src/LinearSolvers/MLMG/AMReX_MLMG.cpp
```diff
@@ -890,6 +900,9 @@ MLMG::actualBottomSolve ()
         else
         {
             MLCGSolver cg_solver(this, linop);
+
+	    if (nghost > 1) cg_solver.setNGhost(nghost);
```
WeiqunZhang: Why not just use `linop.getNGrow()` inside MLCGSolver?

Src/LinearSolvers/MLMG/AMReX_MLMG.cpp
```diff
@@ -633,14 +639,15 @@ MLMG::interpCorrection (int alev)
...
+                Box tmpbx = amrex::refine(cbx,2);
+		if (nghost > 1) tmpbx.grow(2);
                 tmpfab.resize(tmpbx,ncomp);
                 amrex_mlmg_lin_nd_interp(BL_TO_FORTRAN_BOX(cbx),
                                          BL_TO_FORTRAN_BOX(tmpbx),
                                          BL_TO_FORTRAN_ANYD(tmpfab),
                                          BL_TO_FORTRAN_ANYD(cfine[mfi]),
 					 &ncomp);
-                fine_cor[mfi].copy(tmpfab, fbx, 0, fbx, 0, ncomp);
+                fine_cor[mfi].copy(tmpfab, tmpbx, 0, tmpbx, 0, ncomp);
```
WeiqunZhang: Note that even when nghost is zero, tmpbx may be different from fbx.

Src/LinearSolvers/MLMG/AMReX_MLMG.cpp
bsrunnels: Because the original implementation does not use the same number of ghost cells everywhere. In the original implementation:
```
    const int nghost = sol.nGrow(), ncomp = sol.nComp(); //line 84
    ...
    MultiFab ph(ba, dm, ncomp, nghost, MFInfo(), factory); // line 90
    ...
    MultiFab sorig(ba, dm, ncomp, 0, MFInfo(), factory); // line 95```

I did not want to break anything by setting Multifabs `sorig,p,r,w,rh,v,t` to have `nghost=0`. However as long as it won't break anything I would be fine with implementing it that way.

File: Src/LinearSolvers/MLMG/AMReX_MLMG.cpp (hunk 2 above)
Comment (bsrunnels) [reply Src/LinearSolvers/MLMG/AMReX_MLMG.cpp]:
```suggestion
fine_cor[mfi].copy(tmpfab, tmpbx, 0, tmpbx, 0, ncomp);
```
This is what we currently have, so I may not understand the suggestion. Using `fbx` instead of `tmpbox` definitely breaks the code; the correction must be copied to the ghost nodes. What about
```suggestion
Box tmpbx2 = fbx;
if (nghost>1) tmpbx2.grow(nghost);
...
fine_cor[mfi].copy(tmpfab, tmpbx2, 0, tmpbx2, 0, ncomp);
```

WeiqunZhang: That works.

WeiqunZhang:
I like your suggestion of enum type.

bsrunnels:
Closing this PR - see PR for `nodeghostcells-pr` instead.

PR #420, opened 2019-03-07
Catch unhandled RealDescriptor logic in VisMf::Write

drummerdoc:
Not all RealDescriptor cases are caught in defining a pointer to one.  As a result, a nullptr is dereferenced later causing a segfault.  Can see problem if running with fab.format=ASCII, for example.

I think I still to be able to set Fab formats to ASCII at some point, but this way is currently no longer supported, so I added this Abort to say so.


WeiqunZhang:
Looks good to me.

PR #422, opened 2019-03-12
Fix index to make it work in all dimensions.

maikel:
This fixes an issue I had when I tried using EB in 2 dimensions.

### DIFF
--- a/Src/EB/AMReX_EBDataCollection.cpp
+++ b/Src/EB/AMReX_EBDataCollection.cpp
@@ -37,7 +37,8 @@ EBDataCollection::EBDataCollection (const EB2::Level& a_level,
 
     if (m_support == EBSupport::full)
     {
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
This is not a bug.  `Vector<int> a_ngrow` happens to have size of 3, but it is not for three dimensions.  The three integers are for `EBSupport::basic`, `volume` and `full`, respectively.

maikel:
Ok. I changed that because I got an out of bounds error in my code (where I use the full support). I will reinvestigate.

WeiqunZhang:
In that case, you need to pass a Vector of size 3.  For example, https://github.com/AMReX-Codes/amrex/blob/master/Src/Amr/AMReX_AmrLevel.cpp#L116

maikel:
I understand.. it worked. Thank you!

PR #427, opened 2019-03-19
Added plot_log_per

dpgrote:
This adds a capability of making plot files that are spaced logarithmically in time. For example, with plot_log_per = 0.2, five plot files will be mode over each factor of 10 increase in time.


WeiqunZhang:
I changed the merge base to development branch.  There is now a conflict.  Could you fix it?

PR #428, opened 2019-03-20
Random number generator on GPU using CUDA - RJ

RevathiJambunathan:
The Cuda_Allocators.H file required device_malloc_allocator to be included as header. 
I have added curand_kernel.H in Vector.H file. 

The main modifications are made to the AMReX_Utility.H and AMReX_Utility.cpp files. 
Added functions to initialize, resize, and deallocate Seed array, such that, each CUDA thread has a separate unique seed. Also included a function to see if the size of seed array is sufficient for random number array. 
Finally, modified the Random() command to include the device function. 
The changes here only support uniform random generation, from a sequence obtained from the in-built MRG32k3a generator.


atmyers:
It looks like there are some conflicts - could you merge development into your branch and fix them, then push the result to your fork? I can help you do this if you want.

atmyers:
Another thing is that some of these things - the Backtrace file, the profiling data, and the output of the run, don't need to be checked in to the repo. Could you change it so that only the extra source files you included are added?

atmyers:
Also, it looks like the changes to `AMReX_Utility.cpp` don't compile when `USE_CUDA=FALSE`. Could you fix this? You could test by building one of the test problems, e.g. Tests/Particles/AssignDensity

RevathiJambunathan:
Hi Andrew,

Sorry about that. I did not check for compilation with USE_CUDA=FALSE.
Will fix it and remove the other unnecessary files.

Thank you.
Revathi

RevathiJambunathan:
Hi Andrew,

I have fixed the issue with AMReX_Utility.cpp such that it runs with
USE_CUDA=FALSE.
Tested files in Tests/Particles/AssignDensity, and also by calling the
Random() function without CUDA flag.

Please let me know if something is still not quite right.

Thank you
Reva

WeiqunZhang:
Looks like some codes that were removed in development are added back by this pull-request.  For example

    int HashDistributionMap(const DistributionMapping & dm, int hashSize);

Could you fix those?

atmyers:
Could you also update the code in the test to call the functions from AMReX_Utility.cpp, so we'll have an example of that stuff in use?

Thanks,
Andrew

Src/Base/AMReX_Vector.H
```diff
@@ -16,6 +16,7 @@
 #include <thrust/iterator/zip_iterator.h>
 #include <thrust/device_ptr.h>
 #include <thrust/tuple.h>
+#include <curand_kernel.h>
```
atmyers: Does this need to be here? I don't see anything in AMReX_Vector.H that uses this. Maybe it should go in AMReX_Utility.H instead?

RevathiJambunathan:
the curand_kernel.h need not be in AMReX_Vector.H.
Weiqun also suggested moving it somewhere else, probably in
CudaAllocators.H or CudaUtility.H where cuda.h is included.

I am making some modifications based on Weiqun's suggestions and also
adding the other test file that call the Random() function added.

Will send an update once I push these changes.

Thank you.

Src/Base/AMReX_Utility.cpp
```diff
@@ -544,12 +535,11 @@ amrex::ResizeRandomSeed (int N)
...
+  int Nbuffer = N * 2;
 
-  dev_RandStates_Seed.resize(N);
-  curandState *d_RS_Seed = dev_RandStates_Seed.dataPtr();
-  cudaMemcpyToSymbol(glo_RandStates,&d_RS_Seed,sizeof(curandState *));
+  dev_RandStates_Seed.resize(Nbuffer);
+  curandState_t *d_RS_Seed = dev_RandStates_Seed.dataPtr();
+  cudaMemcpyToSymbol(glo_RandStates,&d_RS_Seed,sizeof(curandState_t *));
  
   AMREX_PARALLEL_FOR_1D (N, idx,
```
WeiqunZhang: Nbuffer or N?

RevathiJambunathan:
Wow! Thank you for catching that. While modifying I made a mistake!!!!

Will fix and update.

Tests/GPU/RandomNumberGeneration/main.H
```diff
@@ -0,0 +1,15 @@
+#ifndef MAIN_H
```
atmyers: I think this header file isn't needed any more.

RevathiJambunathan: Should I include AMReX.H and the other headers in the main,cpp file then?

atmyers: Yes, I would just include them directly.

RevathiJambunathan: Will do.

RevathiJambunathan: modified the main.cpp file and deleted main.H

RevathiJambunathan:
Thank you Andrew and Weiqun.

RevathiJambunathan:
Hi Andrew and Weiqun, 

Based on the suggestion given to me this morning, I have made a few commits to the branch - random_xorwow_RJ.

atmyers:
Hi Revathi, 

Thanks! Since this PR has been merged, you'll need to open another one for your new changes. You should sync your development branch with the upstream one and merge development into your random_xorwow_RJ branch first.

RevathiJambunathan:
Will do. Thank you Andrew

RevathiJambunathan:
Thank you.
I have submitted a new pull request : Random xorwow -- modified seed with
mpi rank #432

Please let me know if I missed out anything.

Reva

PR #436, opened 2019-03-27
add SUNDIALS 4 support

balos1:
@mic84 This modifies your work to support SUNDIALS 3 in CMake to support SUNDIALS 4 instead.

I have also added optional support of SUNDIALS 4 via the GNU make build system as demonstrated by the example program `Tutorials/CVODE/SUNDIALS4/EX_ark_analytic`.


mic84:
Merged. Thanks for your contribution!

PR #440, opened 2019-04-03
Slice generation for diagnostics rj

RevathiJambunathan:
Test case for generating slice for diagnostics. 
The slice can be 1D, 2D, or 3D, with cell size >= domain cell size.

If cell size of slice < domain cell size, no operations are performed. 

Assumptions : nghost = 0, no of levels = 1;
dx, dy, dz of the slice can be different and the coarsening ratio is defined by an IntVect. 

The slice is generated in two stages : 

1. Generate slice based on user-defined domain with cell size = domain cell size and perform parallel copy. 
2. Coarsen cell size of the slice based on user-defined cell size for the slice and average data from refined to coarse slice using amrex functions. 
The second stage is performed only if the slice cell size > domain cell size in any dimension (x,y,z)


drummerdoc:
Can you define "cell" in this context?  Typically a cell is something that lives at an (I,j,k) index.  An IntVect would point to a cell, and a Box would define a logically rectangular set of cells (by specifying the lower and upper bounds).  But your use seems different.  Thanks!

RevathiJambunathan:
I think I did not do a great job of describing the code using the AMReX definitions. (Sorry about that :) )

Within the code, I first generate slice which is a Box. 
Then create multifab and access/modify cell (i,j,k) of the FArrayBox using IntVect (i,j,k).

In the description here, I have mentioned cell size, which is the dx, dy, dz of a cell (i,j,k) given by CellSize() within the geometry class assuming only one level of refinement.

The averaging of data is performed only if the dx/dy/dz of the cells within the slice > dx/dy/dz of the cells contained in domain.

drummerdoc:
Does the averaging assume that the final cell size in each dimension is an integer multiple of the original data's cell size - suitably interpreted using the centering of the original data?  These things are tricky to communicate... :)

RevathiJambunathan:
Yes, the averaging does assume that the final cell size must be an integer multiple of original data's cell size. In fact, I have only tested for cases where the final cell size = original data cell size * (multiple of 2) in each dimension. The multiple need not be the same in each dimension, i.e., the coarsening IntVect can be (1,2,4). But I have not tested with (1,3,6), and I am not sure how that would work :) Will test and update here.

drummerdoc:
Ahh, ok.  I believe I understand now.  Thanks for the clarification.

Tests/CreateSliceForDiagnostics/main.cpp
```diff
@@ -0,0 +1,371 @@
...
+    Vector<MultiFab*> mf_ptr = GetVecOfPtrs(mf);
+    Vector<MultiFab*> rho_mf_ptr = GetVecOfPtrs(rho_mf);
+    Vector<MultiFab*> Ex_mf_ptr = GetVecOfPtrs(Ex_mf);
+    Vector<MultiFab*> Ey_mf_ptr = GetVecOfPtrs(Ey_mf);
+    Vector<MultiFab*> Ez_mf_ptr = GetVecOfPtrs(Ez_mf);
+    Vector<MultiFab*> Bx_mf_ptr = GetVecOfPtrs(Bx_mf);
+    Vector<MultiFab*> By_mf_ptr = GetVecOfPtrs(By_mf);
+    Vector<MultiFab*> Bz_mf_ptr = GetVecOfPtrs(Bz_mf);
+
+    InitializeVariables( mf_ptr, geom );
```
atmyers: Minor style note: you can actually do `InitializeVariables(GetVecOfPtrs(mf), geom);` and likewise for the create slice function, and avoid having to create the `mf_ptr` variable beforehand.

Tests/CreateSliceForDiagnostics/main.cpp
```diff
@@ -0,0 +1,371 @@
...
+void InitializeVariables(Vector<MultiFab*> mf, Vector<Geometry> geom)
+{
+    int nlevs = geom.size();
+    const Real *dom_dx = geom[0].CellSize();
+
+    for (int lev = 0; lev < nlevs; ++lev) {
+        for (MFIter mfi(*mf[lev]); mfi.isValid(); ++mfi){
+            const Box& bx = mfi.validbox();
+            FArrayBox& fbox_Dst = (*mf[lev])[mfi];
+            for (IntVect cell=bx.smallEnd(); cell<=bx.bigEnd(); bx.next(cell))
```
atmyers: It doesn't really matter here, but this style of looping over the IntVects in a Box has a fair bit of overhead. In general, for looping over the cells in a FAB and doing stuff in C++, it's better to use the style in the `AMReX_MultiFabUtil_3D.C`, except using the `array` function instead of `view`.

Tests/CreateSliceForDiagnostics/main.cpp
RevathiJambunathan: Thank you. I have made this change.

Tests/CreateSliceForDiagnostics/main.cpp
```diff
@@ -0,0 +1,380 @@
...
+    const Real *dom_dx = geom[0].CellSize();
+    for (int lev = 0; lev < nlevs; ++lev) {
+        for (MFIter mfi(*mf[lev]); mfi.isValid(); ++mfi) {
+
+            MultiFab& mfSrc = *mf[lev];
+            auto const &mf_arr = mfSrc.array(mfi);
+
+            const Box& bx = mfi.validbox();
+            
+            for (int k = bx.smallEnd(2); k < bx.bigEnd(2); ++k ) {
```
WeiqunZhang: `k < bx.bigEnd(2)` should be `k <= bx.bigEnd(2)`. Same for j and i.

RevathiJambunathan:
Thank you Weiqun. Even Andrew pointed it out to me yesterday. 
I have fixed it, but, did not add+commit+push that change yet, since I started working on adding a few more things, such as, ensuring that the values are interpolated if the location of the slice is between the points where the data is available.  
That is, if the domain ranges from [0,1.0] in all three directions and we have a cell-centered multifab, with 10 cells in the three directions. And if a 2D slice is required at point 0.4, which is between cell-centered data at 0.35 and 0.45, my goal is to average the cell-centered data at 0.35 and 0.45 for the slice, instead of just picking the nearest data point.

PR #442, opened 2019-04-04
Fix fcompare3D for the Intel compiler

jrood-nrel:
The Intel compiler fails when trying to compile this file directly (without `cpp`) due to the length of some lines being too long. This should help if `fcompare` is ever added as a target to CMake in AMReX. `fcompare` 1D and 2D compile fine without `cpp`.


WeiqunZhang:
Will merge this.  Thanks!  But are you using fcompare from `Tools/Postprocessing/C_Src/`.  The fcompare there has been removed in development branch.  There is fcompare in `Tools/Plotfile`.  You compile with `DIM=3` and the executable will work for 1d, 2d and 3d plotfiles.

jrood-nrel:
Ok great. Thanks for the information. A dimension agnostic fcompare would be great. I will switch to using `Tools/Plotfile`.

PR #446, opened 2019-04-09
Add missing include for Conduit/Make.package and missing linker optio…

ghweber:
…ns for Conduit/Ascent

### DIFF
--- a/Tools/GNUMake/Make.defs
+++ b/Tools/GNUMake/Make.defs
@@ -733,7 +733,6 @@ endif
 ifeq ($(USE_CONDUIT),TRUE)
   $(info Loading $(AMREX_HOME)/Tools/GNUMake/packages/Make.conduit...)
   include        $(AMREX_HOME)/Tools/GNUMake/packages/Make.conduit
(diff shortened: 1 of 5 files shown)


Tools/GNUMake/Make.defs
```diff
@@ -733,6 +733,7 @@ endif
 ifeq ($(USE_CONDUIT),TRUE)
   $(info Loading $(AMREX_HOME)/Tools/GNUMake/packages/Make.conduit...)
   include        $(AMREX_HOME)/Tools/GNUMake/packages/Make.conduit
+  include        $(AMREX_HOME)/Src/Extern/Conduit/Make.package
```
WeiqunZhang: For consistence, maybe we should include this in the main make file like all other `Make.package` files.

ghweber: Sorry, hadn't seen that. Just committed moving it from Make.defs to Make.conduit.

ghweber: Is the updated version ok? Apart from two commits that I seemed to have accidentally picked up, it only changes package specific files Make.conduit and Make.ascent. I can rebase to avoid picking up the other two commits.

PR #447, opened 2019-04-10
Use ccache if present to speedup build

mwm126:


mwm126:
No need to change build files; we can pass -DCMAKE_CXX_COMPILER_LAUNCHER=ccache to cmake instead

PR #451, opened 2019-04-12
Use "  sed -i''  " for BSD sed on Mac

mwm126:


maxpkatz:
Changed base to development

PR #457, opened 2019-04-19
Remove dependency on sed

mwm126:


WeiqunZhang:
The sed issue has been addressed recently.

PR #471, opened 2019-04-29
bug fixes for particles when BL_FLOAT is used.

atmyers:


GumpXiaoli:
In addition, i can't understand the intent of `PD_fixdenormals`. I am not sure whether there is a need to change `NativeRealDescriptor` to `Native64RealDesciptor` in `PD_fixdenormals` of `convertToNativeDoubleFormat`.

atmyers:
I believe it should indeed be `Native64RealDesciptor` for `convertToNativeDoubleFormat`, and `Native32RealDescriptor` for `convertToNativeFloatFormat`. That function is operating on the bytes of native 'double's and 'float's, not `Real`s, so we want them use the proper descriptor regardless of what `Real` is.

atmyers:
Thanks for reporting this - I'll hold off on merging until after the release tomorrow, to be safe. After that, it will be merged into development.

Issue #468, opened 2019-04-27
Using FillPatchTwoLevels on MultiFabs without EB when EB are enabled


maikel:
Hey,

In my computations, I try to have multiple domains, where one domain makes use of EB and the other domains don't. Each simulation currently works in isolation, but when I want to do an EB-less computation linked against an EB-enabled AMReX, I have to initialize the EB database prior to using FillPatchTwoLevels, with `AllRegularIF` for example. As an example: calling `FillPatchTwoLevels` creates an `FPInfo` object which again queries the EB database to make an EB FAB factory...

Here is a snippet from the constructor of `FPInfo`:
https://github.com/AMReX-Codes/amrex/blob/7252f34ccf209827d430c679330429a84588f14d/Src/Base/AMReX_FabArrayBase.cpp#L1114-L1121

* Now I wonder what implications does it have -- if any -- if I have configured the EB database and the *top* index space does not map the EB geometry of the `MultiFab` which I want to fill (which should be `AllRegularIF`). Shouldn't it take the Factory of the MultiFab which are used as Parameters to FillPatchTwoLevel? 

* Can the interface somehow be changed to tell `FPInfo` which fab factory type to create?

* I do not really understand FPInfo well enough yet. What is its purpose? I tried to search a bit, but couldn't find an explanation and I haven't dug too much into its source code yet.

Thank you,
Maikel Nadolski

WeiqunZhang:
I don't exactly know what you are trying to do and what's the right thing to do in your case.  I will just try to explain what `FillPatchTwoLevels` does.

[...]

PR #472, opened 2019-04-29
EB: Add an additional overload for FillPatchTwoLevels.

maikel:
* This adds an EB2::IndexSpace parameter to FillPatchTwoLevels.
   That way one can specify the exact EB2::IndexSpace for the
   interpolation region.
 * Add a defaulted FabFactory parameter to the constructor of FPInfo.

This is the first attempt for issue #468. Any thoughts and feedback are welcome

### DIFF
--- a/Src/AmrCore/AMReX_FillPatchUtil.H
+++ b/Src/AmrCore/AMReX_FillPatchUtil.H
@@ -7,6 +7,10 @@
 #include <AMReX_Interpolater.H>
 #include <AMReX_Array.H>
 
+#ifdef AMREX_USE_EB
+#include <AMReX_EB2.H>
+#endif
+
 namespace amrex
 {
     class InterpHook {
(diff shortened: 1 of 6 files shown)


maikel:
This somehow compiled on my machine, but should not. It will need some fixes, which I still need to push. Sorry.

WeiqunZhang:
You need to pass the EB index space into FPinfo ctor.  Each factory is associated with a specific BoxArray.  At the time FPinfo is constructed, we don't know the BoxArray yet.

maikel:
> You need to pass the EB index space into FPinfo ctor. Each factory is associated with a specific BoxArray. At the time FPinfo is constructed, we don't know the BoxArray yet.

Absolutely right. I hope I have done so now successfully.

WeiqunZhang:
I will merge this, but will make some changes to remove some duplicated code.

maikel:
> I will merge this, but will make some changes to remove some duplicated code.

Thank you!

PR #474, opened 2019-05-03
[CMake] Require only C bindings to MPI

maikel:
AFAIK, `find_package(MPI REQUIRED CXX)` will fail if it can not find the (deprecated) C++ bindings for MPI. Since AMReX only uses the C interface of MPI a `find_package(MPI REQUIRED C)` call will be enough. **It has the drawback that consumer projects will need to enable the C language in their CMakeLists.**


mic84:
Thank you for reporting this. I merged your pull request.

PR #476, opened 2019-05-08
[WIP] Adding CUDA graph support to FillBoundary

kngott:
First draft of CUDA graph support from CUDA 10 into FillBoundary.
WIP Pull Request for exploration and testing.
-------------------------------------------------
Features:
- Multiple CUDA graph tests are available in Tests/CudaGraphs/GraphBoundary.
- Multiple changes to CUDA Device and other objects to make easier integration.
- Gpu::inGraphRegion static variable added to allow selective graphing. (Also great for testing.)
- Templated AMReX:CudaGraph object to simplify control and implementation.
- FillBoundary integrated CUDA graphs into FB cached object to minimize construction and instantiation costs. copy_to_buffer, local_copy and copy_from_buffer graphs created.

-- Tests/CudaGraphs/GraphBoundary is the most up-to-date and usable direct test on FillBoundary.
-- Also tested on CUDA-based HeatEquation and Tutorials/LinearSolvers/ABecLaplacian_C/.
-- Has shown consistent speedup throughout testing, although thorough scaling tests required.
        (Needs a machine with CUDA 10 bigger than 16 nodes!)

My Current List:
* Opt-out feature for MultiFabs that won't use the graph enough times to see a benefit.
      (Also, store objects in FB cache longer to allow multiple temporaries to use same graph if no current longer-life MultiFab match exists?)
* Tests/CudaGraphs/GraphBoundary segFaults in cudaInstantiateGraph for n_cell = 256^3 & max_grid_size = 8. Why?
* (DONE) Needs additional testing for CUDA 10 proofing. (Local Machines and Summit still run?)
* Clean up CopyMemory object (don't keep number-of-launches copies of scomp & ncomp).
* Possible test to see if same fab being worked on repetitively? Can eliminate remove graph data setup and device copy of pointers before each graph launch. (Cache checks + a few FAB pointer locations sufficient?)
* Scale testing to make sure graphs give consistently better results for all cases of interest.
* Can counting launches be done cleaner/more efficiently? And can the counts be reused elsewhere?

Additional optimizations:
* Test more general graph implementation ideas, then pick a "best" graph creation strategy and remove alternatives. (Ideas: Oversized graph with NULL launches, given graphs should eliminate most launch overhead, or very general graph of a few iterations that is setup and executed many times on any MultiFab.)
* Fused launches: histogram sort launches based on number of threads (or items) and perform multiple identically-sized copies in the same launch?
* Try on ParallelCopy()?


Src/Base/AMReX_CudaAsyncFab.cpp
```diff
@@ -57,10 +59,13 @@ AsyncFab::clear ()
...
-            AMREX_GPU_SAFE_CALL(cudaStreamAddCallback(Device::cudaStream(),
-                                                      amrex_devicefab_delete,
-                                                      m_impl, 0));
+// CUDA 10
+            cudaHostFn_t clear = amrex_devicefab_delete;
+            AMREX_GPU_SAFE_CALL(cudaLaunchHostFunc(Device::cudaStream(), clear, m_impl));
+// CUDA 9
+//            AMREX_GPU_SAFE_CALL(cudaStreamAddCallback(Device::cudaStream(),
+//                                                      amrex_devicefab_delete,
+//                                                      m_impl, 0));
```
WeiqunZhang: Need to make sure it works for CUDA 9.

kngott: Looks like it's fixed now. Tried the LinearSolver tutorial on with CUDA 9 using pgi and gnu. Both run as expected.

PR #488, opened 2019-05-30
Init from random restart gpu friendly

jmsexton03:
The first Redistribute after reading from a binary file in Nyx is on the CPU, so using the HostVector strategy makes this faster. 

Comparing`pdxi == dxi` gave a compiler error in Nyx since GpuArray doesn't have an `operator==`


atmyers:
I'd like to hold off on merging this until after the release - I'm concerned that these changes may slow down the Redistribute in CPU-only mode, so I'd like to do some testing. 

I'll go ahead and fix the `dx == particle_dx` issue separately.

jmsexton03:
I will submit a new pull request that only uses the HostVector when AMREX_USE_CUDA=TRUE

PR #491, opened 2019-05-31
conduit blueprint integration updates

cyrush:
Two small conduit blueprint integration changes: 

* select proper floating point type for particle meshes using BL_SINGLE_PRECISION_PARTICLES
* allow MPI tasks without mesh data


cyrush:
@WeiqunZhang thanks!

PR #511, opened 2019-07-09
Added function for checking RealBox equality

KiranEiden:
Added the AlmostEqual function to the amrex namespace, which checks
that the domains represented by two real boxes are the same within a
tolerance.

### DIFF
--- a/Src/Base/AMReX_RealBox.cpp
+++ b/Src/Base/AMReX_RealBox.cpp
@@ -78,4 +78,17 @@ operator >> (std::istream &is, RealBox& b)
     return is;
 }
 
+bool AlmostEqual (const RealBox& box1,
+                  const RealBox& box2,
+                  Real eps /* = std::numeric_limits<Real>::round_error() */) noexcept
+{
+    bool almostEqual = true;
+    for(int i = 0; i < AMREX_SPACEDIM && almostEqual; ++i)
+    {
+        almostEqual &= abs(box1.lo(i) - box2.lo(i)) < eps;
+        almostEqual &= abs(box1.hi(i) - box2.hi(i)) < eps;
+    }
+    return almostEqual;
+}
+
 }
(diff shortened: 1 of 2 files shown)


Src/Base/AMReX_RealBox.cpp
```diff
@@ -78,4 +78,17 @@ operator >> (std::istream &is, RealBox& b)
...
 
+bool AlmostEqual (const RealBox& box1,
+                  const RealBox& box2,
+                  Real eps /* = std::numeric_limits<Real>::round_error() */) noexcept
+{
+    bool almostEqual = true;
+    for(int i = 0; i < AMREX_SPACEDIM && almostEqual; ++i)
+    {
+        almostEqual &= abs(box1.lo(i) - box2.lo(i)) < eps;
+        almostEqual &= abs(box1.hi(i) - box2.hi(i)) < eps;
```
WeiqunZhang: I don't think `&=` should be used.  https://stackoverflow.com/questions/23107162/do-the-and-operators-for-bool-short-circuit

Another problem is `std::numeric_limits<double>::round_error()` returns something like 0.5, not a small number.
https://en.cppreference.com/w/cpp/types/numeric_limits/round_error

WeiqunZhang: Also, `std::abs` should be used.  `abs` might be in global namespace, but it's not guaranteed.

KiranEiden: Okay, I'll fix those issues.

KiranEiden: Is the issue with &= relating to performance? I think that `a &= b` and `a = a && b` will always yield the same result provided that both a and b are booleans, and short circuiting won't impact the return value here.

WeiqunZhang: I don't think the standard specifies how `bool` is implemented.  So `a & b` may not always be correct.

WeiqunZhang: https://news.ycombinator.com/item?id=6309008

PR #512, opened 2019-07-09
Fix issues RealBox equality check

KiranEiden:
This fixes the issues relating to #511. Doing exact comparison by default looks preferable to attempting to use std::numeric_precision::epsilon(), given the documentation [here](https://en.cppreference.com/w/cpp/types/numeric_limits/epsilon).

### DIFF
--- a/Src/Base/AMReX_RealBox.cpp
+++ b/Src/Base/AMReX_RealBox.cpp
@@ -85,8 +85,8 @@ bool AlmostEqual (const RealBox& box1,
     bool almostEqual = true;
     for(int i = 0; i < AMREX_SPACEDIM && almostEqual; ++i)
     {
-        almostEqual = almostEqual && std::abs(box1.lo(i) - box2.lo(i)) < eps;
-        almostEqual = almostEqual && std::abs(box1.hi(i) - box2.hi(i)) < eps;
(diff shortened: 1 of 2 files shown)


Src/Base/AMReX_RealBox.cpp
```diff
@@ -80,13 +80,13 @@ operator >> (std::istream &is, RealBox& b)
...
-                  Real eps /* = std::numeric_limits<Real>::round_error() */) noexcept
+                  Real eps /* = 0.0 */) noexcept
 {
     bool almostEqual = true;
     for(int i = 0; i < AMREX_SPACEDIM && almostEqual; ++i)
     {
-        almostEqual &= abs(box1.lo(i) - box2.lo(i)) < eps;
-        almostEqual &= abs(box1.hi(i) - box2.hi(i)) < eps;
+        almostEqual = almostEqual && std::abs(box1.lo(i) - box2.lo(i)) < eps;
+        almostEqual = almostEqual && std::abs(box1.hi(i) - box2.hi(i)) < eps;
```
WeiqunZhang: Should it be `<=` instead of `<`?

KiranEiden: Sure, I'll change that.

KiranEiden: That's fixed, thanks for the help!

PR #513, opened 2019-07-10
Makefile tweak for OLCF

jaharris87:
Add CUDA version to PGI compile flags on summit.

### DIFF
--- a/Tools/GNUMake/sites/Make.olcf
+++ b/Tools/GNUMake/sites/Make.olcf
@@ -119,6 +119,7 @@ ifeq ($(which_computer),$(filter $(which_computer),summit))
   # Specify that we want to build for Volta
 
   CUDA_ARCH = 70
+  CUDA_VER = 10.1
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
It seems that you could extract the version number from the output of `nvcc --version` using `shell`, `tail` and `sed`.  That will make it more future proof.

maxpkatz:
I've been trying to avoid the CUDA specifier to the PGI compiler because we already override its choice for the CUDA toolkit by pointing to the system install (rather than whatever it comes with); this helps ensure compatibility with the nvcc part of the build.

jaharris87:
I noticed quite a bit of difference in optimization by specifying 10.1 for OpenACC, but it probably doesn't make a difference elsewhere.

jaharris87:
Given the feedback from @maxpkatz , I'll withdraw this PR since it seems more pervasive than I had originally thought.

PR #526, opened 2019-07-19
CMake install step: make header modification more cross-platform

### LINKED ISSUE #524: Setting AMREX_SPACEDIM via CMake
So we have a project using AMReX and have been having issues getting the correct `#define`s passed through to the compiler. We have been working around by setting `AMREX_SPACEDIM=3` via `target_compile_definitions`.

However as I understand it, the intended behaviour is that this should be done by CMake through `install(EXPORT...)`

https://github.com/AMReX-Codes/amrex/blob/a5da9eafa0bba73bb594239970bb912809b11097/Tools/CMake/AMReXInstallHelpers.cmake#L115

The definitions are added to the amrex target by `add_amrex_define`, which sets the target build definition to appear only if the `BUILD_INTERFACE` genex is true
https://github.com/AMReX-Codes/amrex/blob/a5da9eafa0bba73bb594239970bb912809b11097/Tools/CMake/AMReX_Utils.cmake#L356

My reading of the CMake docs however suggests that this should also appear in the `INSTALL_INTERFACE` genex. 

I used the somewhat clunky equivalent of `if(BUILD_INTERACE || INSTALL_INTERFACE)`
```
target_compile_definitions( amrex PUBLIC $<$<OR:$<BOOL:$<BUILD_INTERFACE:1>>,$<BOOL:$<INSTALL_INTERFACE:1>>>:${new_define}> )
```
but then I realised that this is *almost* equivalent to not having a conditional generator expression

Is there any reason why we can't just have `add_amrex_define` just make the definition a normal public definition?

This seems to work for my simple cases so I've forked and pushed this code:
https://github.com/rupertnash/amrex/tree/development

rupertnash:
This PR will close #524 

The install step of the CMake build uses `sed` to insert `#include <AMReX_Config.H>` at the top of the file. Since GNU sed and BSD sed have different behaviour this fails. Here I instead use pure CMake to do the same thing.

### DIFF
--- a/Tools/CMake/modify_installed_headers.cmake
+++ b/Tools/CMake/modify_installed_headers.cmake
@@ -1,4 +1,4 @@
-#
+r#
 # Add line #include<AMReX_Config.H> to all installed headers
 # 
 message(STATUS "Modifying installed headers")
@@ -6,7 +6,7 @@ file(GLOB _installed_headers "${CMAKE_INSTALL_PREFIX}/include/*.H")
 list(REMOVE_ITEM _installed_headers "${CMAKE_INSTALL_PREFIX}/include/AMReX_Config.H")
 
 foreach( _header IN LISTS _installed_headers)
-   # This command only affect the first line of each file. Therefore, it can be used countless
-   # times on the same file and it always gives the same result (i.e. it does not keep prepending)
-   execute_process(COMMAND bash "-c" "sed -i '1s/^/#include <AMReX_Config.H>\\n\\n/' ${_header}" )
+   file(READ ${_header} _contents)
+   string(PREPEND _contents "#include <AMReX_Config.H>\n")
+   file(WRITE ${_header} "${_contents}")
 endforeach()
--- a/Tools/CMake/modify_installed_headers.cmake
+++ b/Tools/CMake/modify_installed_headers.cmake
@@ -1,4 +1,4 @@
-r#
+#
 # Add line #include<AMReX_Config.H> to all installed headers
 # 
 message(STATUS "Modifying installed headers")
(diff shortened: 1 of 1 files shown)


Tools/CMake/modify_installed_headers.cmake
```diff
@@ -6,7 +6,27 @@ file(GLOB _installed_headers "${CMAKE_INSTALL_PREFIX}/include/*.H")
...
-   execute_process(COMMAND bash "-c" "sed -i '1s/^/#include <AMReX_Config.H>\\n\\n/' ${_header}" )
+  file(READ ${_header} _contents)
+  # Regex looks for:
+  # - Anything before include guard
+  # - #ifndef CPP_IDENTIFIER
+  # - #define CPP_IDENTIFIER
+  # - Rest of file
+  if (_contents MATCHES "^(.*)#ifndef +([A-Za-z0-9_]+)\n#define +([A-Za-z0-9_]+) *\n(.+)")
+    # We have found an include guard
+    if (${CMAKE_MATCH_2} STREQUAL ${CMAKE_MATCH_3})
```
WeiqunZhang: I like this check here.

Tools/CMake/modify_installed_headers.cmake
```diff
@@ -6,7 +6,27 @@ file(GLOB _installed_headers "${CMAKE_INSTALL_PREFIX}/include/*.H")
...
+    else()
+      message(FATAL_ERROR "Include guard is wrong up for header file ${_header}")
+    endif()
+
+    # Insert the #include <AMReX_Config.H> between include guard and the rest of the file
+    string(CONCAT "_contents" "${CMAKE_MATCH_1}" "#ifndef " "${CMAKE_MATCH_2}" "\n#define " "${CMAKE_MATCH_3}" "\n#include <AMReX_Config.H>\n" "${CMAKE_MATCH_4}")
+
+  else()
+    message(WARNING "No include guard in header ${_header}")
+    string(PREPEND _contents "#include <AMReX_Config.H>\n")
```
WeiqunZhang: The warning and the include is unnecessary.  We have header files that are only included by other headers.

rupertnash: I think there's only one `AMReX_FabArrayCommI.H`? Happy to remove the warning if you prefer of course

WeiqunZhang: There are a couple more in `Particles` and maybe other directories.  But I take my comment back.  It is probably better to add include guard to those files.

mic84:
Thank you for your contribution!

PR #527, opened 2019-07-19
bugfixes for conduit blueprint support

cyrush:
This PR provides to small fixes to the AMReX Conduit support:

* fetch the proper pointer when wrapping multiple fields into conduit
* switch to use file-per-domain output when writting blueprint files


cyrush:
@WeiqunZhang thanks!

PR #541, opened 2019-08-07
Allow to create shallow IndexSpace's 

maikel:
Currently, it is not possible to create "thin" IndexSpaces with multiple refinement levels because of a `coarsenable` restriction. These changes enable me to use it as in the example

```cpp
  const int n_cells = 128;
  const std::array<double, 2> xlower{0.0, 0.0};
  const std::array<double, 2> xupper{+1.0, +0.03};
  const std::array<int, 2> periodicity{0, 0};

  const amrex::RealBox xbox(xlower, xupper);
  const amrex::Box domain({0, 0}, {n_cells - 1, 0});

  const amrex::IntVect refine_ratio{2, 1};

  amrex::Geometry geom(domain, &xbox, -1, periodicity.data());
  geom.refine(refine_ratio);
  auto all_regular = amrex::EB2::makeShop(amrex::EB2::AllRegularIF());
  amrex::EB2::Build(all_regular, geom, refine_ratio, 1, 1);
```

The most relevant change is from

```cpp
bool coarsenable = m_geom.back().Domain().coarsenable(2,2);
```

to

```cpp
bool coarsenable = m_geom.back().Domain().coarsenable(refine_ratio);
```

as it might change its meaning. What do you think?

### DIFF
--- a/Src/EB/AMReX_EB2.H
+++ b/Src/EB/AMReX_EB2.H
@@ -58,6 +58,10 @@ class IndexSpaceImp
 {
 public:
 
+    IndexSpaceImp (const G& gshop, const Geometry& geom, const IntVect& refine_ratio,
+                   int required_coarsening_level, int max_coarsening_level,
+                   int ngrow);
+
     IndexSpaceImp (const G& gshop, const Geometry& geom,
                    int required_coarsening_level, int max_coarsening_level,
                    int ngrow);
@@ -78,7 +82,6 @@ public:
     using F = typename G::FunctionType;
 
 private:
-
     Vector<GShopLevel<G> > m_gslevel;
     Vector<Geometry> m_geom;
     Vector<Box> m_domain;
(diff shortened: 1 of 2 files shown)


maikel:
Do you have any opinion on this PR?

WeiqunZhang:
What I am worried about is this can be confusing for people who read the code later.  We build EB at the finest level.  Then we always coarsen it by a factor of 2.  The functions for coarsening are hardwired for coarsening ratio of 2 in each direction.  Now there is a new `Build` function that takes `IntVect` as refinement ratio.  It might work for your specific case, but it does not work in general for an arbitrary IntVect.

Yes, `bool coarsenable = m_geom.back().Domain().coarsenable(refine_ratio);` is not the same as `bool coarsenable = m_geom.back().Domain().coarsenable(2,2);` even when `refine_ratio=IntVect(2)`.  The second parameter 2 is an optional argument to `coarsenable`.  We set it to 2 so that we do not coarsen all the way down to `1x1x1` box.

maikel:
I can see your issues with this. Since I only use this for index spaces which contain no embedded boundaries at all I could also create a specific IndexSpace implementation (RegularIndexSpace?) for that particular case. Would this be a better approach?

PR #544, opened 2019-08-07
RNGs are now assigned a block instead of a thread + a test problem to test locking

ChrisDeGrendele:
This is safer for older gpus. Also I learned how to squash commits :)

### DIFF
--- a/Tests/GPU/Locking/main.cpp
+++ b/Tests/GPU/Locking/main.cpp
@@ -1,0 +1,119 @@
+#include <AMReX.H>
+#include <AMReX_MultiFab.H>
+#include <AMReX_Gpu.H>
+#include <AMReX_Utility.H>
+#include <AMReX_Array.H>
+#include <AMReX_CudaContainers.H>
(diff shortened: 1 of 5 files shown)


Tests/GPU/Locking/main.cpp
```diff
@@ -0,0 +1,119 @@
...
+    double *hzpos = hz.dataPtr();
+
+
+    AMREX_PARALLEL_FOR_1D (Ndraw, idx,
+	 {
+#ifdef AMREX_USE_CUDA
+	   
+	   addone(dxpos[idx]);
+	   addone(dypos[idx]);
+	   addone(dzpos[idx]);
```
atmyers: To me it looks like each thread will update it's own value here, regardless of whether the lock is working correctly. If you comment out the get / set state above, does the test still pass? It seems that what you should do is pass a pointer into the nums into the `addone` function, and then increment the value at index `i`. If the total is correct then, we'll know that the lock is preventing race conditions.

PR #547, opened 2019-08-12
Development

cgilet:


WeiqunZhang:
It looks good to me.  You can clean up the comments.  Also there is no reason to have `pflux`.  You can just use `fluxfab_tmp[idim].array()`.

PR #559, opened 2019-09-11
Make sure Vector bounds check catches negative indices.

cgilet:
in looking at a case of vector[-1] (which didn't throw an error for the debug build), I found 
	std::cout<<" i = "<<i<<"\n";
	std::cout<<" cast i = "<<(static_cast<long>(i))<<"\n";
gives 
        i = 18446744073709551615
        cast i = -1

### DIFF
--- a/Src/Base/AMReX_Vector.H
+++ b/Src/Base/AMReX_Vector.H
@@ -36,13 +36,13 @@ public:
 
     T& operator[] (std::size_t i) noexcept
     {
-	BL_ASSERT(static_cast<long>(i) < this->size() && static_cast<long>(i) >= 0);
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
Maybe the check could be `i < std::vector<T, Allocator>::size()`.

WeiqunZhang:
`std::size_t` is always positive.  So if it's less the size of the vector, we are good.  Note `vector::size` return `size_t` so no casting is needed for the comparison.

WeiqunZhang:
Yes, I think a single comparison is cleaner.

```
AMREX_ASSERT(i < std::vector<T, Allocator>::size());
return this->std::vector<T, Allocator>::operator[](i);
```

PR #560, opened 2019-09-11
Tools: update compiler flags for Cray compiler v9

bcfriesen:
In CCE 9, Cray overhauled their C/C++ compilers and moved to clang/LLVM, so the
flags for C/C++ compilers are totally different than in previous versions of
CCE. Confusingly, the Fortran compiler was *not* changed significantly in CCE
9, so its flags are the same as before. As a result, we need some more complex
logic to get the right flags for the right version of CCE.

### DIFF
--- a/Tools/GNUMake/comps/cray.mak
+++ b/Tools/GNUMake/comps/cray.mak
@@ -13,36 +13,60 @@ F90FLAGS =
 
 ########################################################################
 
-cray_version = $(shell $(CXX) -V 2>&1 | grep 'Version')
-
-COMP_VERSION = $(cray_version)
+# CRAY_CC_VERSION is defined by the 'cce' module. Starting with CCE 9, Cray
+# changed the C and C++ compilers to clang/LLVM based, so all of the
+# options/flags for those compilers changed. But the Cray Fortran compiler is
+# still based on CCE 8 and so it has the same options as before.
+COMP_VERSION = $(shell echo $(CRAY_CC_VERSION) | cut -f 1 -d .)
 
 ########################################################################
 
 ifeq ($(DEBUG),TRUE)
 
-  GENERIC_COMP_FLAGS += -K trap=fp
+  ifeq ($(COMP_VERSION),9)
+    CXXFLAGS += -g -O0
+    CFLAGS   += -g -O0
+    FFLAGS   += -g -O0 -e i -K trap=fp
+    F90FLAGS += -g -O0 -e i -K trap=fp
+  else
+    GENERIC_COMP_FLAGS += -K trap=fp
 
-  CXXFLAGS += -g -O0
-  CFLAGS   += -g -O0
-  FFLAGS   += -g -O0 -e i
-  F90FLAGS += -g -O0 -e i
+    CXXFLAGS += -g -O0
+    CFLAGS   += -g -O0
+    FFLAGS   += -g -O0 -e i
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
Interesting.  Presumably `-X` option is gone and `-Xlinker` will work with CCE 9.

bcfriesen:
> Interesting. Presumably `-X` option is gone and `-Xlinker` will work with CCE 9.

Yes, according to `man craycc`:

```
       -Xanalyzer <arg>
              Pass arg to the static analyzer.

       -Xassembler <arg>
              Pass arg to the assembler.

       -Xlinker <arg>
              Pass arg to the linker.

       -Xpreprocessor <arg>
              Pass arg to the preprocessor.
```

WeiqunZhang:
That's great.  Does it support llvm santizer?  Cori is still down.  So I cannot find this out myself.  But it seems that you have special access.

bcfriesen:
> 
> 
> That's great. Does it support llvm santizer?

Yes it does:

```
friesen@gert01:sanitizer> cat umr.cc
// https://clang.llvm.org/docs/MemorySanitizer.html

#include <stdio.h>

int main(int argc, char** argv) {
  int* a = new int[10];
  a[5] = 0;
  if (a[argc])
    printf("xx\n");
  return 0;
}
friesen@gert01:sanitizer> CC --version
Cray clang version 9.0.0 (44f957486d17952efd83575cc9709118c146a616) (based on LLVM 9.0.0svn)
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /opt/cray/pe/cce/9.0.0/cce-clang/x86_64/bin
friesen@gert01:sanitizer> CC -fsanitize=memory -fno-omit-frame-pointer -g -O2 umr.cc
friesen@gert01:sanitizer> ./a.out
==42137==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x2009ec5f in main /global/homes/f/friesen/tests/clang/sanitizer/umr.cc:8:7
    #1 0x7ffff4f2df89 in __libc_start_main (/lib64/libc.so.6+0x20f89)
    #2 0x200202b9 in _start /home/abuild/rpmbuild/BUILD/glibc-2.26/csu/../sysdeps/x86_64/start.S:120

SUMMARY: MemorySanitizer: use-of-uninitialized-value /global/homes/f/friesen/tests/clang/sanitizer/umr.cc:8:7 in main
Exiting
friesen@gert01:sanitizer>
```

> Cori is still down. So I cannot find this out myself. But it seems that you have special access.

Yes, Cori is on track to come back tomorrow. Right now most of the file systems are unmounted, so logging in is treacherous even for staff.

bcfriesen:
AFAICT the C/C++ compilers in CCE 9 are [mild extensions](http://pubs.cray.com/content/S-5212/9.0/cray-compiling-environment-cce-release-overview/cce-900-software-enhancements) of vanilla clang/clang++. So all flags for vanilla clang/clang++ should work with `cc` and `CC`. The flags listed explicitly in `man craycc` are the additional ones that Cray added, e.g., `-fsave-loopmark` (similar to `-h list=a` from CCE 8) and `-fsave-decompile`.

To run CCE9 on Cori you need to `module load cdt/19.06`. You can't just switch the `cce` module alone, because Cray had to rebuild the entire PE to support the new LLVM backend in CCE 9.

PR #563, opened 2019-09-12
some modifications for the Nodal poisson tutorial program

svater:
- modified plotting routine from ABecLaplacian_C to get proper output
- modified exact solution, so that it is also valid for 2d


WeiqunZhang:
In some codes, when we want to save nodal data directly without averaging them to cell centers, we use `VisMF::Write(const FabArray<FArrayBox>& fa, const std::string& name)`.  We can then use https://github.com/AMReX-Codes/Amrvis to view the data.

svater:
I have added the averaging procedure to accomodate for the problem that only cell values can be written. Thanks for the explanation. I will try `VisMF::Write` for my further debugging.

PR #577, opened 2019-09-26
C++ User Literal for amrex::Real (_rt)

ax3l:
This adds a user literal for `amrex::Real`. This is a [C++11 feature](https://en.cppreference.com/w/cpp/language/user_literal) and allows to type constants properly instead of relying on implicit conversation.

A typical fallacy is the following user code:
```C++
auto const mypi = 3.14;
Real const sphere_volume = 4 / 3 * mypi * pow(r,3);
```

The first term `4/3` makes the whole expression an `int` until the result is casted to `amrex::Real`. Urgh, `4/3 == 1`!

We can do better, with `amrex::Real(4) / amrex::Real(3)` but who wants to write such verbose code, right? Or `4./3.` which is always a `double` operation? No. So C++11 user literals to the rescue!
```C++
auto const mypi = 3.14_rt; // that's an amrex:Real, too!
Real const sphere_volume = 4_rt / 3_rt * mypi * pow(r,3);
```

Note: the naming `rt` is borrowed from the [Fortran equivalent](https://github.com/AMReX-Codes/amrex/search?q=rt+%3D%3E+amrex_real&unscoped_q=rt+%3D%3E+amrex_real) usage in AMReX.

### DIFF
--- a/Src/Base/AMReX_FArrayBox.cpp
+++ b/Src/Base/AMReX_FArrayBox.cpp
@@ -769,7 +769,7 @@ FABio_8bit::write (std::ostream&    os,
 {
     BL_ASSERT(comp >= 0 && num_comp >= 1 && (comp+num_comp) <= f.nComp());
 
-    const Real eps = Real(1.0e-8); // FIXME - whats a better value?
+    const Real eps = 1.0e-8_rt; // FIXME - whats a better value?
     const long siz = f.box().numPts();
 
     unsigned char *c = new unsigned char[siz];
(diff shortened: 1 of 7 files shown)


Src/Base/AMReX_FArrayBox.cpp
```diff
@@ -769,7 +769,7 @@ FABio_8bit::write (std::ostream&    os,
 {
     BL_ASSERT(comp >= 0 && num_comp >= 1 && (comp+num_comp) <= f.nComp());
 
-    const Real eps = Real(1.0e-8); // FIXME - whats a better value?
+    const Real eps = 1.0e-8_rt; // FIXME - whats a better value?
```
ax3l: Btw, a better value might be [`std::numeric_limits<Real>::epsilon`](https://en.cppreference.com/w/cpp/types/numeric_limits/epsilon), without checking what this function actually does.

PR #578, opened 2019-09-26
C++ User Literal for amrex::ParticleReal (_prt)

ax3l:
Same as the literal for `amrex::Real`. Follow-up to #577 

Throughout the code base, I could not find usages of `ParticleReal` literals yet.

### DIFF
--- a/Src/Base/AMReX_REAL.H
+++ b/Src/Base/AMReX_REAL.H
@@ -56,8 +56,8 @@ namespace amrex {
     using Real = amrex_real;
     using ParticleReal = amrex_particle_real;
 
-    /*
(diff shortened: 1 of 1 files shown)


Src/Base/AMReX_REAL.H
```diff
@@ -57,7 +57,7 @@ namespace amrex {
     using ParticleReal = amrex_particle_real;
 
     /*
-      C++ user literal ``_rt`` for short-hand notations
+      C++ user literals ``_rt`` &  ``p_rt`` for short-hand notations
```
WeiqunZhang: `p_rt` should read `_prt`.

ax3l: Dang!

PR #581, opened 2019-10-01
Tracer Particles: Fix Narrow Warnings

ax3l:
Mitigate some narrowing warnings in tracer particles when compiling in single precision:
```
Src/Particle/AMReX_TracerParticle_mod_K.H:86:27:
  warning: narrowing conversion of
  ‘(1.0e+0 - ((double)xint))’ from ‘double’ to
  ‘amrex::Real {aka float}’ inside { } [-Wnarrowing]
```

Follow-up to work in #577


atmyers:
Thanks!

PR #583, opened 2019-10-01
Single precision random

atmyers:


atmyers:
Currently, yes, but if/when WarpX needs that we can add support for it.

ax3l:
Sounds good, we can cross that bridge when we come to it.

PR #584, opened 2019-10-02
Fixes to the spline IF routines

shashankNREL:
This pull request continues the fix introduced by Weiqun  (@WeiqunZhang ) to the distFcnElement2D routine. A couple of fixes to the first derivative calculation in CalcD and solveThomas routine. 

Introduced a test case in the Tutorials/EB/GeometryGeneration. This test case builds a stepped lip piston bowl geometry. Also compared the code against the original python based implementation by Ray Grout (@rgrout).

### DIFF
--- a/Src/EB/AMReX_distFcnElement.cpp
+++ b/Src/EB/AMReX_distFcnElement.cpp
@@ -6,17 +6,11 @@
 
 namespace amrex {
 
-int distFcnElement2d::solve_thomas(std::vector<amrex::Real> ain,
-                                   std::vector<amrex::Real> bin,
-                                   std::vector<amrex::Real> cin,
-                                   std::vector<amrex::Real> din,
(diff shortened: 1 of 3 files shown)


Src/EB/AMReX_distFcnElement.cpp
```diff
@@ -10,7 +10,7 @@ int distFcnElement2d::solve_thomas(std::vector<amrex::Real> ain,
                                    std::vector<amrex::Real> bin,
                                    std::vector<amrex::Real> cin,
                                    std::vector<amrex::Real> din,
-                                   std::vector<amrex::Real>& x) {
+                                   std::vector<amrex::Real> &x) {
   std::vector<amrex::Real> a, b, c, d;
```
WeiqunZhang: `a`, `b`, `c`, and `d` are passed by value.  Thus there are no need to create local copies.  Furthermore, `a` and `c` are not modified, so they can be passed as `const&`.

PR #585, opened 2019-10-05
Reduce number of CUDA streams to 4

maxpkatz:
Empirical testing with Castro shows that 4 CUDA streams is slightly better than the current default of 16, and is also better than 2 and 8. With 4 streams we can achieve concurrency even when there are callbacks (this is generally not enough streams to expose the stream serialization bug in the NVIDIA driver), and my observation is that there are never more than 4 streams used in practice by AMReX for any of its core operations, so there is no harm in this limit.

### DIFF
--- a/Src/Base/AMReX_GpuDevice.H
+++ b/Src/Base/AMReX_GpuDevice.H
@@ -144,7 +144,7 @@ private:
     static int verbose;
 
 #ifdef AMREX_USE_GPU
-    static constexpr int max_gpu_streams = 16;
+    static constexpr int max_gpu_streams = 4;
 #else
     // Equivalent to "single dependent stream". Fits best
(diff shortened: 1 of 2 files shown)


drummerdoc:
When we can get to more testing with chemistry, it might well be that more streams could be better.  @weiqun do we have a way to set the number of streams for an individual loop where 4 may not be the best number?

maxpkatz:
I have not seen any real world use cases in general where more than 4 streams are all running at the same time and where it provides a performance benefit. The MFInfo flag does control the number of streams used, but it can never be more than max_gpu_streams because that is how many we actually create. So an alternate way to do this would be to create more streams but make the default for MFIter be 4 -- but again, I don't think this will ever help.

PR #590, opened 2019-10-10
Linker: Use RPath

ax3l:
Although AMReX does not come with much dependencies besides runtimes, the build system allows to add various libraries for downstream application dependencies.

In the case of WarpX, we link auxiliary libraries for I/O among others. When linking against shared libraries, these dependent libraries (think: `.so`|`.dylib`|`.dll` files) must be present and accessible at runtime again. Usually, one sets their paths in `export LD_LIBRARY_PATH` hints which is both cumbersome for users as well as bad for file-systems which will be heavily loaded with meta-data operations while searching for the corresponding dependencies, especially at scale.

Luckily, all modern Unix systems support RPath hints, which embed the location of a dependent shared library at link time into the generated binary.
  https://en.wikipedia.org/wiki/Rpath
We now enable this feature by default. The linker flag is implemented in a generic way that works with the most common compilers:
- clang :heavy_check_mark:
- gcc :heavy_check_mark:
- nvcc :heavy_check_mark:
- icc :heavy_check_mark:
- pgcc :heavy_check_mark:
- xlC :heavy_check_mark:
- crayc++ "classic" :heavy_multiplication_x:
- crayc++ 9.0+ "LLVM": :heavy_check_mark:

~~In case there are any regressions or one desires not to use RPaths, users can just set `USE_RPATH=FALSE` as `make` option.~~
Update: Downstream projects can set the `USE_RPATH=TRUE` option to enable this new feature. Feedback on any problems that might appear when setting this is very welcome and if we can make this a robust feature we would like to enable this by default in the future.

### DIFF
--- a/Docs/sphinx_documentation/source/BuildingAMReX.rst
+++ b/Docs/sphinx_documentation/source/BuildingAMReX.rst
@@ -43,6 +43,8 @@ list of important variables.
    +------------+-------------------------------------+-------------+
    | USE_OMP    | TRUE or FALSE                       | FALSE       |
    +------------+-------------------------------------+-------------+
+   | USE_RPATH  | TRUE or FALSE                       | TRUE        |
+   +------------+-------------------------------------+-------------+
 
 .. raw:: latex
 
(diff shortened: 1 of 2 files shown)


ax3l:
Follow-up fix in 32af3a7fcc0504c2a6bdd8c8ffe00eeeef55f4b4 0868f803c6899b654a1d62c8f56609fd72f90344

Issue #579, opened 2019-09-27
how to implement hoextrap for cells?


DUGKS:
I found that the default method for hoextrap boundary is designed for face values.

I want to change that function to `amrex_hoextraptocc (q, qlo, qhi, domlo, domhi, dx, xlo)` so that ghost cell values can be extrapolated using high order scheme. 

But a simple substitution of 
```
amrex_hoextraptocc(BL_TO_FORTRAN_N_ANYD(dest,dcomp), &numcomp,
                    BL_TO_FORTRAN_BOX(domain),
                    dx, xlo);
```
for
```
amrex_fab_filcc(BL_TO_FORTRAN_N_ANYD(dest,dcomp), &numcomp,
                       BL_TO_FORTRAN_BOX(domain),
                       dx, xlo, bcr[bcomp].vect());
```
disabled other boundary types such as foextrap, reflect_even, etc..

what's the best method to implement a hoextrap BC?

Thanks a lot!

WeiqunZhang:
I think you can add a new bc type say HOEXTRAPCC to `AMReX_BC_TYPES.H` and related Fortran files `AMReX_bc_types_mod.F90` and `AMReX_bc_types.fi`.  Then add a new option to `amrex_filccn` in `AMReX_filcc_mod.F90`.  If you like to make this feature available for others, you are welcome to submit a pull-request.

DUGKS:
I finally realized that a linear extrapolation is more preferable. So a new bc type called LIEXTRAP is added to AMReX_BC_TYPES.H as well as AMReX_bc_types_mod.F90.

In AMReX_BC_TYPES.H, the changes are
```
enum mathematicalBndryTypes : int {
    ...
    ...
    ...
    liextrap     =  5
};
#ifndef AMREX_XSDK
    ...
    ...
    ...
#define LIEXTRAP 5
    ...
    ...
    ...
#endif
```
In AMReX_bc_types_mod.F90, the newly added term is 
```
integer, parameter, public :: amrex_bc_liextrap     =  5
```

The newly added block in subroutine `amrex_filccn` located in AMReX_filcc_mod.F90 is somehow like
```
else if (bc(?,?,?) .eq. amrex_bc_liextrap) then

      do k = lo(3), hi(3)
          do j = lo(2), hi(2)
             do i = imin, imax
                q(i,j,k,n) = (2*q(ihi,j,k,n) - q(ihi-1,j,k,n))
             end do
          end do
       end do
```
[...]

PR #592, opened 2019-10-11
add mathematical boundary types : HOEXTRAPLI

DUGKS:
add a new BC type to fill ghost CELL value. [#579](https://github.com/AMReX-Codes/amrex/issues/579)


WeiqunZhang:
Thanks for the pull-request!  Could you add something like `TOCC` or `CC` in the BC type name so that it is more explicit that the value is extrapolated to cell center?

DUGKS:
> Thanks for the pull-request! Could you add something like `TOCC` or `CC` in the BC type name so that it is more explicit that the value is extrapolated to cell center?

Done. 
HOEXTRAPCCLI is too wordy so I just use HOEXTRAPCC. But it may raise ambiguity since users would expect this extrapolation scheme the same as the one used for EDGE values.

PR #601, opened 2019-10-18
Abort on unused inputs

atmyers:
This adds an option, requested by @MaxThevenet, to abort at the end of a run if unused ParmParse variables are present in the table. You turn it on by setting `amrex.abort_on_unused_inputs = 1`.

### DIFF
--- a/Src/Base/AMReX.H
+++ b/Src/Base/AMReX.H
@@ -54,6 +54,7 @@ namespace amrex
         extern std::ostream* oserr;
         
         extern ErrorHandler error_handler;
+        extern int abort_on_unused_inputs;
     }
 
     std::string Version ();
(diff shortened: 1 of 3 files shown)


MaxThevenet:
Thanks!

WeiqunZhang:
@MaxThevenet The abort happens at the end of the run when amrex::Finalize is called.  So what advantage does it provide comparing to checking the stdout for warnings about unused parameters?

atmyers:
For automated testing, we thought it would be easier if it failed instead of having to grep the output through the regression testing suite.

WeiqunZhang:
How do you check if a run has crashed in your automated testing?

atmyers:
if it doesn't exit cleanly it will already be reported as a failure.

PR #610, opened 2019-10-29
small bug fix for a three MultiFab ReduceSum

michaeljbrazell:


WeiqunZhang:
Thanks!  PS. for incflo you want to use the ReduceSum function there to compute the total kinetic energy (and precompute and cache the mask).  The function can take a lambda function working on either Array4s or FABs.  It's probably more convenient to use the Array4 version.

PR #613, opened 2019-11-04
Mini PR: Fix XLinker option to allow spaces in directory names

dpgrote:
As the title says, this fix is needed in case the current directory has spaces in the name. If there is another way of fixing this, that would be fine too.

Without this, the final linking fails since the extra spaces mess up the rpath xlinker options. Note that single quotes are used since double quotes mess up makebuildinfo_C.py since the linker_flags is already wrapped in double quotes.


ax3l:
Thanks!

PR #614, opened 2019-11-05
Expose "dense bins" data structure so that WarpX can use it to implement binary collisions

atmyers:
Note that this is tested in the neighbor list test.


Src/Particle/AMReX_DenseBins.H
```diff
@@ -0,0 +1,117 @@
...
+     *
+     * \param nitems the number of items to put in the bins
+     * \param v pointer to the start of the items
+     * \param bx the Box that defines the space over which the bins will be defined
+     * \param f a function object that maps items to bins 
+     */        
+    template <typename N, typename F>
+    void build (N nitems, T const* v, const Box& bx, F f)
+    {
+        BL_PROFILE("DenseBins<T>::build");
```
WeiqunZhang: We may need to explicitly include AMReX_BLProfiler.H.

Src/Particle/AMReX_DenseBins.H
```diff
@@ -0,0 +1,117 @@
...
+            int ny = hi.y-lo.y+1;
+            int nz = hi.z-lo.z+1;
+            index_type uix = min(nx-1,max(0,iv[0]));
+            index_type uiy = min(ny-1,max(0,iv[1]));
+            index_type uiz = min(nz-1,max(0,iv[2]));
+            pcell[i] = (uix * ny + uiy) * nz + uiz;
+            Gpu::Atomic::Add(&pcount[pcell[i]], index_type{ 1 });
+        });
+
+        Gpu::exclusive_scan(m_counts.begin(), m_counts.end()+1, m_offsets.begin());
```
WeiqunZhang: Why `end()+1`?

Src/Particle/AMReX_DenseBins.H
```diff
@@ -0,0 +1,117 @@
...
+ * empty bins will still take up space.
+ *
+ * \tparam The type of items we hold
+ *
+ */    
+template <typename T>
+class DenseBins
+{
+    using bin_type = IntVect;
+    using index_type = unsigned int;
```
WeiqunZhang: Should these two types be exposed?

Src/Particle/AMReX_NeighborList.H
```diff
@@ -98,61 +99,26 @@ public:
...
-        m_offsets.resize(0);
-        m_offsets.resize(bx.numPts()+1, np);
-
-        m_permutation.resize(np);
+        const auto lo = lbound(bx);
+        const auto hi = ubound(bx);
+        m_bins.build(np, pstruct_ptr, bx,
+                     [=] AMREX_GPU_HOST_DEVICE (const ParticleType& p) noexcept -> IntVect
+                     {
+                         return IntVect((p.pos(0)-plo[0])*dxi[0] - lo.x,
```
WeiqunZhang: Use `AMREX_D_DECL`?

WeiqunZhang: It might be needed in other places.

Src/Particle/AMReX_DenseBins.H
```diff
@@ -0,0 +1,117 @@
...
+        
+        Gpu::thrust_copy(m_offsets.begin(), m_offsets.end()-1, m_counts.begin());
+
+        index_type* pperm = m_perm.dataPtr();
+        constexpr index_type max_index = std::numeric_limits<index_type>::max();
+        AMREX_FOR_1D ( nitems, i,
+        {
+            index_type index = Gpu::Atomic::Inc(&pcount[pcell[i]], max_index);
+            pperm[index] = i;
+        });
```
WeiqunZhang: Might want to insert a stream sync here because the scan and copy are blocking anyway.

Src/Particle/AMReX_DenseBins.H
atmyers: offsets is size num_bins + 1, so that the last entry is the total number of items. Since this is an exclusive scan, the last value in the input array is never used. But this way, the last entry of  the output will be written to.

atmyers: Note, this is now implemented by making `m_counts` size `nbins+1 as well,

PR #616, opened 2019-11-08
Add CUPTI trace feature for GPU kernel timing

mrowan137:
This pull request adds a feature to profile execution time of kernels (which may execute asynchronously) using the CUPTI (CUDA Profiling Tools Interface) APIs.  Added a test case CuptiTest to show example usage for timing of kernel execution.  The feature can be enabled or disabled by setting the value of USE_CUPTI (to TRUE or FALSE) at compile time.


kngott:
What do we want to happen in these cases?
TINY_PROFILE=TRUE, CUPTI=TRUE
TINY_PROFILE=FALSE, CUPTI=TRUE
e.g. Do we want to have the option to turn off tiny profiling, but still have CUPTI timers for load balancing, or do we want to force TINY_PROFILE on whenever CUPTI is on?

Tools/GNUMake/Make.defs
```diff
@@ -536,6 +542,14 @@ ifeq ($(USE_ACC),TRUE)
...
 
+ifeq ($(USE_CUPTI),TRUE)
+  ifeq ($(USE_CUDA),TRUE)
+
+  DEFINES += -DAMREX_USE_CUPTI
+
+  endif
+endif
+
 ifeq ($(USE_HIP),TRUE)
```
kngott: Should add an error if USE_CUDA=FALSE

Src/Base/AMReX_TinyProfiler.H
```diff
@@ -24,12 +24,21 @@ class TinyProfiler
...
     TinyProfiler (std::string funcname, bool start_) noexcept;
+#ifdef AMREX_USE_CUPTI
+    TinyProfiler (std::string funcname, bool start_, bool useCUPTI) noexcept;
+#endif // AMREX_USE_CUPTI
     explicit TinyProfiler (const char* funcname) noexcept;
     TinyProfiler (const char* funcname, bool start_) noexcept;
+#ifdef AMREX_USE_CUPTI
+    TinyProfiler (const char* funcname, bool start_, bool useCUPTI) noexcept;
+#endif // AMREX_USE_CUPTI
     ~TinyProfiler ();
```
kngott: Because the CUPTI macros are separate macros, there's no need for an additional function. Can just add a defaulted (useCUPTI = false) to the current ones.

Src/Base/AMReX_BLProfiler.H
```diff
@@ -484,10 +484,17 @@ inline std::string BLProfiler::CommStats::CFTToString(CommFuncType cft) {
...
+#ifdef AMREX_USE_CUPTI
+#define BL_PROFILE_VAR_NS_CUPTI(fname, vname)             amrex::TinyProfiler tiny_profiler__##vname(fname, false, true);
+#endif
+#define BL_PROFILE_VAR_START(vname)                       tiny_profiler__##vname.start();
+#define BL_PROFILE_VAR_STOP(vname)                        tiny_profiler__##vname.stop();
+#ifdef AMREX_USE_CUPTI
+#define BL_PROFILE_VAR_START_CUPTI(vname)                 tiny_profiler__##vname.start();
+#define BL_PROFILE_VAR_STOP_CUPTI(vname, uintID, charID)  tiny_profiler__##vname.stop(uintID, charID);
+#endif
 #define BL_PROFILE_INIT_PARAMS(ptl,wall,wfabs)
```
kngott: I would put all the CUPTIs together so they are easier to find.

Src/Base/AMReX_BLProfiler.H
```diff
@@ -484,10 +484,17 @@ inline std::string BLProfiler::CommStats::CFTToString(CommFuncType cft) {
...
 #define BL_PROFILE_ADD_STEP(snum)
```
kngott: What happens to *_CUPTI macros when TINY_PROFILE is off? Looks like this is only the "on" situation.

Tests/GPU/CuptiTest/Exec/CUDA/mykernel.H
```diff
@@ -0,0 +1,32 @@
...
+
+AMREX_GPU_DEVICE AMREX_FORCE_INLINE
+void init_mf (int i, int j, int k,
+               amrex::Array4<amrex::Real> const& mf)
+{
+  mf(i,j,k) = 0.;
+}
+
+using clock_value_t = long long;
+
```
kngott: Use the AMReX Macros for device and global.

Tests/GPU/CuptiTest/Exec/CUDA/mykernel.H
```diff
@@ -0,0 +1,32 @@
...
+__global__
+void randomSleep ()
+{
+  // Sleep for random number of cycles
+  int cycles;
+  cycles = (int) (100000*amrex::Random());
+  d_sleep(cycles);
+}
+
```
kngott: Should also make a device version of the sleep function and test this capturing an inlined lambda in AMReX fashion, so make sure it doesn't change the output or accuracy. (And with multiple nested device functions.)

Src/Base/AMReX_TinyProfiler.H
```diff
@@ -41,15 +50,49 @@ public:
 
 private:
     //! stats on a single process
+#ifndef AMREX_USE_CUPTI
```
kngott: Could these be combined into a single version with slight differences if AMREX_USE_CUPTI instead of two completely separate implementations?

Tests/GPU/CuptiTest/Source/main.cpp
```diff
@@ -0,0 +1,88 @@
...
+    // How Boxes are distrubuted among MPI processes
+    DistributionMapping dm(ba);
+
+    // Allocate a MultiFab
+    MultiFab mf(ba, dm, Ncomp, Nghost);
+
+    GpuArray<Real,AMREX_SPACEDIM> dx = geom.CellSizeArray();
+
+    init_mf(mf, geom);
+
```
kngott: Looks like this function doesn't need "geom". Also, is this just "setVal"?

Tests/GPU/CuptiTest/Exec/CUDA/myfunc.cpp
```diff
@@ -0,0 +1,65 @@
+#include "myfunc.H"
+#include "mykernel.H"
+
+#ifdef AMREX_USE_CUPTI
+#include <AMReX_ActivityTraceAsync.H>
+#endif // AMREX_USE_CUPTI
+
```
kngott: Depending on how we want to deal with TINY_PROFILE + CUPTI options, we may want to bury this ifdef logic in the macros themselves.

Src/Base/AMReX_TinyProfiler.cpp
```diff
@@ -134,6 +141,225 @@ TinyProfiler::stop () noexcept
...
+        	
+        if (!uCUPTI) {
+	  t = amrex::second();
+	} else {
+	  cudaDeviceSynchronize();
+	  cuptiActivityFlushAll(0);
+	  t = computeElapsedTimeUserdata(activityRecordUserdata);
+	  nKernelCalls = activityRecordUserdata.size();
+	}
+
```
kngott: cuptiActivityFlushAll is called at every start and stop? Is that required?

Src/Base/AMReX_ActivityTraceAsync.cpp
```diff
@@ -0,0 +1,292 @@
...
+ * Downsized version of activity_trace_async.cpp in:
+ *     $(CUDA_HOME)/extras/CUPTI/samples 
+ *
+ * Copyright 2011-2015 NVIDIA Corporation. All rights reserved
+ *
+ * Sample CUPTI app to print a trace of CUDA API and GPU activity
+ * using asynchronous handling of activity buffers.
+ *
+ */
+
```
kngott: Does this cause any problems?

Tools/GNUMake/Make.defs
mrowan137: Added error: 'error CUPTI can only be used with USE_CUDA=TRUE'.

Src/Base/AMReX_TinyProfiler.H
mrowan137: Modified so that 'useCUPTI = false' is the default.

Src/Base/AMReX_BLProfiler.H
mrowan137: CUPTI macros grouped together now.

Src/Base/AMReX_BLProfiler.H
mrowan137: The macros are not defined when TINY_PROFILE is off, but added a CuptiTrace class that can be used for kernel timings even without TINY_PROFILE.

Tests/GPU/CuptiTest/Exec/CUDA/mykernel.H
mrowan137: Now using the AMReX macros.

Tests/GPU/CuptiTest/Exec/CUDA/mykernel.H
mrowan137: Added a test for inlined lambda launch of device sleep, no obvious changes to accuracy/stability. 
 Added also a test with a recursive device sleep.  calling, deviceSleep(n_cycles) or nestedDeviceSleep(n_cycles/n_levels, n_levels), i.e. recursive sleep over n_levels levels, sleeping for n_cycles/n_levels cycles per level, gives comparable results.

Src/Base/AMReX_TinyProfiler.H
mrowan137: I've made an effort here (and elsewhere) to eliminate redundancy.

Tests/GPU/CuptiTest/Source/main.cpp
mrowan137: Yes; replaced all of the init_mf stuff with setVal(0.0).

Tests/GPU/CuptiTest/Exec/CUDA/myfunc.cpp
mrowan137: The import is now moved to the place where CUPTI macros are defined.  If using the CuptiTrace class as a standalone, the header still needs to be imported.

Src/Base/AMReX_TinyProfiler.cpp
mrowan137: As is, this is how we are able to tag records with a loop iteration marker.  It is not ideal, so we can think about a better way to handle this.

Tools/GNUMake/comps/nvcc.mak
```diff
@@ -112,6 +112,12 @@ ifeq ($(CUDA_VERBOSE),TRUE)
   NVCC_FLAGS += --ptxas-options=-v
 endif
 
+ifeq ($(USE_CUPTI),TRUE)
+  INCLUDE_LOCATIONS += $(CUDAPATH)/extras/CUPTI/include
+  LIBRARY_LOCATIONS += ${CUDAPATH}/extras/CUPTI/lib64 
+  LIBRARIES += -Wl,-rpath,${CUDAPATH}/extras/CUPTI/lib64 -lcupti
```
mrowan137: This compiles on Summit but not Cori, is there a system-independent alternative to `$(CUDAPATH)` ?

jmsexton03: In the sites/Make.nersc there is a SYSTEM_CUDA_PATH and COMPILE_CUDA_PATH set, is that the variable you're looking for?
https://github.com/AMReX-Codes/amrex/blob/83e214b59c54649ea70612e040b2d514c43e961f/Tools/GNUMake/sites/Make.nersc#L64

mrowan137: Yes, that's what I needed!  Thanks.

Tools/GNUMake/comps/nvcc.mak
```diff
@@ -112,6 +112,12 @@ ifeq ($(CUDA_VERBOSE),TRUE)
   NVCC_FLAGS += --ptxas-options=-v
 endif
 
+ifeq ($(USE_CUPTI),TRUE)
+  INCLUDE_LOCATIONS += $(SYSTEM_CUDA_PATH)/extras/CUPTI/include
+  LIBRARY_LOCATIONS += ${SYSTEM_CUDA_PATH}/extras/CUPTI/lib64 
+  LIBRARIES += -Wl,-rpath,${SYSTEM_CUDA_PATH}/extras/CUPTI/lib64 -lcupti
```
WeiqunZhang: Can we change this to `$(MAKE_CUDA_PATH)`?

Src/Base/AMReX_BLProfiler.H
```diff
@@ -484,10 +484,17 @@ inline std::string BLProfiler::CommStats::CFTToString(CommFuncType cft) {
...
+#define BL_PROFILE_VAR_NS(fname, vname)                   amrex::TinyProfiler tiny_profiler__##vname(fname, false, false);                                                                                              
+#define BL_PROFILE_VAR_START(vname)                       tiny_profiler__##vname.start();                                                                                                                               
+#define BL_PROFILE_VAR_STOP(vname)                        tiny_profiler__##vname.stop(); 
+#ifdef AMREX_USE_CUPTI
+  #include <AMReX_CuptiTrace.H>
+  #define BL_PROFILE_VAR_NS_CUPTI(fname, vname)             amrex::TinyProfiler tiny_profiler__##vname(fname, false, true);
+  #define BL_PROFILE_VAR_START_CUPTI(vname)                 tiny_profiler__##vname.start();
+  #define BL_PROFILE_VAR_STOP_CUPTI(vname)                  tiny_profiler__##vname.stop();
+  #define BL_PROFILE_VAR_STOP_CUPTI_ID(vname, uintID)       tiny_profiler__##vname.stop(uintID);
```
WeiqunZhang: The standard says `#` has to be the first character of a line.

Src/Base/AMReX_TinyProfiler.H
```diff
@@ -23,13 +23,16 @@ class TinyProfiler
 {
 public:
     explicit TinyProfiler (std::string funcname) noexcept;
-    TinyProfiler (std::string funcname, bool start_) noexcept;
+    TinyProfiler (std::string funcname, bool start_, bool useCUPTI) noexcept;
```
WeiqunZhang: Could we change this to `TinyProfiler (std::string funcname, bool start_, bool useCUPTI=false) noexcept;`?  And the `const char*` version?

Src/Base/AMReX_CuptiTrace.H
```diff
@@ -0,0 +1,80 @@
...
+// CUPTI buffer size, enough for 4096 activity records in a single buffer;
+// `CUpti_Activity` objects are 8 bytes long
+#define BFR_SIZE (32768)
+
+// 8-byte alignment
+#define ALIGNMENT (8)
+
+// Round down to `align` boundary; mask out last 3 bits (address therefore ends in a 0 or 8)
+#define ALIGN_BFR(bfr, align) ( (uint8_t*) (((uintptr_t)bfr + (align - 1)) & ~ (uintptr_t)(align - 1)) )
+
```
WeiqunZhang: Could we move these macros to `AMReX_CuptiTrace.cpp` so that they do not pollute files that include this header?

Src/Base/AMReX_CuptiTrace.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+
+void CUPTIAPI
+bfrRequestCallback (uint8_t* *bfr, size_t* size, size_t* maxNumRecords) noexcept
+{
+  // Allocate a buffer for use by CUPTI; activity records are stored in the buffer
+  uint8_t* buffer = (uint8_t*) malloc(BFR_SIZE + ALIGNMENT);
+  if (buffer == NULL) {
+    amrex::Print() << "Error: CUPTI requested a buffer but memory cannot be allocated\n";
+    exit(EXIT_FAILURE);
```
WeiqunZhang: Could we use `amrex::Abort` instead?

Src/Base/AMReX_CuptiTrace.cpp
```diff
@@ -0,0 +1,294 @@
...
+      }
+      else if (status == CUPTI_ERROR_MAX_LIMIT_REACHED) {
+	// No more records in the buffer
+        break;
+      }
+      else if (status != CUPTI_SUCCESS) {
+        const char* err;
+	cuptiGetResultString(status, &err);
+	amrex::Print() << "Error: " << err <<  "\n";
+	exit(EXIT_FAILURE);
```
WeiqunZhang: Could we use `amrex::Abort`?

Src/Base/AMReX_CuptiTrace.cpp
```diff
@@ -0,0 +1,294 @@
...
+		     size_t size, size_t validSize) noexcept
+{ 
+  CUptiResult status;
+  CUpti_Activity* record = NULL;
+
+  if (validSize > 0) {
+    do {
+      status = cuptiActivityGetNextRecord(bfr, validSize, &record);
+      if (status == CUPTI_SUCCESS) {
+	CUpti_Activity_Userdata* recordUserData = new CUpti_Activity_Userdata();
```
WeiqunZhang: Where is this `delete`d?

Src/Base/AMReX_CuptiTrace.cpp
```diff
@@ -0,0 +1,294 @@
...
+	CUpti_ActivityKernel4* kernel = (CUpti_ActivityKernel4*) record;
+
+	// Save record data
+	recordUserData->setRecord(record);
```
WeiqunZhang: I think `record` is a pointer that points to somewhere in the user provided buffer `bfr`.   Because `bfr` gets freed later in this function, `record` stored in `recordUserData` will become a dangling pointer.

Src/Base/AMReX_CuptiTrace.cpp
```diff
@@ -0,0 +1,294 @@
...
+}
+
+void
+cuptiTraceStop (unsigned boxUintID) noexcept
+{
+  cudaDeviceSynchronize();
+  cuptiActivityFlushAll(0);
+  for (auto record : activityRecordUserdata) {
+    record->setUintID(boxUintID);
+    record->setCharID(("CharID_" + std::to_string(boxUintID)).c_str());
```
WeiqunZhang: `("CharID_" + std::to_string(boxUintID))` is a temporary `std::string`.  The pointer returned by `c_str()` will be invalidated when the temporary is gone.  The `charID_` member should be changed to `std::string`.

Src/Base/AMReX_CuptiTrace.cpp
```diff
@@ -0,0 +1,294 @@
...
+	CUpti_Activity_Userdata* recordUserData = new CUpti_Activity_Userdata();
+	CUpti_ActivityKernel4* kernel = (CUpti_ActivityKernel4*) record;
+
+	// Save record data
+	recordUserData->setRecord(record);
+	recordUserData->setStartTime(kernel->start);
+	recordUserData->setEndTime(kernel->end);
+	recordUserData->setTimeElapsed(kernel->end - kernel->start);
+	recordUserData->setStreamID(kernel->streamId);
+	recordUserData->setName(kernel->name);
```
WeiqunZhang: `const char* name_` member should be changed to `std::string`, otherwise it will be a dangling pointer.

Src/Base/AMReX_TinyProfiler.H
mrowan137: Done.

Src/Base/AMReX_CuptiTrace.H
mrowan137: Done, all these macros moved to `AMReX_CuptiTrace.cpp`.

Src/Base/AMReX_CuptiTrace.cpp
mrowan137: Switched to `amrex::Abort`.

Src/Base/AMReX_CuptiTrace.cpp
mrowan137: Switched to `amrex::Abort`.

Src/Base/AMReX_CuptiTrace.cpp
mrowan137: I had a function `void CuptiTrace::cleanup () noexcept` to either call manually or that is called by the destructor of a `CuptiTrace` object, which clears the `recordUserData`s.  is this not a safe way to do it, since there is no guarantee the `cleanup` function is called?

Src/Base/AMReX_BLProfiler.H
mrowan137: Done.

Tools/GNUMake/comps/nvcc.mak
mrowan137: Done.

Src/Base/AMReX_CuptiTrace.cpp
WeiqunZhang: Yes, it's safe.  For example, you call in `cuptiTraceStart`, `std::vector::clear` is called.  That will not release the memory.  I actually missed the `delete[]` in `cleanup` because I was searching for `delete ` with a space after `delete`.  The reason I searched for deleted was the object was made with `new`, not `new[]`.  Note it is an *error* to call `delete[]` on an object made with `new`.  One can only call `delete[]` on objects made with `new[]`.  My suggestion is to make it `std::vector<std::unique_ptr<CUpti_Activity_UserData>>`.

Src/Base/AMReX_CuptiTrace.cpp
mrowan137: Got it, thanks.  Switched to `std::string`. the only `const char*` in use now is in returning and error before an `amrex::Abort`.

Src/Base/AMReX_CuptiTrace.cpp
mrowan137: I've removed `record` as an object saved to `recordUserData`, now it just gathers the needed information during the buffer completion callback, and the dangling pointer is cleaned up after `bfr` is freed.

Src/Base/AMReX_CuptiTrace.cpp
mrowan137: Thanks for explaining, I have switched to `std::vector<std::unique_ptr<CUpti_Activity_UserData>>` as you suggested.  I hope this and similar issues are now avoided, please let me know if it looks OK.

Src/Base/AMReX_CuptiTrace.cpp
mrowan137: Got it, this is switched now to `std::string`.

PR #623, opened 2019-11-19
added option in libamrex configure script to specify cuda-arch

bsrunnels:
This allows user to specify the cuda architecture using the `./configure` script when building `libamrex`.


WeiqunZhang:
It appears that the script doesn't work.

bsrunnels:
@WeiqunZhang Sorry about that I was using a different version of python. It is fixed now and passes the tests.

PR #626, opened 2019-11-21
sundials 5 interface

balos1:
@mic84


drummerdoc:
@jmsexton03 @EnnaDelfen 

Please review this and verify that you can work with the new scheme (or that your setup is unaffected).  It would be awesome if Cody's work here pays off, but I don't want you guys caught off guard.

Tools/GNUMake/Make.defs
```diff
@@ -814,14 +814,9 @@ ifeq ($(USE_UPCXX),TRUE)
...
-  else
-    $(info Loading $(AMREX_HOME)/Tools/GNUMake/packages/Make.sundials3...)
-    include        $(AMREX_HOME)/Tools/GNUMake/packages/Make.sundials3
-  endif
+ifeq ($(USE_SUNDIALS),TRUE)
+  $(info Loading $(AMREX_HOME)/Tools/GNUMake/packages/Make.sundials...)
+  include        $(AMREX_HOME)/Tools/GNUMake/packages/Make.sundials
 endif
 
 ifeq ($(USE_CVODE),TRUE)
```
jmsexton03: It looks like Tools/GNUMake/packages/Make.cvode was deleted, so code which relies on USE_CVODE=TRUE will fail. The initial purpose of having two separate flags for USE_SUNDIALS and USE_CVODE was to allow backward compatibility, particularly for users who rely on the Cray TPSL (cori's current version of that is 18.6, which I believe corresponds to cvode 2.7).

drummerdoc: which apps are depending on the Crap TPSL?  What if we asked that it be updated to more recent versions of CVODE/SUNDIALS?  I don't know the issues involved.

balos1: I have added the cvode 2.7 related pieces back in.

jmsexton03:
This looks good to me, I'll try to update the documentation on the flags/tutorial sometime in the next week.

PR #627, opened 2019-11-24
Add a MultiFab operation that explicitly swaps dst/src data

maxpkatz:
Also add this as an option for StateData.

Pointer swaps are not sufficient when the data between the source and destination MultiFabs live in different memory spaces, such as pinned and device memory, so this addresses those cases.


WeiqunZhang:
Some information for you to consider.  `MultiFab`'s base class `FabArray` has a function `const Arena* arena()` that returns the Arena used for building Fabs.  So one could test whether two MultiFabs are built with the same arena.  Also `MultiFab` objects (not just pointers) can be swapped too with `std::swap` because they have move operator=.   (Their pointers to arena will be swapped too.  So that's probably not what you want.)

maxpkatz:
Right, I'm specifically trying to handle a case in Castro where I want to copy the data between a pinned buffer and a managed buffer but I _don't_ want to swap the pointers.

WeiqunZhang:
I guess the question is whether it makes sense in StateData's swap to test before doing the deep swap.  Just worried about potential confusion for the users.  There is an extra argument in the StateData swap and there are two ways to swap MultiFab data.

maxpkatz:
Since I named the argument `pointer_swap` I would rather honor what the user asked us to do than attempt to be clever and skip if it we detect that it ought not to be necessary.

But I could make this a separate interface rather than overloading replaceOldData/replaceNewData?

WeiqunZhang:
For StateData, either way is fine with me.  But for MultiFab, should we take a shortcut if the two have the same number of ghost cells and arena?

maxpkatz:
Oh, I understand. Yes, I agree that makes sense and I will make that change.

WeiqunZhang:
And the number of components and starting component need to be same too.  And the factory.

PR #630, opened 2019-12-05
CXXSTD: Project Control of C++ Std

ax3l:
Add a new option, defaulting to C++11, to give projects control over the required C++ standard of a project.

This improves the current logic, which does not cleanly fail if a compiler does not support e.g. C++14 but instead gave hard-to-read error messages such as "hm, dunno what [std::make_unique](https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique) is" :)
Now the compiler will fail with messages such as:
```
gcc: error: unrecognized command line option ‘-std=c++14’
```
which is better and safes us from re-implementing a ton of logic that CMake provides out-of-the-box.

Update: We actually error for GCC now already during the make check of the compiler's major version.

```
Loading ../amrex/Tools/GNUMake/comps/gnu.mak...
../amrex/Tools/GNUMake/comps/gnu.mak:121:
*** C++14 support requires GCC 5 or newer..  Stop.
```


Tools/GNUMake/comps/gnu.mak
```diff
@@ -114,12 +114,8 @@ endif
 
 ########################################################################
 
-ifeq ($(gcc_major_version),4)
```
maxpkatz: If we make this change, it seems appropriate to me for the makefile to throw an error if the user attempts to use C++14 with gcc 4.x, for precisely the reason you're saying about avoiding incomprehensible warnings.

Docs/sphinx_documentation/source/BuildingAMReX.rst
```diff
@@ -34,6 +34,8 @@ list of important variables.
    | AMREX_HOME | Path to amrex                       | environment |
    +------------+-------------------------------------+-------------+
    | COMP       | gnu, cray, ibm, intel, llvm, or pgi | none        |
+   +------------|-------------------------------------+-------------+
+   | USE_CXXSTD | C++ Standard (11, 14, 17, ...)      | 11          |
```
maxpkatz: Can we name the variable CXXSTD instead of USE_CXXSTD? We usually reserve the USE_ prefix for boolean variables that either are TRUE or FALSE.

Tools/GNUMake/comps/gnu.mak
ax3l: Sounds good, yes!

Docs/sphinx_documentation/source/BuildingAMReX.rst
ax3l: absolutely, thanks for the hint! :)

Tools/GNUMake/comps/gnu.mak
ax3l: Hm, on second though: we actually have to do <= and >= comparisons here for major versions and C++ std version strings which are not first-class citizens in GNUMake. Not sure how much of CMake's compiler sanity checks we want to re-implement.

ax3l: But can do shell calls if that's ok. `$(shell expr $(CXXSTD) \> 11)`, etc.

maxpkatz: We don't have to enforce every sanity check, but there is a specific historical reason why we had that particular check and it might be good to keep it. It's unlikely that anyone is using gcc <4, so the check as it currently is seems fine to me.

Tools/GNUMake/comps/pgi.mak
```diff
@@ -82,11 +82,7 @@ else
 endif
 
 # The logic here should be consistent with what's in nvcc.mak
-ifeq ($(shell expr $(gcc_major_version) \>= 5), 1)
```
ax3l: is this correct to check GCC's version here? Does PGI use their stdlib and we rely on that? Does not mean the language features are there, though.

Tools/GNUMake/comps/gnu.mak
ax3l: Alright, updated anyway and tested now.

Docs/sphinx_documentation/source/BuildingAMReX.rst
```diff
@@ -34,6 +34,8 @@ list of important variables.
    | AMREX_HOME | Path to amrex                       | environment |
    +------------+-------------------------------------+-------------+
    | COMP       | gnu, cray, ibm, intel, llvm, or pgi | none        |
+   +------------|-------------------------------------+-------------+
+   | CXXSTD     | C++ Standard (11, 14, 17, ...)      | 11          |
```
ax3l: we could also write the whole string, e.g. `c++11`, and `gnu11` here to be able to express compiler extensions. what do you think?

Tools/GNUMake/comps/pgi.mak
maxpkatz: That code came from @atmyers in commit daa5a07 based on the git log, maybe he can comment on his reasoning for it.

Tools/GNUMake/comps/gnu.mak
```diff
@@ -114,12 +114,14 @@ endif
...
-  CXXFLAGS += -std=c++14
+ifeq ($(shell expr $(gcc_major_version) \< 5),1)
+  ifeq ($(shell expr $(CXXSTD) \> 11),1)
+    $(error C++14 support requires GCC 5 or newer.)
+  endif
 endif
-CFLAGS     += -std=gnu99
+
+CXXFLAGS += -std=c++$(CXXSTD)
+CFLAGS   += -std=gnu99
```
ax3l: not sure how well gnu C extensions and iso c++ mix in corner cases...

WeiqunZhang: This seems to change the make system behavior.  Currently for gcc >= 6, we use whatever the default of gcc is.  But now, we use C++11 by default.

Tools/GNUMake/comps/intel.mak
```diff
@@ -37,11 +37,13 @@ endif
...
 ########################################################################
 
-ifeq ($(firstword $(sort 17.0 $(intel_version))), 17.0) 
-  CXXFLAGS += -std=c++14
-else
-  CXXFLAGS += -std=c++11
+ifneq ($(firstword $(sort 17.0 $(intel_version))), 17.0)
+  ifeq ($(shell expr $(CXXSTD) \> 11),1)
+    $(error C++14 support requires Intel icpc 17.0 or newer.)
+  endif
```
WeiqunZhang: It seems that for Intel >= 18 it now uses C++11 by default.

Tools/GNUMake/comps/gnu.mak
ax3l: Yep, we can also keep the default in `Tools/GNUMake/Make.defs` empty and keep that logic, as you prefer.

Tools/GNUMake/comps/intel.mak
ax3l: yes, as above.

Tools/GNUMake/comps/pgi.mak
WeiqunZhang: PGI uses some gcc headers.  If those headers are C++11, pgc++ -std=c++14 will not work.  I remember at some point we had to do that to be able to compile on both ornl machines and our local machines.  We were not happy with the check.  But didn't find another way.

WeiqunZhang:
A hack Andrew and I sometimes use is `XTRA_CXXFLAGS = --std=c++xxxx`.

Tools/GNUMake/comps/gnu.mak
ax3l: If we want to keep the old behaviour, I would implement blocks like this in every compiler makefile:
```make
ifdef CXXSTD
  CXXSTD := $(strip $(CXXSTD))
  ifeq ($(shell expr $(gcc_major_version) \< 5),1)
    ifeq ($(shell expr $(CXXSTD) \> 11),1)
      $(error C++14 support requires GCC 5 or newer.)
    endif
  endif
  CXXFLAGS += -std=c++$(CXXSTD)
else
  ifeq ($(gcc_major_version),4)
    CXXFLAGS += -std=c++11
  else ifeq ($(gcc_major_version),5)
    CXXFLAGS += -std=c++14
  endif
endif
```

ax3l: The documentation entry for users should then read for the current behavior:
```rst
+------------|-------------------------------------+--------------------+
| CXXSTD     | C++ standard (``c++11``, ``c++14``) | compiler default,  |
|            |                                     | at least ``c++11`` |
+------------+-------------------------------------+--------------------+
```

ax3l:
Yes, I try to avoid this because it's not easy to explain to our users :)

Tools/GNUMake/comps/gnu.mak
WeiqunZhang: Yes, I think that's the way to go.  Should be `11` instead of `c++11`, right?

Tools/GNUMake/comps/pgi.mak
WeiqunZhang: To put it another way.  The issue was a version of pgc++ can support C++14 doesn't mean `pgc++ -std=c++14` will always work.  It actually depends on the gcc headers they use.

Tools/GNUMake/comps/gnu.mak
ax3l: Okay with me.
`11` vs. `c++11`: https://github.com/AMReX-Codes/amrex/pull/630#discussion_r354539768

Docs/sphinx_documentation/source/BuildingAMReX.rst
WeiqunZhang: I actually like `c++11` better than `11` for the reason that we can set `gnu++`.  I made the comment because the code is currently doing CXXFLAGS += -std=c++$(CXXSTD).

Tools/GNUMake/comps/pgi.mak
ax3l: Should we check if the GCC major is a certain minimum and the PGI major is sufficiently high as well?

Docs/sphinx_documentation/source/BuildingAMReX.rst
ax3l: Ok cool, then I'll change this as indicated because it's more general.

Tools/GNUMake/comps/intel.mak
ax3l: updated to old behaviour (which is also closer to CMake's behaviour)

Docs/sphinx_documentation/source/BuildingAMReX.rst
```diff
@@ -28,23 +28,26 @@ list of important variables.
 
 .. table:: Important make variables
```
ax3l: follow-up PR: what's missing here and what's very important: `CC` and `CXX` variables, since environment hints are ignored by AMReX.

WeiqunZhang:
`ifeq ($(shell expr $(CXXSTD) \> 11),1)` won't work anymore because `CXXSTD` is no longer just a number.

ax3l:
cheese, yep. will crop to the last two digits and neglect things like `2a`, `1y`, etc.

ax3l:
`expr` does not support substrings from the back, so I will now just stop implementing CMake functionality in Make and make this a comparison for C++14 safety only, as before :)

ax3l:
Thanks for the great review @maxpkatz and @WeiqunZhang ! :+1:

PR #636, opened 2019-12-16
Add support for CMake to handle PETSc

BarrySmith:
Add missing include files to Src/EB/CMakeLists.txt


mic84:
Thanks for your pull request. I will take care of making some adjustments.

PR #638, opened 2019-12-19
In ParticleToMesh, apply SumBoundary & finalize MultiFab for both GPU and CPU

dwillcox:
This moves the call to `SumBoundary` and the MultiFab copy and cleanup from inside the scope of the `else` condition for `Gpu::inLaunchRegion()` to the outer scope of the `ParticleToMesh` function so we do this independent of whether we're on the GPU.


atmyers:
Thanks!

PR #639, opened 2019-12-22
Added begin and end methods to IntVect and RealVect classes

rporcu:
The begin() and end() methods are needed in order to use STL algorithms such as std::fill, std::transform, etc...


WeiqunZhang:
Line 83 of AMReX_MLCellABecLap.cpp

```
// const auto probhi = m_geom[amrlev][mglev].ProbHiArray(); SET_BUT_NOT_USED 
```

This is used in 1D and 2D.

rporcu:
Thanks for letting me know. I've added a conditional statement for fixing the bug. Is that correction proper?

PR #648, opened 2020-01-05
The defition of single_level() should be 'void' rather 'int', otherwise compilers will report errors.

sudormroot:
amrex/OldTutorials/PIC_C/single_level.cpp

The defition of single_level() should be 'void' rather 'int', otherwise compilers will report errors.


WeiqunZhang:
It has been merged into development branch.

PR #652, opened 2020-01-08
Adding AMREX_GPU_HOST_DEVICE to functions and members of class RealVect

rporcu:
Adding attribute AMREX_GPU_HOST_DEVICE in order to be able to use the class in device code and also be consistent with class IntVect


atmyers:
Thanks!

PR #654, opened 2020-01-08
Added functionality of extract a RealVect from Particle class

rporcu:
In this way, we can extract a RealVect like velocity, angular velocity, etc... without having to call function rdata many times

### DIFF
--- a/Src/Particle/AMReX_Particle.H
+++ b/Src/Particle/AMReX_Particle.H
@@ -6,6 +6,7 @@
 #include <AMReX_REAL.H>
 #include <AMReX_FArrayBox.H>
 #include <AMReX_IntVect.H>
+#include <AMReX_RealVect.H>
 
 #include <AMReX_ParmParse.H>
 #include <AMReX_Geometry.H>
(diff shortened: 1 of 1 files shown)


Src/Particle/AMReX_Particle.H
```diff
@@ -91,6 +95,20 @@ struct Particle
 		AMREX_ASSERT(index < NReal);
 		return m_rdata.arr[AMREX_SPACEDIM + index];
 	}
+    AMREX_GPU_HOST_DEVICE RealVect  rdata(AMREX_D_DECL(int indx, int indy, int indz)) const &
```
atmyers: There is actually a problem with this that I didn't notice yesterday. In 1D, the above function arguments become the same as the `RealType  rdata(int index) const &` method defined above. You aren't allowed to overload functions that differ only in their return type, so this doesn't compile.

What do you think about renaming the `RealVect`-returning versions of `rdata` to something like `rvec`?

rporcu: Right! I did not realize there could be that issue. To me sounds good to rename it from rdata to rvec. How should we proceed? Is it possible to fix the bug on this pull request or I should make another pull request or do you prefer to fix it personally?

atmyers: Another pull request sounds good - thanks!

PR #661, opened 2020-01-13
Added 'TheZeroVector', 'TheUnitVector' and 'floor' functions to class RealVect

rporcu:
All the changes are consistent with the class IntVect. The 'floor' function can help to write more readable code.

### DIFF
--- a/Src/Base/AMReX_RealVect.H
+++ b/Src/Base/AMReX_RealVect.H
@@ -62,6 +62,14 @@ public:
 
   ///
   /**
+     Construct a RealVect setting the coordinates to the corresponding values in
+     the Real array a.
+  */
+  AMREX_GPU_HOST_DEVICE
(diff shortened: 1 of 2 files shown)


Src/Base/AMReX_RealVect.H
```diff
@@ -672,6 +714,36 @@ RealVect::scale (Real s) noexcept
   return *this;
 }
 
+AMREX_GPU_HOST_DEVICE
+inline
+IntVect
+RealVect::floor () const noexcept
+{
```
atmyers: are you sure these functions ('floor', 'ceil', and 'round') have device implementations in the versions of CUDA we care about (say, 9.2 and later)?

rporcu: I found this in the documentation
https://docs.nvidia.com/cuda/archive/9.2/cuda-math-api/group__CUDA__MATH__DOUBLE.html

PR #673, opened 2020-01-25
New version of setBndry for GPUs. 

kngott:


kngott:
A few questions on this same topic:

1) What about setDomainBndry? On GPUs, there are so few big boxes, there's probably some work to do on every box, so maybe this will also work there, just with fewer positive results? (I imagine the most likely difference would be a ton of interior boxes that do nothing but slow down the total runtime by having to check every cell anyway.)

2) Can we reduce the observed overhead in the fused launch and/or the BoxDiff and/or the AsyncArray to make the other options more feasible? (Or just make those options better?)

PR #674, opened 2020-01-25
VisMF: Remove Shadow Warning

ax3l:
Change a private static interface to use the only-usage, same-class static control variable it passes to avoid variable name shadowing.


ax3l:
The public `RemoveFiles` member also has a `verbose` option that shadows a member, not sure if we can remove that. Only used once in `IOTest.cpp`.

PR #682, opened 2020-01-31
amrex::Vector implicit conversion from std::vector

maikel:
Hi,

this change allows to pass `std::vector<T, Allocator>` objects directly to functions which take an  `amrex::Vector<T, Allocator>` argument. Currently one need to call another constructor as in `amrex::Vector<T>(v.begin(), v.end())`

What do you think?

### DIFF
--- a/Src/Base/AMReX_Vector.H
+++ b/Src/Base/AMReX_Vector.H
@@ -27,6 +27,12 @@ public:
     using std::vector<T, Allocator>::vector;
     using typename std::vector<T, Allocator>::size_type;
 
+    Vector(const std::vector<T, Allocator>& other)
+      : vector(other) {}
+
+    Vector(std::vector<T, Allocator>&& other) noexcept
+      : vector(std::move(other)) {}
+
     T& operator[] (size_type i) noexcept
     {
         BL_ASSERT( i < (this->std::vector<T, Allocator>::size()) );
--- a/Src/Base/AMReX_Vector.H
+++ b/Src/Base/AMReX_Vector.H
@@ -27,6 +27,8 @@ public:
     using std::vector<T, Allocator>::vector;
     using typename std::vector<T, Allocator>::size_type;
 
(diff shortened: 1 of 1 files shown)


maxpkatz:
I noticed a need last month for something which takes only an `amrex::Vector<T>` as an argument (e.g. `CoordSys::GetCellLoc`) but that I wanted to fill with an `amrex::Vector<T, allocator>` (i.e. a `DeviceVector`). Is that what this would allow? It wasn't clear to me from the description.

maikel:
No this is merely a conversion from the standard vector `std::vector<T, Allocator>` to `amrex::Vector<T, Allocator>`. I think you ask for a copy when the Allocator type changes. But this depends very heavily on the allocators in question. In this case I think you do not come around the copy constructor `amrex::Vector<T, Allocator2>(v.begin(), v.end())`, because even the notion of memory and pointers could be completely different.

WeiqunZhang:
It's probably not good that the deep copy is hidden from the user.  Maybe make it explicit?

WeiqunZhang:
`Vector() = default;` Is that necessary given that there is already `using std::vector<T, Allocator>::vector;`?

WeiqunZhang:
Interesting.  Seems that once there is a user provided ctor, the default one is no longer generated even with `using std::vector::vector` and `std::vector` has a default ctor.

WeiqunZhang:
I guess the explanation is the user deleted default ctor blocks the default ctor of base class because they have the same signature.

WeiqunZhang:
If we make it explicit, you have to type more and it's not clear it's better than simply passing `{v.begin(),v.end()}` without any changes.

maikel:
The hidden copy is a valid point. But instead of an explicit constructor I prefer no change at all.

maikel:
Btw, I think the need for the default constructor is a compiler issue and not mandated by the standard.

PR #684, opened 2020-02-05
Add Poisson random distribution

lucafedeli88:
Hi,

I would like to propose to add Poisson distribution to random number distributions available in AMReX .
For the CPU version we can rely on the C++ standard template library, while for the GPU version we can use the function provided by cuRAND.

My editor has also removed some trailing spaces. I can revert these changes if you want.

### DIFF
--- a/Src/Base/AMReX_Random.H
+++ b/Src/Base/AMReX_Random.H
@@ -18,11 +18,22 @@ namespace amrex
     /**
     * \brief Generate a psuedo-random double using C++11's mt19937.
     *
-    *  Generates one pseudorandom real number (double) which is
-    *  uniformly distributed on [0,1)-interval for each call.
+    *  Generates one pseudorandom real number (double) from a normal
+    *  distribution with mean 'mean' and standard deviation 'stddev'.
     *
     */
-    AMREX_GPU_HOST_DEVICE Real Random (); // [0,1)
+    AMREX_GPU_HOST_DEVICE Real Random ();
+
+    /**
+    * \brief Generate a psuedo-random integer from a Poisson distribution
+    *
+    *  Generates one pseudorandom positive integer number (double)
+    *  extracted from a Poisson distribution, given the Real parameter lambda.
+    *  The CPU version of this function relies on the standard Template Library
+    *  The GPU version of this function relies on the cuRAND library
+    *
+    */
+    AMREX_GPU_HOST_DEVICE unsigned int RandomPoisson (Real lambda);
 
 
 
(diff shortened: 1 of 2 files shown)


WeiqunZhang:
I have some comments.

The CUDA documentation says

```
__device__ unsigned int 
curand_poisson (curandState_t *state, double lambda)
```

> This function returns a single Poisson-distributed unsigned int based on a Poisson distribution with the given lambda. The algorithm used to derive a Poisson result from a uniformly distributed result varies depending on the value of lambda and the type of generator. Some algorithms draw more than one sample for a single output. Also note that this distribuition requires pre-processing on the host. See the description of curandCreatePoissonDistribution() below. 

So it seems that it requires a pre-processing on the host.

I also have concern about the performance on CPU.  For Poisson distribution, it might be faster to save the distribution instead of constructing one for every call.  The constructor might be slow.  Here is a comment in gcc source code.

> // XXX Is the constructor too slow?
> std::poisson_distribution<result_type> __poisson(__y);

https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/include/bits/random.tcc#L1116

Another thing is the implementation might generate several numbers.  They give out one and save the rest.

lucafedeli88:
Thanks for your comments!
Concerning curand_poisson`,  the documentation (e.g. https://docs.nvidia.com/cuda/curand/device-api-overview.html#poisson-api-example ) is a bit ambiguous... indeed, while it mentions what you reported, it also says:

> This example shows the differences between the 3 API types for the Poisson distribution. [...]
 The host API is the most robust for generating large vectors of Poisson-distributed random numbers. (i.e. it has the best statistical properties across the full range of lambda values) The discrete Device API is almost as robust as the HOST API and allows Poisson-distributed random numbers to be generated inside a kernel. The simple Device API is the least robust but is more efficient when generating Poisson-distributed random numbers for many different lambdas.

and then it shows an example where, for the `simple Device API`, no particular pre-processing seems to be done. This is encouraging. However, I've not been able to find a clear explanation on how the API works.
Concerning the possible performance issues related to `std::poisson_distribution`, you may be right.
However, the `lambda` parameter might change from call to call (at least this is the use case that I have in mind). So if this case it's not really possible to save the distribution, since it depends on lambda...

WeiqunZhang:
@maxpkatz Looks like the CUDA documentation is inaccurate on `curand_poisson` requiring host side preparation.

maxpkatz:
@WeiqunZhang Indeed, the documentation on this was incorrect. It will be corrected in the next CUDA release.

PR #687, opened 2020-02-05
Conduit Blueprint Update to support Multiple Topologies

cyrush:
Ascent now supports publishing multiple topologies via a single publish call.
(This allows you to render images that contain both the AMReX particles and AMReX mesh)

This PR makes a small change to the blueprint wrapping functions to avoid resetting the output node, which would undermine this new support.


cyrush:
Thanks @WeiqunZhang !

PR #690, opened 2020-02-09
simplify write_probin + add C++ parameter support

zingale:
This simplifies write_probin.py to admit only a single parameter group + updates it to allow for C++ headers to be output. 

As far as I know only MAESTROeX was using multiple parameter groups, and it has been updated already.


Tools/F_scripts/write_probin.py
```diff
@@ -385,21 +364,111 @@ def write_probin(probin_template, param_A_files, param_B_files,
...
+        fout.write(CXX_HEADER)
+
+        fout.write("  void init_{}_parameters();\n\n".format(cxx_prefix))
+
+        for p in params:
+            if p.dtype == "character":
+                # we don't support character
+                continue
+
+            fout.write("  extern {} {};\n\n".format(p.get_cxx_decl(), p.var))
```
maxpkatz: We'll want to add a __managed__ qualifier so we can access these on the GPU, but we can do that in a later commit.

maxpkatz: Done.

PR #700, opened 2020-02-21
FAB: Fix Constructor Calls

ax3l:
This fixes a bug in `Array4CopyTag` construction while pushing back to a vector. The current call does not do what one expects, as hinted by the nvcc warning:
```
Src/Base/AMReX_FBI.H(920): warning: variable "zero" was set but never used
          detected during:
            instantiation of "void amrex::FabArray<FAB>::FBEP_nowait(int, int, const amrex::IntVect &, const amrex::Periodicity &, __nv_bool, __nv_bool) [with FAB=amrex::FArrayBox, F=amrex::FArrayBox, <unnamed>=void]"
Src/Base/AMReX_FabArray.H(1993): here
            instantiation of "void amrex::FabArray<FAB>::FillBoundary_nowait(int, int, const amrex::IntVect &, const amrex::Periodicity &, __nv_bool) [with FAB=amrex::FArrayBox]"
Src/Base/AMReX_FabArray.H(1870): here
            instantiation of "void amrex::FabArray<FAB>::FillBoundary(const amrex::Periodicity &, __nv_bool) [with FAB=amrex::FArrayBox]"

Src/Base/AMReX_FBI.H(955): warning: variable "zero" was set but never used
          detected during:
            instantiation of "void amrex::FabArray<FAB>::FillBoundary_finish() [with FAB=amrex::FArrayBox, F=amrex::FArrayBox, <unnamed>=void]"
Src/Base/AMReX_FabArray.H(1871): here
            instantiation of "void amrex::FabArray<FAB>::FillBoundary(const amrex::Periodicity &, __nv_bool) [with FAB=amrex::FArrayBox]"
```

Basically the last arguments were not passed to object construction of `amrex::Array4CopyTag`, because a default constructor with four args is not implicitly created (likely due to some non-trivial types).

Further fixes:
- add missing includes
- move created r-values in (`emplace_back`)

CC @WeiqunZhang and @kngott (inline HIP notes)

### DIFF
--- a/Src/Base/AMReX_FBI.H
+++ b/Src/Base/AMReX_FBI.H
@@ -1,6 +1,12 @@
 #ifndef AMREX_FBI_H_
 #define AMREX_FBI_H_
 
+#include <AMReX_Array4.H>
+#include <AMReX_Box.H>
+#include <AMReX_Dim3.H>
+#include <AMReX_IntVect.H>
+
+
 template <class FAB>
 struct FabCopyTag {
     FAB const* sfab;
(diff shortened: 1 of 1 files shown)


Src/Base/AMReX_FBI.H
```diff
@@ -1,6 +1,12 @@
 #ifndef AMREX_FBI_H_
 #define AMREX_FBI_H_
 
+#include <AMReX_Array4.H>
+#include <AMReX_Box.H>
+#include <AMReX_Dim3.H>
+#include <AMReX_IntVect.H>
+
+
```
WeiqunZhang: This is meant to be a private header that is included by AMReX_FabArrayCommI.H`, which is also meant to be private.  But it's OK to include these.

WeiqunZhang:
Thanks!

Src/Base/AMReX_FBI.H
ax3l: but... but...: https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/WhyIWYU.md
:)

PR #705, opened 2020-02-26
Fix Warning in AMREX_DEVICE_COMPILE

ax3l:
Adding `defined()` in a macro is raising warnings of undefined behavior (clang) or non-portable behavior (gcc).

Modern versions of GCC (used: v7.4.0) issue a `-Wexpansion-to-defined` warning on this.
```
Src/Base/AMReX_GpuAtomic.H:390:26:
  warning: this use of "defined" may not be portable
  [-Wexpansion-to-defined]
 #if AMREX_DEVICE_COMPILE()
```

Clang's warning:
```
  warning: macro expansion producing 'defined' has undefined behavior
    [-Wexpansion-to-defined] 
```

This tries to fix 2133x compile warnings in WarpX.

### To Do

- [x] it's late at night, verify again the warning is actually gone now
- [x] test-compile with nvcc to see everything is still working
- [x] feedback if DPC++ & HIP compile paths still work

### DIFF
--- a/Src/Base/AMReX.H
+++ b/Src/Base/AMReX.H
@@ -91,7 +91,7 @@ namespace amrex
     namespace detail { void Error_host_doit (const char * msg); }
     AMREX_GPU_HOST_DEVICE inline
     void Error (const char * msg = 0) {
-#if AMREX_DEVICE_COMPILE()
+#if AMREX_DEVICE_COMPILE
         if (msg) AMREX_DEVICE_PRINTF("Error %s\n", msg);
         AMREX_DEVICE_ASSERT(0);
 #else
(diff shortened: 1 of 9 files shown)


WeiqunZhang:
Could you make the following change?

```
diff --git a/Src/Base/AMReX_GpuQualifiers.H b/Src/Base/AMReX_GpuQualifiers.H
index 06418da..dadc689 100644
--- a/Src/Base/AMReX_GpuQualifiers.H
+++ b/Src/Base/AMReX_GpuQualifiers.H
@@ -33,6 +33,6 @@

 #endif

-#define AMREX_DEVICE_COMPILE() defined(__CUDA_ARCH__) || defined(__HIP_DEVICE_COMPILE__) || defined(__SYCL_DEVICE_ONLY__)
+#define AMREX_DEVICE_COMPILE (__CUDA_ARCH__ || __HIP_DEVICE_COMPILE__ || __SYCL_DEVICE_ONLY__)
```

This seems to work on for CUDA, DPC++ and non-gpu.  Didn't test HIP, but I expect it to work because its compiler is similar to nvcc in terms of host and device compilation processes.

ax3l:
Thank you for testing this @WeiqunZhang! I added your suggestion and made sure that WarpX OpenMP and CUDA builds with the patch.

ax3l:
Update: the issue arises from defining a macro with `defined(...)` in it. This patch is not yet fixing the warning -.-

PR #724, opened 2020-03-03
Conduit Blueprint: fix case needed for single prec particles

cyrush:
Use `ParticleReal` instead of `Real` when wrapping used defined Particle SOA values

(thanks to @jmsexton03 for finding this issue)


cyrush:
@atmyers thanks!

PR #725, opened 2020-03-04
Add efficiency argument to makeSFC, makeKnapSack

mrowan137:
This pull request adds function overloads for `DistributionMapping::MakeKnapSack` and `DistributionMapping::MakeSFC` which allow to return the efficiency of the newly computed distribution mapping.

### DIFF
--- a/Src/Base/AMReX_DistributionMapping.H
+++ b/Src/Base/AMReX_DistributionMapping.H
@@ -229,7 +229,7 @@ private:
                               const std::vector<long>& wgts,
                               int                      nprocs,
                               bool                     sort=true,
-			      Real*                    efficiency=nullptr);
+                              Real*                    efficiency=nullptr);
 
     void RRSFCDoIt           (const BoxArray&          boxes,
                               int                      nprocs);
--- a/Src/Base/AMReX_DistributionMapping.H
+++ b/Src/Base/AMReX_DistributionMapping.H
@@ -132,6 +132,8 @@ class DistributionMapping
 
     void SFCProcessorMap(const BoxArray& boxes, const std::vector<long>& wgts, int nprocs,
                          bool sort=true);
+    void SFCProcessorMap(const BoxArray& boxes, const std::vector<long>& wgts, int nprocs,
+                         Real& efficiency, bool sort=true);
     void KnapSackProcessorMap(const std::vector<long>& wgts, int nprocs,
                               Real* efficiency = 0,
 			      bool do_full_knapsack = true,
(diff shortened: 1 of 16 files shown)


Src/Base/AMReX_DistributionMapping.H
```diff
@@ -159,13 +161,20 @@ class DistributionMapping
...
+    static DistributionMapping makeKnapSack   (const MultiFab& weight, Real& eff,
+                                               int nmax=std::numeric_limits<int>::max());
     static DistributionMapping makeKnapSack   (const Vector<Real>& rcost,
                                                int nmax=std::numeric_limits<int>::max());
+    static DistributionMapping makeKnapSack   (const Vector<Real>& rcost, Real& eff,
+                                               int nmax=std::numeric_limits<int>::max());
 
     static DistributionMapping makeRoundRobin (const MultiFab& weight);
     static DistributionMapping makeSFC        (const MultiFab& weight, bool sort=true);
+    static DistributionMapping makeSFC        (const MultiFab& weight, Real& eff, bool sort=true);
```
mrowan137: It is sort of weird to return the efficiency for `makeSFC` since it doesn't incorporate any communications cost.  Just  a comment.

Src/Base/AMReX_DistributionMapping.cpp
```diff
@@ -1196,18 +1197,21 @@ DistributionMapping::SFCProcessorMapDoIt (const BoxArray&          boxes,
...
         Real sum_wgt = 0, max_wgt = 0;
         for (int i = 0; i < nteams; ++i)
         {
             const long W = LIpairV[i].first;
-            if (W > max_wgt)
-                max_wgt = W;
+            if (W > max_wgt) max_wgt = W;
             sum_wgt += W;
         }
+        *eff = (sum_wgt/(nteams*max_wgt));
```
atmyers: I think you need to check if `eff` is `nullptr` here before writing to the loction.

mrowan137: Thanks, I fixed the logic here.

Src/Base/AMReX_DistributionMapping.cpp
```diff
@@ -630,11 +630,18 @@ knapsack (const std::vector<long>&         wgts,
 
     std::sort(wblv.begin(), wblv.end());
 
+    // amrex::Print() << "wblv: ";
+    // for (auto const& x : wblv) {
+    //   amrex::Print() << x.weight() << " ";
+    // }
+    // amrex::Print() << "\n";
+    // amrex::Print() << "Doing a load balance" << "\n";
+    amrex::Print() << "efficiency (amrex): " << efficiency<< ", max_efficiency: "<< max_efficiency<<"\n";
```
atmyers: Could you remove this print statement?

mrowan137: It is removed.

atmyers:
Thanks!

Issue #688, opened 2020-02-06
Enhanced FlashFluxRegister methods


kweide:
Thank you @WeiqunZhang for adding the `amrex_flash_fluxregister` Fortran type and its `store`, `communicate`, and `load` methods.

Based on several use cases for flux register functionality in FLASH code, I would like to request three  enhancements to the functionality (roughly in order of priority):
1. Additional arithmetic (in `load`)
2. Area factors (in `store` and `load`)
3. `isFluxDensity` flags (in `store` and `load`)

I think these will be straightforward to implement (especially 1 and 2), and I am willing to try that and provide PRs. But you may have a different preferred way of doing something like this, and have much better understanding of performance or portability aspects, so I am going to write down the requirements (in the process also clarifying them for myself) and hope for comments or even code.

Details will follow in a series of comments.

kweide:
### Remarks (maybe return to this later)
* __Motivation:__ All the things requested could be done with the existing `load` and `store` methods, by pre- and postprocessing the fluxes appropriately. The main reason for wanting new features inside those methods: the `FlashFluxRegister` code keeps track of where fine-coarse boundaries occur, and `amrex_flash_fluxregister` cleverly uses this information to restrict its actions (especially in `load`) to exactly the right cell face locations. We have to keep track of this metainformation separately in FLASH code (or we have to repeatedly invoke auxiliary functions to enquire about it), if we want some actions only to be performed on fluxes on the fine side of a f-c boundary (as we do).
* Nothing should have to change in the `communicate` method.
* __Implementation__ could be done by just adding to the existing `store` and `load` methods, by introducing optional arguments, by providing variant methods under different names (together with Fortran `GENERIC`), or a combination thereof. A similar choice has to be made at the level of C++ methods. That's one reason I'm asking for input.

kweide:
### 0. Existing Functionality
(Inner code shown is for the 2D case.)

`call flux_reg%store(flux=src, flo, fhi, igrd, dir[, sf])` with partial interface
```Fortran
integer, intent(in) :: flo(*), fhi(*), igrd, dir  ! not repeated below
real, intent(in) :: flux(flo(1):fhi(1),flo(2):fhi(2),flo(3):fhi(3),flo(4):fhi(4))
``` 
leads to

```C++
     dest(i,j,0,n) = (src(2*i,2*j  ,0,n) +
                      src(2*i,2*j+1,0,n)) * (0.5*sf);
```


[...]

PR #728, opened 2020-03-04
Additional variant of FlashFluxRegister load method (untested)

kweide:
1. Add another variant of load method, amrex_flash_fluxregister_load_1_area,
   Since we will frequently use this for FLASH. This is for when we
   have area factors, but do not use an additional (cflux) input as in
   amrex_flash_fluxregister_load_2_area. It turns out we never need a
   separate scalar scaling factor in this case, so I also removed that.

2. Rename previous
     amrex_flash_fluxregister_load_area -> amrex_flash_fluxregister_load_2_area
   and
     amrex_fi_flash_fluxregister_load_area -> amrex_fi_flash_fluxregister_load_2_area

3. Fix: Replace sf_f -> my_sf_f, sf_c -> my_sf_c in a couple
   amrex_fi_flash_fluxregister_load_foo calls, as apparently intended.

### DIFF
--- a/Src/F_Interfaces/AmrCore/AMReX_FlashFluxRegister.H
+++ b/Src/F_Interfaces/AmrCore/AMReX_FlashFluxRegister.H
@@ -36,7 +36,7 @@ public:
     void store (int fine_global_index, int dir, FArrayBox const& fine_flux, FArrayBox const& area,
                 Real scaling_factor);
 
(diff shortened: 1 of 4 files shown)


kweide:
Cf. #688

@WeiqunZhang Does this look ok?

If yes, would you please approve the merge, and take it from there?

WeiqunZhang:
Looks good.  Thanks for fixing the bugs.

PR #734, opened 2020-03-06
conduit blueprint: add topo name option to particle container wrapper

cyrush:
This adds a new optional argument to ParticleContainerToBlueprint that allows you to explicitly set the topology named used to wrap a ParticleContainer.

This is necessary for codes like WarpX that can use multiple ParticleContainer to represent different species.


cyrush:
@atmyers thanks!

Issue #441, opened 2019-04-04
binary output format (HDF5?)


darylbond:
The native output format for AMReX generates an excessive number of files and uses non-binary format for the header files. This leads to unnecessary overhead in both storage and time when undertaking data management tasks. For example, it took over an hour for me to delete the results from a small AMR simulation (<80GB) simply due to the number of files involved. Transferring results to my local machine (rsync) also takes excessive amounts of time when compared to previous transfers of equivalent simulation data using the HDF5 output of Chombo.

While there is an example hdf5 output included in the source this has not been integrated into the core functionality and is thus only useful for specific use cases rather than as a general option for plot and checkpoint output. Using this example as a template I have implemented hdf5 plot output for my own use but this required modification of core AMReX files (AMReX_Amr and AMReX_AmrLevel).  Checkpointing is a somewhat more involved task although not particularly difficult to implement.

Essentially I am asking that HDF5 output (or similar) be implemented in the core code such that there is a built in option for single file binary output with checkpointing. If this already exists and I have simply missed it then please enlighten me!

asalmgren:
Daryl -- if I read your comment correctly you have implemented an improved
hdf plotfile writer -- would you be willing to share the source files so we
could look at integrating those into the repo?

We haven't had the resources to date to improve the hdf5 functionality but
we would certainly like to have better/faster I/O.

Thanks

darylbond:
I'm happy to share my work.

This should have all of the functionality for plot file writing as a hdf5 file. The hdf5 library wrapper is incomplete and a bit rough and has compound datatypes for use with the Chombo file format. I believe the hdf5 example in amrex builds these in a different manner. Also note that I encountered issues with writing attributes to file for nProcs > 2 when using the below options, I'm sure a HDF5 guru will know why.

```
        ret = H5Pset_all_coll_metadata_ops(file_access, true);
        ret = H5Pset_coll_metadata_write(file_access, true);
```

I also tried to keep all of the data structure stuff hidden from the user by getting them to just supply a MultiFab. That is, rather than getting them to override the writePlotFile() routine, get them to override a getPlotData() routine instead which just compiles a MultiFab. If the user wants to do anything else they can do it in pre or post write.

I'm sure you have your own ideas/constraints on how to implement a comprehensive binary I/O solution so please take the following and use as little or as much of it as you like.

[share.zip](https://github.com/AMReX-Codes/amrex/files/3041382/share.zip)

darylbond:
[...]

PR #745, opened 2020-03-23
HDF5 integration

darylbond:
Integrates HDF5 file operations into the plot and checkpoint methods. I have tried to implement these features unobtrusively and existing I/O should still work without modification.

The EB/CNS tutorial has been modified to provide an example and to serve as a test of the file I/O and restart capability.

Note that I have not made any changes to the hdf5 routines in `AMREX_PlotFileUtil` or `AMREX_ParticleIO` although all of the HDF5 data operations should probably be consolidated under `AMREX_HDF5` at some point. I am not advocating for my particular flavour of implementation, but having multiple implementations is clearly non-optimal.

See Issue #441 for the origin of this work.

### DIFF
--- a/Src/Amr/AMReX_Amr.H
+++ b/Src/Amr/AMReX_Amr.H
@@ -355,6 +355,9 @@ public:
   H5& getRestartHDF5() { return restart_h5; }
 #endif
 
+  std::string& getPlotType() {return plot_file_type;}
+  std::string& getCheckType() {return check_file_type;}
+
 protected:
 
     //! Initialize grid hierarchy -- called by Amr::init.
--- a/Src/Amr/AMReX_Amr.H
+++ b/Src/Amr/AMReX_Amr.H
@@ -12,12 +12,17 @@
 #include <AMReX_Array.H>
 #include <AMReX_Vector.H>
 #include <AMReX_BCRec.H>
+#include <AMReX_VisMF.H>
 
 #include <AMReX_AmrCore.H>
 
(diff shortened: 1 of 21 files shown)


atmyers:
Hi Daryl,

Thanks for this PR sharing your HDF5 work! 

I'd like to test this out and see how it compares to the HDF5 stuff in `AMReX_PlotFileUtils.H`. At first glance, this PR seems to be better integrated into AMRLevel, which I like. I think that ideally, we would reuse the the writers and readers in `AMReX_PlotFileUtil.H` and `AMReX_ParticleIO.H`, while maintaining the ability to easily get an HDF5 plotfile from an AMRLevel code, which I don't think is this case with our current capabilities. I'll give this a try and let you know how it goes.

atmyers:
Are you using a standard viz tool to look at the HDF5 plotfiles?

darylbond:
@atmyers I'm using [VisIt](https://wci.llnl.gov/simulation/computer-codes/visit/) for the HDF5 plotfiles

atmyers:
Hi Daryl,

I've tried out your pull request and it works well. I think I'd like to try merge your work with the existing HDF5 writers in `AMReX_PlotFileUtil.H` and 'AMReX_ParticleIO.H`. What I'll do is first merge your pull request onto an intermediate branch in the AMReX repo. Then, I'll work a bit on consolidating the two different writers. Later, we can merge the intermediate branch into development, and ultimately master. 

Thanks for your PR,
Andrew

PR #753, opened 2020-04-03
Cori-gpu make update for testing USE_MPI=TRUE USE_CUDA=FALSE

jmsexton03:
Added which_computer=cgpu
Added test in AMReX_Machines.cpp for false flag that assumes cgpu## instead of .*nid(\d+)
Moved flags for pthreads from Make.nersc to Make.defs
Add error message for cgpu build on cori login node

### DIFF
--- a/Src/Base/AMReX_Machine.cpp
+++ b/Src/Base/AMReX_Machine.cpp
@@ -306,6 +306,7 @@ class Machine
 
     std::string hostname;
     std::string nersc_host;
+    std::string cluster_name;
     std::string partition;
     std::string node_list;
     std::string topo_addr;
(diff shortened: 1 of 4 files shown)


maxpkatz:
Some notes: we have tested this to ensure that it works with OpenMPI, MVAPICH2, and MPICH. AMReX CNS still works, but this also allows Castro to work with OpenMPI (thanks to the addition of the -pthread fix in Make.defs).

The ban on building with USE_CUDA=TRUE on the Cori login nodes will save us a lot of user support trouble in the future, and in any case is required for at least MVAPICH2 as per the Cori docs.

PR #761, opened 2020-04-16
makeKnapSack sort argument

mrowan137:
This pull request adds optional sort (of process ids from least heavily to more heavily used), analogous to the `sort` argument in `makeSFC`.  By default, `sort=true` to keep prior code behavior.  The reason to add the sort argument is that if `makeKnapSack` is called from only one process on a multiprocess run, the code will stall due to `ParallelGather` in `LeastUsedCPUs` or `LeastUsedTeams`.  Setting `sort=false`, `makeKnapSack` could now be called from a single process on a multiprocess run (albeit without sorting).


Src/Base/AMReX_DistributionMapping.cpp
```diff
@@ -766,14 +767,24 @@ DistributionMapping::KnapSackDoIt (const std::vector<long>& wgts,
     Vector<Vector<int> > wrkerord;
 
     if (nteams == nprocs) {
-	LeastUsedCPUs(nprocs,ord);
-	wrkerord.resize(nprocs);
-	for (int i = 0; i < nprocs; ++i) {
-	    wrkerord[i].resize(1);
-	    wrkerord[i][0] = 0;
```
mrowan137, line 773: I believe this is unneeded because `wrkerord` is unused  in the remainder of the function for `nteams == nprocs`

PR #762, opened 2020-04-17
CI: macOS

ax3l:
Since more than half of our developers use macOS for development, its time to check how well we support native compilers.

Adds CMake checks.

### DIFF
--- a/.github/workflows/cmake/dependencies_mac.sh
+++ b/.github/workflows/cmake/dependencies_mac.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+#
+# Copyright 2020 Axel Huebl
+#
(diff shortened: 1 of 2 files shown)


.github/workflows/cmake/dependencies_mac.sh
```diff
@@ -0,0 +1,15 @@
...
+
+# search recursive inside a folder if a file contains tabs
+#
+# @result 0 if no files are found, else 1
+#
+
+set -eu -o pipefail
+
+brew update
+brew install openmpi
```
ax3l, line 15: Let's try for now to install dependencies with homebrew.
if this should be too unstable, we will find another solution. but I am using it in CI elsewhere and it's actually quite good.

PR #764, opened 2020-04-17
remove timestramp files from Tools

WeiqunZhang:
Could we remove these two files?


zingale:
That's probably my handywork originally.  I actually suspect that all of this is obsolete, since it was for the FBoxLib particle timestamp format.  Fine we me to remove.

zingale:
oh... I'm not an approving reviewer :(

PR #770, opened 2020-04-18
Add a CONTRIBUTING.md

atmyers:
This adds a file to the repo explaining how to use forks and pull requests to contribute to AMReX. 

The text heavily "borrows" from the corresponding file in WarpX.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -70,6 +70,9 @@ git clone --branch development https://github.com/<myGithubUsername>/amrex.git
 # Add a new remote for the main AMReX repo, to remain up-to-date.
 git remote add upstream https://github.com/AMReX-Codes/amrex
 
+# We recommend setting your development branch to track the upstream one instead of your fork:
+git branch -u upstream/development
+
 # For safety, disable pushing directly to upstream (only relevant if you have write access)
 git remote set-url --push upstream no_push
 ```
@@ -81,7 +84,7 @@ Now you are free to play with your fork (for additional information, you can vis
 > on your fork with
 > ```
 > git checkout development
-> git pull upstream development
+> git pull development
 > ```
 
 Make sure you are on the `development` branch with
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,111 @@
+# Contribute to AMReX
+
+Here the workflow we use for AMReX development.
+
+## Git workflow
+
+AMReX uses [git](https://git-scm.com) for version control. If you
+are new to git, you can follow one of these tutorials:
+- [Learn git with bitbucket](https://www.atlassian.com/git/tutorials/learn-git-with-bitbucket-cloud)
+- [git - the simple guide](http://rogerdudler.github.io/git-guide/)
+
+### Make your own fork and create a branch on it
+
+The basic workflow is:
+- fork the main repo (or update it if you already created it);
+- Implement your changes and push them on a new branch `<branch_name>` on
+your fork;
+- Create a Pull Request from branch `<branch_name>` on your fork to branch
+`development` on the main AMReX repository.
+
+First, let us setup your local git repo. Make your own fork of the main
+(`upstream`) repository:
+on the [AMReX Github page](https://github.com/atmyers/amrex), press the
+fork button. Then, you can execute:
+```
+# Clone your fork on your local computer. You can get this address on your fork's Github page.
+git clone --branch development https://github.com/<myGithubUsername>/amrex.git
+
+# Add a new remote for the main AMReX repo, to remain up-to-date.
+git remote add upstream https://github.com/AMReX-Codes/amrex
+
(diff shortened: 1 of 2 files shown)


atmyers:
Also, should we make `development` the default branch for amrex?

maxpkatz:
The problem with making development the default branch is that the default branch is what is shown to new users when they look at the code, but in general we want users to be on a stable release tag unless they know what they're doing.

CONTRIBUTING.md
```diff
@@ -0,0 +1,111 @@
...
+
+# For safety, disable pushing directly to upstream (only relevant if you have write access)
+git remote set-url --push upstream no_push
+```
+Now you are free to play with your fork (for additional information, you can visit the
+[Github fork help page](https://help.github.com/en/articles/fork-a-repo)).
+
+> Note: you do not have to re-do the setup above every time.
+> Instead, in the future, you need to update the `development` branch
+> on your fork with
```
maxpkatz, line 40: Perhaps we can consider recommending that they make their development branch on their local system point to `upstream` instead of their fork, so that they don't have to remember to add that `upstream` source branch every time (I know I would constantly forget).

zingale:
instead of [WIP], you could just make it a draft PR

maxpkatz:
Now that this will exist, I think we can merge the discussion of the git-flow model, and move the core developer discussion, into this file. Then the README can just tell you to check out CONTRIBUTING.md for information about development.

atmyers:
How many of our users actually use `master`, in practice? None of the codes I work on do.

maxpkatz:
Well, that's a valid point, and perhaps we just move to the GitHub flow model where we only have `master` and then we periodically tag it when we want to tag it. But as long as we're maintaining the git flow model I think `master` ought to be the default.

CONTRIBUTING.md
atmyers: Meaning, on the `development` branch, do `git branch -u upstream/development`

maxpkatz: Indeed. This would make your recommendation equivalent to Weiqun's (he advocated cloning upstream so you get the right development and master, and then pushing feature branches to one's fork), but yours is a little easier.

atmyers:
I'd kinda prefer that personally, but I think you're right that this a separate discussion.

CONTRIBUTING.md
atmyers: Okay, I agree and have made this change.

atmyers:
I also moved the discussion of the development model to CONTRIBUTING.md.

re: core contributors, it might be hard for people to rack up 100 commits now that we are going to squash-and-merge.

CONTRIBUTING.md
```diff
@@ -0,0 +1,172 @@
...
+
+Please DO NOT write large Pull Requests, as they are very difficult and
+time-consuming to review. As much as possible, split them into small,
+targeted PRs.
+For example, if find typos in the documentation open a pull request that only fixes typos.
+If you want to fix a bug, make a small pull request that only fixes a bug.
+If you want to implement a large feature, write helper functionality first, test it and submit those as a first pull request.
+If you want to implement a feature and are not too sure how to split it, just open an issue about your plans and ping other AMReX developers on it to chime in.
+
+Even before your work is ready to merge, it can be convenient to create a PR
```
maxpkatz, line 153: Agreed with @zingale that we should recommend Draft Pull Requests here as an alternative option.

zingale:
For core developers, maybe switch to a yt model where current core developers nominate someone and others chime in with support?

CONTRIBUTING.md
WeiqunZhang: I agree.

README.md
```diff
@@ -25,66 +25,7 @@ publicly and display publicly, and to permit other to do so.
...
-
-  * addition of a new algorithm / module  *and/or*
-
-  * substantial input into the code design process or testing
-
-If a core developer is inactive for multiple years, we may reassess their
-status as a core developer.
-
-The current list of core developers is: Ann Almgren (LBNL), Vince Beckner, John Bell (LBNL), Johannes Blaschke (LBNL), Cy Chan (LBNL), Marcus Day (LBNL), Brian Friesen (NERSC), Kevin Gott (NERSC), Daniel Graves (LBNL), Max Katz (NVIDIA), Andrew Myers (LBNL), Tan Nguyen (LBNL), Andrew Nonaka (LBNL), Michele Rosso (LBNL), Sam Williams (LBNL), Weiqun Zhang (LBNL), Michael Zingale (Stonybrook University).
+Please see CONTRIBUTING.md for details on how to contribute to AMReX.
```
WeiqunZhang, line 28: Could we make `CONTRIBUTING.md` clickable like `LICENSE`?

drummerdoc:
In the CONTRIBUTING.md the instructions are missing a couple things.  After the first clone, need to cd into the repo before running git commands.  Also, I need to run ``git fetch remote`` before I could set the upstream.

PR #778, opened 2020-04-20
Fix logic for 3D interpolation of cell centroids to face centroids

asalmgren:
My logic for how to avoid using covered cell data in the 3D interpolation from cell centroids to face centroids was wrong ... I believe this is correct.


Src/EB/AMReX_EBMultiFabUtil_3D_C.H
```diff
@@ -636,6 +636,10 @@ void eb_interp_centroid2facecent (Box const& bx,
   //
   amrex::Loop(ubx, ncomp, [=] (int i, int j, int k, int n) noexcept
     {
+      if (i == 3 && j == 9 && k == 0) 
+      {
+          std::cout << "DOING FX(3,9,0): APX = " << apx(i,j,k) << std::endl;
+      }
```
zingale, line 642: can you remove these prints?

PR #780, opened 2020-04-20
Launch macros simplification

WeiqunZhang:
Simplify GPU launch macros by calling GPU launch functions.

### DIFF
--- a/Src/Base/AMReX_GpuLaunchMacrosG.H
+++ b/Src/Base/AMReX_GpuLaunchMacrosG.H
@@ -439,399 +439,58 @@
 
 // FOR_1D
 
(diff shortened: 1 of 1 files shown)


kngott:
Should the macros / funcs separate files be removed? Just put the macro versions into the funcs file as well?

WeiqunZhang:
`AMRex_GpuLaunchMacrosG.H` still has more than 500 lines.  I don't putting them into `AMReX_GpuLaunchFunctsG.H`, which has about 1000 lines, will help anything.

kngott:
Oh, I see. The `LAMBDA_RANGE` version don't have a function equivalent. Okay then.

PR #782, opened 2020-04-20
Bugfix after PR#780

atmyers:
I need to use `return` instead of `continue` to break out of AMREX_PARALLEL_FOR, which does not work now that the macro calls the function version of `ParallelFor`.

### DIFF
--- a/Src/Particle/AMReX_ParticleCommunication.H
+++ b/Src/Particle/AMReX_ParticleCommunication.H
@@ -158,11 +158,13 @@ struct ParticleCopyPlan
                 AMREX_FOR_1D ( num_copies, i,
                 {
                     int dst_box = p_boxes[i];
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
I wonder if one should avoid `return` and `continue`.  `return` might give a wrong message indicating return from the function.

atmyers:
See update. I'm probably doing this elsewhere, but will change that in a separate PR.

maxpkatz:
Kind of like with `go to`, sometimes it's just easier and cleaner to have an early return than to have complicated conditional branching. I think we'll just have to get used to the new normal where we are mostly working in lambdas.

atmyers:
I think that in lambdas it's sometimes better to return early, but with the launch macros, where you can't see the explicit lambda declaration, it's confusing to have a return statement.

PR #789, opened 2020-04-21
add host device for functions

WeiqunZhang:
We have discussed that we could use type traits to overload `ParallelFor`.  However,  there is no distinction between host device and device lambda in DPC++.  It's also awkward to use optional template parameter or function parameter.  So I have chosen to add a set of functions with different names.

### DIFF
--- a/Src/Base/AMReX_GpuLaunchFunctsG.H
+++ b/Src/Base/AMReX_GpuLaunchFunctsG.H
@@ -997,6 +997,147 @@ void For (Box const& box1, T1 ncomp1, L1&& f1,
                 box3,ncomp3,std::forward<L3>(f3));
 }
 
+template <typename T, typename L, typename M=amrex::EnableIf_t<std::is_integral<T>::value> >
+void HostDeviceParallelFor (T n, L&& f) noexcept
+{
+    if (Gpu::inLaunchRegion()) {
+        ParallelFor(n,std::forward<L>(f));
+    } else {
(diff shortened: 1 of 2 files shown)


WeiqunZhang:
There will be another PR to make the compilation fail if wrong types of lambdas are passed to the parallel for family of functions.

Src/Base/AMReX_GpuLaunchFunctsG.H
```diff
@@ -997,6 +997,146 @@ void For (Box const& box1, T1 ncomp1, L1&& f1,
                 box3,ncomp3,std::forward<L3>(f3));
 }
 
+template <typename T, typename L, typename M=amrex::EnableIf_t<std::is_integral<T>::value> >
+void HostDeviceParallelFor (T n, L&& f) noexcept
+{
+    if (Gpu::inLaunchRegion()) {
+        ParallelFor(n,std::forward<L>(f));
+    } else {
+        for (T i = 0; i < n; ++i) f(i);
```
atmyers, line 1006: since this is `ParallelFor`, do we want `AMREX_PRAGMA_SIMD` here for the CPU version?

WeiqunZhang: Yes, I will modify it.

WeiqunZhang: updated

PR #790, opened 2020-04-21
Parallel for type traits

WeiqunZhang:


Src/Base/AMReX_TypeTraits.H
```diff
@@ -77,23 +77,47 @@ namespace amrex
...
-#endif
+
+#elif defined(AMREX_USE_HIP)
+
+    // xxxxx HIP todo
+    template <class T, class Enable = void>
+    struct HostRunnable : std::true_type {};
+
+    template <class T, class Enable = void>
+    struct DeviceRunnable : std::true_type {};
```
maxpkatz, line 112: Won't this break compilation for HIP and DPC++ since e.g. HostDeviceParallelFor now uses HostDeviceRunnable but that is only defined for CUDA?

WeiqunZhang: Yes, you are right.  I have update the PR moving `HostDeviceRunnable` out of `AMREX_USE_CUDA` into `AMREX_USE_GPU` block.

maxpkatz:
Thanks, this is very useful!

PR #793, opened 2020-04-22
Support for DPCPP USM prefetch and mem_advise

abagusetty:
1. Adds support for DPCPP USM's `prefetch()`
2. Support for DPCPP USM's `mem_advise()`
3. Adding a couple of device_prop attributes to DPCPP mode to mimic CUDA's `memoryManaged` and `concurrentManagedAccess`

### DIFF
--- a/Src/Base/AMReX_BaseFab.H
+++ b/Src/Base/AMReX_BaseFab.H
@@ -1647,9 +1647,10 @@ void
 BaseFab<T>::prefetchToHost () const noexcept
 {
 #if defined(AMREX_USE_DPCPP)
-    std::size_t s = sizeof(T)*this->nvar*this->domain.numPts();
-    auto& q = Device::streamQueue();
-    q.submit([&] (sycl::handler& h) { h.prefetch(this->dptr, s); });
+    // xxxxx DPCPP todo: prefetchToHost
+    // std::size_t s = sizeof(T)*this->nvar*this->domain.numPts();
+    // auto& q = Gpu::Device::streamQueue();
+    // q.submit([&] (sycl::handler& h) { h.prefetch(this->dptr, s); });
 #elif defined(AMREX_USE_CUDA)
     std::size_t s = sizeof(T)*this->nvar*this->domain.numPts();
     AMREX_CUDA_SAFE_CALL(cudaMemPrefetchAsync(this->dptr, s,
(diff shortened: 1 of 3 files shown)


Src/Base/AMReX_GpuDevice.cpp
```diff
@@ -740,6 +744,12 @@ Device::mem_advise_set_preferred (void* p, const std::size_t sz, const int devic
     {
         AMREX_CUDA_SAFE_CALL(cudaMemAdvise(p, sz, cudaMemAdviseSetPreferredLocation, device));
     }
+#elif defined(AMREX_USE_DPCPP)
+    if (device_prop.managedMemory == 1 && device_prop.concurrentManagedAccess == 1)
+    {
+        auto& q = streamQueue();
+        q.mem_advise(p, sz, PI_MEM_ADVICE_SET_PREFERRED_LOCATION);
```
WeiqunZhang, line 751: When I compile this, I get,
```
error: no member named 'mem_advise' in 'cl::sycl::ordered_queue'
        q.mem_advise(p, sz, PI_MEM_ADVICE_SET_PREFERRED_LOCATION);
```

Src/Base/AMReX_BaseFab.H
```diff
@@ -1647,7 +1647,9 @@ void
 BaseFab<T>::prefetchToHost () const noexcept
 {
 #if defined(AMREX_USE_DPCPP)
-    // xxxxx DPCPP todo: how to prefetchToHost
+    std::size_t s = sizeof(T)*this->nvar*this->domain.numPts();
+    auto& q = Device::streamQueue();
+    q.submit([&] (sycl::handler& h) { h.prefetch(this->dptr, s); });
```
WeiqunZhang, line 1652: Does this prefetch from device to host or from host to device?  Note that `q` is built with `sycl::gpu_selector`.

Src/Base/AMReX_GpuDevice.cpp
abagusetty: I believe this error is because of this issue https://github.com/intel/llvm/pull/1035. Taking a look at it.

WeiqunZhang: It seems that the `in_order` property is not available in beta5 yet.  Otherwise we could switch to use `queue` with `in_order` property.

abagusetty: Yes and it is also not available in any internal beta versions yet. Do you suggest to comment out the `mem_advise` for DPCPP block for time being. As with `queue q{property::queue::in_order()};` we will also have the support for `mem_advise` in the future. Please let me know.

WeiqunZhang: That sounds good.  Please comment out the `mem_advise` code for now.  Please also add a comment with words "xxxxx DPCPP todo".  I use that for grep.

Src/Base/AMReX_BaseFab.H
abagusetty: Thanks for pointing it out. `prefetch to host` is not yet fully working when the queue or handler is based on device. I have disabled it for now with a todo note.

PR #794, opened 2020-04-22
CMake: get rid of zero size arrays in AMReX_buildInfo.cpp

mic84:


atmyers:
In the future could you make a new branch for each PR?

PR #799, opened 2020-04-24
Call the same code for building neighbor lists whether we are on the CPU or GPU

atmyers:
This is part of a broader project to have the same implementation for many cpu / gpu particles functions. 

Edit: ready for review.


ajnonaka:
All FHDeX tests pass using amrex branch atmyers:consolidation with FHDeX branch atmyers/consolidation

PR #803, opened 2020-04-24
EnableIf for ParticleContainer plotfile methods

atmyers:
This fixes an issue @JBlaschke was seeing with certain overloads.


WeiqunZhang:
Could you give more details on the issue?

atmyers:
There are currently a lot of overloads of `WritePlotFile` that let you toggle on or off writing various components, specify component names, and filter a subset of the particles for output. The number of arguments alone is not enough to distinguish these. For example, without the EnableIf, the compiler doesn't know whether you mean [this](https://github.com/AMReX-Codes/amrex/blob/6d458300dbf28e9796db065edd2b7d7442dfbabb/Src/Particle/AMReX_Particles.H#L815) or [this](https://github.com/AMReX-Codes/amrex/blob/6d458300dbf28e9796db065edd2b7d7442dfbabb/Src/Particle/AMReX_Particles.H#L856) when you call the three-component version.

WeiqunZhang:
I will look at the `IsVector` stuff, but I think a simpler solution for this type of issue might be
```
template <>
void WritePlotFile (const std::string& dir, const std::string& name,
                        const Vector<std::string>& real_comp_names) const;
```

WeiqunZhang:
Or even better would to enable_if on F has certain property.

atmyers:
> Or even better would to enable_if on F has certain property.

I tried that first, but it turns out not to be easy in `< C++17` to come up with something that will work for a general callable.

WeiqunZhang:
https://stackoverflow.com/questions/51187974/can-stdis-invocable-be-emulated-within-c11

WeiqunZhang:
But it seems that the solution above requires C++14.

atmyers:
See the note about requiring this though: https://cplusplus.github.io/LWG/issue2132

We currently don't require C++14, so I'm not sure that solution will work unless we change that. 

Let me try your first suggestion.

JBlaschke:
> But it seems that the solution above requires C++14.

@WeiqunZhang I did test with c++14 on gcc-9.3

atmyers:
> But it seems that the solution above requires C++14.

Yeah, crossed messages with you.

WeiqunZhang:
If you want to go the not vector way, I would suggest `std::is_same` instead of introducing `IsVector`.  My reasoning for that is `IsVector` is different from `IsBaseFab` because `FArrayBox` is a `BaseFab`. But I don't think there is a similar case for `Vector`.

atmyers:
I am okay with using `std::is_same` instead of `IsVector`.

@JBlaschke could you please verify that this still fixes your issue?

JBlaschke:
> I am okay with using `std::is_same` instead of `IsVector`.
> 
> @JBlaschke could you please verify that this still fixes your issue?

@atmyers works for me (I just pulled the latest version of this branch)

WeiqunZhang:
https://godbolt.org/z/ujd8LA
You can also do something like this.  The only thing there that is not C++11 is `enable_if_t`.  You can use `amrex::EnableIf_t`.

PR #804, opened 2020-04-24
make ParallelFor work for functors

WeiqunZhang:
Unfortunately #790 broke WarpX because it passes functors to `ParallelFor`.  If we want to support both lambda and functor, the only thing we can be sure is CUDA extended device lambda does not run on host.


atmyers:
So if we support functors, the only thing we can really catch is if you mark a callable as `AMREX_GPU_DEVICE`, but try to run it on the host?

WeiqunZhang:
I think that's right.

WeiqunZhang:
More specifically it can only catch running cuda extended device lambda on host.

PR #805, opened 2020-04-25
Add `makeKnapSack` and `makeSFC` overloads with alternate to `ParallelAllReduce::Sum`

mrowan137:
This PR adds overloads to `makeKnapSack` and `makeSFC` functions.  The communications is different for these, compared to the existing versions: instead of `ParallelAllReduce::Sum` to give all ranks the same global vector of cost (which potentially sends many unneeded values), the new overloads gather costs to just one rank, collecting only the values needed from each rank.  The processor map for the new distribution mapping can (optionally) be computed only on a selected rank, which can save a broadcast in the case that the newly computed distribution mapping is not an improvement.


Src/Base/AMReX_DistributionMapping.cpp
```diff
@@ -1410,6 +1410,183 @@ DistributionMapping::makeKnapSack (const Vector<Real>& rcost, Real& eff, int nma
...
+    //             &root_recvbuf[0], root_recvcount,
+    //             root)
+    amrex::Vector<int> index_to_send = rcost_local.IndexArray();
+    amrex::Vector<Real> cost_to_send;
+    cost_to_send.resize(rcost_local.local_size());
+    
+    for (auto i : rcost_local.IndexArray())
+    {   
+        cost_to_send[rcost_local.localindex(i)] = rcost_local[i];
+    }
```
WeiqunZhang, line 1447: Both `locaindex(i)` and `rcost_local[i]` needs to do a bisection.  So I think it's more efficient to use `MFIter` and `push_back`.

```suggestion
    cost_to_send.reserve(rcost_local.local_size());
    
    for (MFIter mfi(rcost_local); mfi.isValid(); ++mfi
    {   
        cost_to_send.push_back(rcost_local[mfi]);
    }
```
This suggestion applied to the other function too.

Src/Base/AMReX_DistributionMapping.cpp
```diff
@@ -1410,6 +1410,183 @@ DistributionMapping::makeKnapSack (const Vector<Real>& rcost, Real& eff, int nma
...
+
+    amrex::Vector<long> recvcount, disp;
```
WeiqunZhang, line 1449: Can we use `Long` instead?  It's an alias to `long` on Unix and `long long` on Windows.

Src/Base/AMReX_DistributionMapping.cpp
```diff
@@ -1410,6 +1410,183 @@ DistributionMapping::makeKnapSack (const Vector<Real>& rcost, Real& eff, int nma
...
+    recvcount.resize(ParallelDescriptor::NProcs(), 0);
+    disp.resize(ParallelDescriptor::NProcs(), 0);
+
+    long n_cost = cost_to_send.size();
+    ParallelDescriptor::Gather(&n_cost, 1,
+                               &recvcount[0], 1,
+                               root);
```
WeiqunZhang, line 1456: There is no need to gather (or rather it should be allgather).  `DistributionMapping` has all the information we need.  Some like
```
const Vector<int>& old_pmap = rcost_local.DistributionMapping().ProcessorMap();
for (int rank : old_pmap) {
    ++recvcount[rank];
}
```

Src/Base/AMReX_DistributionMapping.cpp
```diff
@@ -1410,6 +1410,183 @@ DistributionMapping::makeKnapSack (const Vector<Real>& rcost, Real& eff, int nma
...
+        
+    for (int i=1; i<disp.size(); i++)
+    {
+        disp[i] = disp[i-1] + recvcount[i-1];
+    }
```
WeiqunZhang, line 1461: We could use `std::partial_sum`.

Src/Base/AMReX_DistributionMapping.cpp
```diff
@@ -1410,6 +1410,183 @@ DistributionMapping::makeKnapSack (const Vector<Real>& rcost, Real& eff, int nma
...
+                                recvcount,
+                                disp,
+                                root);
+
+    ParallelDescriptor::Gatherv(&index_to_send[0],
+                                index_to_send.size(),
+                                &new_index_to_old_index[0],
+                                recvcount,
+                                disp,
+                                root);
```
WeiqunZhang, line 1480: This `gatherv` is not needed because `LayoutData`'s `DistributionMapping` contains all the information we need.

Src/Base/AMReX_DistributionMapping.cpp
```diff
@@ -1410,6 +1410,183 @@ DistributionMapping::makeKnapSack (const Vector<Real>& rcost, Real& eff, int nma
...
+DistributionMapping::ComputeDistributionMappingEfficiency (const DistributionMapping& dm,
+                                                           const Vector<Real>& cost,
+                                                           Real* efficiency)
+{
+    const Real nprocs = ParallelDescriptor::NProcs();
+
+    // This will store mapping from processor to the costs of FABs it controls,
+    // (proc) --> ([cost_FAB_1, cost_FAB_2, ... ]),
+    // for each proc
+    std::map<int, Vector<Real>> rankToCosts;
```
WeiqunZhang, line 1566: Maybe we can use `Vector<Vector<Real> > rankToCosts(nprocs);`.

Src/Base/AMReX_DistributionMapping.cpp
```diff
@@ -1410,6 +1410,183 @@ DistributionMapping::makeKnapSack (const Vector<Real>& rcost, Real& eff, int nma
...
+    for (int i=0; i<cost.size(); ++i)
+    {
+        rankToCosts[dm[i]].push_back(cost[i]);
+    }
+
+    Real maxCost = -1.0;
+
+    // This will store mapping from (proc) --> (sum of cost) for each proc
+    Vector<Real> rankToCost = {0.0};
+    rankToCost.resize(nprocs);
```
WeiqunZhang, line 1577: It seems that this can be `Vector<Real> rankToCost(nprocs);`.

Src/Base/AMReX_DistributionMapping.cpp
```diff
@@ -1641,6 +1818,145 @@ DistributionMapping::makeSFC (const Vector<Real>& rcost, const BoxArray& ba, Rea
     return r;
 }
 
+DistributionMapping
+DistributionMapping::makeSFC (const LayoutData<Real>& rcost_local,
```
WeiqunZhang, line 1822: The first part of this function is very similar to the first part of the first new function.  Maybe this part can be put into a new function.

Src/Base/AMReX_DistributionMapping.cpp
mrowan137: Long works, switched to that.

Src/Base/AMReX_DistributionMapping.cpp
mrowan137: Changed to this.

Src/Base/AMReX_DistributionMapping.cpp
mrowan137: Thanks, switched to that.

Src/Base/AMReX_DistributionMapping.cpp
mrowan137: Switched to that.

Src/Base/AMReX_DistributionMapping.cpp
mrowan137: Got it, switched to this.  Saves another expensive operation!

Src/Base/AMReX_DistributionMapping.cpp
mrowan137: I've added a function ParallelDescriptor::GatherLayoutData which collections LayoutData local values to a vector on root process, this eliminate the code repetition.

Src/Base/AMReX_DistributionMapping.cpp
mrowan137: Good point!  I did not notice this initially, it is now changed to get the information from the distribution mapping.

Src/Base/AMReX_ParallelDescriptor.H
```diff
@@ -1005,6 +1013,79 @@ ParallelDescriptor::Gatherv (const T* send, Long sc,
...
+#ifdef BL_USE_MPI
+    // Gather prelims
+    Vector<int> index_to_send = sendbuf.IndexArray();
+    Vector<T> T_to_send;
+    T_to_send.reserve(sendbuf.local_size());
+    
+    for (auto i : sendbuf.IndexArray())
+    {
+        T_to_send.push_back(sendbuf[i]);
+    }
```
mrowan137, line 1032: I switched back to the way as written from original revision you suggested:

> for (MFIter mfi(sendbuf); mfi.isValid(); ++mfi)
> {   
>     T_to_send.push_back(sendbuf[mfi]);
> }

after moving the function to AMReX_ParallelDescriptor.H.  Is this way OK, or do you recommend that we make the necessary changes to allow using the MFIter loop?

Src/Base/AMReX_DistributionMapping.cpp
mrowan137: I added this change initially, but then switched to something slightly different (see further below)- is the change OK?

Src/Base/AMReX_ParallelDescriptor.H
```diff
@@ -1005,6 +1013,79 @@ ParallelDescriptor::Gatherv (const T* send, Long sc,
...
+
+    int nprocs = ParallelContext::NProcsSub();
```
WeiqunZhang, line 1034: Let's change this to `ParalleDescriptor::NProcs()`, because this function is meant to use the global communicator.

Src/Base/AMReX_ParallelDescriptor.H
```diff
@@ -1005,6 +1013,79 @@ ParallelDescriptor::Gatherv (const T* send, Long sc,
...
+    Vector<Long> recvcount(nprocs), disp(nprocs);
```
WeiqunZhang, line 1035: We need to initialize `recvcount` to zero.  Maybe `Vector<Long> recvcount(nprocs,0), disp(nprocs);`.

Src/Base/AMReX_ParallelDescriptor.H
```diff
@@ -1005,6 +1013,79 @@ ParallelDescriptor::Gatherv (const T* send, Long sc,
...
+
+    const Vector<int>& old_pmap = sendbuf.DistributionMap().ProcessorMap();
+    for (int rank : old_pmap)
+    {
+        ++recvcount[rank];
+    }
+
+    std::partial_sum(recvcount.begin(), --recvcount.end(), ++disp.begin());
```
WeiqunZhang, line 1043: I would use `std::partial_sum(recvcount.begin(), recvcount.back(), disp.begin()+1);`

WeiqunZhang: And `disp[0] = 0;` before this line.

Src/Base/AMReX_ParallelDescriptor.H
```diff
@@ -1005,6 +1013,79 @@ ParallelDescriptor::Gatherv (const T* send, Long sc,
...
+                                &new_index_to_T[0],
+                                recvcount,
+                                disp,
+                                root);
+
+    Vector<int> new_index_to_old_index(sendbuf.size());
+    std::iota(new_index_to_old_index.begin(), new_index_to_old_index.end(), 0);
+    
+    // stable_sort to prevent reorderings of index with same rank
+    std::stable_sort(new_index_to_old_index.begin(), new_index_to_old_index.end(),
```
WeiqunZhang, line 1057: Why do we need to sort?

WeiqunZhang: To clarify, I understand how you do this with sort.  However, we could avoid sort because it's expensive.  Like in your previous implementation, we could instead build `Vector<Vector<int> >`, where the outer vector has the size of `nprocs`.  When we loop over `old_pmap`, we can build a vector like `IndexArray()` for each process.

Src/Base/AMReX_ParallelDescriptor.H
mrowan137: Got it, switched to that.

Src/Base/AMReX_ParallelDescriptor.H
mrowan137: `recvcount` initializes to zero now.

Src/Base/AMReX_ParallelDescriptor.H
mrowan137: Switched to this.

Src/Base/AMReX_ParallelDescriptor.H
mrowan137: Thanks for catching this again, that is a better way to do it.

PR #807, opened 2020-04-26
Fix NeighborParticlesTest

JBlaschke:
Simple fix -- probably to do with shifting APIs

1. Added `lev` when accessing `m_neighbor_list`


atmyers:
Thanks!

PR #808, opened 2020-04-26
Async I/O

kngott:
Async I/O
========
Infrastructure for multiple simultaneous asynchronous outputs of MultiFabs and Plotfiles. Should be easily extendable to other outputs, app specific outputs, and under useful circumstances, inputs (e.g. time-varying BCs, stored in input files?) Includes tests, confirmed working on CPU and GPU using fcompare and diffmultifab, and example integration into GPU HeatEquation tutorial.

Requires MPI_THREAD_MULTIPLE. Uses static comm pattern for now: simplest possible MPI comm pattern for minimal THREAD_MULTIPLE overhead. Dynamic I/O should be possible, with additional work and study, but good load-balancing may alleviate the need.

Decisions we should make before merging:
===============================
* Decide on nOutFiles methodology. Previously, looks like it was fully adjustable, maybe even expected to be set immediately before each Write with VisMF::SetNOutFiles. Also initialized in AMReX_Amr.cpp with inputs file. Async I/O makes this call much heavier, so clean pattern should be decided on.
* Write function interface: e.g. 3 writes: standard "Write*" function that adjusts based on build flag, "WriteAsync*" and "WriteSync*" functions for fixed write types regardless?
* Change future return type? WriteAsyncStatus seems unnecessarily heavy. Just nbytes?

Important:
========
* CMake -- added MPI_MULTIPLE build flag.
* Documentation
* Async Particle PlotFile - Build with Andrew.
* Test on CPU with OpenMP threading. (CHECK)
* Add tag to the lambda function for future portability.
* `Vector<MultiFab>` print.

Performance:
==========
* Move "Header" write into Plotfile lambda.
* Test moving min/max calc into lambda? (Is THREAD_MULTIPLE overhead negligible?)
* Thread the Async I/O based on Brian and Vince's design: (https://github.com/AMReX-Codes/amrex/blob/master/Tests/IOBenchmark/IOTestDriver.cpp)


drummerdoc:
Very cool...bonus points for fancy formatting above!

What is MPI_THREAD_MULTIPLE and how would a user know to turn that on?

Why is the CMake flag different (MPI_MULTIPLE vs MPI_THREAD_MULTIPLE)?

If I have a section of code calling MultiFab::Write that under the covers is async, do I have to worry about pointers leaving scope?  That is, do I need an explicit sync somewhere afterward?

kngott:
* The fancy formatting was an accident. I never looked at it formatted. Guess I'm just naturally stylish. 😉

* THREAD_MULTIPLE is one of the MPI initialization options. Best description I ever heard was it's your "contract" with the MPI library: you guarantee you'll interact with the threading in a specific way. If you don't follow your chosen contract, the output is no longer guaranteed correct. 

THREAD_MULTIPLE means each thread can make unique, independent MPI calls. Sounds good in principle, but it has to introduce additional thread locks to guarantee accuracy, so most MPI_THREAD_MULTIPLE implementations add overhead. You shouldn't turn it on unless you have a good reason (based on Summit tests, I expect Async I/O will always be a win. Other systems & MPI types may behave different, but I'm seeing a LOT of runtime improvement, so I'm hopeful).

In this specific case, the user will know they need to turn it on because of the documentation we'll (read `I'll`) write when we're happy with it. 😄 If apps let AMReX initialize MPI, the MPI_MULTIPLE flag will do it. For apps that initialize MPI themselves, they'd have to adjust to use Async I/O (apps could just copy it right out of ParallelDescriptor::Initialize. It's a very standardized method at this point.)

* No reason why it's called MPI_MULTIPLE. I tested all MPI_Thread_Init options for overhead, and MULTIPLE was easier to type when I was typing 4 different long names at every #ifdef. I'm not against any name changes anywhere.

* In this implementation, you do not need any additional explicit synchronizations anywhere, but they are available if you want additional control (e.g. you make your own branching, asynchronous CPU work). The built-in lambda capture makes a copy of the data to write, so everything the write needs is in scope and properly deleted, and there are `WaitAll` calls, which finishes all remaining writes, both in AMReX::Finalize, to finish writing before the code ends and in BLBacktrace, to finish writing even if the app throws some types of errors (both tested and working).

What you will have to watch out for is your use of threads on the CPU. Async I/O spawns C++ threads to do the writing, so you need to make sure if you're running on the CPU, you don't oversubscribe threads through OpenMP. (This reminds me, I need to test on the CPU with OpenMP on.)

Src/Base/AMReX_BLBackTrace.cpp
```diff
@@ -32,6 +33,12 @@ std::stack<std::pair<std::string, std::string> >  BLBackTrace::bt_stack;
 void
 BLBackTrace::handler(int s)
 {
+
+#ifdef AMREX_MPI_MULTIPLE
+    amrex::Print() << "Aborting. Completing Asynchronous Prints." << std::endl;
+    VisMF::asyncWaitAll();
+#endif
+
```
WeiqunZhang, line 41: I think this should be moved after `signal(s,SIG_DFL)` in case `asyncWaitAll` raises a signal too.  And maybe print statement should be moved into `asyncWaitAll` so that we don't have to print if there is nothing in the asyn i/o queue or verbosity is off.

Src/Base/AMReX_ParallelDescriptor.cpp
```diff
@@ -305,7 +305,23 @@ ParallelDescriptor::StartParallel (int*    argc,
...
+        int requested = MPI_THREAD_MULTIPLE;
+        int provided = -1;
+
+        MPI_Init_thread(argc, argv, requested, &provided);
+
+        if (provided < requested)
+        {
+            std::cout << "MPI provided < requested: " + std::to_string(provided) + 
+                                                " < " + std::to_string(requested);
+            abort();
```
WeiqunZhang, line 319: We should add `std::`.

Src/Base/AMReX_ParallelDescriptor.cpp
```diff
@@ -305,7 +305,23 @@ ParallelDescriptor::StartParallel (int*    argc,
...
+
+#ifdef AMREX_MPI_MULTIPLE
+        int requested = MPI_THREAD_MULTIPLE;
+        int provided = -1;
+
+        MPI_Init_thread(argc, argv, requested, &provided);
+
+        if (provided < requested)
+        {
+            std::cout << "MPI provided < requested: " + std::to_string(provided) +
```
WeiqunZhang, line 317: This works of course.  But why not simply `std::cout << "xxx " << int << "yyy "?

Src/Base/AMReX_ParallelDescriptor.cpp
```diff
@@ -339,6 +355,15 @@ ParallelDescriptor::StartParallel (int*    argc,
...
 
+#ifdef AMREX_MPI_MULTIPLE
+    int provided = -1;
+    MPI_Query_thread(&provided);
+    amrex::Print() << "MPI Thread Initialized with support level " << provided << "." << std::endl;
+
+#else
+    amrex::Print() << "MPI Initialized." << std::endl;
+#endif
+
```
WeiqunZhang, line 366: In `amrex::Initialize`, we already print mpi information.  So maybe this can be merged into there.  And it should not print unless verbosity is on for compliance with XSDK.

Src/Base/AMReX_ParallelDescriptor.cpp
```diff
@@ -397,6 +422,30 @@ ParallelDescriptor::Barrier (const MPI_Comm &comm, const std::string &message)
...
 }
 
+
+ParallelDescriptor::Message
+ParallelDescriptor::Abarrier ()
+{
+    MPI_Request req;
+    BL_MPI_REQUIRE( MPI_Ibarrier(ParallelDescriptor::Communicator(), &req) );
+
+    // Use a char/(byte) as a faux-type for compatibility.
```
WeiqunZhang, line 432: Can it be MPI_DATATYPE_NULL?

Src/Base/AMReX_VisMF.H
```diff
@@ -382,6 +421,17 @@ std::ostream& operator<< (std::ostream& os, const VisMF::Header& hd);
...
 std::ostream& operator<< (std::ostream& os, const WriteAsyncStatus& status);
 
+template <typename Func, typename... Params>
+inline void VisMF::asyncAddWrite (Func&& f, Params&&... args)
+{
+#ifdef AMREX_MPI_MULTIPLE
+    asyncWaitUntilFree();
+
+    current_comm = ((current_comm + 1) % nAsyncWrites);
+    future_list.push( std::async(std::launch::async, f, async_comm[current_comm], std::forward<Params>(args)...) );
```
WeiqunZhang, line 431: We can `forward` `f` as well.

Src/Base/AMReX_VisMF.cpp
```diff
@@ -86,20 +95,81 @@ VisMF::Initialize ()
...
 }
 
 void
 VisMF::Finalize ()
 {
+
+#ifdef AMREX_MPI_MULTIPLE
+    VisMF::asyncWaitAll();
+
+    for (int i=0; i<nAsyncWrites; ++i)
```
WeiqunZhang, line 115: `nAsyncWrites` -> `async_comm.size()`?

Src/Base/AMReX_VisMF.cpp
```diff
@@ -86,20 +95,81 @@ VisMF::Initialize ()
...
+
+#ifdef AMREX_MPI_MULTIPLE
+
+    VisMF::asyncWaitAll();
+    current_comm = 0;
+
+    // Function is being recalled, so free MPI objects. 
+    if (initialized)
+    {
+        for (int i=0; i<nAsyncWrites; ++i)
```
WeiqunZhang, line 146: If the user calls SetNOutFiles with the same `noutfiles`, are there any reasons to free the comms and recreate them?

Src/Base/AMReX_VisMF.cpp
```diff
@@ -2471,6 +2518,828 @@ VisMF::WriteAsync (const FabArray<FArrayBox>& mf, const std::string& mf_name)
...
+    int ifile = std::get<0>(myinfo);   // file #
+    int ispot = std::get<1>(myinfo);   // spot #
+
+    constexpr int sizeof_int64_over_real = sizeof(int64_t) / sizeof(Real);
+    const int n_local_fabs = mf.local_size();
+    const int n_global_fabs = mf.size();
+    const int ncomp = mf.nComp();
+    const long n_fab_reals = 2*ncomp;
+    const long n_fab_int64 = 1;
+    const long n_fab_nums = n_fab_reals*sizeof_int64_over_real + n_fab_int64;
```
WeiqunZhang, line 2567: It should be `/sizeof_int64_ove_real` instead of `*sizeof_int64_over_real`.

Src/Base/AMReX_VisMF.cpp
```diff
@@ -2471,6 +2518,828 @@ VisMF::WriteAsync (const FabArray<FArrayBox>& mf, const std::string& mf_name)
...
+
+        for (int icomp = 0; icomp < ncomp; ++icomp) {
+            if (runon == RunOn::Host)
+            {
+                Real cmin = fab.min<RunOn::Host>(bx,icomp);
+                Real cmax = fab.max<RunOn::Host>(bx,icomp);
+                std::memcpy(pld, &cmin, sizeof(Real));
+                pld += sizeof(Real);
+                std::memcpy(pld, &cmax, sizeof(Real));
+                pld += sizeof(Real);
```
WeiqunZhang, line 2614: Maybe we can move this and the similar code in `else if` out of the if block to remove duplication.

Src/Base/AMReX_VisMF.cpp
```diff
@@ -2471,6 +2518,828 @@ VisMF::WriteAsync (const FabArray<FArrayBox>& mf, const std::string& mf_name)
...
+
+    auto myinfo = StaticWriteInfo(myproc);
+    int ifile = std::get<0>(myinfo);   // file #
+    int ispot = std::get<1>(myinfo);   // spot #
+
+    constexpr int sizeof_int64_over_real = sizeof(int64_t) / sizeof(Real);
+    const int n_local_fabs = mf.local_size();
+    const int n_global_fabs = mf.size();
+    const int ncomp = mf.nComp();
+    const long n_fab_reals = 2*ncomp;
```
WeiqunZhang, line 2565: Let's use `Long`.

Src/Base/AMReX_VisMF.cpp
```diff
@@ -2471,6 +2518,828 @@ VisMF::WriteAsync (const FabArray<FArrayBox>& mf, const std::string& mf_name)
...
+        Vector<MPI_Request> waiting_reqs(ispot);
+        Vector<MPI_Status> waiting_status(ispot);
+        Vector<MPI_Request> syncing_reqs(ranks_in_file - ispot);
+        Vector<MPI_Status> syncing_status(ranks_in_file - ispot);
+
+        Real t0 = amrex::second();
+
+        // If not the first MPI writing on this rank,
+        // block until it is your turn.
+        while (barrier_count < ispot)
```
WeiqunZhang, line 2797: Can we write this as a simple for loop?

Src/Base/AMReX_VisMF.cpp
```diff
@@ -2471,6 +2518,828 @@ VisMF::WriteAsync (const FabArray<FArrayBox>& mf, const std::string& mf_name)
...
+        {
+            waiting_reqs[barrier_count++] =
+                ParallelDescriptor::Abarrier(io_comm).req();
+        }
+        barrier_count = 0; // reset while waiting.
+        ParallelDescriptor::Waitall(waiting_reqs, waiting_status);
```
WeiqunZhang, line 2803: Maybe we should only call this if `ispot > 0`.  It seems that MPI standard does not say what to do if the size of waiting_reqs is zero.

Src/Base/AMReX_VisMF.cpp
```diff
@@ -2471,6 +2518,828 @@ VisMF::WriteAsync (const FabArray<FArrayBox>& mf, const std::string& mf_name)
...
+                     : (std::ios::binary | std::ios::app));
+            if (!ofs.good()) amrex::FileOpenFailed(file_name);
+            ofs.write(d.get(), total_bytes);
+            ofs.close();
+        }
+
+        Real t2 = amrex::second();
+
+        // If not the last, release the next rank to write
+        while (barrier_count < ranks_in_file-ispot)
```
WeiqunZhang, line 2821: Could be a simple for loop.

WeiqunZhang: Or range for 
```
for (auto& req : sycing_reqs) req = ParallelDesscriptor::Abarrier(io_comm).req();
```

Src/Base/AMReX_VisMF.cpp
```diff
@@ -2471,6 +2518,828 @@ VisMF::WriteAsync (const FabArray<FArrayBox>& mf, const std::string& mf_name)
...
+                }
+            }
+
+            if (doConvert) {
+                RealDescriptor::convertFromNativeFormat(p, nreals, ptmp, whichRD);
+                The_Pinned_Arena()->free(ptmp);
+            }
+            p += nreals * whichRD.numBytes();
+        }
+        local_plotfile_data[level].swap(level_data);
```
WeiqunZhang, line 3079: We could do `local_plotfile_data[level] = std::move(level_data);` instead.

WeiqunZhang:
Looks good to me.  Thanks!  We can get this merged first and then worry about other things.  For example, if we believe that MPI thread multiple does not cause any performance issue when all mpi calls are made from non-threaded region, we can turn mpi thread on by default and make async i/o fall back sync i/o if mpi multiple thread is not supported.  If we do that, we do not need to modify cmake for the new macro.  Maybe we can just ignore the user set nOutFiles for async i/o.

Tools/GNUMake/Make.defs
```diff
@@ -538,10 +545,14 @@ ifeq ($(USE_ARRAYVIEW), TRUE)
 endif
 
 ifeq ($(USE_MPI),TRUE)
-    MPISuffix	:= .MPI
-    CPPFLAGS	+= -DBL_USE_MPI -DAMREX_USE_MPI
-else
-    MPISuffix	:=
```
WeiqunZhang, line 544: Can we put this back?

Src/Base/AMReX_VisMF.cpp
kngott: I believe this is correct as written. sizeof_int64_over_real is the ratio of (int64 size / real size). So, reals * int64s_per_real = int64s. It's also a direct copy from the file async version. Do you agree?

WeiqunZhang: No I don't agree.  I sometimes write buggy code.  So you copied the bug.

WeiqunZhang: Just thing it this way.  If we switch from double to float, we need less space in `Vector<int64_t> localdata`.  Right?  But currently, `n_fab_nums` is bigger for float than double, because 8/4 is bigger thatn 8/8.

kngott: I gotcha:
The units are `sizeof_int64_over_real = (bytes/int64)/(bytes/real)` = `reals/int64`.
`number of reals / (reals/int64) = number of int64s`
👍 
I'll adjust and rerun to double check.

kngott:
* On Thread Multiple: I'm wary of just turning it on everywhere. Every MPI implementation is different and I don't believe most MPI teams care much about thread multiple performance. It felt like they were starting to, and then GPU Direct came along. I mean, the latest version of Spectrum, with a late January 2020 date,  is the first version to confirm full THREAD_MULTIPLE support. Not performance, just support. Plus, older MPI implementations are definitely going to take a hit with thread multiple.

I imagine eventually we'll just pick one, but for now it seems like it should be an option, especially as we move to these new systems and MPI implementations.

* On NOutFiles: Before Async, it looks like this got changed by a ParmParse "amr" variable, or calling SetNOutFiles at runtime. For Async, SetNOutFiles get more complex, so minimizing the number of times it's called makes sense. That's easiest with a ParmParse variable. If a code doesn't have AMR, you need a new VisMF input, which leaves a very weird interaction with the the one in "amr" (e.g. which comes first?). That's my first concern. A separate asyncNFilesOut variable would solve that, but that could make a bigger mess of things that need to be maintained. (I imagine a SetAsyncNOutFiles function might be needed.) Any ideas?

kngott:
This update should cover everything I didn't comment on. (In all cases, I performed the requested change.)

PR #835, opened 2020-05-01
CMake: make AMReX_buildInfo.cpp always out of date

mic84:
buildInfoGetBuildDate() will now return the date and time of the build. 
Before this change, it returned the date and time of the CMake configuration step.


Tools/CMake/AMReXBuildInfo.cmake
```diff
@@ -67,8 +67,8 @@ function (generate_buildinfo _target _git_dir)
    # Set variables to be used in generated source file
    #
 
-   # Date and time
-   string(TIMESTAMP BUILD_DATE "%Y-%m-%d %H:%M:%S"  )
+   # Date and time -- C++ preprocessor macros
+   set(BUILD_DATE "__DATE__ \" \"  __TIME__")
```
ax3l, line 71: Can we make this a cache-eable variable? At least we do not need to actively counter efforts for reproducible builds (Debian et al.) :-)

mic84: Just added some codee to take care of that

ax3l:
I just realized the `add_custom_command(TARGET ... PRE_BUILD ...)` does not work with `OBJECTS` libraries.

Can you recall why is the configure time is not sufficient?
Example: `file(TOUCH_NOCREATE [<files>...])`

PR #837, opened 2020-05-02
DPC++ scan

WeiqunZhang:


Src/Base/AMReX_Scan.H
```diff
@@ -102,11 +122,19 @@ struct BlockStatus<T, false>
...
+#if defined(AMREX_USE_DPCPP)
+    STVA<T> wait (sycl::nd_item<1> const& item) volatile {
+#else
     STVA<T> wait () volatile {
+#endif
         STVA<T> r;
         do {
             r = read();
+#if defined(AMREX_USE_DPCPP)
+            item.mem_fence(); // xxxxx DPCPP todo: This is at block level, but needs to be device level fence, which is currently a PR in intel/llvm
```
atmyers, line 134: Does this work in its current form, without the device level fence?

WeiqunZhang: My understanding is it's not guaranteed to work.  However, this is only used when if the type in scan is long.

PR #842, opened 2020-05-05
prefetch only if it's managed memory

WeiqunZhang:


Src/Base/AMReX_BaseFab.H
```diff
@@ -1646,37 +1646,41 @@ template <class T>
 void
 BaseFab<T>::prefetchToHost () const noexcept
 {
+    if (this->arena() == The_Arena() or this->arena() == The_Managed_Arena()) {
 #if defined(AMREX_USE_DPCPP)
     // xxxxx DPCPP todo: prefetchToHost
     // std::size_t s = sizeof(T)*this->nvar*this->domain.numPts();
     // auto& q = Gpu::Device::streamQueue();
     // q.submit([&] (sycl::handler& h) { h.prefetch(this->dptr, s); });
```
atmyers, line 1654: Why doesn't this work?

WeiqunZhang: https://github.com/intel/llvm/blob/sycl/sycl/include/CL/sycl/handler.hpp#L1485

In SYCL, both gpu and cpu are "device".  `Gpu::Device::streamQueue()` returns a `queue` for gpu device.  We need to build a `queue` for cpu device and then use that to call prefetch.  We haven't done that yet.

PR #843, opened 2020-05-05
Move prefetch from AmrLevel WritePlotfile to VisMF::Write.  Avoid unn…

WeiqunZhang:
…ecessary work if no ghost cells.

### DIFF
--- a/Src/Base/AMReX_VisMF.cpp
+++ b/Src/Base/AMReX_VisMF.cpp
@@ -1002,21 +1002,34 @@ VisMF::Write (const FabArray<FArrayBox>&    mf,
     }
     bool doConvert(*whichRD != FPC::NativeRealDescriptor());
 
-    if(set_ghost) {
+    if(set_ghost and mf.nGrowVect() != 0) {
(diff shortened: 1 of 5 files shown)


Src/Base/AMReX_VisMF.cpp
```diff
@@ -1152,6 +1165,10 @@ VisMF::Write (const FabArray<FArrayBox>&    mf,
 
     delete whichRD;
 
+    if (Gpu::inLaunchRegion()) {
+        amrex::prefetchToDevice(mf);
```
atmyers, line 1169: is `prefetchToDevice` a no-op if the mf uses host memory, or do we need to check for that here?

WeiqunZhang: Yes, see #842, which fixed this issue.  Prefetch is only supposed to be called on managed memory.

WeiqunZhang: I will modify this PR slightly.  For non-managed, we don't even need to start MFIter.

PR #844, opened 2020-05-05
For cuda > 10, use cudaLaunchHostFunc instead of cudaStreamAddCallback, which has been marked as deprecated

atmyers:
There is a chance this will fix the problems @jordanm304 and @rporcu observed on Joule, but I think we should do it anyway since the stream callback way is deprecated.


atmyers:
As a test, I ran the MFIX tests using this branch on garuda. The results are unchanged. 

https://ccse.lbl.gov/pub/GpuRegressionTesting/MFIX-Exa/2020-05-05-002/index.html

PR #845, opened 2020-05-05
Async out

WeiqunZhang:
See the commit message for more details.  The previous implementations have not been deleted.  I plan to do it in another PR.

### DIFF
--- a/Src/Base/AMReX_AsyncOut.H
+++ b/Src/Base/AMReX_AsyncOut.H
@@ -1,0 +1,36 @@
+#ifndef AMREX_ASYNCOUT_H_
+#define AMREX_ASYNCOUT_H_
+
+#include <AMReX_ParallelDescriptor.H>
+#include <functional>
+
+namespace amrex {
+namespace AsyncOut {
+
+struct WriteInfo {
+    int ifile;
+    int ispot;
+    int nspots;
+};
+
+void Initialize ();
+void Finalize ();
+
+bool UseAsyncOut ();
+
+WriteInfo GetWriteInfo (int rank);
+
+void Submit (std::function<void()>&& a_f);
+void Sumbit (std::function<void()> const& a_f);
+
+void Finish (); // If you want to wait for jobs submitted to finish
+
+//
+// These functions are used inside user's job funciton.
+//
+void Wait ();   // Wait for my turn to write file.  This is not for waiting for job to finish.
+void Notify (); // Notify next MPI process in the same file.
+
+}}
+
+#endif
--- a/Src/Base/AMReX_AsyncOut.H
+++ b/Src/Base/AMReX_AsyncOut.H
@@ -21,7 +21,7 @@ bool UseAsyncOut ();
 WriteInfo GetWriteInfo (int rank);
 
 void Submit (std::function<void()>&& a_f);
-void Sumbit (std::function<void()> const& a_f);
+void Submit (std::function<void()> const& a_f);
 
 void Finish (); // If you want to wait for jobs submitted to finish
 
--- a/Src/Base/AMReX_AsyncOut.cpp
+++ b/Src/Base/AMReX_AsyncOut.cpp
@@ -1,0 +1,163 @@
+#include <AMReX_AsyncOut.H>
+#include <AMReX_Vector.H>
+#include <AMReX_ParmParse.H>
+#include <AMReX.H>
+#include <algorithm>
+#include <condition_variable>
+#include <memory>
+#include <mutex>
+#include <queue>
+#include <thread>
+
+namespace amrex {
+namespace AsyncOut {
+
+namespace {
+
+int s_asyncout = false;
+int s_noutfiles = 64;
+MPI_Comm s_comm = MPI_COMM_NULL;
+
+std::unique_ptr<std::thread> s_thread;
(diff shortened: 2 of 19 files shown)


Src/Base/AMReX_AsyncOut.H
```diff
@@ -0,0 +1,36 @@
...
+
+void Initialize ();
+void Finalize ();
+
+bool UseAsyncOut ();
+
+WriteInfo GetWriteInfo (int rank);
+
+void Submit (std::function<void()>&& a_f);
+void Sumbit (std::function<void()> const& a_f);
```
atmyers, line 24: Typo.

Src/Base/AMReX_FArrayBox.H
```diff
@@ -346,7 +380,7 @@ public:
     * It must be the case the comp >= 0 && num_comp >= 1 &&
     * (comp+num_comp) <= nComp().  The FAB is written out in
     * whatever format you've set, with the default format being
-    * ASCII.  The FAB that is written to disk will be an
+    * NATIVE.  The FAB that is written to disk will be an
     * num_comp component FAB.
```
atmyers, line 384: The above overload of this function should also have a default on NATIVE, not ASCII.

Src/Base/AMReX_VisMF.cpp
```diff
@@ -2217,6 +2245,248 @@ void VisMF::CloseAllStreams() {
...
+        if (is_rvalue) {
+            myfabs->emplace_back(std::move(const_cast<FArrayBox&>(mf[mfi])));
+        } else {
+            myfabs->emplace_back(mf[mfi].box(), mf.nComp(), The_Cpu_Arena());
+            auto& new_fab = myfabs->back();
+            new_fab.copy<RunOn::Host>(mf[mfi]);
+        }
+    }
+
+    std::shared_ptr<FABio> fabio(new FABio_binary(FPC::NativeRealDescriptor().clone()));
```
atmyers, line 2384: What keep this from going out of scope when the async call finishes? Is it that the shared pointer gets captured in the lamba, so it doesn't get destroyed until the lambda call finishes, which would be when the IO thread is done writing?

Src/Base/AMReX_AsyncOut.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+void Wait ()
+{
+#ifdef AMREX_USE_MPI
+    const int N = s_info.ispot;
+    if (N > 0) {
+        Vector<MPI_Request> reqs(N);
+        Vector<MPI_Status> stats(N);
+        for (int i = 0; i < N; ++i) {
+            reqs[i] = ParallelDescriptor::Abarrier(s_comm).req();
```
atmyers, line 141: So the rank with `ispot` 1 will wait for the one with 0 to call IBarrier, which happens below on line 156. The one with `ispot` 2 will wait for 0 and 1, etc. Then they will all wait at 158 until everybody gets there. The notification happens through the `s_comm`?

Src/Base/AMReX_FArrayBox.H
atmyers: I mean that the comment should be updated, the code already uses NATIVE

Src/Base/AMReX_FabConv.cpp
```diff
@@ -524,7 +524,7 @@ permute_real_word_order (void*       out,
                          const int*  inord, 
                          int         REALSIZE)
 {
-    BL_PROFILE("permute_real_word_order");
+//    BL_PROFILE("permute_real_word_order");
```
atmyers, line 527: This is intentional I take it, because these functions are called too many times?

Src/Base/AMReX_AsyncOut.H
WeiqunZhang: Thanks.  Fixed.

Src/Base/AMReX_FArrayBox.H
WeiqunZhang: That's right.  Fixed.

Src/Base/AMReX_VisMF.cpp
WeiqunZhang: Yes, exactly.  The `shared_ptr` is captured by value by the lambda function provided to `AsyncOut::Submit`.  The local copy gets destroyed, but the copy in the lambda function (wrapper in `std::function` for type erasure) is pushed on to a `queue` by the main thread.  When the IO thread wakes up, it will pop the function from the queue and do work.  After the work is one, it's destroyed.

Src/Base/AMReX_AsyncOut.cpp
WeiqunZhang: Yes.  Kevin has tested a bunch of communication patterns.  This one is the cleanest and the fastest.  Yes the first `nspots-1` ranks all wait at line 158.  But they don't wait for the last process because there is no more work after the last one get its turn.

Src/Base/AMReX_FabConv.cpp
WeiqunZhang: These functions are now used by the background thread.  It messes up the profiling tools.  Since they don't take much time anyway, I took them out.

PR #847, opened 2020-05-06
fix a corner case in which the object is fully contained within a coa…

WeiqunZhang:
…rse cell without touching its surface.  this could happen during the coarsening of fine geometry.


WeiqunZhang:
This fixes a bug Emmanuel noticed.

PR #860, opened 2020-05-11
use more relaxed memory order for BaseFab stats because we need them …

WeiqunZhang:
…to be atomic but we do not require a specific synchronization order across threads


kngott:
If memory ordering is changed relaxed, does this mean individual atomic-style operations may give different (granted, slightly) results if ran multiple times? If so, is that okay for debugging, or do we want to be able to enforce a specific order?

WeiqunZhang:
We keep track of four stats for Fabs, total bytes, total cells, and high-water-mark of these.  But we do not require a specific order that one should be updating these variables, and we do not require all threads observe them in the same order.  As long as they are atomic, we are safe.

PR #866, opened 2020-05-11
CMake: add helper function setup_target_for_cuda_compilation()

mic84:
This function ensures that the CUDA compilation of a dependent
target is consistent with amrex CUDA build.

### DIFF
--- a/Tools/CMake/AMReXTargetHelpers.cmake
+++ b/Tools/CMake/AMReXTargetHelpers.cmake
@@ -114,6 +114,8 @@ endfunction ()
 # is compatible with amrex CUDA build.
 #
 function (setup_target_for_cuda_compilation _target)
+   # separable compilation:
+   #   mainly due to amrex::Random which uses global device variables
    set_target_properties( ${_target}
       PROPERTIES
       CUDA_SEPARABLE_COMPILATION ON      # This adds -dc
--- a/Tools/CMake/AMReXTargetHelpers.cmake
+++ b/Tools/CMake/AMReXTargetHelpers.cmake
@@ -119,7 +119,6 @@ function (setup_target_for_cuda_compilation _target)
    set_target_properties( ${_target}
       PROPERTIES
       CUDA_SEPARABLE_COMPILATION ON      # This adds -dc
-      CUDA_RESOLVE_DEVICE_SYMBOLS OFF
       )
    set_cpp_sources_to_cuda_language(${_target})
 endfunction ()
--- a/Tools/CMake/AMReXTargetHelpers.cmake
+++ b/Tools/CMake/AMReXTargetHelpers.cmake
@@ -107,3 +107,17 @@ function (set_cpp_sources_to_cuda_language _target)
    list(FILTER _sources INCLUDE REGEX "\\.cpp")
    set_source_files_properties(${_sources} PROPERTIES LANGUAGE CUDA )
 endfunction ()
+
+#
+# Setup an amrex-dependent target for cuda compilation.
+# This function ensures that the CUDA compilation of _target
+# is compatible with amrex CUDA build.
+#
+function (setup_target_for_cuda_compilation _target)
+   set_target_properties( ${_target}
+      PROPERTIES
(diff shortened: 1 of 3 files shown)


mic84:
@ax3l try this out and see if it helps. If it does, we can go ahead and update the amrex doc on CMaka+CUDA builds before merging.

Tools/CMake/AMReXTargetHelpers.cmake
```diff
@@ -107,3 +107,17 @@ function (set_cpp_sources_to_cuda_language _target)
...
    set_source_files_properties(${_sources} PROPERTIES LANGUAGE CUDA )
 endfunction ()
+
+#
+# Setup an amrex-dependent target for cuda compilation.
+# This function ensures that the CUDA compilation of _target
+# is compatible with amrex CUDA build.
+#
+function (setup_target_for_cuda_compilation _target)
+   set_target_properties( ${_target}
```
ax3l, line 117: I added the following comment in WarpX to recall why we have to set this at all:
```suggestion
   # mainly due to amrex::Random which uses global device variables
   set_target_properties( ${_target}
```
https://github.com/ECP-WarpX/WarpX/pull/759

Tools/CMake/AMReXTargetHelpers.cmake
```diff
@@ -107,3 +107,17 @@ function (set_cpp_sources_to_cuda_language _target)
...
+      PROPERTIES
+      CUDA_SEPARABLE_COMPILATION ON      # This adds -dc
+      CUDA_RESOLVE_DEVICE_SYMBOLS OFF
```
ax3l, line 122: I think this depends, if I am passing the WarpX executable target in here this will be problematic: https://cmake.org/cmake/help/v3.17/prop_tgt/CUDA_RESOLVE_DEVICE_SYMBOLS.html
```suggestion
```
But will test.

mic84: We can toggle that property on and off depending on the nature of the target by checking:
```
get_target_property(_target_type ${_target} TYPE)
if (_target_type STREQUAL "EXECUTABLE")
  ...
endif ()
```

mic84: I will also double check why we use it in mfix

ax3l: Yet... the actual default of this property that does exactly that.

mic84: yeah...weird

ax3l: > I will also double check why we use it in mfix.

Just in case this helps: Looks like these defaults were not documented prior to CMake 3.14: https://cmake.org/cmake/help/v3.14/prop_tgt/CUDA_RESOLVE_DEVICE_SYMBOLS.html

ax3l: Tried now, definitely does not work on our executable and does not look safe to me to be changed from the default at this point.

mic84:
@ax3l if you are satisfied with the latest changes, we can go ahead and update the documentation on CUDA builds with CMake

Tools/CMake/AMReX_Config.cmake
```diff
@@ -104,12 +104,6 @@ function (configure_amrex)
          AMREX_NVCC_MAJOR_VERSION=${NVCC_VERSION_MAJOR}
          AMREX_NVCC_MINOR_VERSION=${NVCC_VERSION_MINOR} )
 
-      set_target_properties( amrex
-         PROPERTIES
-         CUDA_SEPARABLE_COMPILATION ON      # This adds -dc
-         CUDA_RESOLVE_DEVICE_SYMBOLS OFF
```
ax3l, line 110: Summary:
- we intentionally remove `CUDA_RESOLVE_DEVICE_SYMBOLS OFF` for the `amrex` target since @mic84 tested this and it is superfluous.
- we intentionally do not set it downstream either in `setup_target_for_cuda_compilation` because this did not build for WarpX

PR #867, opened 2020-05-11
Templating the PhysBCBase class 

bsrunnels:
This adds a template parameter to the `PhysBCBase` class, and (for derived classes in AMReX) sets the template parameter to `MultiFab`.

This allows for the templated `FillPatchUtil` routines to handle template parameters other than `amrex::MultiFab`. (Currently, it will not compile with a template parameter other than `MultiFab` since the `physbc.FillBoundary` call always expects a `MultiFab`.


WeiqunZhang:
Thanks for fixing this!  I wonder if we should use a default template parameter in PhysBCFunctBase to maintain backward compatibility, or we should simply make it a template parameter of the fillpatch functions instead of being virtual.  We used virtual in the past because the fillpatch functions were not template.  But now we do not have any reasons for it to be virtual anymore.

bsrunnels:
Good idea, I just added a default template parameter - although that will only make a difference for those compiling with `g++17` I think. But in my code it was very easy to just add `<MultiFab>` to every instance of `PhysBCFuncBase`, so the fallout should be minimal.

Moving to templated FIllPatch functions sounds fine, although it might create more backwards compatibility issues. So maybe do that separately from this?

WeiqunZhang:
Oh, that's right.  Class template argument deduction is a C++17 feature.  In that case, I feel we definitely should make it a template parameter and rename FillBoundary function to operator() so that any callable object works.

bsrunnels:
Sure - that makes sense. I like the idea of a boundary condition functor better anyway, makes it more flexible. 

I added the BC template option and switched from `bc.FillBoundary(...)` to `bc(...)` in the FillPatch functions. I added the `operator ()` to `PhysBCFunctBase`, but left the pure virtual `FillBoundary` function there - it is now called by the `operator ()` function.

PR #869, opened 2020-05-12
add inline namespace literals and use in CNS and Heat Equation tutorials

WeiqunZhang:
Users can use `using namespace amrex::literals;` to import amrex literals without importing others.


Tutorials/GPU/CNS/Source/hydro/CNS_hydro_K.H
```diff
@@ -6,6 +6,8 @@
 #include <AMReX_FArrayBox.H>
 #include <cmath>
 
+using namespace amrex::literals;
```
ax3l, line 9: the only tiny comment: I would generally move those into function body scopes in header files. It's not too much harm here, but people will copy-paste from tutorials...

WeiqunZhang: Yes, will update it tonight.

PR #870, opened 2020-05-12
Fix bug in GpuComplex substraction

MaxThevenet:
With the current code, I believe that
```c++
Real x = 1.;
Real y = 2;
Complex I{0.,1.};
Print()<<x - I*y;
// returns
-1 + 2i
// instead of
1 + -2i
```
This PR proposes a fix.

### DIFF
--- a/Src/Base/AMReX_GpuComplex.H
+++ b/Src/Base/AMReX_GpuComplex.H
@@ -200,7 +200,7 @@ GpuComplex<T> operator- (const T& a_x, const GpuComplex<T>& a_y) noexcept
 {
     GpuComplex<T> r = a_y;
     r -= a_x;
-    return r;
+    return -r;
 }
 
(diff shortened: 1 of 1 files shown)


MaxThevenet:
Good question. Note that in the initial case, `r -= a_x` is a difference `complex - real`, whereas in your solution the difference is `complex - complex`, so I am not sure. I could check on compiler explorer.

drummerdoc:
ha!  Good point!

atmyers:
Thanks!

WeiqunZhang:
I was about to suggest we do something I think it's more straightforward and easier to read.  And not just for this function.
```
return GpuComplex<T>{a_x-a_y.real(), -a_y.imag()};
```

I think it's also faster.

https://godbolt.org/z/6tLNUz

PR #877, opened 2020-05-12
Teach the gpu particle locators about lev_min, lev_max, and ngrow

atmyers:
This follows the API we use for the CPU version, but maybe it would be better to make people do:

`assign_grid(p, Level(1), Level(2), nGrow(0))`  

instead of 

`assign_grid(p, 1, 2, 0)`

especially since these are optional arguments so it's pretty easy to do the wrong thing. What do you think?

### DIFF
--- a/Src/Particle/AMReX_ParticleLocator.H
+++ b/Src/Particle/AMReX_ParticleLocator.H
@@ -48,7 +48,7 @@ struct AssignGrid
     int operator() (const P& p, int nGrow=0) const noexcept
     {
         const auto iv = getParticleCell(p, m_plo, m_dxi, m_domain);
-        return this->operator()(iv);
+        return this->operator()(iv, nGrow);
     }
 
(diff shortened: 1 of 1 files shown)


Src/Particle/AMReX_ParticleLocator.H
```diff
@@ -45,14 +45,14 @@ struct AssignGrid
 
     template <typename P>
     AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
-    int operator() (const P& p) const noexcept
+    int operator() (const P& p, int nGrow=0) const noexcept
     {
         const auto iv = getParticleCell(p, m_plo, m_dxi, m_domain);
         return this->operator()(iv);
```
WeiqunZhang, line 51: missing `nGrow`?

atmyers: Thanks.

PR #882, opened 2020-05-13
use explicit casts to silence warnings

atmyers:


Src/Particle/AMReX_NeighborList.H
```diff
@@ -108,9 +108,9 @@ public:
         m_bins.build(np_total, pstruct_ptr, bx,
                      [=] AMREX_GPU_HOST_DEVICE (const ParticleType& p) noexcept -> IntVect
                      {
-                         return IntVect(AMREX_D_DECL((p.pos(0)-plo[0])*dxi[0] - lo.x,
-                                                     (p.pos(1)-plo[1])*dxi[1] - lo.y,
-                                                     (p.pos(2)-plo[2])*dxi[2] - lo.z));
+                         return IntVect(AMREX_D_DECL(static_cast<int>((p.pos(0)-plo[0])*dxi[0] - lo.x),
+                                                     static_cast<int>((p.pos(1)-plo[1])*dxi[1] - lo.y),
+                                                     static_cast<int>((p.pos(2)-plo[2])*dxi[2] - lo.z)));
```
WeiqunZhang, line 113: It might be better to do `static_cast<int>((p.pos(0)-plo[0])*dxi) - lo.x`, because `lo.x` is already int.  Looks like currently it casts `lo.x` to `amrex::Real` and then casts the final result to int.

atmyers: I think you are right.

PR #883, opened 2020-05-13
Templating FillPatch and Interpolator

bsrunnels:
This pull request fixes templating issues with some of the FillPatch functions, and templates the `nodebilin` interpolation functions.


WeiqunZhang:
We can now remove the call to `mf_crse_patch.setDomainBndry` and the `Geometry const&` argument in `make_mf_crse_patch`, because it's called after `make_mf_crse_patch`.

bsrunnels:
Done.

PR #885, opened 2020-05-14
Fix bug I introduced in PR 882 and add an assertion because...

atmyers:
this function assumes the particles are all inside the grown box.

### DIFF
--- a/Src/Particle/AMReX_NeighborList.H
+++ b/Src/Particle/AMReX_NeighborList.H
@@ -108,9 +108,9 @@ public:
         m_bins.build(np_total, pstruct_ptr, bx,
                      [=] AMREX_GPU_HOST_DEVICE (const ParticleType& p) noexcept -> IntVect
                      {
-                         AMREX_D_DECL(AMREX_ASSERT((p.pos(0)-plo[0])*dxi[0] - lo.x),
-                                      AMREX_ASSERT((p.pos(1)-plo[1])*dxi[1] - lo.y),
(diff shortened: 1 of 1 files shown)


Src/Particle/AMReX_NeighborList.H
```diff
@@ -108,9 +108,13 @@ public:
         m_bins.build(np_total, pstruct_ptr, bx,
                      [=] AMREX_GPU_HOST_DEVICE (const ParticleType& p) noexcept -> IntVect
                      {
-                         return IntVect(AMREX_D_DECL(static_cast<int>((p.pos(0)-plo[0])*dxi[0]) - lo.x,
-                                                     static_cast<int>((p.pos(1)-plo[1])*dxi[1]) - lo.y,
-                                                     static_cast<int>((p.pos(2)-plo[2])*dxi[2]) - lo.z));
+                         AMREX_D_DECL(AMREX_ASSERT((p.pos(0)-plo[0])*dxi[0] - lo.x),
```
WeiqunZhang, line 111: Missing `>= 0`?  Also it might be better to do `AMREX_D_TERM(AMREX_ASSERT();, AMREX_ASSERT();, AMREX_ASSERT());`.  `AMREX_D_DECL` always work, but it depends on the `,` has lower priority than `?:`.  Without looking up the rules, it's hard to be sure.

asalmgren: i defer to weiqun as always :-)

atmyers: You are right, fixed.

PR #890, opened 2020-05-16
fix division by zero

WeiqunZhang:


mic84:
It works perfectly for me. @emotheau wanna try this out and see if it solves your issue too?

PR #891, opened 2020-05-17
get HelloWorld to compile on tulip

WeiqunZhang:


WeiqunZhang:
Note that long __shfl_down is available now.

PR #906, opened 2020-05-19
make device version random compile with dpc++, though it doesn't work…

WeiqunZhang:
… yet.

### DIFF
--- a/Src/Base/AMReX_GpuQualifiers.H
+++ b/Src/Base/AMReX_GpuQualifiers.H
@@ -12,6 +12,7 @@
 #define AMREX_GPU_GLOBAL __global__
 #define AMREX_GPU_HOST_DEVICE __host__ __device__
 #define AMREX_GPU_CONSTANT __constant__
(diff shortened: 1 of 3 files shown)


Src/Base/AMReX_GpuQualifiers.H
```diff
@@ -30,6 +31,7 @@
 #define AMREX_GPU_CONSTANT
 #define AMREX_GPU_MANAGED
 #define AMREX_GPU_DEVICE_MANAGED
+#define AMREX_GPU_EXTERNAL SYCL_EXTERNAL
```
atmyers, line 34: why this?

WeiqunZhang:
There is an error I need to fix.

Src/Base/AMReX_GpuQualifiers.H
WeiqunZhang: In sycl, to use an external function on device, one must mark it SYCL_EXTERNAL.

PR #913, opened 2020-05-20
add HIP issues we have

WeiqunZhang:


WeiqunZhang:
@kngott I know you have seen many issues in the past.  Could you verify they still exist on tulip with ROCm 3.3?  If so, could you add them to the list?

PR #927, opened 2020-05-22
Atomics: Constify Wrapper

ax3l:
Just improves [const correctness](https://isocpp.org/wiki/faq/const-correctness) in wrapper functions for atomics.

~~Add restrict hints as well as const correctness to wrappers for atomics.~~

See if this addresses a performance issue we see in WarpX depositions for single precision.
Result: with `cuda/10.1.243` on Summit, it did not.

### DIFF
--- a/Src/Base/AMReX_GpuAtomic.H
+++ b/Src/Base/AMReX_GpuAtomic.H
@@ -22,13 +22,13 @@ namespace detail {
 
     template <typename R, typename I, typename F>
     AMREX_GPU_DEVICE AMREX_FORCE_INLINE
-    R atomic_op (R* address, R val, F f) noexcept
+    R atomic_op (R* const address, R const val, F const f) noexcept
     {
 #if defined(__SYCL_DEVICE_ONLY__)
         constexpr auto mo = sycl::memory_order::relaxed;
         constexpr auto as = sycl::access::address_space::global_space;
         static_assert(sizeof(R) == sizeof(I), "sizeof R != sizeof I");
-        I* add_as_I = reinterpret_cast<I*>(address);
+        I* const add_as_I = reinterpret_cast<I*>(address);
         sycl::atomic<I,as> a{sycl::multi_ptr<I,as>(add_as_I)};
         I old_I = a.load(mo), new_I;
         do {
@@ -47,10 +47,10 @@ namespace detail {
 
     template <typename R, typename I, typename F>
     AMREX_GPU_DEVICE AMREX_FORCE_INLINE
-    R atomic_op (R* address, R val, F f) noexcept
+    R atomic_op (R* const address, R const val, F const f) noexcept
     {
         static_assert(sizeof(R) == sizeof(I), "sizeof R != sizeof I");
-        I* add_as_I = reinterpret_cast<I*>(address);
+        I* const add_as_I = reinterpret_cast<I*>(address);
         I old_I = *add_as_I, assumed_I;
         do {
             assumed_I = old_I;
@@ -72,7 +72,7 @@ namespace detail {
 
     template<class T>
     AMREX_GPU_DEVICE AMREX_FORCE_INLINE
-    T Add_device (T* sum, T value) noexcept
+    T Add_device (T* const sum, T const value) noexcept
     {
 #if defined(__CUDA_ARCH__) || defined(__HIP_DEVICE_COMPILE__)
         return atomicAdd(sum, value);
@@ -89,13 +89,13 @@ namespace detail {
 #ifdef AMREX_USE_DPCPP
 
     AMREX_GPU_DEVICE AMREX_FORCE_INLINE
-    float Add_device (float* sum, float value) noexcept
+    float Add_device (float* const sum, float const value) noexcept
     {
         return detail::atomic_op<float, int>(sum, value, amrex::Plus<float>());
     }
 
     AMREX_GPU_DEVICE AMREX_FORCE_INLINE
-    double Add_device (double* sum, double value) noexcept
+    double Add_device (double* const sum, double const value) noexcept
     {
         return detail::atomic_op<double, unsigned long long>(sum, value, amrex::Plus<double>());
     }
@@ -103,7 +103,7 @@ namespace detail {
 #elif defined(AMREX_USE_CUDA) || defined(AMREX_USE_HIP)
 
     AMREX_GPU_DEVICE AMREX_FORCE_INLINE
-    Long Add_device (Long* sum, Long value) noexcept
+    Long Add_device (Long* const sum, Long const value) noexcept
     {
         return detail::atomic_op<Long, unsigned long long>(sum, value, amrex::Plus<Long>());
     }
@@ -114,7 +114,7 @@ namespace detail {
 
     template<class T>
     AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
-    T Add (T* sum, T value) noexcept
+    T Add (T* const sum, T const value) noexcept
     {
 #if AMREX_DEVICE_COMPILE
         return Add_device(sum, value);
@@ -133,7 +133,7 @@ namespace detail {
 
     template<class T>
     AMREX_GPU_DEVICE AMREX_FORCE_INLINE
-    T Min_device (T* m, T value) noexcept
+    T Min_device (T* const m, T const value) noexcept
     {
 #if defined(__CUDA_ARCH__) || defined(__HIP_DEVICE_COMPILE__)
         return atomicMin(m, value);
@@ -148,13 +148,13 @@ namespace detail {
     }
 
     AMREX_GPU_DEVICE AMREX_FORCE_INLINE
-    float Min_device (float* m, float value) noexcept
+    float Min_device (float* const m, float const value) noexcept
     {
         return detail::atomic_op<float,int>(m,value,amrex::Less<float>());
     }
 
     AMREX_GPU_DEVICE AMREX_FORCE_INLINE
-    double Min_device (double* m, double value) noexcept
+    double Min_device (double* const m, double const value) noexcept
     {
         return detail::atomic_op<double, unsigned long long int>(m,value,amrex::Less<double>());
     }
@@ -162,7 +162,7 @@ namespace detail {
 #if defined(AMREX_USE_CUDA) || defined(AMREX_USE_HIP)
 
     AMREX_GPU_DEVICE AMREX_FORCE_INLINE
-    Long Min_device (Long* m, Long value) noexcept
+    Long Min_device (Long* const m, Long const value) noexcept
     {
         return detail::atomic_op<Long, unsigned long long int>(m,value,amrex::Less<Long>());
     }
@@ -173,12 +173,12 @@ namespace detail {
 
     template<class T>
     AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
-    T Min (T* m, T value) noexcept
+    T Min (T* const m, T const value) noexcept
     {
 #if AMREX_DEVICE_COMPILE
         return Min_device(m, value);
 #else
-        auto old = *m;
+        auto const old = *m;
         *m = (*m) < value ? (*m) : value;
         return old;
 #endif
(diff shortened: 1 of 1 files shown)


Src/Base/AMReX_GpuAtomic.H
```diff
@@ -22,13 +22,13 @@ namespace detail {
 
     template <typename R, typename I, typename F>
     AMREX_GPU_DEVICE AMREX_FORCE_INLINE
-    R atomic_op (R* address, R val, F f) noexcept
+    R atomic_op (R* AMREX_RESTRICT address, R const val, F const f) noexcept
```
WeiqunZhang, line 25: `const` in `R const` is certainly superfluous.

ax3l: are you sure? what's the type of `R`?

WeiqunZhang: Then what's `R*`?

ax3l: usually a `double*` in DP

ax3l:
Andrew did run the same tests again (SP current deposition in WarpX) and we did not see a performance difference. One could merge this PR for sake of const-correctness and minimal wrapper-confusion for the compiler, but it does not seem to matter either.

WeiqunZhang:
I don't think it makes a difference to the compiler, and it's confusing to some people.  For some types, it's just a thin wrapper around say CUDA's `double atomicAdd(double*, double)`.  If the CUDA function takes value but our function takes `double const`, that's confusing to me.  And if we our function take `double const`, why stop there?  Why not add `const` to the address to and  make it `double our_atomicadd(double * const, double const)`.

As for `restrict`, the function has only one argument of the pointer type,  so I don't see the compiler can be confused, because there is nothing for it to be aliased to.  But it's confusing at least to me, because I am wondering if it is OK to `reinterpret_cast<U* __restrict__>(T* __restrict__)`.  It's probably OK because they are different types and restrict only applies to the same type, but it's not all clear to me.

ax3l:
> For some types, it's just a thin wrapper around say CUDA's double atomicAdd(double*, double). If the CUDA function takes value but our function takes double const, that's confusing to me.

A wrapper is not the same thing as an API interface we call. The ISOC++ core guidelines on Const-Correctness explain this a bit: Why, where, when: https://isocpp.org/wiki/faq/const-correctness

> I don't think it makes a difference to the compiler.

I guess you mean "with respect to performance"? Exactly, as reported it does not change the performance in the specific case we tested and with the specific compiler (nvcc) we used. It increases intent of what happens to the parameters, see link above.

> And if we our function take double const, why stop there? Why not add const to the address to and make it double our_atomicadd(double * const, double const).

You are right, it should even read:
```C++
double our_atomicadd(double const * const, double const);
```

Imho, default mutability is a defect in C/C++ since its inception. Modern programming languages like Rust default to const unless otherwise specified. C++ is verbose because its very much backwards compatible with most C code from the 70s.

> As for restrict, the function has only one argument of the pointer type, so I don't see the compiler can be confused, because there is nothing for it to be aliased to.

Thanks that's a good point, superfluous here!

ax3l:
Back to topic: with `cuda/10.1.243` on Summit the performance did not change.

This PR "only" improves maintainability and API expressiveness now: https://isocpp.org/wiki/faq/const-correctness

Src/Base/AMReX_GpuAtomic.H
```diff
@@ -22,13 +22,13 @@ namespace detail {
...
     AMREX_GPU_DEVICE AMREX_FORCE_INLINE
-    R atomic_op (R* address, R val, F f) noexcept
+    R atomic_op (R* AMREX_RESTRICT const address, R const val, F const f) noexcept
     {
 #if defined(__SYCL_DEVICE_ONLY__)
         constexpr auto mo = sycl::memory_order::relaxed;
         constexpr auto as = sycl::access::address_space::global_space;
         static_assert(sizeof(R) == sizeof(I), "sizeof R != sizeof I");
-        I* add_as_I = reinterpret_cast<I*>(address);
+        I* AMREX_RESTRICT add_as_I = reinterpret_cast<I*>(address);
```
ax3l, line 31: since we handle more than one pointer in this kernel, I was not sure if it might be useful to add the restrict here to be very explicit.

Src/Base/AMReX_GpuAtomic.H
```diff
@@ -232,12 +232,12 @@ namespace detail {
 
     template<class T>
     AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
-    T Max (T* m, T value) noexcept
+    T Max (T* const m, T value) noexcept
```
WeiqunZhang, line 235: T value -> T const value for consistence?

ax3l: thank you, that slipped through!

Src/Base/AMReX_GpuAtomic.H
```diff
@@ -22,13 +22,13 @@ namespace detail {
 
     template <typename R, typename I, typename F>
     AMREX_GPU_DEVICE AMREX_FORCE_INLINE
-    R atomic_op (R* address, R val, F f) noexcept
+    R atomic_op (R* const address, R const val, F const f) noexcept
```
ax3l, line 25: From
```C++
#include <type_traits>
```
one should probably rather take:
```suggestion
    R atomic_op (std::add_const<R*>::type address, std::add_const<R>::type val, std::add_const<F>::type f) noexcept
```

Src/Base/AMReX_GpuAtomic.H
```diff
@@ -47,10 +47,10 @@ namespace detail {
...
```
ax3l, line 50:
```suggestion
    R atomic_op (std::add_const<R*>::type address, std::add_const<R>::type val, std::add_const<F>::type f) noexcept
```

Src/Base/AMReX_GpuAtomic.H
```diff
@@ -47,10 +47,10 @@ namespace detail {
...
     {
         static_assert(sizeof(R) == sizeof(I), "sizeof R != sizeof I");
-        I* add_as_I = reinterpret_cast<I*>(address);
+        I* const add_as_I = reinterpret_cast<I*>(address);
```
ax3l, line 53:
```suggestion
        auto const add_as_I = reinterpret_cast<I*>(address);
```

Src/Base/AMReX_GpuAtomic.H
```diff
@@ -72,7 +72,7 @@ namespace detail {
 
     template<class T>
     AMREX_GPU_DEVICE AMREX_FORCE_INLINE
-    T Add_device (T* sum, T value) noexcept
+    T Add_device (T* const sum, T const value) noexcept
```
ax3l, line 75:
```suggestion
    T Add_device (std::add_const<T*>::type sum, std::add_const<T>::type value) noexcept
```

Src/Base/AMReX_GpuAtomic.H
```diff
@@ -114,7 +114,7 @@ namespace detail {
 
     template<class T>
     AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
-    T Add (T* sum, T value) noexcept
+    T Add (T* const sum, T const value) noexcept
```
ax3l, line 117:
```suggestion
    T Add (std::add_const<T*>::type sum, std::add_const<T>::type value) noexcept
```

Src/Base/AMReX_GpuAtomic.H
```diff
@@ -133,7 +133,7 @@ namespace detail {
 
     template<class T>
     AMREX_GPU_DEVICE AMREX_FORCE_INLINE
-    T Min_device (T* m, T value) noexcept
+    T Min_device (T* const m, T const value) noexcept
```
ax3l, line 136:
```suggestion
    T Min_device (std::add_const<T*>::type m, std::add_const<T>::type value) noexcept
```

Src/Base/AMReX_GpuAtomic.H
```diff
@@ -173,12 +173,12 @@ namespace detail {
 
     template<class T>
     AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
-    T Min (T* m, T value) noexcept
+    T Min (T* const m, T const value) noexcept
```
ax3l, line 176:
```suggestion
    T Min (std::add_const<T*>::type m, std::add_const<T>::type value) noexcept
```

Src/Base/AMReX_GpuAtomic.H
```diff
@@ -192,7 +192,7 @@ namespace detail {
 
     template<class T>
     AMREX_GPU_DEVICE AMREX_FORCE_INLINE
-    T Max_device (T* m, T value) noexcept
+    T Max_device (T* const m, T const value) noexcept
```
ax3l, line 195:
```suggestion
    T Max_device (std::add_const<T*>::type m, std::add_const<T>::type value) noexcept
```

Src/Base/AMReX_GpuAtomic.H
```diff
@@ -232,12 +232,12 @@ namespace detail {
 
     template<class T>
     AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
-    T Max (T* m, T value) noexcept
+    T Max (T* const m, T const value) noexcept
```
ax3l, line 235:
```suggestion
    T Max (std::add_const<T*>::type m, std::add_const<T>::type value) noexcept
```

Src/Base/AMReX_GpuAtomic.H
```diff
@@ -341,7 +341,7 @@ namespace detail {
 
     template <typename T>
     AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
-    T Exch (T* address, T val) noexcept
+    T Exch (T* const address, T const val) noexcept
```
ax3l, line 344:
```suggestion
    T Exch (std::add_const<T*>::type address, std::add_const<T>::type val) noexcept
```

Src/Base/AMReX_GpuAtomic.H
```diff
@@ -363,7 +363,7 @@ namespace detail {
 
     template <typename T>
     AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
-    T CAS (T* address, T compare, T val) noexcept
+    T CAS (T* const address, T const compare, T const val) noexcept
```
ax3l, line 366:
```suggestion
    T CAS (std::add_const<T*>::type address, std::add_const<T>::type compare, std::add_const<T>::type val) noexcept
```

Src/Base/AMReX_GpuAtomic.H
```diff
@@ -385,7 +385,7 @@ namespace HostDevice { namespace Atomic {
 
     template <class T>
     AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
-    void Add (T* sum, T value) noexcept
+    void Add (T* const sum, T const value) noexcept
```
ax3l, line 388:
```suggestion
    void Add (std::add_const<T*>::type sum, std::add_const<T>::type value) noexcept
```

ax3l:
Ah no, what I wrote above is just a late night moment of confusion. Can be ignored.

Src/Base/AMReX_GpuAtomic.H
WeiqunZhang: Let's not try to use everything in the standard.  What do we gain?  If `T` is say `double const`, `double const const` will become simply `double const`.  Furthermore, if `T` is `double const`, the code won't compile anyway because `atomicAdd`, `atomicCAS` cannot take `const*` and `retinterpret_cast` cannot cast `const` away.  So why not keeping things clean?

WeiqunZhang:
Only saw your latest comment, after seeing your earlier ones (on slack).

Src/Base/AMReX_GpuAtomic.H
```diff
@@ -374,7 +374,7 @@ namespace detail {
         a.compare_exchange_strong(compare, val, mo);
```
ax3l, line 374: Uff, need DPC++ CI :D
```
../amrex/Src/Base/AMReX_GpuAtomic.H:374:35: error: binding reference of type 'int' to value of type 'const int' drops 'const' qualifier
         a.compare_exchange_strong(compare, val, mo);
                                   ^~~~~~~
 ../amrex/Src/Base/AMReX_FBI.H:416:52: note: in instantiation of function template specialization 'amrex::Gpu::Atomic::CAS<int>' requested here
             int msk = (m && to_try) ? Gpu::Atomic::CAS(m, 0, mypriority) : 0;
```

Will revert

PR #937, opened 2020-05-25
CMake: minimum C++ standard required is now C++14

mic84:


mic84:
@ax3l  can you fix the gitlab CI? We need to support C++14 from now on.

maxpkatz:
I don't believe we should have the CMake requirements differ from the GNU Make requirements; at the moment the GNU Make build allows C++11, and this PR isn't changing that.

The main problem with requiring C++14 is that gcc 4.8.5 is still the default system compiler on RHEL 7.x which is relevant for systems we care about like Summit. Unfortunately, the answer here is not quite as simple as "just `module load` a more recent gcc" because we may be using PGI to compile the Fortran code (to get access to OpenACC or CUDA Fortran) and that means we need to have the `pgi` module loaded (which conflicts with the `gcc` module), dropping us down to the system gcc as the nvcc host compiler. This is not an insurmountable problem but it is one that we would have to address.

mic84:
@asalmgren  what do you think?

asalmgren:
I think we don’t want to require all amrex-based codes to use C++14, but we
do want to put 14 features into mfix itself

Sent from my iPhone

On May 24, 2020, at 11:37 PM, mic84 <notifications@github.com> wrote:

﻿

@asalmgren <https://github.com/asalmgren> what do you think?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/AMReX-Codes/amrex/pull/937#issuecomment-633404746>, or
unsubscribe

WeiqunZhang:
I am closing this PR because MFiX could build with C++14 without changing the minimum requirement in AMReX.

PR #938, opened 2020-05-25
disable unqualified abs function with amrex namespace

WeiqunZhang:
Using unqualified `abs` function is often a bug due implicit conversion of `double` to `int`.  This disables the use within `namespace amrex`, and with `using namespace amrex;` or `using namespace amrex::disabled;`.


maxpkatz:
A user who makes this mistake might run into this, wonder what is going on here, and consult this header file when the compile fails -- can you add a comment above the disabled functions explaining what is going on here and what they should do to fix their compilation failure?"

WeiqunZhang:
Sure.  I just added more comments.

PR #942, opened 2020-05-27
Revert "Atomics: Constify Wrapper"

ax3l:
Reverts AMReX-Codes/amrex#927

DPC++ not happy (see inline comment) :)


ax3l:
addressed by #941

PR #943, opened 2020-05-27
CI: DPC++

ax3l:
Prepares Intel DPC++ CI builds, backported from WarpX.

Let's intentionally not set the "required" option yet to this one, because we depend on a public beta release for CI that is a bit slower than other releases we have access to for development.

### DIFF
--- a/.github/workflows/cmake/dependencies_dpcpp.sh
+++ b/.github/workflows/cmake/dependencies_dpcpp.sh
@@ -1,0 +1,23 @@
+#!/usr/bin/env bash
+#
+# Copyright 2020 The AMReX Community
+#
+# License: BSD-3-Clause-LBNL
+# Authors: Axel Huebl
+
+set -eu -o pipefail
+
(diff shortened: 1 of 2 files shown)


WeiqunZhang:
@mic84 is currently working on adding DPC++ support to cmake build system.  Maybe you two can coordinate.

ax3l:
Yes, already in contact. I leave this open and rebase when the new `ENABLE_DPP` option was added.

ax3l:
DPC++ is now public on beta7 and we are well set to roll :)

One tutorial still has a hick-up:
```
Tutorials/LinearSolvers/MultiComponent/MCNodalLinOp.cpp:46:25: error: SYCL kernel cannot call a virtual function
                                for (int m = 0; m < getNComp(); m++)
                                                    ^
```

And it turns out `dpcpp` builds slower than `nvcc`.

WeiqunZhang:
I will fix that tutorial.

ax3l:
Thanks! Cross-ref: #1038

PR #947, opened 2020-05-29
Windows

WeiqunZhang:
- If it's Windows, use C++17 for files system operations, and avoid posix functions.

- Avoid gcc extension that allows embedded prepressor directive in function macro.

- Get around some MSVC template argument deduction issues.

- Rename amrex::USleep to amrex:Sleep and use C++ thread for sleep.  The reason for renaming is it would be very confusing that `usleep` in `unistd` sleeps for given microseconds whereas `amrex::USleep` sleep for given seconds.

- Windows CI


ax3l:
If you like to cherry-pick a remote CI build, I drafted the support in https://github.com/ax3l/amrex/commit/99fae0bbc10af35d6e692751efea8448ddfe209e :)

WeiqunZhang:
Great! I will.

ax3l:
here is another one for commonly needed MSVC flags: https://github.com/ax3l/amrex/commit/d60a8d820d6e5a6041c890f0e209d9cd54011089

.github/workflows/windows.yml
```diff
@@ -0,0 +1,18 @@
...
+    name: MSVC C++17 w/o Fortran w/o MPI
+    runs-on: windows-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Build & Install
+      run: |
+        mkdir build
+        cd build
+        cmake .. -DENABLE_FORTRAN=OFF -DENABLE_MPI=OFF -DCMAKE_CXX_STANDARD=17
+        cmake --build . --config Release
```
ax3l, line 17: Let's try to build tutorials as well:
```suggestion
        cmake .. -DENABLE_TUTORIALS=ON -DENABLE_FORTRAN=OFF -DENABLE_MPI=OFF -DCMAKE_CXX_STANDARD=17
        cmake --build . --config Release --target tutorials
```

WeiqunZhang:
This is ready for review.

Src/Base/AMReX_Arena.cpp
```diff
@@ -11,7 +11,17 @@
 #include <AMReX_ParmParse.H>
 #include <AMReX_Gpu.H>
 
+#ifdef _WIN32
```
ax3l, line 14: I think it's the right way to support the system compilers (XCode's Clang fork on macOS and MSVC on Windows).

Nevertheless, where specific to a stdlib & runtime system, I would replace `__APPLE__` and `_WIN32` macros with slightly more specific Apple+Clang guards and MSVC guards (`if defined(_MSC_VER) && !defined(__INTEL_COMPILER)`).

Background: there are GNU-based compilers for Windows, too such as Cygwin and MSYS2 that some scientific projects use which rely on e.g. a full POSIX.

On macOS, one might need to potentially differentiate between vanilla clang, xcode clang and gcc, too - but let's just add this when it's actually needed.

WeiqunZhang: ```In gcc for Cygwin, _WIN32 is only defined when you use the -mwin32 gcc command line options. This is because Cygwin is supposed to be a POSIX emulation environment in the first place and defining _WIN32 confuses some programs which think that they have to make special concessions for a Windows environment which Cygwin handles automatically.```

This is from https://cygwin.com/faq/

WeiqunZhang: As for Mac, there are a few places with `__APPLE__` for two purposes.  One is backtrace support.  I don't have access to Mac.  So I used someone's Mac and added backtrace support.  I think it works with gcc and I don't know if it works with clang.  The other is some cuda stuff does not work on Mac.

ax3l: Sounds good then! CUDA is deprecated on macOS anyway and removed in CUDA11+.

WeiqunZhang:
I have run CPU and GPU regression testing using several codes.  And beside the CI, I have tested PGI, Cray, Intel, and IBM compilers.

PR #951, opened 2020-05-31
set up DPC++ regression test suite

WeiqunZhang:


atmyers:
The "Tracers" test works too I think.

WeiqunZhang:
Yes, I will add it.

atmyers:
There is a setup for this with CUDA (not checked in) on garuda: https://ccse.lbl.gov/pub/GpuRegressionTesting/AMReX/2020-05-29/Tracers.html

PR #953, opened 2020-06-01
turn tiling off by default for gpu

WeiqunZhang:
Tiling is turned off by default for GPU.  There are still two ways tiling can be turned on.  Users can change the default tiling size in their inputs files.  If explicit tile size is passed to `MFIter`, it will be chosen over the default value.

### DIFF
--- a/Src/Base/AMReX_FabArrayBase.cpp
+++ b/Src/Base/AMReX_FabArrayBase.cpp
@@ -25,7 +25,7 @@ namespace amrex {
 //
 int     FabArrayBase::MaxComp;
 
-#if defined(AMREX_USE_GPU) && defined(AMREX_USE_GPU_PRAGMA)
+#if defined(AMREX_USE_GPU)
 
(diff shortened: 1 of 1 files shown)


maximumcats (approved):

atmyers:
I should make `ParticleContainer` have the same behavior. 

EDIT - nevermind, it already off by default.

atmyers (approved):

maximumcats:
Perhaps rather than have this determined by the compile options, it should be determined at runtime? i.e. if we're in a GPU launch region, then set the tile size to a large number, otherwise choose the default CPU numbers.

WeiqunZhang:
That's a possibility.  However, one problem is if an `MFIter` is constructed with an explicit tile size, what should we do?  I think the explicit size should be respected.  Then the problem is how we can determine if an explicit size has been given.  `MFIter` has several constructors.  So it's actually hard to do.

drummerdoc (approved):

PR #958, opened 2020-06-01
ignore unused variables in the cpu versions of these gpu functions

atmyers:


atmyers:
I am closing this for now because there may be a better to fix these.

PR #977, opened 2020-06-04
Modify HeatEquation dts to account for non-cubic cases.

kngott:
The rest of the code correctly accounts for non-cubic already.

### DIFF
--- a/Tutorials/Basic/HeatEquation_EX1_C/Source/main.cpp
+++ b/Tutorials/Basic/HeatEquation_EX1_C/Source/main.cpp
@@ -90,7 +90,7 @@ void main_main ()
     init_phi(phi_new, geom);
     // ========================================
 
-    Real dt = 0.9_rt*dx[0]*dx[0] / (2.0_rt*AMREX_SPACEDIM);
+    Real dt = 0.9_rt*(dx[0]*dx[0] + dx[1]*dx[1] + dx[2]*dx[2])/ (2.0_rt*AMREX_SPACEDIM);
 
     // time = starting time in the simulation
     Real time = 0.0_rt;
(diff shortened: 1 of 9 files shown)


Tutorials/Basic/HeatEquation_EX1_C/Source/main.cpp
```diff
@@ -90,7 +90,7 @@ void main_main ()
     init_phi(phi_new, geom);
     // ========================================
 
-    Real dt = 0.9_rt*dx[0]*dx[0] / (2.0_rt*AMREX_SPACEDIM);
+    Real dt = 0.9_rt*(dx[0]*dx[0] + dx[1]*dx[1] + dx[2]*dx[2])/ (2.0_rt*AMREX_SPACEDIM);
```
WeiqunZhang: There are two issues with this.  It breaks 2d.  It changes dt to much smaller than necessary.

I think it's supposed to be `dx**2/3` -> `1./((1./dx)**2+(1./dy)**2+(1./dz)**2)` if dx, dy and dz are not equal to each other, for the 3d case.

kngott: Is there a macro that'll clean this up? I don't believe D_TERM will work:

```
#if (AMREX_SPACEDIM==1)
    Real coeff = (1./dx[0]**2);
#elif (AMREX_SPACEDIM==2)
    Real coeff = (1./dx[0]**2) + (1./dx[1]**2);
#else 
    Real coeff = (1./dx[0]**2) + (1./dx[1]**2) + (1./dx[2]**2);
#endif

    Real dt = 0.9_rt/ (2.0_rt*length);
```

WeiqunZhang:
AMREX_D_TERM should work.

WeiqunZhang (approved):

PR #980, opened 2020-06-06
Make: use -M to generate dependencies

### LINKED ISSUE #930: mkdep: does not support relative includes
In C/C++, the pre-processor checks `#include "..."` files relative to the current file's path first, *independent* of the include path options: https://gcc.gnu.org/onlinedocs/cpp/Search-Path.html

The perl script [`mkdep`](https://github.com/AMReX-Codes/amrex/blob/development/Tools/C_scripts/mkdep) does not support this and consequently forgets includes.

Result: if a developer intentionally or accidentally forgets to set `INCLUDE_LOCATIONS` in the GNU-make build script collection for a sub directory, then a program *will still compile properly* (most of the time) because it's valid. On rebuilds, dependencies will not be spotted and will be forgotten, e.g. if a file was modified.

Note that it is *not required* to have those paths in `-I`.

Example:
```c++
// main.cpp
#include "group/smth.H"

int main(){ return foo(); }
```

```c++
// group/smth.H
#ifndef SMTH_H
#define SMTH_H
inline int foo() { return 42; }
#endif
```

```bash
g++ main.cpp
```

This is also problematic if a project using the GNU-make build scripts depends on more projects than AMReX alone. Many C/C++ projects provide facade headers with a single include path these days (this allows to change things over time without breaking user-include logic). If one of the transitive headers changes this will not be caught.

## Possible Solutions

- Clean: CMake approach (IIRC): instead of parsing files with scripts, run the pre-compiler with `$CXX -M` to generate dependencies (including define code branching). Be warned, this is tricky for nvcc and host/device splitting. For some compilers, one might have to parse the output of `$CXX -E` instead.

- Hackity-hack: unify paths in `INCLUDE_LOCATIONS` with `VPATH_LOCATIONS` and pass those together as `-I` options to `mkdep`. Solves only the in-code but not the third party include problem.

- Half'n'half: brave perl programmer wanted :)

WeiqunZhang:


WeiqunZhang:
I have tested on all major compilers we use with several codes on several computers.

maximumcats (approved):

PR #982, opened 2020-06-06
fix a bug in norm2 where there are overlaps

WeiqunZhang:


ajnonaka (approved):
Looks good, and I tested it with a face-centered MultiFab and it checks out.  Clever avoidance of taking the square root of mask, like I was thinking.

PR #983, opened 2020-06-07
Compare single-level plotfiles; diff and norms; all nodalities

ajnonaka:
This utility performs a diff operation between two
plotfiles that have the same geometrical domain and nodality
(supports all nodality types; cell, face, edge, node)
but possibly a factor of refinement between the cells,
and outputs the L0, L1, and L2 norms
L1 = sum(|diff_ijk|)/npts_coarsedomain
L2 = sqrt[sum(diff_ijk^2)]/sqrt(npts_coarsedomain)
(only single-level supported)


Tools/C_util/Convergence/ComparePlotfiles.cpp
```diff
@@ -0,0 +1,253 @@
...
+    exit(1);
+}
+
+int
+main (int   argc,
+      char* argv[])
+{
+    amrex::Initialize(argc,argv);
+
+    if (argc == 1) {
```
WeiqunZhang, line 50: Maybe only the I/O process should call `PrintUsage`.

Tools/C_util/Convergence/ComparePlotfiles.cpp
```diff
@@ -0,0 +1,253 @@
...
+    // AMR comes later, where we iterate over each level in isolation
+
+    // check to see whether the user pointed to the plotfile base directory
+    // or the data itself
+    std::ifstream xxx;
+    xxx.open(iFile1+"/Level_0/Cell_H");
+    if (xxx) {
+       iFile1 += "/Level_0/Cell";
+        xxx.close();
+    }
```
WeiqunZhang:
```suggestion
    if (amrex::FileExists(iFile1+"/Level_0/Cell_H")) {
       iFile1 += "/Level_0/Cell";
    }
```

WeiqunZhang (approved):

Tools/C_util/Convergence/ComparePlotfiles.cpp
ajnonaka: done

Tools/C_util/Convergence/ComparePlotfiles.cpp
ajnonaka: done.  I changed std::cout to Print(); wrapping the function call in IOProcessor caused the remaining MPI ranks to continue with the program and spit out the wrong error messages about not specifying infile1

PR #986, opened 2020-06-08
Fix a missing semicolon

WeiqunZhang:
The (DPC++ only) bug was in
```
if (....) return
ParallelFor(...);
ParallelFor(...);
```

A semicolon was missing and the code became
```
if(...) {
    return ParallelFor(...);
}
ParallelFor(...);
```

Maybe we should always use `{}` with `if` no matter how short the block is.


atmyers (approved):

atmyers:
:(

PR #988, opened 2020-06-08
use __CUDACC_VER_MAJOR__ instead of AMREX_NVCC_MAJOR_VERSION, which i…

### LINKED ISSUE #987: AMREX_NVCC_MAJOR_VERSION issue
Seems like AMREX_NVCC_MAJOR_VERSION is not longer defined but is still being used in some places.

https://github.com/AMReX-Codes/amrex/blob/86a361c8856bdda455e3fe7d2c01b4d7ce35261a/Src/Base/AMReX_GpuDevice.cpp#L281

WeiqunZhang:
…s gone

### DIFF
--- a/Src/Base/AMReX_GpuDevice.cpp
+++ b/Src/Base/AMReX_GpuDevice.cpp
@@ -278,7 +278,7 @@ Device::Initialize ()
     // is only available starting from CUDA 10.0, so we will
     // leave num_devices_used as 0 for older CUDA toolkits.
 
-#if AMREX_NVCC_MAJOR_VERSION >= 10
+#if (__CUDACC_VER_MAJOR__ >= 10)
     size_t uuid_length = 16;
     size_t recv_sz = uuid_length * ParallelDescriptor::NProcs();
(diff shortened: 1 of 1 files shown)


Src/Base/AMReX_GpuDevice.cpp
```diff
@@ -278,7 +278,7 @@ Device::Initialize ()
     // is only available starting from CUDA 10.0, so we will
     // leave num_devices_used as 0 for older CUDA toolkits.
 
-#if AMREX_NVCC_MAJOR_VERSION >= 10
+#if (__CUDACC_VER_MAJOR__ >= 10)
```
maximumcats, line 281: Are we sure this works in all compilers we work with, or do we need to guard this with an #ifdef?

maximumcats: Oh nevermind, we only get here inside AMREX_USE_CUDA, so I think it's fine.

maximumcats (approved):

PR #964, opened 2020-06-02
CMake CUDA: Cleanups & CCache

ax3l:
- gencode: use no-space syntax; this confuses/breaks with CCache<4.0 (all available releases today) and is complicating deduplication operations on lists (see e.g. https://github.com/alpaka-group/alpaka/pull/991)
- min. CXX standard in CUDA: technically the right thing to set this as well for the CMake CUDA C++ language

Now AMReX + CMake 3.17 + CUDA works with [CCache 3.7.9](https://ccache.dev/releasenotes.html#_ccache_3_7_9) ([more cross-links](https://github.com/spack/spack/pull/16902), [more](https://github.com/ccache/ccache/issues/560#issuecomment-597621694)). Since this is a severe productivity plus I label this "install" + "performance" ;-)

Note: downstream projects might want to add this to their `CMakeLists.txt`:
```cmake
find_program(CCACHE_PROGRAM ccache)
if(CCACHE_PROGRAM)
    set(CMAKE_CXX_COMPILER_LAUNCHER "${CCACHE_PROGRAM}")
    if(ENABLE_CUDA)
        set(CMAKE_CUDA_COMPILER_LAUNCHER "${CCACHE_PROGRAM}")
    endif()
endif()
```

CCache does [not support Fortran](https://ccache.dev/platform-compiler-language-support.html).

PR #990, opened 2020-06-08
CMake: Guard 3.17+ Feature

ax3l:
We still support CMake 3.14+

Follow-up to #964


ax3l:
oh wait, that's guarded already.

PR #993, opened 2020-06-09
Non-const neighbor data iterator

JBlaschke:
This change lets you iterate over neighbors in the neighbor list, where the neighbor data is _non_-`const`

```c++
    auto nbor_data = m_neighbor_list[lev][index].data();
    for (auto & p2 : nbor_data.getNeighbors(part_index.first)) {
    // Modify p2
    }
```

This is necessary for immersed-boundary calculations, where forces are added onto the neighbor particles. In theory this can be done without modifying the neighbor particles -- but this get complex very fast, because immersed-boundary markers don't interact only via pairwise interactions.

### DIFF
--- a/Src/Particle/AMReX_NeighborList.H
+++ b/Src/Particle/AMReX_NeighborList.H
@@ -14,7 +14,7 @@ struct Neighbors
     struct iterator
     {
         AMREX_GPU_HOST_DEVICE
-        iterator (int start, int stop, const unsigned int * nbor_list_ptr, const ParticleType* pstruct) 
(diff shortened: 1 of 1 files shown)


atmyers:
Thanks for this pull request. I prefer to do this slightly differently, so I opened another PR based on this one, so that your commits are still there. Could you verify that this PR also fixes the issue you were seeing please?

Thanks.

JBlaschke:
Hey Andrew,

https://github.com/AMReX-Codes/amrex/pull/994 makes sense -- closing this PR in favor of yours.

Thanks

PR #994, opened 2020-06-09
Non const iterator neighbor list

atmyers:
Iterate over neighbors in both `const` and non-`const` fashion.

### DIFF
--- a/Src/Particle/AMReX_NeighborList.H
+++ b/Src/Particle/AMReX_NeighborList.H
@@ -14,7 +14,7 @@ struct Neighbors
     struct iterator
     {
         AMREX_GPU_HOST_DEVICE
-        iterator (int start, int stop, const unsigned int * nbor_list_ptr, const ParticleType* pstruct) 
+        iterator (int start, int stop, const unsigned int * nbor_list_ptr, ParticleType* pstruct) 
             : m_index(start), m_stop(stop), m_nbor_list_ptr(nbor_list_ptr), m_pstruct(pstruct)   
         {}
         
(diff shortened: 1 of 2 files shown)


Src/Particle/AMReX_NeighborList.H
```diff
@@ -10,77 +10,127 @@ namespace amrex
...
         bool operator!= (iterator const& rhs) const { return m_index < m_stop; }
-        
+
         AMREX_GPU_HOST_DEVICE
-        ParticleType operator* () const { return m_pstruct[m_nbor_list_ptr[m_index]];  }
-        
+        const ParticleType& operator* () const { return m_pstruct[m_nbor_list_ptr[m_index]];  }
+
+        AMREX_GPU_HOST_DEVICE
+        ParticleType& operator* () { return m_pstruct[m_nbor_list_ptr[m_index]];  }
```
WeiqunZhang: I think there should be only one version `ParticleType& operator* () const`.  `Iterator const` should still return non-const reference.  User could assign to to `const&` if they want.

See https://en.cppreference.com/w/cpp/iterator/iterator

atmyers: Okay, done.

Src/Particle/AMReX_NeighborList.H
```diff
@@ -10,77 +10,124 @@ namespace amrex
...
         AMREX_GPU_HOST_DEVICE
         void operator++ () { ++m_index;; }
-        
+
         AMREX_GPU_HOST_DEVICE
         bool operator!= (iterator const& rhs) const { return m_index < m_stop; }
-        
+
+        AMREX_GPU_HOST_DEVICE
+        ParticleType& operator* () { return m_pstruct[m_nbor_list_ptr[m_index]];  }
```
WeiqunZhang:
```suggestion
        ParticleType& operator* () const { return m_pstruct[m_nbor_list_ptr[m_index]];  }
```

WeiqunZhang (approved):

JBlaschke (approved):
All checks out from my end

atmyers:
Thanks!

JBlaschke:
I just ran 2 FHDeX execs (both slightly different uses of immersed boundaries) on CPU -- looks fine from my end @atmyers

JBlaschke:
+1 🍺

PR #996, opened 2020-06-10
CMake: improve MSVC+CUDA support

robertmaynard:
The primary issue now is that nvcc doesn't seem to be parsing with
/Za enabled, so alternate tokens are causing compile failures.

### DIFF
--- a/Tools/CMake/AMReXFlagsTargets.cmake
+++ b/Tools/CMake/AMReXFlagsTargets.cmake
@@ -94,10 +94,17 @@ add_library(AMReX::Flags_CXX_REQUIRED ALIAS Flags_CXX_REQUIRED)
 
 target_compile_options( Flags_CXX_REQUIRED
    INTERFACE
-   $<${_cxx_msvc}:/Za /bigobj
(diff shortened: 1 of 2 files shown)


Tools/CMake/AMReXFlagsTargets.cmake
```diff
@@ -94,10 +94,17 @@ add_library(AMReX::Flags_CXX_REQUIRED ALIAS Flags_CXX_REQUIRED)
 
 target_compile_options( Flags_CXX_REQUIRED
    INTERFACE
-   $<${_cxx_msvc}:/Za /bigobj
```
robertmaynard, line 97: FYI: CMake generator expressions can't span multiple lines so these two lines had to change no matter what

ax3l (approved):
Welcome to AMReX Robert!
Thank you for these fixes! :rocket: :sparkles:

ax3l:
Rebased my [GH Action test](https://github.com/ax3l/amrex/tree/ci-winCUDA) now, let's see: https://github.com/ax3l/amrex/actions

PR #1001, opened 2020-06-11
`and` -> `&&`, `or` -> `||`, and `not` -> `!` for MSVC

WeiqunZhang:
Do not merge both #1001 and #1003.

PR #1003, opened 2020-06-11
use iso646 to provide alternative operator names for Windows

WeiqunZhang:
Do not merge both #1003 and #1001.


WeiqunZhang:
Let's take this approach.  The CUDA build on Windows go much further than #1001.

atmyers (approved):

ax3l:
Glad this helps :+1:

PR #1004, opened 2020-06-11
fix Abs for FabArray

WeiqunZhang:


WeiqunZhang:
`macos/AppleClang@11.0 Gfortran@9.3` seems to fail quite often.  It seems that sometimes it fails to comile Fortran module.  Maybe we should disable Fortran.  @ax3l

ax3l:
Interesting, is that a logic race condition somewhere?
```
[ 65%] Linking CXX static library libamrex.a
/Applications/Xcode_11.5.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: libamrex.a(AMReX_constants_mod.f90.o) has no symbols
/Applications/Xcode_11.5.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: libamrex.a(AMReX_bc_types_mod.F90.o) has no symbols
/Applications/Xcode_11.5.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: libamrex.a(AMReX_lo_bctypes_mod.F90.o) has no symbols
/Applications/Xcode_11.5.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: libamrex.a(AMReX_constants_mod.f90.o) has no symbols
/Applications/Xcode_11.5.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: libamrex.a(AMReX_bc_types_mod.F90.o) has no symbols
/Applications/Xcode_11.5.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: libamrex.a(AMReX_lo_bctypes_mod.F90.o) has no symbols
```

ax3l:
Yes, looks like some Fortran modules only have private declarations that probably do not create symbols at all - so I guess they do not need to be packed into `libamrex.a`. Everything else is then probably linker order that sometimes causes this.

There are also linker-dependent flags to suppress this, as a less-clean workaround.

ax3l:
Hm, that's only a warning...
Let's try to trigger it again, I did not save the complete log and it passes now. Generally, on all CI services, macOS is a bit more flaky. Maybe rerunning the check from the `Checks` tabs when this happens is okay if this is sporadic.

WeiqunZhang:
This happens again in #1024.

WeiqunZhang:
And after Ann pushed a new commit, it failed again.  That was twice in a row.  Is this a cmake issue?

ax3l:
Ouch, can you please save the log file before you restart next time?
In the "Checks" tabs, e.g. [here](https://github.com/AMReX-Codes/amrex/pull/1004/checks?check_run_id=763167004), is a `...` symbol that has the option "View raw logs". Maybe just download and add as a text file to the comment so we can find out what's going on.

I suspect it's either a network split, flaky macOS VM runner, VM filesystem issue, brew issue (dependency download for gfortran et al.), or CMake dependency race if we forget something... but that's all speculative until we can see the logs.

Generally speaking, if we are unlucky its just the flakyness of the VMs we run on. In my experience, 1:20 on bad days to 1:50 runners on Travis CI error due to network/VM issues. There is a lot of booting and downloading from third parties going on for partitioning the runners, it's way more than starting a local docker image. GitHub CI is significantly better on Linux than travis, but maybe the macOS images are flaky. Restarting or ignoring in such a case and checking for maintenance windows on the respective status pages would then be okay, as well as excluding (or ignoring states of) known unstable images if we cannot figure out what's off from the logs in the long term.

WeiqunZhang:
The log is still there. https://github.com/AMReX-Codes/amrex/runs/777512636?check_suite_focus=true

ax3l:
Oh awesome, thanks! I collect more in #1025.
Will take a quick look, I agree this looks a bit like a missing declared CMake dependency for Fortran modules in some way or something macOS specific with Fortran modules.

atmyers (approved):

PR #1006, opened 2020-06-11
add -DNDEBUG for non-debug build

WeiqunZhang:


maximumcats (approved):

atmyers:
I think that if `USE_ASSERTION` is `TRUE`, we should also not give the `-DNDEBUG` flag.

WeiqunZhang:
yes

atmyers (approved):

PR #1009, opened 2020-06-11
CMake: MSVC Pre-Processor Flags (Fix Old/New)

mic84:
Follow-up to #996

Fix the experimental preprocessor flag for older MSVC and the new one for newer versions.

### DIFF
--- a/Tools/CMake/AMReXFlagsTargets.cmake
+++ b/Tools/CMake/AMReXFlagsTargets.cmake
@@ -99,12 +99,20 @@ target_compile_options( Flags_CXX_REQUIRED
 
 # Currently can't make this a generator expression as amrex_evaluate_genex fails
 # to parse it and propagate the flags to cuda
-if(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC" AND CXX_COMPILER_VERSION VERSION_LESS 19.26)
-   target_compile_options( Flags_CXX_REQUIRED
-      INTERFACE
-      $<${_cxx_msvc}:/experimental:preprocessor /Zc:preprocessor>
-   )
+if (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
+   if (CXX_COMPILER_VERSION VERSION_LESS 19.26)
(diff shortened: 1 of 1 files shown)


WeiqunZhang (approved):

ax3l (approved):

Tools/CMake/AMReXFlagsTargets.cmake
```diff
@@ -99,12 +99,20 @@ target_compile_options( Flags_CXX_REQUIRED
...
-if(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC" AND CXX_COMPILER_VERSION VERSION_LESS 19.26)
-   target_compile_options( Flags_CXX_REQUIRED
-      INTERFACE
-      $<${_cxx_msvc}:/experimental:preprocessor /Zc:preprocessor>
-   )
+if (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
+   if (CXX_COMPILER_VERSION VERSION_LESS 19.26)
+      target_compile_options( Flags_CXX_REQUIRED
+         INTERFACE
+         $<${_cxx_msvc}:/experimental:preprocessor>
```
ax3l, line 106: is the generator expression still needed when we have a `CMAKE_CXX_COMPILER_ID STREQUAL "MSVC"` guard?

mic84: No it's not, strictly speaking. However, CXX_COMPILER_VERSION does not have the notion of compiler id so I would say it makes it clearer. Also, without checking the compiler ID, we would be adding MSVC flags based on the version of a compiler which could potentially be something other than MSVC

ax3l: But don't we `if` the `CMAKE_CXX_COMPILER_ID` for MSVC as well (line 102)?

mic84: The IF statement is a workaround for a deficiency in the current genex evaluation function. It will be removed when we fix that. The generator expression is what we want to keep so there is no point in modifying it now.

PR #1011, opened 2020-06-12
Single Precision: INVALID_TIME Value OoR

ax3l:
AppleClang 11.0.3 with single precision C++14 builds errors with the following in WarpX:
```
error: constexpr variable 'INVALID_TIME' must be initialized by a
       constant expression
static constexpr Real INVALID_TIME = -1.0e200;
                      ^              ~~~~~~~~
WarpX/build_sp/_deps/fetchedamrex-src/Src/Amr/AMReX_StateData.cpp:17:38:
  note: value -1.0E+200 is outside the range of representable values of
        type 'const amrex::Real' (aka 'const float')
static constexpr Real INVALID_TIME = -1.0e200;
```

This changes the numeric value of this constant to the lowest in the supported range of the underlying data type.

### DIFF
--- a/Src/Amr/AMReX_StateData.cpp
+++ b/Src/Amr/AMReX_StateData.cpp
@@ -1,5 +1,6 @@
 
 #include <iostream>
+#include <limits>
 #include <algorithm>
 
 #include <AMReX_RealBox.H>
@@ -14,7 +15,7 @@
 
 namespace amrex {
 
-static constexpr Real INVALID_TIME = -1.0e200;
+static constexpr Real INVALID_TIME = std::numeric_limits<Real>::lowest();
 
 static constexpr int MFNEWDATA = 0;
 static constexpr int MFOLDDATA = 1;
--- a/Src/Amr/AMReX_StateData.cpp
+++ b/Src/Amr/AMReX_StateData.cpp
@@ -15,7 +15,11 @@
 
 namespace amrex {
 
-static constexpr Real INVALID_TIME = std::numeric_limits<Real>::lowest();
+#ifdef AMREX_USE_FLOAT
+static constexpr Real INVALID_TIME = -1.0e20;
+#else
+static constexpr Real INVALID_TIME = -1.0e200; 
+#endif
 
 static constexpr int MFNEWDATA = 0;
 static constexpr int MFOLDDATA = 1;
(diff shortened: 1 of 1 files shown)


Src/Amr/AMReX_StateData.cpp
```diff
@@ -14,7 +15,7 @@
 
 namespace amrex {
 
-static constexpr Real INVALID_TIME = -1.0e200;
+static constexpr Real INVALID_TIME = std::numeric_limits<Real>::min();
```
WeiqunZhang: lowest?

ax3l: Oh right, good catch!
https://en.cppreference.com/w/cpp/types/numeric_limits/lowest

WeiqunZhang:
I am not entirely sure about this.  But I think setting it the lowest might break some codes that need this to be a low, but not the lowest number, in some initial iteration before the normal time stepping starts.  Rather than merging this to find it out, I suggest we keep it at `1.e-200` for double and have it be `1.e-30` for float.

ax3l:
Sounds good to me, will adjust accordingly :)

Src/Amr/AMReX_StateData.cpp
```diff
@@ -14,7 +15,7 @@
 
 namespace amrex {
 
-static constexpr Real INVALID_TIME = -1.0e200;
+static constexpr Real INVALID_TIME = std::numeric_limits<Real>::lowest();
```
ax3l:
```suggestion
#ifdef AMREX_USE_FLOAT
static constexpr Real INVALID_TIME = -1.0e20;
#else
static constexpr Real INVALID_TIME = -1.0e200; 
#endif
```

Src/Amr/AMReX_StateData.cpp
```diff
@@ -14,7 +15,11 @@
 
 namespace amrex {
 
-static constexpr Real INVALID_TIME = -1.0e200;
+#ifdef AMREX_USE_FLOAT
+static constexpr Real INVALID_TIME = -1.0e20;
```
ax3l:
```suggestion
static constexpr Real INVALID_TIME = -1.0e30;
```

ax3l:
Oh, actually it was at a huge `-1.e200` not a small `(-)1.e-200`.

Probably lowest is quite safe then as a magic number, cannot imagine that someone depends on such a number (it's not close to zero and quite far out in the range). But will do as you suggest.

WeiqunZhang (approved):

PR #1016, opened 2020-06-12
CMake: add support for DPCPP

mic84:


Tools/CMake/AMReX_Options.cmake
```diff
@@ -90,15 +90,33 @@ endif ()
...
-option( ENABLE_MPI  "Enable MPI"  ON)
+
+# For the time being ENABLE_DPCPP is defined before project() is called
+# Check whether the C++ compiler is dpcpp
+print_option(ENABLE_DPCPP)
+if (ENABLE_DPCPP AND (NOT (CMAKE_CXX_COMPILER MATCHES "dpcpp") ) )
+   message(FATAL_ERROR "\nENABLE_DPCPP=${ENABLE_DPCPP} but CXX compiler is not dpcpp\n")
+endif ()
+
+cmake_dependent_option( ENABLE_DPCPP_AOT  "Enable DPCPP AOT"  OFF
```
ax3l: I think this does not really work with the compiler yet:
```suggestion
cmake_dependent_option( ENABLE_DPCPP_AOT  "Enable DPCPP ahead-of-time compilation (WIP)"  OFF
```

Tools/CMake/AMReX_Options.cmake
```diff
@@ -90,15 +90,33 @@ endif ()
...
+   "ENABLE_DPCPP" OFF)
+print_option( ENABLE_DPCPP_AOT )
+
+cmake_dependent_option( ENABLE_DPCPP_SPLIT_KERNEL "Enable DPCPP kernel splitting"  ON
+   "ENABLE_DPCPP" OFF)
+print_option(  ENABLE_DPCPP_SPLIT_KERNEL )
+
+cmake_dependent_option( ENABLE_MPI  "Enable MPI"  ON
```
ax3l, line 111: Does DPC++ not work with MPI?

Tools/CMake/AMReX_Options.cmake
```diff
@@ -90,15 +90,33 @@ endif ()
...
+   "NOT ENABLE_DPCPP" OFF)
 print_option( ENABLE_MPI )
 
 option( ENABLE_OMP  "Enable OpenMP" OFF)
 print_option( ENABLE_OMP )
 
-option( ENABLE_CUDA "Enable GPU support via CUDA" OFF )
+cmake_dependent_option( ENABLE_CUDA "Enable GPU support via CUDA" OFF
```
ax3l, line 118: I think this is more of a use-case for a conflict/fatal error than a dependent option, but am happy to be convinced otherwise with example workflows.

Tools/CMake/AMReX_Options.cmake
```diff
@@ -90,15 +90,33 @@ endif ()
 print_option( DIM )
 
 #
-# Programming models  ========================================================
+# Parallel backends    ========================================================
```
ax3l, line 93:
```suggestion
# Parallel, on-node compute backends    =======================================
```

Tools/CMake/AMReX_Config.cmake
```diff
@@ -155,6 +155,54 @@ function (configure_amrex)
...
+
+         string(STRIP "${_cpu_long_name}" _cpu_long_name)
+         if (_cpu_long_name STREQUAL "skylake")
+            set(_cpu_short_name "skl")
+         elseif (_cpu_long_name STREQUAL "kabylake")
+            set(_cpu_short_name "kbl")
+         elseif (_cpu_long_name STREQUAL "cascadelake")
+            set(_cpu_short_name "cfl")
+         else ()
+            message(FATAL_ERROR "\n AOT CPU ${_cpu_long_name} not yet supported\n")
```
ax3l: Users love this trick ;-)
```suggestion
            message(FATAL_ERROR "\n Ahead-of-time compilation for CPU ${_cpu_long_name} is not yet supported\n"
            "Maybe set ENABLE_DPCPP_AOT to OFF?\n")
```

Tools/CMake/AMReX_Config.cmake
```diff
@@ -155,6 +155,54 @@ function (configure_amrex)
...
+
+      if (ENABLE_DPCPP_AOT)
+         if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
+            ## TODO: use file(READ)
+            execute_process( COMMAND cat /sys/devices/cpu/caps/pmu_name OUTPUT_VARIABLE _cpu_long_name )
+         else ()
+            message(FATAL_ERROR "\nENABLE_DPCPP_AOT is not supported on ${CMAKE_SYSTEM_NAME}\n")
+         endif ()
+
+         string(STRIP "${_cpu_long_name}" _cpu_long_name)
```
ax3l: you could make this an advanced cached var to overwrite the logic from the cmd line

ax3l (commented):
Looks great, thanks a lot!
I added a few suggestions & questions inline.

Tools/CMake/AMReX_Options.cmake
mic84: Should we issue a warning in case a user sets it? This variable is present in the GNUMake so I would like to keep it in the CMake too even if not yet usable.

Tools/CMake/AMReX_Options.cmake
mic84: In the GNUMake MPI is turned off if DPC++ is on, so I guess it does not.

Tools/CMake/AMReX_Options.cmake
mic84: As for MPI, In the GNUMake CUDA is turned off if DPC++ is on. From the CMake doc: ```[cmake_dependent_option]  presents an option to the user only if a set of other conditions are true.``` We don't want to preset ENABLE_CUDA as an option if DPC++ is enabled, so I would argue this falls into the cmake_dependent_option usage

Tools/CMake/AMReX_Options.cmake
mic84: I would rather keep it more general: that comment marks the beginning of the section devoted to every parallel backend, including MPI

Tools/CMake/AMReX_Config.cmake
mic84: Which variable are you referring to? _cpu_long_name?

ax3l: Yes, thinking the following situations:
- new CPU on an experimental system with a not-yet-updated AMReX lookup table
- testing with some generic GPU name

Tools/CMake/AMReX_Options.cmake
ax3l: But how would a user then switch back with `ccmake` from DPC++ to CUDA? Would probably need 2x reconfig (disable DPC++, reconfig, enable CUD, reconfig, generate), which is a bit too involved, I think.

Tools/CMake/AMReX_Options.cmake
ax3l: @WeiqunZhang can you please clarify this aspect? Do MPI and DPC++ work together already or is there another reason we miss?

Tools/CMake/AMReX_Options.cmake
ax3l: Good idea, let's issue a warning for now.

Tools/CMake/AMReX_Options.cmake
ax3l: Okay, I would separate on-node computing backends and intra-node comm - alone to avoid the confusion which level of parallelism we refer to, but this is no big deal.

Tools/CMake/AMReX_Options.cmake
WeiqunZhang: We have never tested it and don't really have a place to test it.  So I simply disable it for now in GNU make.  However, there is no reason it won't work.

Tools/CMake/AMReX_Config.cmake
WeiqunZhang: We probably want to suggest to Intel for their compiler to take the long name from `pmu_name`.  Then we don't need to have that table.

Another issue is in the future, the login and compute nodes might have different cpus.

WeiqunZhang: Currently AOT is not useful to us because it does not support device math functions.

ax3l: Sounds great, thank you for the details.
Can one of you please open an issue at https://github.com/intel/llvm for the flags?

Let's issue a fatal error (or at least warning) if AOT is selected in our build logic, because we know it cannot work yet.

Tools/CMake/AMReX_Options.cmake
mic84: You can set ENABLE_CUDA an any time but it will not be used until you switch off ENABLE_DPCPP.
[https://cmake.org/cmake/help/v3.14/module/CMakeDependentOption.html](https://cmake.org/cmake/help/v3.14/module/CMakeDependentOption.html)

Tools/CMake/AMReX_Config.cmake
mic84: I commented out the AOT support for now and added an error message if ENABLE_AOT is on. We will re-enable AOT support when the time comes.

WeiqunZhang (approved):

WeiqunZhang:
I have tested this on Iris. It works.

WeiqunZhang:
I will merge this so that people can use it.

PR #1017, opened 2020-06-12
CMake: set minimum C++ standard to 17 for Windows builds

mic84:


WeiqunZhang (approved):

ax3l (approved):
LGTM, but I would probably differentiate by compiler ID instead (think: MSYS2/Cygwin/WSL).

Please remove the last argument, `-DCMAKE_CXX_STANDARD=17`, in CI: https://github.com/AMReX-Codes/amrex/blob/1d75ba05f084cd3c1d91dacefdfedc6c2e85ab7f/.github/workflows/windows.yml#L16
This is now no longer needed.

PR #1023, opened 2020-06-15
Support for overset in cell-centered MLABecLaplacian

WeiqunZhang:
The user needs to pass a mask the the constructor of `MLABecLapalcian`.   The mask is 1 for overset region and 0 otherwise.  The overset region's solution is passed in the `MulitFab` argument to `MLMG::solve`.


drummerdoc:
What is "overset"?

WeiqunZhang:
I don't exactly know how it works.  Suppose there is unstructured mesh around an object and a structured mesh around the unstructured grids.  The two grids have overlapped regions.   One can interpolate unstructured data onto structured grids and then solve the problem on structured grids using the overlapped region as boundary conditions.

asalmgren (approved):

PR #1024, opened 2020-06-16
Update how we write out EB data so it correctly handles problo !=0

asalmgren:
Currently the EB data is written out assuming problo = 0.   This PR passes problo into the writing routines so the locations are written correctly when problo != 0.

### DIFF
--- a/Src/EB/AMReX_eb_to_pvd.F90
+++ b/Src/EB/AMReX_eb_to_pvd.F90
@@ -641,10 +641,10 @@ contains
     if(allocated(lines)) deallocate(lines)
     allocate(lines(0:lnodes))
 
-    grid_start = problo + real(lo)*ldx
+    grid_start = problo + real(lo,amrex_real)*ldx
 
     do llc = 0, lnodes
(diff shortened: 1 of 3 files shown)


Src/EB/AMReX_eb_to_pvd.F90
```diff
@@ -627,21 +626,25 @@ subroutine amrex_eb_grid_coverage (myID, dx, lo, hi, flag, fglo, fghi)&
...
     integer :: llc
     real(amrex_real), allocatable :: lines(:)
+    real(amrex_real)              :: grid_start 
 
     if(allocated(lines)) deallocate(lines)
     allocate(lines(0:lnodes))
 
-    do llc=0, lnodes
-       lines(llc) = real(lo + llc)*ldx
+    grid_start = problo + real(lo)*ldx
```
WeiqunZhang: I think `real` here (and three lines below probably) should be `dble(lo)` or `real(lo,amrex_real)`.

WeiqunZhang (approved):

PR #1027, opened 2020-06-16
Tools: remove Make.cscs since Piz Daint is not used anymore

bcfriesen:


WeiqunZhang:
Is that the machine is gone or we are not using it?

bcfriesen:
Apparently nobody in CCSE uses it. I added it originally because NERSC wanted access to some GPUs before we had Cori GPU. So we got accounts (briefly) on Piz Daint. But now that we have Cori GPU there is less value in having access to Piz Daint.

asalmgren (approved):

PR #1038, opened 2020-06-19
fix a tutorial for dpc++ (and cuda)

WeiqunZhang:


Tutorials/LinearSolvers/MultiComponent/MCNodalLinOp.cpp
```diff
@@ -62,6 +66,10 @@ void MCNodalLinOp::Diag (int amrlev, int mglev, MultiFab& a_diag)
 	domain.grow(-1); // Shrink domain so we don't operate on any boundaries
 	const Real* DX = m_geom[amrlev][mglev].CellSize();
 
+        amrex::AsyncArray<Real> aa(coeff.data(), coeff.size());
+        amrex::Real const* dcoeff = aa.data();
+        const int N = getNComp();
+
```
ax3l: Alignment looks a bit off:
```suggestion
      amrex::AsyncArray<Real> aa(coeff.data(), coeff.size());
      amrex::Real const* dcoeff = aa.data();
      const int N = getNComp();

```

Tutorials/LinearSolvers/MultiComponent/MCNodalLinOp.cpp
```diff
@@ -31,6 +31,10 @@ void MCNodalLinOp::Fapply (int amrlev, int mglev, MultiFab& a_out,const MultiFab
 	domain.grow(-1); // Shrink domain so we don't operate on any boundaries
 	const Real* DX = m_geom[amrlev][mglev].CellSize();
 
+        amrex::AsyncArray<Real> aa(coeff.data(), coeff.size());
+        amrex::Real const* dcoeff = aa.data();
+        const int N = getNComp();
```
ax3l:
```suggestion
     amrex::AsyncArray<Real> aa(coeff.data(), coeff.size());
     amrex::Real const* dcoeff = aa.data();
     const int N = getNComp();
```

ax3l (approved):

ax3l:
`make` tries to run the `HeatEquation` example. We can just add a `AND NOT ENABLE_DPCPP` [here](https://github.com/AMReX-Codes/amrex/blob/development/Tutorials/Basic/HeatEquation_EX1_CF/CMakeLists.txt#L47).

@mic84 I would propose that we remove that the "run" target is part of the default target in the tutorial(s?): https://github.com/AMReX-Codes/amrex/blob/cc65bd6f21d65e6d3183ebc8b875252b67b23a15/Tutorials/Basic/HeatEquation_EX1_CF/CMakeLists.txt#L49-L61

We can add them to e.g. `ctest`/`make test` targets but we should not run the example executables with the default target of `make` (build), I think.

Tutorials/LinearSolvers/MultiComponent/MCNodalLinOp.cpp
WeiqunZhang: The problem is that file a full of tabs.  The old spacing of these lines with no tabs actually looked fine in my emacs, but not the new spacing.  Anyway we don't want remove all the tabs just for the sake of removing.

PR #1044, opened 2020-06-21
Fix for hypre

asalmgren:
When using hypre for an EB nodal solve, restrict the number of multigrid levels based on how much the EB can be coarsened.


WeiqunZhang:
We can now remove the assertion from `MLNodeLinOp::makeHypreNodeLap`
```
    AMREX_ALWAYS_ASSERT_WITH_MESSAGE(NMGLevels(0) == 1, 
                                     "MLNodeLaplacian: To use hypre, max_coarsening_level must be 0");
```

WeiqunZhang (approved):

PR #1049, opened 2020-06-24
Porting fortran files to cpp in Src/Extern/HYPRE

OscarAntepara:
I have replaced AMReX_HABEC_2D.F90 with AMReX_Habec_2D_K.H and AMReX_Habec_2D.cpp
I have replaced AMReX_HABEC_3D.F90 with AMReX_Habec_3D_K.H and AMReX_Habec_3D.cpp


WeiqunZhang:
Thanks for the PR!  There are two issues need to be fixed.  First, this breaks PETSc because `Src/Extern/PETSc` uses some files in `Src/Extern/HYPRE`.  So you need to convert `PETSc` too.  Second, it breaks when hypre uses long integer.  For example, line 285 of `AMReX_Habec_2D.cpp` should use `HYPRE_int` instead of `int`.

OscarAntepara:
There are still some issues that I have to check

Src/Extern/HYPRE/AMReX_HypreABecLap.cpp
```diff
@@ -195,24 +200,16 @@ HypreABecLap::prepareSolver ()
     std::iota(stencil_indices.begin(), stencil_indices.end(), 0);
     const Real* dx = geom.CellSize();
     const int bho = (m_maxorder > 2) ? 1 : 0;
-    FArrayBox rfab;
+    BaseFab<GpuArray<Real,regular_stencil_size>> rfab;
     for (MFIter mfi(acoefs); mfi.isValid(); ++mfi)
     {  
         const Box &reg = mfi.validbox();
 
         rfab.resize(reg,regular_stencil_size);
```
WeiqunZhang:
```suggestion
        rfab.resize(reg);
```
because each element of the fab is already an array.

Src/Extern/HYPRE/AMReX_Habec_2D_K.H
```diff
@@ -0,0 +1,778 @@
...
+#include <AMReX_EBMultiFabUtil.H>
+#include <AMReX_MultiCutFab.H>
+#include <AMReX_EBFabFactory.H>
+#include <AMReX_MLEBABecLap_K.H>
+#endif
+
+namespace amrex {
+
+AMREX_FORCE_INLINE
+void amrex_hpacoef (Box const& box,
```
WeiqunZhang: This 2d version of `amrex_hpacoef` is identical to the 3d version.  Could we move them to a new file `AMReX_Habec_K.H`?  Maybe move the following block in `AMReX_HypreABecLap.cpp` to `AMReX_Habec_K.H` too?

```
#if (AMREX_SPACEDIM == 2)                                                                                                                                                                                  
#include <AMReX_Habec_2D_K.H>
#elif (AMREX_SPACEDIM == 3)
#include <AMReX_Habec_3D_K.H>
#endif
```

OscarAntepara: sure!!

WeiqunZhang:
Currently it does not compile with `USE_CUDA=TRUE`.  Some kernels are not GPU ready yet.  To make it fully work with GPU is beyond the scope of this PR.  But at least we should make it compile, and we should disable GPU launch region for functions that are not ready for GPU.   To disabe launch region, you can `git grep LaunchSafeGuard` for examples.  You will need to use `AMREX_HOST_DEVICE_PARALLEL_FOR_3D` instead of `AMREX_PARALLEL_FOR_3D`, so that the launch region flag is respected.

OscarAntepara:
Now it compiles with USE_CUDA=TRUE

Src/Extern/HYPRE/AMReX_Habec_2D_K.H
```diff
@@ -0,0 +1,673 @@
...
+        }
+
+        if (j == hi.y && cell_id_arr(i,j+1,k)<0) {
+           mat_tmpPtr[0] += bf1[3]*by_arr(i,j+1,k);
+           mat_tmpPtr[3] += bf2[3]*by_arr(i,j+1,k);
+        }
+
+        diag_arr(i,j,k) = 1.0/mat_tmpPtr[0];
+
+        for (int ic=0; ic<=5; ic++){
```
WeiqunZhang: Should this be `4` instead of `5`?

OscarAntepara: ohh yeah, now it is fixed! Thanks.

Src/Extern/HYPRE/AMReX_Habec_2D_K.H
```diff
@@ -0,0 +1,673 @@
...
+        bf1 = -fac;
+        bf2 = 0.0;
+    }else{
+        Abort("hpmat: unsupported boundary type");
+    }
+
+    if (cdir == 0){
+        AMREX_HOST_DEVICE_PARALLEL_FOR_3D (box, i, j, k,
+        {
+            if (msk_arr(lo.x-1,j,k) > 0 && i==lo.x){
```
WeiqunZhang:
```suggestion
        Box tbx = box;
        tbx.setSmall(0,lo.x).setBig(0,lo.x); // 0: x-direction
        AMREX_HOST_DEVICE_PARALLEL_FOR_3D (tbx, i, j, k,
        {
```
The looping is on a slice only.  Could you fix in other place too including 3D?

Src/Extern/HYPRE/AMReX_Habec_2D_K.H
```diff
@@ -0,0 +1,673 @@
...
+              Array4<Real const> const& apy_arr,
+              Array4<Real const> const& fcy_arr,
+              Array4<Real const> const& bcen_arr,
+              Array4<Real const> const& vfrc_arr,
+              Array4<Real const> const& ba_arr,
+              Array4<Real const> const& beb_arr) noexcept
+{
+        Array2D<Real,-1,1,-1,1> mat_tmp;
+        Array1D<Real,0,4> phig1;
+        Array1D<Real,0,4> feb;
```
WeiqunZhang:
```suggestion
        Array1D<Real,0,3> phig1;
        Array1D<Real,0,3> feb;
```

WeiqunZhang (approved):

PR #1052, opened 2020-06-24
CMake: simplify configuration option summary

mic84:
At configuration time, list the enabled options only.


ax3l (approved):
Good idea! :)

PR #1054, opened 2020-06-24
[WIP] CMake: Assertions & FPEs in Subprojects

ax3l:
Make sure `ENABLE_ASSERTIONS` and `ENABLE_FPE` work well with sub projects: Now show up as advanced option if set in a downstream project.
Passed to `cmake <path> -DCMAKE_BUILD_TYPE=Release -DENABLE_ASSERTIONS=ON` as an argument, one can enable assertions (of floating point exceptions with `-DENABLE_FPE=ON`) in super-builds without enabling the `Debug` build type.

Also, enables FPEs by default in Debug builds now as in GNUmake.

### DIFF
--- a/Tools/CMake/AMReX_Options.cmake
+++ b/Tools/CMake/AMReX_Options.cmake
@@ -206,17 +206,19 @@ endif ()
 option( ENABLE_PIC "Build position-independent code" OFF)
 print_option( ENABLE_PIC )
 
-option(ENABLE_FPE "Enable Floating Point Exceptions checks" OFF)
-print_option( ENABLE_FPE )
-
+set(ENABLE_FPE_DEFAULT OFF)
 set(ENABLE_ASSERTIONS_DEFAULT OFF)
 if ( "${CMAKE_BUILD_TYPE}" MATCHES "Debug" )
    set(ENABLE_ASSERTIONS_DEFAULT ON)
+   set(ENABLE_FPE_DEFAULT ON)
 endif ()
 cmake_dependent_option(ENABLE_ASSERTIONS "Enable assertions" ENABLE_ASSERTIONS_DEFAULT
     "ON" ENABLE_ASSERTIONS_DEFAULT)
+cmake_dependent_option(ENABLE_FPE "Enable Floating Point Exceptions checks" ENABLE_FPE_DEFAULT
+    "ON" ENABLE_FPE_DEFAULT)
 
 print_option( ENABLE_ASSERTIONS )
+print_option( ENABLE_FPE )
 
 
 #
--- a/Tools/CMake/AMReX_Options.cmake
+++ b/Tools/CMake/AMReX_Options.cmake
@@ -209,11 +209,12 @@ print_option( ENABLE_PIC )
 option(ENABLE_FPE "Enable Floating Point Exceptions checks" OFF)
 print_option( ENABLE_FPE )
 
+set(ENABLE_ASSERTIONS_DEFAULT OFF)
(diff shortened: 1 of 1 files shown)


mic84 (approved):

Tools/CMake/AMReX_Options.cmake
```diff
@@ -206,16 +206,19 @@ endif ()
 option( ENABLE_PIC "Build position-independent code" OFF)
 print_option( ENABLE_PIC )
 
-option(ENABLE_FPE "Enable Floating Point Exceptions checks" OFF)
-print_option( ENABLE_FPE )
-
+set(ENABLE_FPE_DEFAULT OFF)
+set(ENABLE_ASSERTIONS_DEFAULT OFF)
 if ( "${CMAKE_BUILD_TYPE}" MATCHES "Debug" )
```
mic84, line 211:
```suggestion
option(ENABLE_FPE "Enable Floating Point Exceptions checks" OFF)
print_option( ENABLE_FPE )
```

I would leave FPE options as an opt out no matter what the build type is since we experienced considerable slowdowns in certain codes where they were used. AMReX has fpe-catching capabilities even without those flags so I am inclined to keep them out by default.

Tools/CMake/AMReX_Options.cmake
```diff
@@ -206,16 +206,19 @@ endif ()
...
+   set(ENABLE_ASSERTIONS_DEFAULT ON)
+   set(ENABLE_FPE_DEFAULT ON)
 endif ()
+cmake_dependent_option(ENABLE_ASSERTIONS "Enable assertions" ENABLE_ASSERTIONS_DEFAULT
+    "ON" ENABLE_ASSERTIONS_DEFAULT)
+cmake_dependent_option(ENABLE_FPE "Enable Floating Point Exceptions checks" ENABLE_FPE_DEFAULT
+    "ON" ENABLE_FPE_DEFAULT)
 
 print_option( ENABLE_ASSERTIONS )
+print_option( ENABLE_FPE )
```
mic84, line 221:
```suggestion
cmake_dependent_option( ENABLE_ASSERTIONS "Enable assertions" ON
   "CMAKE_BUILD_TYPE MATCHES Debug" OFF)
print_option( ENABLE_ASSERTIONS )
```
I think this is cleaner and uses cmake_dependent_options as it is intended to be used.

ax3l: I think this hides the asserts as an options for all non-debug builds, which is probably not what we want in some debugging situations with compiler optimizations on?

Tools/CMake/AMReX_Options.cmake
ax3l: Yeah, I don't know - maybe something you and @WeiqunZhang can weight the pro's and cons in your developer meetings, I have no real opinion on it.

I am okay to keep it enabled automatically in Debug, as done in GNUmake:
https://github.com/AMReX-Codes/amrex/blob/7e5c619d5ce21907e6f4dbcff17fef3c3a10d992/Tools/GNUMake/comps/gnu.mak#L77-L80

ax3l:
I just realized that setting those vars via
```cmake
set(ENABLE_ASSERTIONS OFF CACHE BOOL "")
```
instead of
```cmake
set(ENABLE_ASSERTIONS OFF CACHE INTERNAL "")
```

does not hide them :) (I am relying on [CMP0007](https://cmake.org/cmake/help/git-stage/policy/CMP0077.html) logic `NEW`).

PR #1055, opened 2020-06-25
remove compile time BACKTRACE flag for simplicity

WeiqunZhang:


Docs/Notes/Readme.backtrace
```diff
@@ -19,27 +19,23 @@ with TEST=TRUE or DEBUG=TRUE in C++ AMReX, or TEST=t or NDEBUG= in
...
 the runtime parameter, fab.init_snan.
 
-If it is compiled with BACKTRACE=TRUE, one can get more information
-than the backtrace of the call stack by instrumenting the code.  (This
-is in C++ code only.)  Here is an example.  You know the line "Real
-rho = state(cell,0);" is causing a segfault.  You could add a print
-statement before that.  But it might print out thousands (or even
-millions) of line before it hits the segfault.  With BACKTRACE, you
-could do
+One can get more information than the backtrace of the call stack by
```
ax3l: Looks good! Can we either link this file from the `.rst` manual or convert it to `.rst` and include it? This is good to know but I could not find it initially without a hint :)

WeiqunZhang: OK.  I will move it to the manual.

mic84 (approved):

WeiqunZhang:
Could @mic84 or @ax3l check if what I say on cmake is correct?

ax3l (approved):
Yep, looks all good - thanks for the ping!

PR #1058, opened 2020-06-25
Async IO for Particles

atmyers:


Src/Particle/AMReX_ParticleIO.H
```diff
@@ -400,302 +402,19 @@ ParticleContainer<NStructReal, NStructInt, NArrayReal, NArrayInt>
...
-                }                
-            }            
-        }
-    }
-    
-    if (ParallelDescriptor::IOProcessor())
+    if (AsyncOut::UseAsyncOut()) {
+        WriteBinaryParticleDataAsync(*this, dir, name,
+                                     write_real_comp, write_int_comp,
+                                     real_comp_names, int_comp_names);
```
WeiqunZhang, line 408: Any reason why the async version does not take `F`?

atmyers: I plan to remove `F` soon. WarpX will implement this differently, by copying to a tmp particle container and applying a filter during the copy. That functionality will be useful in other places, and it will let us remove all those extra overloads of `WritePlotFile` and `Checkpoint`.

Src/Particle/AMReX_WriteBinaryParticleData.H
```diff
@@ -0,0 +1,688 @@
...
+            LevelDir = amrex::Concatenate(LevelDir + "Level_", lev, 1);
+
+            if (MyProc == IOProcNumber)
+            {
+                if ( ! amrex::UtilCreateDirectory(LevelDir, 0755))
+                {
+                    amrex::CreateDirectoryFailed(LevelDir);
+                }
+            }
+            ParallelDescriptor::Barrier();
```
WeiqunZhang: Can we try to merge this with lines 340-347?

Src/Particle/AMReX_WriteBinaryParticleData.H
```diff
@@ -0,0 +1,688 @@
...
+        }
+    }
+    ParallelDescriptor::Barrier();
+
+    Long total_np = 0;
+    Vector<Long> np_per_level(pc.finestLevel()+1);
+    Vector<Vector<Long> > np_per_grid(pc.finestLevel()+1);
+    for (int lev = 0; lev <= pc.finestLevel(); lev++)
+    {
+        np_per_grid[lev] = pc.NumberOfParticlesInGrid(lev);
```
WeiqunZhang: We could also do local reduction only and then do mpi_allreduce after the for loop.

WeiqunZhang: And for the following code in `NumberOfParticlesInGrid`
```    if (!only_local) {                                                                                                                                                                                     
        ParallelAllReduce::Sum(&nparticles[0], ngrids, ParallelContext::CommunicatorSub());                                                                                                         
    }
```
I think @mrowan137 has shown it's much faster to gather and broadcast.  @mrowan137 Can you comment on that?  There is a function for the gather `ParallelDescriptor::GatherLayoutDataToVector`.

Src/Particle/AMReX_WriteBinaryParticleData.H
```diff
@@ -0,0 +1,688 @@
...
+                new_ptile.resize(ptile.numParticles());
+                amrex::copyParticles(new_ptile, ptile);
+            }
+        }
+    }
+
+    auto wrc = std::make_shared<Vector<int> >(write_real_comp);
+    auto wic = std::make_shared<Vector<int> >(write_int_comp);
+    auto rcn = std::make_shared<Vector<std::string> >(real_comp_names);
+    auto icn = std::make_shared<Vector<std::string> >(int_comp_names);
```
WeiqunZhang: Are they needed?  Can we just use `write_real_comp` etc. directly in the lambda below?

WeiqunZhang:
It's a big PR.  Although I made some comments, I can approve it and then you can decide whether you want to improve it later in follow-up PRs.

atmyers:
Thanks for your comments - there's no hurry to merge this right now. I'll try your suggestions.

Src/Particle/AMReX_WriteBinaryParticleData.H
atmyers: Indeed they are not. Fixed.

Src/Particle/AMReX_WriteBinaryParticleData.H
mrowan137: @atmyers if you look at `DistributionMapping::makeKnapSack (const LayoutData<Real>& rcost_local, ...` in AMReX_DistributionMapping.cpp you can see the alternate gather/broadcast sequence, it should probably be faster than the `ParallelAllReduce::Sum` here.

WeiqunZhang (approved):

PR #1061, opened 2020-06-25
Add mpi reduction to fix a bug in overset linear solver

WeiqunZhang:


WeiqunZhang:
@michaeljbrazell Thanks for noticing the bug!

kngott (approved):

PR #1062, opened 2020-06-26
Conduit: Avoid const_cast

ax3l:
Follow-up to #1057: Would this potentially work? Just trying to avoid `const_cast` across library boundaries.

cc @cyrush @WeiqunZhang


WeiqunZhang:
Probably not because `ptile` is `const`.

WeiqunZhang:
I think we should keep the way it is till Conduit fixes the constness in its API.

ax3l:
Urgh, you are right - otherwise we would need to un-const the `ptile`, right.

I tried reworking the public API the other day, bit it's still a bit to go there: https://github.com/LLNL/conduit/pull/578

PR #1063, opened 2020-06-26
MFIter reducer belongs only to AMREX_USE_GPU_PRAGMA

maximumcats:
Only AMReX-Astro uses this reduction code inside MFIter (when using the CUDA build), so this restricts the ifdef appropriately. Eventually this code will be removed.

### DIFF
--- a/Src/Base/AMReX_MFIter.cpp
+++ b/Src/Base/AMReX_MFIter.cpp
@@ -216,11 +216,11 @@ MFIter::~MFIter ()
     if (device_sync) Gpu::synchronize();
 #endif
 
-#ifdef AMREX_USE_GPU
+#ifdef AMREX_USE_GPU_PRAGMA
     reduce();
 #endif
 
-#ifdef AMREX_USE_GPU
+#ifdef AMREX_USE_GPU_PRAGMA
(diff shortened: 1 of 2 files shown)


Src/Base/AMReX_MFIter.cpp
```diff
@@ -496,6 +496,7 @@ MFIter::operator++ () noexcept
 #endif
         
         bool use_gpu = (numOmpThreads == 1) && Gpu::inLaunchRegion();
```
WeiqunZhang: I think we should move this line into the pragma region.  `use_gpu` is used a few lines below.  That should be changed to `Gpu::inLaunchRegion()`.

WeiqunZhang: We will make gpu streams thread safe in a different PR.

maximumcats: Agreed and fixed. Actually `use_gpu` didn't really express what I wanted to do (and that code won't work either way when using OpenMP) so I just made it `Gpu::inLaunchRegion()` as well.

WeiqunZhang (approved):

PR #1064, opened 2020-06-26
Gpu stream thread safety

WeiqunZhang:


Src/Base/AMReX_OpenMP.H
```diff
@@ -0,0 +1,31 @@
...
+#ifdef _OPENMP
+#include <omp.h>
+
+namespace amrex {
+namespace OpenMP {
+
+    inline int get_num_threads () { return omp_get_num_threads(); }
+    inline int get_max_threads () { return omp_get_max_threads(); }
+    inline int get_thread_num  () { return omp_get_thread_num();  }
+    inline int in_parallel     () { return omp_in_prallel();      }
```
kngott: `omp_in_parallel` is spelled incorrectly.

maximumcats: Typo here

WeiqunZhang: Good catch.  Fixed now.

kngott (approved):

Src/Base/AMReX_OpenMP.H
WeiqunZhang: I think that was fixed already.

kngott: Yeah. I believe I made this around the same time as Max. I'm not sure why, but my approval seemed to reannounce this comment.

maximumcats (approved):

PR #1066, opened 2020-06-26
Manage memory

WeiqunZhang:
Add a parameter `amrex.the_arena_is_managed` so that one can use device memory by default for Fabs.

### DIFF
--- a/Docs/sphinx_documentation/source/GPU.rst
+++ b/Docs/sphinx_documentation/source/GPU.rst
@@ -288,17 +288,17 @@ specific type of GPU memory:
 
 .. table:: Memory Arenas
 
-    +---------------------+------------------+
-    | Arena               |    Memory Type   |
-    +=====================+==================+
-    | The_Arena()         |  unified memory  |
-    +---------------------+------------------+
-    | The_Device_Arena()  |  device memory   |
-    +---------------------+------------------+
-    | The_Managed_Arena() |  unified memory  |
-    +---------------------+------------------+
-    | The_Pinned_Arena()  |  pinned memory   |
-    +---------------------+------------------+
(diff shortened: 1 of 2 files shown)


Docs/sphinx_documentation/source/GPU.rst
```diff
@@ -315,11 +315,13 @@ to two functions:
    void free (void* p);
 
 :cpp:`The_Arena()` is used for memory allocation of data in
-:cpp:`BaseFab`.  Therefore the data in a :cpp:`MultiFab` is placed in
-unified memory and is accessible from both CPU host and GPU device.
+:cpp:`BaseFab`.  By default, it allocates managed memory.  This can be changed with
+a boolean runtime parameter ``amrex.the_arena_is_managed``.
+Therefore the data in a :cpp:`MultiFab` is placed in
+managed memory by default and is accessible from both CPU host and GPU device.
```
kngott, line 321: Do we also want to add here what memory is used if the_arena_is_managed is false? Only describes what happens if true and "it can be changed with a boolean flag".

WeiqunZhang: We could.  But the table above already shows it's either managed or device.

kngott: Yeah, I know. That's why I left it as a question. I'm on the fence. When I first read the change, I was expecting to see the alternative also described and when I didn't find it, it took me a minute to find the table and think through what false would be. Then, I imagined users reading this and not looking at the chart, or asking what "device memory" means, since it's not spelled out right here.

Generally, I'm 50-50 on it. I'll leave it up to you.

kngott (approved):

PR #1084, opened 2020-07-02
Makes `>>` operators for reading checkpoint files dimension agnostic

dwillcox:
Previously, we needed to compile AMReX with the same dimensionality of the checkpoint files when using `Amr::restart()` because some of the classes define `>>` operators that used `AMREX_D_EXPR` to read data from the checkpoint header.

This PR is to support compiling in 3D and loading a checkpoint file from 1D or 2D, needed for postprocessing where we want to read a 2D checkpoint file and convert it to a 3D checkpoint file that we can restart from.

This will let us create an `Amr` object for the 2D checkpoint and a second `Amr` object for the 3D domain we want to fill.

### DIFF
--- a/Src/Base/AMReX_Utility.cpp
+++ b/Src/Base/AMReX_Utility.cpp
@@ -145,6 +145,34 @@ amrex::Concatenate (const std::string& root,
     return result.str();
 }
 
+bool
+amrex::is_next_non_space(std::istream& is, char c)
+{
+    // save current position
+    const int pos = is.tellg();
+    const int cseek = static_cast<int>(c);
+
+    bool found = false;
+    int ic;
+
+    // search for first non-whitespace char and test if c
+    while (! is.eof()) {
+      ic = is.peek();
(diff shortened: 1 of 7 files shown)


Src/Base/AMReX_Utility.cpp
```diff
@@ -145,6 +145,34 @@ amrex::Concatenate (const std::string& root,
     return result.str();
 }
 
+bool
+amrex::is_next_non_space(std::istream& is, char c)
```
WeiqunZhang, line 149: I am sure this works.  But C++ provides a simpler solution,  `std::ws`.  https://en.cppreference.com/w/cpp/io/manip/ws

You want want too look at how we do a similar thing here.  https://github.com/AMReX-Codes/amrex/blob/development/Src/Base/AMReX_IntVect.cpp#L42

dwillcox:
Thank you for your recommendation @WeiqunZhang ! Closing this draft PR for now and will re-open again when needed.

PR #1088, opened 2020-07-03
fix rereflux at domain boundary: ghost cells outside non-periodic dom…

WeiqunZhang:
…ain should not participate rereflux


WeiqunZhang:
@jrood-nrel

asalmgren (approved):

drummerdoc (commented):
Looks good

PR #1094, opened 2020-07-04
For nvcc >= 10.2, tell it to forward unknowns to host compiler, and we…

### LINKED ISSUE #1093: nvcc and mpicxx
I don't see a way to fix this WarpX issue without major changes to the make system. https://github.com/ECP-WarpX/WarpX/issues/1144  The issue is nvcc does not recognize the `-pipe` flags passed by mpicxx to gcc. 

Currently we are setting environment variables such as `OMPI_CXX` and `MPICH_CXX` so that the `mpicxx` will use nvcc as underlying compilers.  For OpenMPI family, we further set `OMPI_CXXFLAGS` to substitute `-pthread` by `-Xcompiler -pthread` for nvcc.  However, we do not have a way to do that for MPICH and MVAPICH.  Looking through mvapich's mpicxx bash script, it seems that the compiler flags cannot be altered through an environment variable.

Maybe we need to use our own nvcc wrapper so that the flags passed from mpicxx can be sanitized for nvcc.

P.S.  CMake does not work either for WarpX.

WeiqunZhang:
… no longer need to sanitize options for nvcc.

### DIFF
--- a/Tools/GNUMake/Make.defs
+++ b/Tools/GNUMake/Make.defs
@@ -1036,40 +1036,43 @@ else ifeq ($(USE_CUDA),TRUE)
 
     DEFINES += -DAMREX_GPU_MAX_THREADS=$(CUDA_MAX_THREADS)
 
-    comm := ,
-    space :=
-    space +=
-
     ifneq ($(LINK_WITH_FORTRAN_COMPILER),TRUE)
       LINKFLAGS = $(NVCC_FLAGS)
-      # we are using nvcc for linking
-      libraries := $(subst -Wl$(comm),-Xlinker=,$(libraries))
-      ifeq ($(FIX_NVCC_PTHREAD),TRUE)
-        libraries := $(subst -pthread,-Xcompiler$(space)-pthread,$(libraries))
-      endif
       AMREX_LINKER = nvcc
     endif
 
-    ifeq ($(USE_MPI),TRUE)
+    ifeq ($(nvcc_forward_unknowns),0)
 
-      ifneq ($(findstring Open MPI, $(shell mpicxx -showme:version 2>&1)),)
+      comm := ,
+      space :=
+      space +=
 
+      ifeq ($(AMREX_LINKER),nvcc)
+        libraries := $(subst -Wl$(comm),-Xlinker=,$(libraries))
         ifeq ($(FIX_NVCC_PTHREAD),TRUE)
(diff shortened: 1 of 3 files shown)


WeiqunZhang:
It seems that we only need to use `--forward-unknown-to-host-compiler`, but not `--forward-unknown-to-host-linker`.  If the latter is used, unknown options at link time such as `-pthread` will be forwarded to `ld` instead of `gcc`.  That actually causes `ld` to fail.

Tools/GNUMake/comps/nvcc.mak
```diff
@@ -11,6 +11,13 @@ ifeq ($(nvcc_major_lt_8),1)
   $(error Your nvcc version is $(nvcc_version). This is unsupported. Please use CUDA toolkit version 8.0 or newer.)
 endif
 
+nvcc_forward_unknowns = 0
+ifeq ($(shell expr $(nvcc_major_version) \>= 10),1)
```
maximumcats: I think this check needs to be fixed so it's correct for 11.0+.

WeiqunZhang: You are absolutely right.

Tools/GNUMake/Make.defs
```diff
@@ -1036,40 +1036,43 @@ else ifeq ($(USE_CUDA),TRUE)
...
+          ifeq ($(FIX_NVCC_PTHREAD),TRUE)
+            OMPI_CXXFLAGS_ORIG = $(shell mpicxx -showme:compile)
+            export OMPI_CXXFLAGS := $(subst -pthread,-Xcompiler$(space)-pthread,$(OMPI_CXXFLAGS_ORIG))
 
-      endif
+            OMPI_CFLAGS_ORIG = $(shell mpicc -showme:compile)
+            export OMPI_CFLAGS := $(subst -pthread,-Xcompiler$(space)-pthread,$(OMPI_CXXFLAGS_ORIG))
 
-    endif
+            OMPI_FCFLAGS_ORIG = $(shell mpif90 -showme:compile)
```
maximumcats: Unfortunately, it seems that this is still necessary. I tested this and things work fine now with nvcc/mpicxx, but on cgpu I now have this problem with mpif90 and pgfortran. It seems FIX_NVCC_PTHREAD was inadvertently saving us there.

WeiqunZhang: OK. Then we still have to do that for Fortran.

maximumcats (approved):

PR #1097, opened 2020-07-04
Document GPUS_PER_SOCKET and GPUS_PER_NODE

maximumcats:


zingale (approved):

Docs/sphinx_documentation/source/GPU.rst
```diff
@@ -1347,6 +1347,32 @@ AMReX for GPUs:
...
 
+* Pay attention to what GPUs your job scheduler is assigning to each MPI
+  rank. In most cases you'll achieve the best performance when a single
+  MPI rank is assigned to each GPU, and has boxes large enough to saturate
+  that GPU's compute capacity. While there are some cases where multiple
+  MPI ranks per GPU can make sense (typically this would be when you have
+  some portion of your code that is not GPU accelerated and want to have
+  many MPI ranks to make that part faster), this is probably the minority
+  of cases. For example, on OLCF Summit you would want to ensure that your
+  resource sets contain one MPI rank and GPU each, using `jsrun -n N -a 1 -c 1 -g 1`,
```
WeiqunZhang: Do we want to use `-c 7` instead of `-c 1`?  This is `-c 1` https://jsrunvisualizer.olcf.ornl.gov/?s1f0o01n6c1g1r11d1b1l0= and this is `-c 7`  https://jsrunvisualizer.olcf.ornl.gov/?s1f0o01n6c7g1r11d1b1l0=

Also, I usually run with `jsrun -r 6 -a 1 -g 1 -c 7 -b rs` without specifying `-n`.  It seems to work.  Is this an undocumented feature?  Cannot find it anything on this in the manual.  The advantage of this is one only needs to modify in one place, `#BSUB -nnodes xxx`, to run a job with a different number of processes.

maximumcats: `-c 7` would benefit you if you were using OpenMP threads, sure. I am only commenting on MPI<->GPU mappings here, but could modify this however you want me to.

It's intended to work that you can use `-r 6` instead of `-n`.

WeiqunZhang: Thanks.

```suggestion
  resource sets contain one MPI rank and GPU each, using `jsrun -n N -a 1 -c 7 -g 1`,
```

PR #1103, opened 2020-07-06
CMake: re-factor parallel backends setup

### LINKED ISSUE #944: AMReX CMake error with MPI::MPI_CXX IMPORTED_GLOBAL
When using AMReX as a sub-package within an application's `CMakeLists.txt` (e.g., for amr-wind), CMake generates an error if the application does `find_package(MPI REQUIRED)` before adding AMReX. The exact error is

```
CMake Error at submods/amrex/Src/Base/CMakeLists.txt:15 (set_target_properties):
  Attempt to promote imported target "MPI::MPI_CXX" to global scope (by
  setting IMPORTED_GLOBAL) which is not built in this directory.
```

mic84:
Centralize setup of all parallel backends dependencies. 

Note that:
1) "IMPORTED_GLOBAL" property is not forced on imported targets anymore: this resolve #944
2)  Because of 1), get_target_properties_flattened() may return junk when trying to recursively inquire imported targets. This will need to be fixed somehow


ax3l (approved):
Thank you! This LGTM, anything still open with respect to comment 2) in the PR description?

mic84:
2) will still need to be addressed somehow but it's not high priority (it affects terminal output and `--describe` feature only).

PR #1104, opened 2020-07-07
CMake: remove spurious endif()

mic84:


Src/Extern/Conduit/CMakeLists.txt
```diff
@@ -34,5 +34,4 @@ elseif(ENABLE_CONDUIT)
     else ()
     target_link_libraries(amrex PUBLIC conduit::conduit)
     set_target_properties(conduit::conduit PROPERTIES IMPORTED_GLOBAL True )
-    endif ()
```
ax3l, line 37: isn't this needed... in line 34 is the `else()`

PR #1116, opened 2020-07-08
Nodal bndryfunc

WeiqunZhang:
CpuBndryFunc and GpuBndryFunc for nodal data.


emotheau (approved):

emotheau:
It seems to work with PeleLM with CUDA-TRUE, so I suppose the PR is ok

drummerdoc (approved):

PR #1127, opened 2020-07-09
AppleClang, MSVC: Avoid Malformed Config.H

### LINKED ISSUE #1124: AMReX CMake: Bad AMReX_Config.H with AppleClang
Executing `make install` with a CMake configured project using AppleClang ends up with a malformed `AMReX_Config.H` header file. 

When building against MacOS default LLVM/Clang, CMake sets `CMAKE_CXX_COMPILER_ID = AppleClang` this is not checked in the install helpers https://github.com/AMReX-Codes/amrex/blob/fd4e0d1078a489fe9f07f0d4aa9de77f1680ed2c/Tools/CMake/AMReXInstallHelpers.cmake#L26-L36 and  `COMPILER_ID_MACRO` variable ends up empty, and the `AMReX_Config.H` file that is generated ends up with this 

```C++
#ifdef __cplusplus
#ifndef
static_assert(false,"libamrex was built with AppleClang");
#endif
#ifdef _OPENMP
static_assert(false,"libamrex was built without OpenMP");
#endif
#endif
```

and ends up with a compile time error when this file is included in another project

```
In file included from /exawind/install/clang/amrex/include/AMReX.H:4:
/exawind/install/clang/amrex/include/AMReX_Config.H:46:9: error: macro name missing
#ifndef
        ^
```

cc: @ax3l @mic84

ax3l:
Identify AppleClang as LLVM for AMReX compiler defines. Also fix broken header file generated for MSVC.

Fix #1124 cc @sayerhs (thank you for the report!)

### DIFF
--- a/Tools/CMake/AMReXInstallHelpers.cmake
+++ b/Tools/CMake/AMReXInstallHelpers.cmake
@@ -39,9 +39,8 @@ function ( generate_amrex_config_header )
       set(COMPILER_ID_MACRO  _MSC_VER)
    else ()
       # avoid malforming the AMReX_Config.H file by setting something valid
-      set(COMPILER_ID_MACRO  __cplusplus)
-      message(WARNING "Compiler '${CMAKE_CXX_COMPILER_ID}' is not tested by AMReX developers! "
-                      "Will not be able to validate that the same compiler is picked in user-code.")
+      message(FATAL_ERROR "Compiler '${CMAKE_CXX_COMPILER_ID}' not supported by AMReX developers! "
+                      "Try to configure with -DALLOW_DIFFERENT_COMPILER=ON")
    endif ()
 
    if (ENABLE_OMP)
--- a/Tools/CMake/AMReXInstallHelpers.cmake
+++ b/Tools/CMake/AMReXInstallHelpers.cmake
@@ -33,6 +33,15 @@ function ( generate_amrex_config_header )
       set(COMPILER_ID_MACRO  __PGI)
    elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" )
       set(COMPILER_ID_MACRO  __llvm__)
+   elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "AppleClang" )
+      set(COMPILER_ID_MACRO  __llvm__)
+   elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC" )
+      set(COMPILER_ID_MACRO  _MSC_VER)
+   else ()
+      # avoid malforming the AMReX_Config.H file by setting something valid
+      set(COMPILER_ID_MACRO  __cplusplus)
+      message(WARNING "Compiler '${CMAKE_CXX_COMPILER_ID}' is not tested by AMReX developers! "
+                      "Will not be able to validate that the same compiler is picked in user-code.")
    endif ()
 
    if (ENABLE_OMP)


Tools/CMake/AMReXInstallHelpers.cmake
```diff
@@ -33,6 +33,13 @@ function ( generate_amrex_config_header )
       set(COMPILER_ID_MACRO  __PGI)
    elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" )
       set(COMPILER_ID_MACRO  __llvm__)
+   elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "AppleClang" )
+      set(COMPILER_ID_MACRO  __llvm__)
+   else ()
+      # avoid malforming the AMReX_Config.H file by setting something
+      set(COMPILER_ID_MACRO  ${CMAKE_CXX_COMPILER_ID})
```
ax3l, line 40: I guess if we fallback to guessing `__llvm__` for new compilers we are 90% of the time right. Shall I set accordingly?

Tools/CMake/AMReXInstallHelpers.cmake
```diff
@@ -33,6 +33,13 @@ function ( generate_amrex_config_header )
...
+      message(WARNING "Compiler '${CMAKE_CXX_COMPILER_ID}' is not tested by AMReX developers! "
+                      "Here be dragons.")
```
ax3l: I guess if we fallback to guessing `__llvm__` for new compilers we are 90% of the time right (DPC++, HIP, other SYCL compilers, ...).

Shall I set accordingly?
```suggestion
      set(COMPILER_ID_MACRO __llvm__)
      message(WARNING "Compiler '${CMAKE_CXX_COMPILER_ID}' is not tested by AMReX developers! "
                      "Assuming some kind of LLVM fork...")
```

sayerhs: @ax3l I don't think Line 40 will work. For example, if lines 36-37 weren't present then for AppleClang this would end up as

```C++
#ifndef AppleClang
static_assert(false,"libamrex was built with AppleClang");
#endif
```

While the file is not malformed, since it is not defined it will end up with a `static_assert` failure?

ax3l: Oh that's right, thanks!
will then apply the LLVM fallback, more discussion below in the main thread.

ax3l:
I have to admit, the check here is a bit of snakeoil anyway:

- it is overly strict: some compilers are intentionally ABI compatible, e.g. with a system GCC
- it is not safe: a major version change in the same compiler can still create ABI problems
- it is incomplete: we have to update AMReX for every new compiler users, including compiler vendors in e.g. SPEC, want to use [some defines here]( https://sourceforge.net/p/predef/wiki/Compilers/)

It's a bit of a workflow/policy decision what we want to do here... cc @WeiqunZhang @asalmgren @mic84 

One workflow I could think of:
Skip the check in `Config.H` altogether when the compiler is unknown but throw a `#warning` string instead and also warn at configuration time?

WeiqunZhang:
The `configure` script has an option
```
  --allow-different-compiler {yes,no}
                        Allow an application to use a different compiler than
                        the one used to build libamrex [default=no]
```

I don't know if that is implemented in cmake system.

ax3l:
Okay, that answers the question. We can just implement such a switch and error out if unset - telling a user that switch exists.

Documented as potential improvement in #1129

sayerhs (approved):

Tools/CMake/AMReXInstallHelpers.cmake
```diff
@@ -33,6 +33,15 @@ function ( generate_amrex_config_header )
...
       set(COMPILER_ID_MACRO  __llvm__)
+   elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "AppleClang" )
+      set(COMPILER_ID_MACRO  __llvm__)
+   elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC" )
+      set(COMPILER_ID_MACRO  _MSC_VER)
+   else ()
+      # avoid malforming the AMReX_Config.H file by setting something valid
+      set(COMPILER_ID_MACRO  __cplusplus)
+      message(WARNING "Compiler '${CMAKE_CXX_COMPILER_ID}' is not tested by AMReX developers! "
+                      "Will not be able to validate that the same compiler is picked in user-code.")
```
mic84:
```suggestion
      # avoid malforming the AMReX_Config.H file by setting something valid
      message(FATAL_ERROR "Compiler '${CMAKE_CXX_COMPILER_ID}' not supported by AMReX developers! "
                      "Try to configure with -DALLOW_DIFFERENT_COMPILER=ON")
```
Let's do this to be consistent with what `configure,py` does. I will shortly push a PR to enable the ALLOW_DIFFERENT_COMPILER feature.

mic84 (approved):

Tools/CMake/AMReXInstallHelpers.cmake
ax3l: Thanks! :+1:

WeiqunZhang:
It's also not as restrict as it sounds like.  If amrex is compiled with GCC, the check will be on `__GNUC__`.  Most compilers define that.

ax3l:
That's true on top. All in all it is not that effective in what it tries to achieve.

PR #1130, opened 2020-07-09
CMake: allow an application to use a different compiler than the one used to build AMReX

### LINKED ISSUE #1129: CMake: Allow Unknown Compiler
Follow-up to https://github.com/AMReX-Codes/amrex/pull/1127#issuecomment-656325272:

We should implement a `AMReX_ALLOW_DIFFERENT_COMPILER`/`AMReX_UNKNOWN_COMPILER` switch (or similar) that skips the checks in `AMReX_Config.H`. This workflow is relevant for compiler vendors and other experimental folks trying, testing and developing new compilers against AMReX.

Correspondingly, the `configure` script provides the option:
```
  --allow-different-compiler {yes,no}
                        Allow an application to use a different compiler than
                        the one used to build libamrex [default=no]
```

mic84:
Resolves #1129


Tools/CMake/AMReXInstallHelpers.cmake
```diff
@@ -23,25 +23,28 @@ function ( generate_amrex_config_header )
...
+           set(COMPILER_ID_MACRO  __llvm__)
+       elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "AppleClang" )
+           set(COMPILER_ID_MACRO  __llvm__)
+       elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC" )
+           set(COMPILER_ID_MACRO  _MSC_VER)
+       else ()
+           message(FATAL_ERROR "Compiler '${CMAKE_CXX_COMPILER_ID}' not supported by AMReX developers! "
+               "Try to configure with -DALLOW_DIFFERENT_COMPILER=ON")
+       endif ()
+       set(COMP_DECLS "\n#ifndef ${COMPILER_ID_MACRO}\nstatic_assert(false,\"libamrex was built with ${CMAKE_CXX_COMPILER_ID}\");\n#endif")
```
WeiqunZhang: Can we inform the user the allow_different_compiler option in static assert's error message?

mic84: Sure. Should we change that in the configure/GNUMake as well?

mic84: Actually there is no need to inform the user of the option in the static assert since the configure step will exit with a fatal error if ALLOW_DIFFERENT_COMPILER=off when an unknown compiler is used. This is consistent with what `configure.py` does

WeiqunZhang: But if the user is not using cmake to build app but using cmake to build amrex, they will not see it.

If you can modify the `mkconfig.py`, that would be great.  There the message is different. `./configure --allow_different_compiler=yes`.

mic84: I modified the error message according to your suggestion.

WeiqunZhang: Thanks!  Looks like the modification of `mkconfig.py` is not finished yet.

mic84: Oops. Should be fixed now

Tools/libamrex/mkconfig.py
```diff
@@ -29,36 +29,35 @@ def doit(defines, undefines, comp, allow_diff_comp, use_omp):
...
-            print("#ifndef _CRAYC")
-            print('static_assert(false,"libamrex was built with Cray");')
-            print("#endif")
+            comp_macro = "_CRAYC"
+            comp_id    = Cray
         elif comp == "pgi":
-            print("#ifndef __PGI")
-            print('static_assert(false,"libamrex was built with PGI");')
-            print("#endif")
+            comp_macro = "_PGI"
```
WeiqunZhang: Should have two underscores here.

WeiqunZhang (approved):

Tools/CMake/AMReXInstallHelpers.cmake
```diff
@@ -23,25 +23,29 @@ function ( generate_amrex_config_header )
 
    endforeach ()
 
-   if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" )
```
ax3l, line 26: If you want to avoid C++ code generated from vars and nested of `if`s it could be so easy:
```cmake
       if (ALLOW_DIFFERENT_COMPILER)
           set(COMPILER_ID_MACRO __cplusplus)
       elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" )
           set(COMPILER_ID_MACRO __GNUC__)
```

PR #1131, opened 2020-07-09
Tools: add NVIDIA HPC SDK as a compiler

bcfriesen:
HPC SDK replaced PGI after PGI 20.4, so we should get it working now.


kngott (approved):
This looks like a thorough first pass. I think we should approve it and give it a few test compiles on CoriGPU.

PR #1133, opened 2020-07-10
remove ifdef AMREX_USE_GPU from nodal GpuBndryFunc so that it works f…

WeiqunZhang:
…or CPU build


emotheau (approved):
Now GpuBndryFuncFab enters into the user function for cell-center and nodal data, and for both GPU and CPU, so I think we are good

drummerdoc (approved):

PR #1135, opened 2020-07-12
Fused GPU kernel launches

WeiqunZhang:


WeiqunZhang:
This is ready for review.  Note that this is not turned on any kernels yet.  We will do that and integration with `MFIter` later.  For a test at `amrex/Tests/GPU/Fuse`, a 10x speedup is seen on both 2d and 3d on v100.

Src/Base/AMReX_GpuFuse.H
```diff
@@ -0,0 +1,267 @@
...
+{
+    new (buf) Lambda(f);
+    return Arena::align(sizeof(Lambda));
+}
+
+template <typename Lambda>
+void
+PutLambda1DLauncher (FuseHelper* helper)
+{
+    static Lambda1DLauncher fp = nullptr;
```
maximumcats, line 81: This will not be thread-safe if we use OpenMP, right? We might want to have some sort of assert that prevents us from making that mistake.

WeiqunZhang: Yes, we need more assertions.  There is only one in the constructor.  But that's not enough especially because we are going to use a singleton.  I will add more assertions to `Register` and `Launch`.

WeiqunZhang:
I did more performance comparison of `MultiFab a = MultiFab b + MultiFab c` on a single v100.  Unsurprisingly, fused launch is always losing if there is only one kernel.  For Box size of 32^3 cells and below,  fused launch wins unless there is only one kernel.  Even with two 32^3 cells kernels, fused launch wins.
```
| total # cells | Box size | fused time | unfused time | Speedup |
|---------------+----------+------------+--------------+---------|
|         256^3 |    256^3 |  0.0005288 |    0.0005082 |    0.96 |
|               |    128^3 |  0.0005668 |    0.0005158 |    0.91 |
|               |     64^3 |   0.000655 |    0.0005971 |    0.91 |
|               |     32^3 |  0.0008538 |     0.002013 |     2.4 |
|               |     16^3 |   0.001343 |      0.01572 |    11.7 |
|               |      8^3 |   0.005214 |       0.1279 |    24.5 |
|---------------+----------+------------+--------------+---------|
|         128^3 |    128^3 |   8.815e-5 |     7.593e-5 |    0.86 |
|               |     64^3 |    9.52e-5 |     8.094e-5 |    0.85 |
|               |     32^3 |  0.0001192 |    0.0002641 |     2.2 |
|               |     16^3 |  0.0001873 |     0.002002 |    10.7 |
|               |      8^3 |  0.0005555 |      0.01589 |    28.6 |
|---------------+----------+------------+--------------+---------|
|          64^3 |     64^3 |   2.581e-5 |     1.698e-5 |    0.65 |
|               |     32^3 |   2.975e-5 |     4.133e-5 |     1.3 |
|               |     16^3 |   3.626e-5 |    0.0002594 |     7.2 |
|               |      8^3 |   9.284e-5 |     0.002012 |    21.7 |
|---------------+----------+------------+--------------+---------|
|          32^3 |     32^3 |   1.971e-5 |     1.347e-5 |    0.68 |
|               |     16^3 |   2.169e-5 |     4.116e-5 |     1.9 |
|               |      8^3 |   2.835e-5 |    0.0002592 |     9.1 |
|---------------+----------+------------+--------------+---------|
|      32x32x64 |     32^3 |   2.114e-5 |     2.435e-5 |     1.2 |
```

Src/Base/AMReX_GpuFuse.cpp
```diff
@@ -0,0 +1,241 @@
...
+        std::size_t offset_objects = Arena::align(offset_helpers+sizeof_helpers);
+        std::size_t sizeof_objects = m_nbytes_used_lambda_buf;
+        std::size_t total_buf_size = offset_objects + sizeof_objects;
+
+        char* h_buffer = (char*)The_Pinned_Arena()->alloc(total_buf_size);
+        char* d_buffer = (char*)The_Device_Arena()->alloc(total_buf_size);
+
+        std::memcpy(h_buffer, nwarps, sizeof_nwarps);
+        std::memcpy(h_buffer+offset_helpers, m_helper_buf, sizeof_helpers);
+        std::memcpy(h_buffer+offset_objects, m_lambda_buf, sizeof_objects);
```
atmyers, line 78: As @WeiqunZhang and I discussed offline, this seems to work even though the extended host / device lambda is not technically trivial copyable. @WeiqunZhang convinced me that this will probably have to keep working in the future, but maybe it's something we should keep an eye on.

atmyers (approved):

Src/Base/AMReX_GpuFuse.cpp
maximumcats: I talked to our compiler team about this and let them know that they should keep an eye on this too. I think this particular usage is probably OK.

WeiqunZhang: Thanks!

Src/Base/AMReX_GpuFuse.H
```diff
@@ -0,0 +1,280 @@
...
+
+        static_assert(std::alignment_of<Lambda>::value <= Arena::align_size,
+                      "GpuFuse: alignment error");
+        if (m_nbytes_used_lambda_buf + Arena::align(sizeof(Lambda)) > m_nbytes_lambda_buf) {
+            resize_lambda_buf();
+        }
+        char* p_lambda = m_lambda_buf + m_nbytes_used_lambda_buf;
+        std::size_t sz_lambda = PutLambda<Lambda>(f, p_lambda);
+        m_nbytes_used_lambda_buf += sz_lambda;
+        AMREX_ASSERT(m_nbytes_used_lambda_buf <= m_nbytes_lambda_buf);
```
kngott, line 225: If I'm reading this correctly, this assertion is unnecessary, given the if statement check above and the resize_lambda_buf function does what it's designed to do. If it's possible the resize has the possibility of not increasing the buffer enough, would it make sense to pass the resize a minimum size to add to guarantee success?

Src/Base/AMReX_GpuFuse.cpp
```diff
@@ -0,0 +1,241 @@
...
+#include <AMReX_OpenMP.H>
+
+namespace amrex {
+namespace Gpu {
+
+#ifdef AMREX_USE_CUDA
+
+namespace {
+    bool s_in_fuse_region = false;
+    // Fusing kernels with elements greater than this are not recommended based tests on v100
```
kngott, line 13: Does inFuseRegion do anything yet? Searching s_in_fuse_region and inFuseRegion, as well as checking Register and Launch, I'm not seeing where it's used.

Also, with the explicit writing style shown in the example, do we want the default to be off?

Src/Base/AMReX_GpuFuse.cpp
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    The_Pinned_Arena()->free(m_lambda_buf);
+    The_Pinned_Arena()->free(m_helper_buf);
+}
+
+void Fuser::Launch ()
+{
+    BL_PROFILE("Fuser::Launch()");
+
```
kngott, line 42: I wouldn't leave this timer here. In an actual use case with potentially tons of these calls, I'm not sure what it captures of interest. It might also be deep enough that it could get called an inordinate number of times.

Src/Base/AMReX_GpuFuse.cpp
```diff
@@ -0,0 +1,241 @@
...
+                        }
+                    }
+                }
+            }
+        });
+        Gpu::synchronize();
+        The_Pinned_Arena()->free(nwarps);
+        The_Pinned_Arena()->free(h_buffer);
+        The_Device_Arena()->free(d_buffer);
+
```
kngott, line 149: Would it make sense to try to remove this synch and pushback the cleanup into a callback function? Or at least make synching optional, to match the MFIter loop?

Src/Base/AMReX_GpuFuse.H
WeiqunZhang: Yes, you could argue the assertion is unnecessary.  Like a lot of assertions we do, it's just there to rule out possibilities if the program does not actually do what we think it is supposed to do.   And the assertion is there for debug build only.  Yes, it's pretty much impossible the resize does not increase the buffer enough.

Src/Base/AMReX_GpuFuse.cpp
WeiqunZhang: Not yet.  I have a branch using it waiting on this PR.

Src/Base/AMReX_GpuFuse.cpp
```diff
@@ -0,0 +1,241 @@
...
+int getFuseNumKernelsThreshold () { return s_fuse_numkernels_threshold; }
+
+int
+setFuseNumKernelsThreshold (int new_threshold)
+{
+    int old = s_fuse_numkernels_threshold;
+    s_fuse_numkernels_threshold = new_threshold;
+    return old;
+}
+
```
kngott, line 228: I assume these get and set Threshold functions, plus the FuseSafeGuard are for an up-and-coming integration with the MFIter loop (add kernel launches that fit the parameters and launch in the destructor).

For users to be able to control the launches themselves, access to the current contents of the Fused Launch would be needed. (Number of launches awaiting launch in the fused region + some size metric, such as total number of points/threads in the launch thus far?)

Src/Base/AMReX_GpuFuse.cpp
WeiqunZhang: For the Castro test I did, it got called 7878 times, but `FillBoundary` got called `17816` times. `Fsmooth` got called 16664 times.  `getFB` got called 35593 times.  I think it is of interest to know how much time we spend in those fused launches.

Src/Base/AMReX_GpuFuse.cpp
WeiqunZhang: In the branch that has not been submitted for PR, the `Launch` is called in the destructor of MFIter.  This PR is to get the basic fuse capability into development.  We can worry about how it's integrated into MFIter etc. later.

Src/Base/AMReX_GpuFuse.cpp
WeiqunZhang: We can discuss in the later PR on integrating fuse into MFIter.

kngott:
I think Weiqun answered my questions on this one. Are we just waiting on @maxpkatz? Or is he also content with what's here?

Src/Base/AMReX_GpuFuse.cpp
WeiqunZhang: Thanks for your comments.  But I want this PR to be focused on whether the approach of sending user's lambdas to gpu ourselves and calling them through function pointers is implemented correctly.

kngott (approved):

Src/Base/AMReX_GpuFuse.cpp
kngott: Oh, that was my first pass through this. It looks solid to me. I learned some things too. (e.g. I don't think I've ever seen std::decay in action before).  :)
This might be going in a direction that there aren't a lot of experts for (except, maybe the RAJA developers who are doing something similar), so we may only really be able to determine if there's a problem (some edge case in the standard or compiler bugs) by solidly testing it.

WeiqunZhang: We use `std::decay` a lot in `AMReX_EB2_IF_*.H` and `amrex::GpuTuple`.  :)

Src/Base/AMReX_GpuFuse.cpp
maximumcats: Yeah, I'd like to leave this timer in here for now -- if it gets to be too much we can always remove it later.

maximumcats:
The register and launch functions have no implementation yet when `USE_CUDA = FALSE`; is your intent that we would only ever use these functions in GPU code, or should we have a CPU implementation that is basically the equivalent of LoopConcurrent?

WeiqunZhang:
My intention is we only need to use register and launch functions in GPU code only, and even in GPU code we usually should not need to use register and launch directly.  I have added new versions of ParallelFor and macros (not in this PR) that users can provide a hint that the kernels are fusable.  The runtime can then decide whether to fuse them or not depending on the size.

maximumcats (approved):

PR #1137, opened 2020-07-13
Link with MPI compiler, not nvcc

maximumcats:
When we're building for MPI, we should retain the strategy of linking with the C++ wrapper (e.g. mpicxx) just as we compile with with that wrapper. The wrapper will still call nvcc under the hood for linking. This will help on issues like ECP-WarpX/WarpX#1132.


maximumcats:
Closing stale PR, this was intended to address issues on Cori GPU but we can revisit this if there are still issues on Perlmutter.

PR #1138, opened 2020-07-14
Fix compilation of AMReX_CommProfStats.cp for non-3D builds

mic84:
When building with DIM other then 3, compilation of AMReX_CommProfStats.cpp fails because IntVect is always used in 3D form.
Fix that by wrapping IntVec arguments in AMREX_D_DECL.


atmyers (approved):

kngott:
Why did the BL_SPACEDIM above each of those locations not catch it? Shouldn't that section of code simply not be compiled when SPACEDIM != 3? Does that ifdef need to be changed to (AMREX_SPACEDIM != 3)?

Or is the error only showing up when SPACEDIM == 1 (or 4)?

PR #1126, opened 2020-07-09
FORTLINK: Guess if Unknown

ax3l:
Guess the FORTLINK API convention is `UNDERSCORE` for unknown values, unless it's the classic (non-clang) IBM Fortran compiler, where it's `LOWERCASE`.

@JBlaschke encountered this on macOS 10.14.6 with GNU 8.4.0 (from homebrew).

PR #1140, opened 2020-07-15
work with ax3l: fail when fortran c interface could not be found

JBlaschke:
Following #1126, worked with @ax3l to make cmake give a informative error message and then fail (always a good thing to check return codes)

### DIFF
--- a/Tools/CMake/AMReX_Defines.cmake
+++ b/Tools/CMake/AMReX_Defines.cmake
@@ -93,6 +93,9 @@ function ( set_amrex_defines )
       # Fortran/C mangling scheme
       #
       include( FortranCInterface )
+      if(NOT FortranCInterface_GLOBAL_FOUND)
+          message(FATAL_ERROR "Failed to find the Fortan C Interface -- check the CMakeError.log")
+      endif()
       include( ${FortranCInterface_BINARY_DIR}/Output.cmake )
 
       set( FORTLINK "" )


ax3l (approved):
I am @ax3l and I approve this message.

mic84 (approved):

Tools/CMake/AMReX_Defines.cmake
```diff
@@ -93,6 +93,9 @@ function ( set_amrex_defines )
       # Fortran/C mangling scheme
       #
       include( FortranCInterface )
+      if(NOT FortranCInterface_GLOBAL_FOUND)
+          message(FATAL_ERROR "Failed to find the Fortan C Interface -- check the CMakeError.log")
```
ax3l, line 97: Oh dang @JBlaschke: Fort**r**an

JBlaschke: Eh @ax3l -- if they wanted people to spell the name right, they would have made it a single-letter

ax3l: as long as it's all caps I am in.
F

JBlaschke: As long as J will be "Julia" instead of "Java", then count me in!

ax3l:
Founds something random about IBM XL today with respect to this. Just to document here:
> IBM xlf does not mangle Fortran symbols by default,use -qextnameto add trailing underscores.

https://www.olcf.ornl.gov/wp-content/uploads/2018/12/summit_workshop_20181203_programming_environment.pdf

PR #1142, opened 2020-07-15
fix clamping in enforcePeriodic

atmyers:
@jmsexton03 could you test this and see if it fixes your lost particle issue?

The enforcePeriodic function should always return values in the range [plo, phi). This is not the case in the current implementation, because the machine epsilon refers to the difference between 1.0 and the next representable number for a given floating point type, but that does not mean that, say, 1400.0 - eps is distinct from 1400.0. 

The proposed implementation uses `std::nextafter` to obtain the closest representable number less than phi. I tested this on Summit and it does seem to work in device code, even with cuda 9.2. I also did a test (using clang) that verified the proposed implementation does the right thing for all representable floats in the range plo - 1.0 to phi + 1.0;


jmsexton03:
This fixes the lost particle in my small reproducer.

WeiqunZhang (approved):

PR #1145, opened 2020-07-16
Allow specification of number of amr levels for MLLinOp and MLNodeLinOp

bsrunnels:
Currently `m_num_amr_levels` is determined by the size of the `geom` object. However, sometimes fewer AMR levels might be needed, for instance, if this is used within an AmrCore where `finest_level` is lower than `num_amr_levels`.


ChrisDeGrendele:
This will help fix https://github.com/AMReX-Astro/MAESTROeX/issues/203

bsrunnels:
Updated to include the same fix for `MLCellLinOp`.

WeiqunZhang:
I think people should use these functions in AmrMesh, which is the base of AmrCore (and Maestro I think), `Vector<Geometry> Geom (int a_coarsest_lev, int a_finest_lev)`, `Vector<BoxArray> boxArray (int a_coarsest_lev, int a_finest_lev)` and `Vector<DistributionMapping> DistributionMap (int a_coarsest_lev, int a_finest_lev)`.

Note that sometimes we need to solve multi-level problems not starting from level 0.  The PR will not solve that problem.

bsrunnels:
Oh ok, so in this case use

     linop.define(Geom(0,finest_level), grids, DistributionMap(0,finest_level), info).

That works for me - thanks.

PR #1148, opened 2020-07-16
Error on implicit capture of this by extended lambda by default

WeiqunZhang:


maximumcats (approved):

Tools/GNUMake/comps/nvcc.mak
```diff
@@ -134,6 +134,16 @@ ifeq ($(nvcc_forward_unknowns),1)
   NVCC_FLAGS += --forward-unknown-to-host-compiler
 endif
 
+ifeq ($(shell expr $(nvcc_major_version) \>= 11),1)
+ifeq ($(GPU_ERROR_CAPTURE_THIS),TRUE)
```
maximumcats, line 138: This won't work for the 11.0 RC, but that's fine with me.

WeiqunZhang: Yes, I assume nobody is going to use 11.0 RC.

PR #1149, opened 2020-07-17
CMake: implement ability to export build tree

mic84:


mic84:
@sayerhs try this out and let me know if it works for you.

sayerhs (approved):

ax3l (approved):
Thank you both, great work! :sparkles:

Tools/CMake/AMReXInstallHelpers.cmake
```diff
@@ -80,20 +80,30 @@ function (install_amrex)
    # Write and install configure file
    include(CMakePackageConfigHelpers)
 
+   #
+   # Setup for install and export of build tree
+   #
+   set(CMAKE_FILES_DIR   lib/cmake/AMReX)   # Relative path to top level installation/build-tree
```
ax3l: for installation, that path is only valid on Unix:
https://cmake.org/cmake/help/latest/command/find_package.html#search-procedure
```suggestion
   # Relative path to top level installation/build-tree
   if(WIN32)
       set(CMAKE_FILES_DIR   cmake)   
   else()
       set(CMAKE_FILES_DIR   lib/cmake/AMReX)
   endif()
```

ax3l: I think one usually sets them via changed defaults in a cached `CMAKE_INSTALL_CMAKEDIR` variable:
https://github.com/openPMD/openPMD-api/blob/ddc673bcd96d28293a11cfdbbc8fdcd57c154385/CMakeLists.txt#L38

ax3l (approved):

PR #1150, opened 2020-07-17
Add interfaces for bottom solve Hypre configuration

zjibben:
We use MLMG as an interface to Hypre to use as a preconditioner. This means we want to set up Hypre and then apply the solver to many different vectors, and only pay for a new Hypre setup call when we change the matrix. We also want a slightly different Hypre configuration than what is currently in AMReX.

This PR introduces some MLMG interfaces for a user to decide whether to preserve the Hypre solver (don't delete it at every solve call) and change the Hypre configuration. I've tried to maintain AMReX's default behavior, so if none of the new MLMG methods are called, the current behavior remains the same.

Please let me know if this looks alright / if you would like any changes. Thanks!

### DIFF
--- a/Src/LinearSolvers/MLMG/AMReX_MLMG.cpp
+++ b/Src/LinearSolvers/MLMG/AMReX_MLMG.cpp
@@ -1132,9 +1132,11 @@ MLMG::prepareForSolve (const Vector<MultiFab*>& a_sol, const Vector<MultiFab con
     }
 
 #ifdef AMREX_USE_HYPRE
-    hypre_solver.reset();
-    hypre_bndry.reset();
-    hypre_node_solver.reset();
+    if (!preserve_hypre_solver) {
+      hypre_solver.reset();
+      hypre_bndry.reset();
+      hypre_node_solver.reset();
+    }
 #endif
 
 #ifdef AMREX_USE_PETSC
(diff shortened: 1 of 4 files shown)


Src/LinearSolvers/MLMG/AMReX_MLMG.cpp
```diff
@@ -1132,9 +1132,11 @@ MLMG::prepareForSolve (const Vector<MultiFab*>& a_sol, const Vector<MultiFab con
...
 
 #ifdef AMREX_USE_HYPRE
-    hypre_solver.reset();
-    hypre_bndry.reset();
-    hypre_node_solver.reset();
+    if (!preserve_hypre_solver) {
+      hypre_solver.reset();
+      hypre_bndry.reset();
+      hypre_node_solver.reset();
+    }
```
WeiqunZhang: I believe we can move this block and the petsc one below into the `else if (linop.needsUpdate())` branch above.  Then we do not need the `preserve_hypre_solver flag`.

WeiqunZhang (changes requested):
Looks good to me.

Src/LinearSolvers/MLMG/AMReX_MLMG.cpp
zjibben: Thanks, I just committed this change and tested. Seems it's working as I'd expect.

WeiqunZhang (approved):

zjibben:
Thanks @WeiqunZhang for the review and merging!

PR #1155, opened 2020-07-20
Bugfix/amr mesh constructor

svater:
These resizes are not needed and lead to an error if this constructor is used.


WeiqunZhang:
What error are you seeing?  Could you give more details?

svater:
The runtime error happens for a single-level grid at line 40 in AMReX_AmrMesh.cpp. I think the problem is that amr_info.ref_ratio has size zero in this case. Here is the trace:

```
==11259==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x602000033268 at pc 0x000000732e47 bp 0x7fff36583000 sp 0x7fff36582ff8
READ of size 8 at 0x602000033268 thread T0
    #0 0x732e46 in void std::_Construct<amrex::IntVect, amrex::IntVect const&>(amrex::IntVect*, amrex::IntVect const&) /usr/include/c++/10/bits/stl_construct.h:109
    #1 0x73158d in amrex::IntVect* std::__uninitialized_fill_n<false>::__uninit_fill_n<amrex::IntVect*, unsigned long, amrex::IntVect>(amrex::IntVect*, unsigned long, amrex::IntVect const&) /usr/include/c++/10/bits/stl_uninitialized.h:236
    #2 0x72f526 in amrex::IntVect* std::uninitialized_fill_n<amrex::IntVect*, unsigned long, amrex::IntVect>(amrex::IntVect*, unsigned long, amrex::IntVect const&) /usr/include/c++/10/bits/stl_uninitialized.h:289
    #3 0x72c813 in amrex::IntVect* std::__uninitialized_fill_n_a<amrex::IntVect*, unsigned long, amrex::IntVect, amrex::IntVect>(amrex::IntVect*, unsigned long, amrex::IntVect const&, std::allocator<amrex::IntVect>&) /usr/include/c++/10/bits/stl_uninitialized.h:402
    #4 0xad404f in std::vector<amrex::IntVect, std::allocator<amrex::IntVect> >::_M_fill_insert(__gnu_cxx::__normal_iterator<amrex::IntVect*, std::vector<amrex::IntVect, std::allocator<amrex::IntVect> > >, unsigned long, amrex::IntVect const&) /usr/include/c++/10/bits/vector.tcc:566
    #5 0xad359b in std::vector<amrex::IntVect, std::allocator<amrex::IntVect> >::resize(unsigned long, amrex::IntVect const&) /usr/include/c++/10/bits/stl_vector.h:960
    #6 0xac7f55 in amrex::AmrMesh::AmrMesh(amrex::Geometry const&, amrex::AmrInfo const&) /home/svater/rechnen/amrex/Src/AmrCore/AMReX_AmrMesh.cpp:40
    #7 0xa9138a in amrex::AmrCore::AmrCore(amrex::Geometry const&, amrex::AmrInfo const&) /home/svater/rechnen/amrex/Src/AmrCore/AMReX_AmrCore.cpp:41
    #8 0x714b25 in fub::amrex::GriddingAlgorithm::GriddingAlgorithm(fub::amrex::GriddingAlgorithm const&) /home/svater/rechnen/FiniteVolumeSolver/src/AMReX/GriddingAlgorithm.cpp:34
[...]

0x602000033268 is located 8 bytes to the left of 8-byte region [0x602000033270,0x602000033278)
allocated by thread T0 here:
    #0 0x7f7cb5c48657 in operator new(unsigned long) (/usr/lib64/libasan.so.6+0xae657)
    #1 0x5e2316 in __gnu_cxx::new_allocator<amrex::IntVect>::allocate(unsigned long, void const*) /usr/include/c++/10/ext/new_allocator.h:115
    #2 0x5dcee9 in std::allocator_traits<std::allocator<amrex::IntVect> >::allocate(std::allocator<amrex::IntVect>&, unsigned long) /usr/include/c++/10/bits/alloc_traits.h:460
    #3 0x5d34ef in std::_Vector_base<amrex::IntVect, std::allocator<amrex::IntVect> >::_M_allocate(unsigned long) /usr/include/c++/10/bits/stl_vector.h:346
    #4 0xad3b0e in void std::vector<amrex::IntVect, std::allocator<amrex::IntVect> >::_M_range_initialize<amrex::IntVect const*>(amrex::IntVect const*, amrex::IntVect const*, std::forward_iterator_tag) /usr/include/c++/10/bits/stl_vector.h:1582
    #5 0xad33b4 in std::vector<amrex::IntVect, std::allocator<amrex::IntVect> >::vector(std::initializer_list<amrex::IntVect>, std::allocator<amrex::IntVect> const&) /usr/include/c++/10/bits/stl_vector.h:629
    #6 0xad2ca6 in amrex::Vector<amrex::IntVect, std::allocator<amrex::IntVect> >::vector(std::initializer_list<amrex::IntVect>, std::allocator<amrex::IntVect> const&) /home/svater/rechnen/amrex/Src/Base/AMReX_Vector.H:28
    #7 0xad2f6e in amrex::AmrInfo::AmrInfo() /home/svater/rechnen/amrex/Src/AmrCore/AMReX_AmrMesh.H:15
    #8 0xac7be9 in amrex::AmrMesh::AmrMesh(amrex::RealBox const*, int, amrex::Vector<int, std::allocator<int> > const&, int, amrex::Vector<amrex::IntVect, std::allocator<amrex::IntVect> >, int const*) /home/svater/rechnen/amrex/Src/AmrCore/AMReX_AmrMesh.cpp:21
    #9 0xa912ab in amrex::AmrCore::AmrCore(amrex::RealBox const*, int, amrex::Vector<int, std::allocator<int> > const&, int, amrex::Vector<amrex::IntVect, std::allocator<amrex::IntVect> >, int const*) /home/svater/rechnen/amrex/Src/AmrCore/AMReX_AmrCore.cpp:26
    #10 0x71512a in fub::amrex::GriddingAlgorithm::GriddingAlgorithm(fub::amrex::PatchHierarchy, fub::AnyInitialData<fub::amrex::GriddingAlgorithm>, fub::AnyTaggingMethod<fub::amrex::GriddingAlgorithm>, fub::AnyBoundaryCondition<fub::amrex::GriddingAlgorithm>) /home/svater/rechnen/FiniteVolumeSolver/src/AMReX/GriddingAlgorithm.cpp:61
[...]
SUMMARY: AddressSanitizer: heap-buffer-overflow /usr/include/c++/10/bits/stl_construct.h:109 in void std::_Construct<amrex::IntVect, amrex::IntVect const&>(amrex::IntVect*, amrex::IntVect const&)
[...]
```
I hope that helps.

WeiqunZhang:
When you pass `AmrInfo` to `AmrMesh` constructor, what's the size of the `ref_ratio` vector?  The default constructor of `AmrInfo` will make it 1.  Are you using the default one, or are you making some changes?  If you are passing zero size vector, that could break it, and we can fix it for that case, but we will keep the resize there for other cases.

maikel:
What is the content of `ref_ratio` in case of only one level, i. e. no local refinement?

WeiqunZhang:
If `max_level` is zero, in principle `ref_ratio` should not be used.  But since `ref_ratio` has been a size 1 vector in that case, we probably should keep the behavior.  As for the issue you have, it seems that it is because in AmrMesh constructor the size of the `AmrInfo` argument's members is not checked resulting in `back()` being called on an empty vector.  We could turn the resize into something like
```
if (amr_info.ref_ratio.empty()) {
    ref_ratio.resize(nlev, IntVect{2});
} else {
    ref_ratio.resize(nlev, amr_info.ref_ratio.back());
}
```
And we probably should do that for others too.  If this solves your issue, could you update the PR.  We need the calls to resize because we always allowed the users to just provide a scalar (or size 1 vector) and we want to keep that.

WeiqunZhang:
The fix should be easy.  If you don't mind, I can push to your fork, and then have the PR merged.

maikel:
I am fully aware that this is not a complicated issue. I can also rearrange things in our code. That ref_ratio is not allowed to be used in case of no finer levels is an interpretation that didn't come to me from the docs, but I might have overlooked things. There could be at least an assertion to check for ref_ratios size being non-zero.

We have a copy-assignable grid type in our code that wraps `AmrCore`. This issue was raised in a copy operation of a simple grid. We filled `ref_ratio` wrongly at construction. If you want to keep the current semantics adding an assertion or throwing an exception might be better than silently changing the ref_ratio from the given amr_info object. But that's just my opinion.

WeiqunZhang:
We do need to resize, because if the argument contains fewer elements, we want to fill the rest of the vector with the back, just like what we do with inputs parameters.  And if the argument has more elements, we want to remove the extra elements, just like what we do with inputs parameters.  For inputs parameters, this is done silently, so that the user does not have to change ref_ratio, or max_grid_size, if they simply change max_level.  I think it's better for this to be consistent with how we treat inputs parameters.  As for `ref_ratio`, we should resize it to `nlev-1` to be consistent with other constructors.  If you could work around this in your code, that would be great.
```
if (amr_info.ref_ratio.empty()) {
    ref_ratio.resize(nlev-1, IntVect{2});
} else {
    ref_ratio.resize(nlev-1, amr_info.ref_ratio.back());
}
```

maikel:
InitAmrMesh sets the local `ref_ratio` array to size 0 if `max_level` is zero. This leads to complications in cases where I don't want a refined hierarchy. Using the `amr_info` constructor crashes in those cases. So something is still fishy here, no?

Edit:
I think we should add the safeguard that you propose.

WeiqunZhang (approved):

PR #1163, opened 2020-07-22
Update the documentation for NSight tools.

kngott:
Created a skeleton in External Profiling for documenting AMReX specific implementations of the NSight package. 

Also updated the run.corigpu script with new data from the Hackathon.

Should we eliminate nvprof documentation entirely?


drummerdoc:
Thanks for writing this us, Kevin!

Docs/sphinx_documentation/source/External_Profiling_Tools.rst
```diff
@@ -279,3 +279,167 @@ generated figures. Some examples are shown here.
...
+
+.. highlight:: c++
+
+    nsys profile -o <file_name> ${EXE} ${INPUTS}
+
+AMReX's lambda-based launch system often makes these timelines difficult to parse, as the kernel
+are mangled and are difficult to decipher. AMReX's Tiny Profiler includes NVTX region markers,
+which can be used to mark each section of the NSight Systems timeline and.  To include AMReX's
+built-in Tiny Profiler NVTX regions in NSight Systems outputs, compile AMReX 
+``with TINY_PROFILE=TRUE`` and run nsys with ``-e NSYS_NVTX_PROFILER_REGISTER_ONLY=0``:
```
WeiqunZhang: with should be outside.

WeiqunZhang (approved):

Docs/sphinx_documentation/source/External_Profiling_Tools.rst
```diff
@@ -279,3 +279,167 @@ generated figures. Some examples are shown here.
 
 .. [5]
    https://www.nersc.gov/users/software/performance-and-debugging-tools/ipm/
+
+NSight Systems
+==============
+
+The NSight Systems tool provides a high-level overview of a GPU code, displaying the kernel
```
maximumcats: Nsight Systems can also be used for CPU-only code, it's a useful tool (when combined with NVTX) if you want to see on your graphical timeline where the time is spent in the CPU code so you know what things to prioritize for acceleration. I'd prefer if that were at least mentioned here.

Docs/sphinx_documentation/source/External_Profiling_Tools.rst
```diff
@@ -279,3 +279,167 @@ generated figures. Some examples are shown here.
 
 .. [5]
    https://www.nersc.gov/users/software/performance-and-debugging-tools/ipm/
+
+NSight Systems
```
maximumcats: The official names of the tools are "Nsight Systems" and "Nsight Compute."

Docs/sphinx_documentation/source/External_Profiling_Tools.rst
```diff
@@ -279,3 +279,167 @@ generated figures. Some examples are shown here.
...
+Cori in a system module.
+
+NSight Systems provides a variety of profiling options.  This documentation will cover the
+most commonly used options for AMReX users to keep track of useful flags and analysis
+patterns. For the complete details of using NSight Systems, refer to the official documentation.
+
+Profile Analysis
+----------------
+
+The most common use case of NSight Systems for AMReX users is the creation of a qdstrm file
```
maximumcats: The creation of a `qdrep` file is the goal. The `qdstrm` file is an intermediate internal representation which should not be used by users in recent versions of Nsight Systems.

Docs/sphinx_documentation/source/External_Profiling_Tools.rst
```diff
@@ -279,3 +279,167 @@ generated figures. Some examples are shown here.
...
+
+.. highlight:: c++
+
+    nsys profile -o <file_name> ${EXE} ${INPUTS}
+
+AMReX's lambda-based launch system often makes these timelines difficult to parse, as the kernel
+are mangled and are difficult to decipher. AMReX's Tiny Profiler includes NVTX region markers,
+which can be used to mark each section of the NSight Systems timeline and.  To include AMReX's
+built-in Tiny Profiler NVTX regions in NSight Systems outputs, compile AMReX with 
+``TINY_PROFILE=TRUE`` and run nsys with ``-e NSYS_NVTX_PROFILER_REGISTER_ONLY=0``:
```
maximumcats: This registered string environment variable is only needed if you're trying to use the `--capture-range` functionality. You don't need it if you're just trying to collect a profile with NVTX regions.

Docs/sphinx_documentation/source/External_Profiling_Tools.rst
```diff
@@ -279,3 +279,167 @@ generated figures. Some examples are shown here.
...
+are mangled and are difficult to decipher. AMReX's Tiny Profiler includes NVTX region markers,
+which can be used to mark each section of the NSight Systems timeline and.  To include AMReX's
+built-in Tiny Profiler NVTX regions in NSight Systems outputs, compile AMReX with 
+``TINY_PROFILE=TRUE`` and run nsys with ``-e NSYS_NVTX_PROFILER_REGISTER_ONLY=0``: 
+
+.. highlight:: c++
+
+    nsys profile -o <file_name> -e NSYS_NVTX_PROFILER_REGISTER_ONLY=0 ${EXE} ${INPUTS}
+
+The additional environment flag is needed because AMReX's NVTX region names currently do not use
```
maximumcats: One thing to add in this section would be to note that recent versions of Nsight Systems have the option "Rename CUDA Kernels by NVTX" (Tools -> Options -> Environment) which should be set to "Yes" for optimal usage with TinyProfiler (after #1177 is merged). Then we'll get kernel rows corresponding to the innermost BL_PROFILE region, which is usually much more useful than the compiler name.

Docs/sphinx_documentation/source/External_Profiling_Tools.rst
```diff
@@ -279,3 +279,167 @@ generated figures. Some examples are shown here.
...
+flag specifies the total number of kernels to be analyzed. For example:
+
+ .. highlight:: c++
+
+    ncu --nvtx --nvtx-include "GravitySolve()" -c 10 -o kernels ${EXE} ${INPUTS} amrex.fpe_trap_invalid=0
+
+will only analyze the first ten kernels inside of the ``GravitySolve()`` NVTX region.
+
+For further details on how to choose a subset of CUDA kernels to analyze, or to run a more detailed
+analysis, including CUDA hardware counters, refer to the NSight Compute official documentation.
```
maximumcats: You might want to link directly to NVIDIA documentation in the places where you mention it.

Docs/sphinx_documentation/source/External_Profiling_Tools.rst
kngott: Just to be clear, I would need the environment variable when using `-c nvtx`, not when using `-c cudaProfilerApi`, correct?

Docs/sphinx_documentation/source/External_Profiling_Tools.rst
kngott: I was considering this. But, do we want to maintain potentially changing documentation links? This issue comes up a lot and different groups have their own preference.

If no one minds (e.g. @WeiqunZhang, @atmyers), I'm game to put in the links. It'll just require a little more maintenance.

Docs/sphinx_documentation/source/External_Profiling_Tools.rst
maximumcats: Correct.

kngott: 👍

Docs/sphinx_documentation/source/External_Profiling_Tools.rst
maximumcats: Fixing broken links is one way we can continue inflating our PR count :-)

WeiqunZhang:
Does Max approve Kevin's changes?

maximumcats (approved):

PR #1164, opened 2020-07-22
update CONTRIBUTING.md on how to fix git commits to have a clean history

WeiqunZhang:


atmyers (approved):

CONTRIBUTING.md
```diff
@@ -168,6 +168,57 @@ and you can delete the remote one on your fork with
...
+After this, the local `development` should be in sync with AMReX `development` and your recent
+commits have been saved in `new_branch` branch.
+
+If for some reason your PR branch has diverged from AMReX, you can try to fix it as follows.  Before
+you try it, you should back up your code in case things might go wrong.
+```
+git fetch upstream   # assuming upstream is the remote name for the official amrex repo
+git checkout -b xxx upstream/development  # replace xxx with whatever name you like
+git branch -D development
+git checkout -b development upstream/development
```
ax3l, line 198: The parts around look good, but why do we pro-actively recommend to remove and re-check out the local `development` branch? :)

CONTRIBUTING.md
```diff
@@ -168,6 +168,57 @@ and you can delete the remote one on your fork with
...
+merged, you can drop them by replace `pick` with `drop`,
+```
+drop 7451d9d commit message a
+drop c4c2459 commit message b
+pick 6fj3g90 commit message c
+```
+After saving and then exiting the editor, `git log` should show a clean history based on top of
+`development` branch.  You can also do `git diff yyy..xxx` to make sure nothing new was dropped.  If
+all goes well, you can submit a PR using `xxx` branch.
+
```
ax3l: I usually add something like:
```suggestion
Don't worry, if something goes wrong during the rebase, you an always `git rebase --abort` and start over.
```

CONTRIBUTING.md
WeiqunZhang: In my observation of PRs with dozens of already merged commits but only one real commit, the root cause is almost always development branch being diverged.  Yes, it's not necessary in some situations, but people who are in those usually do not need the instructions here.

ax3l: I see. Just wondering because if you show them how to rebase against the remote development branch it does not matter what the status of the local development branch is.

One could potentially split the 2nd part with that rationale in a smaller chunk after that, just to make it appear less magic to the reader.

PR #1166, opened 2020-07-22
Add alternative error tagging class

drummerdoc:
This PR introduces an alternative to ErrorRec/ErrorList for managing error tagging.  It supports building (GPU-SAFE) tagging functions on the fly via ParmParse inputs, something like:
```
amrex::Vector<amrex::AMRErrorTag> errtags;
errtags.push_back(AMRErrorTag(value,AMRErrorTag::GREATER,field,info));
```
Then later, they are called to tag as:
```
  for (int j=0; j<errtags.size(); ++j) {
    std::unique_ptr<MultiFab> mf;
    if (errtags[0].Field() != std::string()) {
      mf = std::unique_ptr<MultiFab>(derive(errtags[j].Field(), time, errtags[j].NGrow()));
    }
    errtags[j](tags,mf.get(),clearval,tagval,time,level,geom);
  }
```
At the moment, standard support is provided for BOX (over physical region), LESS, GREATER, GRAD, as well as a special one, VORT, for vorticity over AMR levels. and USER.  USER is close to AMReX-classic error tagging functions. In the initial implementation, the inputs can specify a start and stop time, max level for each criteria to apply to, the field to base the decision on, and a physical box to limit the checking. ParmParse input can be organized as follows, e.g.:
```
amr.refinement_indicators = flame_tracer lo_temp
amr.refine.flame_tracer.max_level = 3
amr.refine.flame_tracer.value_greater = 1.e-6
amr.refine.flame_tracer.field_name = Y(H)

amr.refine.lo_temp.max_level = 2
amr.refine.lo_temp.value_less = 1000.
amr.refine.lo_temp.field_name = temp
```
This will tag regions up to level 3 where Y(H)>1.e-6 and up to level 2 where temp<1000.

### DIFF
--- a/Src/AmrCore/AMReX_ErrorList.H
+++ b/Src/AmrCore/AMReX_ErrorList.H
@@ -8,8 +8,12 @@
 #include <AMReX_Vector.H>
 #include <AMReX_REAL.H>
 #include <AMReX_ArrayLim.H>
+#include <AMReX_MultiFab.H>
+#include <AMReX_TagBox.H>
+#include <AMReX_Geometry.H>
 
 namespace amrex {
+
 
 extern "C"
 {
@@ -372,6 +376,86 @@ private:
 
 std::ostream& operator << (std::ostream& os, const ErrorList& elst);
 
+  struct AMRErrorTagInfo
+  {
+    int            m_max_level;
+    amrex::Real    m_min_time;
+    amrex::Real    m_max_time;
+    amrex::RealBox m_realbox;
+
+    AMRErrorTagInfo () noexcept : m_max_level(-1), m_min_time(-1), m_max_time(-1), m_realbox(amrex::RealBox()) {}
+
+    AMRErrorTagInfo& SetMaxLevel (int max_level) noexcept {
+      m_max_level = max_level;
+      return *this;
+    }
+    AMRErrorTagInfo& SetMinTime (amrex::Real min_time) noexcept {
+      m_min_time = min_time;
+      return *this;
+    }
+    AMRErrorTagInfo& SetMaxTime (amrex::Real max_time) noexcept {
+      m_max_time = max_time;
+      return *this;
+    }
+    AMRErrorTagInfo& SetRealBox (const amrex::RealBox& realbox) noexcept {
+      m_realbox = realbox;
+      return *this;
+    }
+  };
+
+  class AMRErrorTag
+  {
+  public:
+
+    enum TEST {GRAD=0, LESS, GREATER, VORT, BOX, USER};
+
+    struct UserFunc
+    {
+      virtual void operator() (const amrex::Box&                       bx,
+                               amrex::Array4<const amrex::Real> const& dat,
+                               amrex::Array4<char> const&              tag,
+                               amrex::Real                             time,
+                               int                                     level,
+                               int                                     tagval,
+                               int                                     clearval) = 0;
+    };
+
+    AMRErrorTag (const AMRErrorTagInfo& info = AMRErrorTagInfo()) noexcept
+      : m_test(BOX), m_field(std::string()), m_info(info) {SetNGrow();}
+
+    AMRErrorTag (amrex::Real            value,
+                 AMRErrorTag::TEST      test,
+                 const std::string&     field,
+                 const AMRErrorTagInfo& info = AMRErrorTagInfo()) noexcept
+      : m_value(value), m_test(test), m_field(field), m_info(info) {SetNGrow();}
+
+    AMRErrorTag (AMRErrorTag::UserFunc* userfunc,
+                 const std::string&     field,
+                 int                    ngrow,
+                 const AMRErrorTagInfo& info = AMRErrorTagInfo()) noexcept
+      : m_userfunc(userfunc), m_field(field), m_ngrow(ngrow), m_info(info) {}
+
+    virtual void operator() (amrex::TagBoxArray&    tb,
+                             const amrex::MultiFab* mf,
+                             int                    clearval,
+                             int                    tagval,
+                             amrex::Real            time,
+                             int                    level,
+                             const amrex::Geometry& geom) const noexcept;
+
+    int NGrow() const noexcept {return m_ngrow;}
+    const std::string& Field () const noexcept {return m_field;}
+
+  protected:
+    int SetNGrow () const noexcept;
+
+    Real m_value;    
+    TEST m_test;
+    UserFunc* m_userfunc = nullptr;
+    std::string m_field;
+    int m_ngrow;
+    AMRErrorTagInfo m_info;
+  };
 }
 
 #endif
--- a/Src/AmrCore/AMReX_ErrorList.H
+++ b/Src/AmrCore/AMReX_ErrorList.H
@@ -378,12 +378,10 @@ std::ostream& operator << (std::ostream& os, const ErrorList& elst);
 
   struct AMRErrorTagInfo
   {
-    int            m_max_level;
-    amrex::Real    m_min_time;
-    amrex::Real    m_max_time;
-    amrex::RealBox m_realbox;
-
-    AMRErrorTagInfo () noexcept : m_max_level(-1), m_min_time(-1), m_max_time(-1), m_realbox(amrex::RealBox()) {}
+    int m_max_level = 100000;
+    Real m_min_time = std::numeric_limits<Real>::lowest();
+    Real m_max_time = std::numeric_limits<Real>::max();
+    RealBox m_realbox;
 
     AMRErrorTagInfo& SetMaxLevel (int max_level) noexcept {
       m_max_level = max_level;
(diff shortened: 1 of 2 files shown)


Src/AmrCore/AMReX_ErrorList.H
```diff
@@ -372,6 +376,86 @@ private:
...
 std::ostream& operator << (std::ostream& os, const ErrorList& elst);
 
+  struct AMRErrorTagInfo
+  {
+    int            m_max_level;
+    amrex::Real    m_min_time;
+    amrex::Real    m_max_time;
+    amrex::RealBox m_realbox;
+
+    AMRErrorTagInfo () noexcept : m_max_level(-1), m_min_time(-1), m_max_time(-1), m_realbox(amrex::RealBox()) {}
```
WeiqunZhang: I think it's cleaner to remove the constructor and do
```
int m_max_level = 100000;
Real m_min_time = std::numeric_limits<Real>::lowest();
Real m_max_time = std::numeric_limits<Real>::max();
RealBox m_realbox;
```
But it's just my personal taste.

Src/AmrCore/AMReX_ErrorList.H
```diff
@@ -372,6 +376,86 @@ private:
...
+    enum TEST {GRAD=0, LESS, GREATER, VORT, BOX, USER};
+
+    struct UserFunc
+    {
+      virtual void operator() (const amrex::Box&                       bx,
+                               amrex::Array4<const amrex::Real> const& dat,
+                               amrex::Array4<char> const&              tag,
+                               amrex::Real                             time,
+                               int                                     level,
+                               int                                     tagval,
```
WeiqunZhang: Maybe we should use `char tagvel`?

Src/AmrCore/AMReX_ErrorList.H
```diff
@@ -372,6 +376,86 @@ private:
...
+                               int                                     clearval) = 0;
+    };
+
+    AMRErrorTag (const AMRErrorTagInfo& info = AMRErrorTagInfo()) noexcept
+      : m_test(BOX), m_field(std::string()), m_info(info) {SetNGrow();}
```
WeiqunZhang: `SetNGrow` does not actually anything.  It just returns a number, and `m_ngrow` is not set.

Src/AmrCore/AMReX_ErrorList.H
```diff
@@ -372,6 +376,86 @@ private:
...
+      virtual void operator() (const amrex::Box&                       bx,
+                               amrex::Array4<const amrex::Real> const& dat,
+                               amrex::Array4<char> const&              tag,
+                               amrex::Real                             time,
+                               int                                     level,
+                               int                                     tagval,
+                               int                                     clearval) = 0;
+    };
+
+    AMRErrorTag (const AMRErrorTagInfo& info = AMRErrorTagInfo()) noexcept
```
WeiqunZhang: Should we add `explicit`?

Src/AmrCore/AMReX_ErrorList.cpp
```diff
@@ -228,4 +228,196 @@ operator << (std::ostream&    os,
...
+  {
+    const Real* dx = geom.CellSize();
+    const Real* plo = geom.ProbLo();
+    class RealBox trb(bx,dx,plo);
+    if (tag_rb.intersects(trb))
+    {
+      amrex::ParallelFor(bx,
+      [=] AMREX_GPU_HOST_DEVICE (int i, int j, int k) noexcept
+      {
+        GpuArray<Real,AMREX_SPACEDIM> pt = {D_DECL(plo[0]+(i+0.5)*dx[0],
```
WeiqunZhang, line 318: We need to use GpuArray returned by `CellSizeArray` on GPU.

Src/AmrCore/AMReX_ErrorList.cpp
```diff
@@ -228,4 +228,196 @@ operator << (std::ostream&    os,
...
+                   Array4<const Real> const& dat,
+                   Array4<char> const&       tag,
+                   int                       level,
+                   Real                      threshold,
+                   int                       tagval)
+  {
+    amrex::ParallelFor(bx,
+    [=] AMREX_GPU_HOST_DEVICE (int i, int j, int k) noexcept
+    {
+      if (dat(i,j,k) >= threshold * std::pow(2,level)) {
```
WeiqunZhang: For performance, we should compute `threshold * std::pow(2,level)` outside `ParallelFor`.

Src/AmrCore/AMReX_ErrorList.cpp
```diff
@@ -228,4 +228,196 @@ operator << (std::ostream&    os,
...
+      for (MFIter mfi(tba,TilingIfNotGPU()); mfi.isValid(); ++mfi)
+      {
+        const auto& bx    = mfi.tilebox();
+        auto const& dat   = mf->array(mfi);
+        Vector<int> itags = tba[mfi].tags();
+        auto tag          = tba.array(mfi);
+
+        (*m_userfunc)(bx,dat,tag,time,level,tagval,clearval);
+
+        tba[mfi].tags(itags);
```
WeiqunZhang: This line and `itags` should be deleted.

Src/AmrCore/AMReX_ErrorList.cpp
```diff
@@ -228,4 +228,196 @@ operator << (std::ostream&    os,
...
+          else if (m_test == VORT)
+          {
+            AMRErrorTag_VORT(bx, dat, tag, level, m_value, tagval);
+          }
+          else
+          {
+            Abort("Bad AMRErrorTag test flag");
+          }
+
+          tba[mfi].tags(itags);
```
WeiqunZhang: This should be deleted.

WeiqunZhang (changes requested):
The functionality looks great.  I have some minor comments.

Src/AmrCore/AMReX_ErrorList.H
```diff
@@ -372,6 +376,86 @@ private:
...
+
+  class AMRErrorTag
+  {
+  public:
+
+    enum TEST {GRAD=0, LESS, GREATER, VORT, BOX, USER};
+
+    struct UserFunc
+    {
+      virtual void operator() (const amrex::Box&                       bx,
```
WeiqunZhang, line 412: Is it simpler for the user to use a function pointer instead of a class derive from this?

maximumcats: I would really like to avoid that path if possible, I'm not a fan of passing around function pointers if there's a more C++ way to do it.

drummerdoc: The interface I created had in mind several use cases I've had over the past, where a class was able to hold state info that was useful later.  The overhead of doing this is minimal in C++ and the flexibility it buys I think is worth it.

Src/AmrCore/AMReX_ErrorList.H
drummerdoc: That would work, as it allows for negative sim times as well.

Src/AmrCore/AMReX_ErrorList.H
drummerdoc: Yeah, probably.

Src/AmrCore/AMReX_ErrorList.H
drummerdoc: Oops!  It's supposed to be used to set grow for the provided test types...will fix.

Src/AmrCore/AMReX_ErrorList.H
drummerdoc: I don't think so, as that would require an info object, correct?  If the defaults are ok, this is the one type of case I'm ok with default arguments.... if I'm wrong about ``explicit`` let me know

Src/AmrCore/AMReX_ErrorList.cpp
drummerdoc: Ahh, maybe that explains a core dump I was seeing in DEBUG mode this morning?  Will check.

Src/AmrCore/AMReX_ErrorList.cpp
drummerdoc: Will do.

Src/AmrCore/AMReX_ErrorList.cpp
drummerdoc: I need to learn about how this is done now.

Src/AmrCore/AMReX_ErrorList.H
WeiqunZhang: That's true.

Src/AmrCore/AMReX_ErrorList.H
WeiqunZhang: explicit just means if there is a function that takes `AMRErrorTag` as argument, one cannot just pass `AMRErrorTagInfo` and let compiler to implicit convert it.

WeiqunZhang: https://godbolt.org/z/6hh89h

WeiqunZhang: You can still have the default argument to the constructor.

Src/AmrCore/AMReX_ErrorList.cpp
WeiqunZhang: In the old days, we call a Fortan 77 function that cannot work on C++ char.  So it is converted to int first, pass the int for Fortran to set, and then update char with int.  But now, we do not need to convert back and forth between char and int (even with Fortan).

WeiqunZhang (approved):

Src/AmrCore/AMReX_ErrorList.cpp
```diff
@@ -228,4 +228,190 @@ operator << (std::ostream&    os,
...
+  {
+    auto plo = geom.ProbLoArray();
+    auto dx  = geom.CellSizeArray();
+    class RealBox trb(bx,dx.data(),plo.data());
+    if (tag_rb.intersects(trb))
+    {
+      amrex::ParallelFor(bx,
+      [=] AMREX_GPU_HOST_DEVICE (int i, int j, int k) noexcept
+      {
+        GpuArray<Real,AMREX_SPACEDIM> pt = {D_DECL(plo[0]+(i+0.5)*dx[0],
```
ax3l, line 318: @drummerdoc @WeiqunZhang Getting a AppleClang (SP) and DPC++ (DP) compile error on this of the form:
```
Src/AmrCore/AMReX_ErrorList.cpp:318:52: error: non-constant-expression cannot be narrowed from type 'double' to 'float' in initializer list [-Wc++11-narrowing]
```

in WarpX.

ax3l: maybe we can cast the `i`, `j`, `k` with `Real(i)` etc. and add `_rt` in `+ 0.5` as `+ 0.5_rt` to fix it.

ax3l: Let's try this: #1180

PR #1169, opened 2020-07-22
Fix bug in DistributionMapping::gather_weights

rporcu:


atmyers:
Hi @rporcu, thanks for this PR. This was same fix was made in #1168, which I saw first.

Issue #568, opened 2019-09-16
Updating FillBoundary for multiple layers of ghost nodes.


bsrunnels:
When treating fabs that have multiple layers of ghost nodes (i.e. more than 1) the existing `FillBoundary` routine does not work since it requires filling ghost nodes with other ghost nodes. (This is important when using the "reflux-free" version of the linear operator).

I've been doing the following instead:

    void realFillBoundary(MultiFab &phi, const Geometry &geom) 
    {
    	for (int i = 0; i < 2; i++)
    	{
    		MultiFab & mf = phi;
    		mf.FillBoundary(geom.periodicity());
    		const int ncomp = mf.nComp();
    		const int ng1 = 1;
    		const int ng2 = 2;
    		MultiFab tmpmf(mf.boxArray(), mf.DistributionMap(), ncomp, ng1);
    		MultiFab::Copy(tmpmf, mf, 0, 0, ncomp, ng1); 
    		mf.ParallelCopy   (tmpmf, 0, 0, ncomp, ng1, ng2, geom.periodicity());
    	}
    }

which works ok. But I have two questions:
(1) is there a better way to do this? This is quite inefficient.
(2) can we build this into AMReX? If so, can we simply add a `MultiFab::RealFillBoundary`? Or update the existing `FillBoundary`? Or something else...?

Thanks!

PR #1171, opened 2020-07-22
Fillboundary

bsrunnels:
This pull request fixes the problem of edge nodes not getting filled properly when filling multiple layers of ghost nodes. This occurs when there are overlapping ghost nodes on a coarse/fine boundary. (See issue #568)

Summary of changes to AMreX:
- Added a `BoxArray::complementIn` function that takes a box _list_ and returns the complement box list. 
- Added an additional boolean argument `multi_ghost` to all `FillBoundary` functions that gets passed (eventually) to `define_fb`. In all cases, a default value of `false` is supplied.
- Modified `define_fb` to include the missing ghost nodes in the list of `send_tags`. (This uses the new `complementIn` function.)

Summery of changes to `LinearSolver/MultiComponent` tutorial:
- Removed the `realFillBoundary` function, replaced with `FillBoundary(false,true)
- Modified `main.cpp` so that it can be compiled and run (non-trivially) in 2D as well as 3D.

Diagram of the problem that this PR fixes:
FillBoundaryIllustration


Src/Base/AMReX_FabArrayBase.cpp
```diff
@@ -638,25 +639,30 @@ FabArrayBase::FB::FB (const FabArrayBase& fa, const IntVect& nghost,
...
+    if (multi_ghost)             ba_ng    = BoxArray(ba.boxList().accrete(1));
     const DistributionMapping& dm       = fa.DistributionMap();
     const Vector<int>&         imap     = fa.IndexArray();
 
     // For local copy, all workers in the same team will have the identical copy of tags
     // so that they can share work.  But for remote communication, they are all different.
     
     const int nlocal = imap.size();
-    const IntVect& ng = m_ngrow;
+    const IntVect ng(m_ngrow[0]);
```
WeiqunZhang: This change should be reverted.

Src/Base/AMReX_FabArrayBase.cpp
```diff
@@ -665,24 +671,35 @@ FabArrayBase::FB::define_fb(const FabArrayBase& fa)
...
 	const int ksnd = imap[i];
-	const Box& vbx = ba[ksnd];
-	
+	const Box& vbx    = ba[ksnd];
+	const Box& vbx_ng = ba_ng[ksnd];
+
 	for (auto pit=pshifts.cbegin(); pit!=pshifts.cend(); ++pit)
 	{
 	    ba.intersections(vbx+(*pit), isects, false, ng);
+	    if (multi_ghost) ba_ng.intersections(vbx_ng+(*pit), isects_ng, false, ng_ng);
```
WeiqunZhang: I don't think it's necessary to define `ba_ng` and do the intersections here.  And it's probably wrong.  Because the code below assumes  isects and isects_ng have the same size and the same `first` for each item.  Because `ba_ng` is a completely different BoxArray and the intersection uses unordered_map internally, the order of the items might be different even though the size will be the same.  I think in the code below you can simply intersect the send box grown by one with the receive box grown by ng.

WeiqunZhang (changes requested):

Src/Base/AMReX_FabArrayBase.cpp
bsrunnels: Done

Src/Base/AMReX_FabArrayBase.cpp
bsrunnels: Good point, I didn't realize that the map was unordered. So at the very least we need to iterate through `isects_ng` to ensure the keys match. That would be straightforward to implement.

I don't understand your solution though - wouldn't the send and receive boxes be too large in that case?

WeiqunZhang: I don't think you need to use `ba_ng` to get `isects_ng` at all.  When you iterate through `isects`, you know which two boxes gives that element of `isects`.  So you can easily compute what the corresponding element of `isects_ng` would be without having `isects_ng`.  We could have a zoom meeting next week to discuss this, if you want.

Src/Base/AMReX_FabArrayBase.cpp
```diff
@@ -1063,7 +1090,7 @@ FabArrayBase::getFB (const IntVect& nghost, const Periodicity& period,
     }
 
     // Have to build a new one
-    FB* new_fb = new FB(*this, nghost, cross, period, enforce_periodicity_only);
+    FB* new_fb = new FB(*this, nghost, cross, period, enforce_periodicity_only,multi_ghost);
```
WeiqunZhang: Before this, we look up in the FB cache.  That code needs modification too to include test on `multi_ghost`.

WeiqunZhang: And we need to save a copy of `multi_ghost` in `FB`.

Src/Base/AMReX_FabArrayBase.cpp
bsrunnels: @WeiqunZhang Ah, I understand now, thanks. I have implemented the change and removed `isects_ng`.

Src/Base/AMReX_FabArrayBase.cpp
bsrunnels: Done. I ended up actually storing `multi_ghost` in `FabArrayBase` as well as `FB`. Now `multi_ghost` will get set as a property of the `FabArrayBase` object and should be used consistently every time `getFB` is called. (Consequently I also removed the `multi_ghost` argument from all the `FillBoundary` calls.)

Eventually it would be nice to remove the `multi_ghost` entirely and just do this for all multifabs that have more than one ghost node. But I don't know what that would end up breaking.

Src/Base/AMReX_FabArrayBase.cpp
```diff
@@ -661,7 +663,9 @@ FabArrayBase::FB::define_fb(const FabArrayBase& fa)
     
     const int nlocal = imap.size();
     const IntVect& ng = m_ngrow;
+    const IntVect ng_ng(AMREX_D_DECL(m_ngrow[0]-1,m_ngrow[1]-1,m_ngrow[1]-1));
```
WeiqunZhang: There is an index bug here.  The last one should be `[2]`.  You could also do `const IntVect ng_ng = m_ngrow - 1;`.

Src/Base/AMReX_FabArrayBase.cpp
```diff
@@ -661,7 +663,9 @@ FabArrayBase::FB::define_fb(const FabArrayBase& fa)
...
     std::vector< std::pair<int,Box> > isects;
+    std::vector< std::pair<int,Box> > isects_ng;
```
WeiqunZhang: Please remove unused variable.

Src/Base/AMReX_FabArrayBase.cpp
```diff
@@ -671,7 +675,8 @@ FabArrayBase::FB::define_fb(const FabArrayBase& fa)
     {
 	const int ksnd = imap[i];
 	const Box& vbx = ba[ksnd];
-	
+	const Box& vbx_ng  = amrex::grow(ba[ksnd],1);
```
WeiqunZhang: It is slightly more efficient to do `const Box& vbx_ng = amrex::grow(vbx,1);` to avoid the cost of `BoxArray::operator[]`.

Src/Base/AMReX_FabArrayBase.cpp
WeiqunZhang: I don't think we want to do that by default for all MulitiFabs.  There is cost associated with doing this.

Src/Base/AMReX_FabArrayBase.cpp
```diff
@@ -718,7 +733,9 @@ FabArrayBase::FB::define_fb(const FabArrayBase& fa)
     {
 	const int   krcv = imap[i];
 	const Box& vbx   = ba[krcv];
+	const Box& vbx_ng  = amrex::grow(ba[krcv],1);
```
WeiqunZhang: We can change this to `const Box& vbx_ng = amrex::grow(vbx,1);`.

Src/Base/AMReX_FabArrayBase.cpp
```diff
@@ -718,7 +733,9 @@ FabArrayBase::FB::define_fb(const FabArrayBase& fa)
...
 	const Box& bxrcv = amrex::grow(vbx, ng);
+	const Box& bxrcv_ng = amrex::grow(vbx_ng, ng_ng);
```
WeiqunZhang: Looks to me `bxrcv_ng == bxrcv`.  So we can remove this.

Src/Base/AMReX_FabArrayBase.cpp
```diff
@@ -740,7 +757,19 @@ FabArrayBase::FB::define_fb(const FabArrayBase& fa)
...
+		
+        if (m_multi_ghost) 
+        {
+            // In the case where ngrow>1, augment the send/rcv box list
+            // with boxes for overlapping ghost nodes.
+            Box ba_ksnd = ba[ksnd];
+            ba_ksnd.grow(1);
+            const Box dst_bx_ng = (ba_ksnd & (bxrcv_ng + (*pit))) - (*pit);
+		    const BoxList& bl_ng = amrex::boxDiff(dst_bx_ng, vbx_ng);
+            bl.join(ba.complementIn(bl_ng));
```
WeiqunZhang: I don't think this works for periodic boundary.  The points in `bl_ng` might overlap with valid points periodically.  But if you don't have periodic boundaries.  We could add an assertion and have this PR merged.

WeiqunZhang: Also, thinking more about the approach in this PR, I don't think it will scale well because of the line above even without periodic boundaries.  I believe it will be slower than doing two communication passes unless the number of component is huge.  The two passes can be done as follows.  First, a modified FillBoundary is called to fill the second layer (or beyond) of ghost cells with the valid cells and the first layer of ghost cells.  Then a normal FillBoundary is called.  In the first pass, one can use this BoxArray member function for intersections
```
    void intersections (const Box& bx, std::vector< std::pair<int,Box> >& isects,
                        bool first_only, const IntVect& ng) const;
```
Here the effect of `ng` is to effectively grow BoxArray by `ng`.  Using this, we can treat the first layer of ghost cells as valid cells.

WeiqunZhang: In the first pass, the communication can be restricted to sending from first layer of ghost cells to the outer layers.

Src/Base/AMReX_FabArrayBase.cpp
bsrunnels: Done

Src/Base/AMReX_FabArrayBase.cpp
bsrunnels: Done

Src/Base/AMReX_FabArrayBase.cpp
bsrunnels: Done

Src/Base/AMReX_FabArrayBase.cpp
bsrunnels: Done

Src/Base/AMReX_FabArrayBase.cpp
bsrunnels: Done

Src/Base/AMReX_FabArrayBase.cpp
bsrunnels: > We could add an assertion and have this PR merged.

Assertion added.

>  I don't think it will scale well because of the line above even without periodic boundaries

Because of the `ba.complementIn(bl_ng)` call? 

> First, a modified FillBoundary is called to fill the second layer (or beyond) of ghost cells with the valid cells and the first layer of ghost cells. Then a normal FillBoundary is called. In the first pass, one can use this BoxArray member function for intersections

Ok, so in practice, one would pass an argument (say, `ng`) to `FillBoundary`, so that the total call would look like

    FillBoundary(ng) // modified fill boundary
    FillBoundary() // regular fill boundary

The `ng` argument for the modified `FillBoundary` call makes its way to `define_fb` where it is used in the `intersections` call. Is that you had in mind? If so, I'm happy to give it a try but I'm doubtful that it will work - I found that the bad `FillBoundary` behavior happened even when the nodes in question were already filled with the proper values.

Src/Base/AMReX_FabArrayBase.cpp
```diff
@@ -765,7 +764,7 @@ FabArrayBase::FB::define_fb(const FabArrayBase& fa)
             // with boxes for overlapping ghost nodes.
             Box ba_ksnd = ba[ksnd];
             ba_ksnd.grow(1);
-            const Box dst_bx_ng = (ba_ksnd & (bxrcv_ng + (*pit))) - (*pit);
+            const Box dst_bx_ng = (ba_ksnd & (bxrcv + (*pit))) - (*pit);
 		    const BoxList& bl_ng = amrex::boxDiff(dst_bx_ng, vbx_ng);
             bl.join(ba.complementIn(bl_ng));
```
WeiqunZhang:
```suggestion
            const BoxList& bltmp = ba.complementIn(dst_bx_ng);
            for (auto const& btmp : bltmp) {
                bl.join(amrex::boxDiff(btmp,vbx_ng));
           }
```
I think this is better for performance.  The reason is that `BoxArray::complemenIn` is expensive, so we want to avoid doing it on a BoxList.  We can then also remove the new `BoxArray::complementIn(BoxList const&)` function.  If it's there, people might use it without realizing the performance issue.

bsrunnels: Ok, makes sense. I've updated the change here and for the parallel part.

WeiqunZhang (approved):

PR #1184, opened 2020-07-24
fix more warnings and add CI for building 3d with configure

WeiqunZhang:


.github/workflows/linux.yml
```diff
@@ -179,5 +179,19 @@ jobs:
     - name: Build & Install
       run: |
         ./configure --dim 2
-        make -j2
+        make -j2 WARN_ALL=TRUE
```
ax3l, line 208: do you also want to add this to the 1D case above? :)

ax3l (approved):
Awesome, thanks a lot!

.github/workflows/linux.yml
ax3l: Another idea: we can add `-Werror` in CI through XTRA flags to check PRs do not introduce warnings for selected compilers :)

Src/Base/AMReX.H
```diff
@@ -89,6 +89,10 @@ namespace amrex
     void ExecOnFinalize (PTR_TO_VOID_FUNC);
     void ExecOnInitialize (PTR_TO_VOID_FUNC);
 
+    //! This shuts up the compiler about unused variables
+    template <class... Ts> AMREX_GPU_HOST_DEVICE
```
ax3l: I usually also put a force inline on that one and make it a `constexpr`.
Took inspiration from boost:
- https://github.com/alpaka-group/alpaka/blob/0.5.0/include/alpaka/core/Unused.hpp

.github/workflows/linux.yml
WeiqunZhang: There are too many warnings in 1D we need to fix. :(  We do not fully support 1D.  So in many places we just have a function with empty body.  As for `-Werror`, we still have some warnings.  One we get rid of them, we can add that to a few CI tests.

Src/Base/AMReX.H
WeiqunZhang: Sounds good.  Will update.

WeiqunZhang: Looks like we have take `constexpr`, because conexpr void is not supported in C++11.

ax3l: Ah right, in C++11 mode I used a const fallback originally: https://github.com/alpaka-group/alpaka/pull/582

atmyers (approved):

PR #1191, opened 2020-07-24
Remove anonymous struct, which is not standard. Note that this is a b…

atmyers:
…reaking change if anyone is using these data members directly.


WeiqunZhang (approved):
Castro, Maui, Nyx, and WarpX are using `m_idata.id`.  Should we wait till they are fixed?

atmyers:
Yes, I will fix those first.

Src/Particle/AMReX_Particle.H
```diff
@@ -55,7 +55,7 @@ struct Particle
     */
     union im_t
     {
-      struct {int id; int cpu;};
+      int ids[2];
```
WeiqunZhang: Maybe we should remove the union entirely.  Technically it's UB, if one tries to access both `ids` and `arr`.

atmyers: I agree and we should do the same for the real data, but I'd like that to be a follow-up PR since I'll also have to deal with codes using `m_rdata.pos` directly.

atmyers:
These been removed from all the applications I'm aware of, so I will merge this now. If I have missed any, the fix is to replace `p.m_idata.id` with `p.id()` and the equivalent for the cpu number.

PR #1196, opened 2020-07-25
Gcc7 no array bounds

WeiqunZhang:


Tools/GNUMake/comps/gnu.mak
```diff
@@ -105,6 +105,10 @@ ifeq ($(WARN_ALL),TRUE)
     warning_flags += -Wshadow
   endif
 
+  ifeq ($(gcc_major_version),7)
+    warning_flags += -Wno-array-bounds
```
ax3l, line 109: offline discussed: GCC 5-6 fine, GCC 7 potential false positive, GCC9+ fine

ax3l (approved):

Tools/GNUMake/comps/gnu.mak
WeiqunZhang: I also tested GCC 8.4.  It works fine.  Note that it is an known issue that GCC 8.1-8.3 crashes when compiling AMReX.  So we need to avoid GCC 8.1-8.3.

PR #1208, opened 2020-07-28
CI: Fix oneAPI Activation

ax3l:
now located at `/opt/intel/oneapi/setvars.sh`


ax3l:
User-defined literals broken since beta08: https://github.com/intel/llvm/issues/2187

WeiqunZhang:
For the `_rt` literal, maybe we can use `double` when `AMREX_USE_DPCPP`.

ax3l:
Yes, but this should be a temporary fix because it's explicitly not what the C++ standard defines: https://en.cppreference.com/w/cpp/language/user_literal

```
Only the following parameter lists are allowed on literal operators : 
( const char * ) | (1) |  
( unsigned long long int ) | (2) |  
( long double ) | (3)
...
```

Will push a work-around on this PR.

ax3l:
Dang, we cannot trick C++:
```
/home/runner/work/amrex/amrex/Src/Base/AMReX_REAL.H:91:21: error: invalid literal operator parameter type 'double', did you mean 'long double'?
    operator"" _rt( AMREX_TMP_LONGDBL x )
                    ^~~~~~~~~~~~~~~~~~~
/home/runner/work/amrex/amrex/Src/Base/AMReX_REAL.H:74:30: note: expanded from macro 'AMREX_TMP_LONGDBL'
#   define AMREX_TMP_LONGDBL double
                             ^
```

ax3l:
Found a work-around: [`-mlong-double-64`](https://reviews.llvm.org/D6260)
Update: ha, flag gets consumed but is ignored in `dpcpp`.

WeiqunZhang:
We might have a workaround.  For DPC++, we use `const char*` instead of `long double`.  But we will need to implement our own `atof` because the one in `cstdlib` does not work on device.
```
constexpr Real 
operator"" _rt( const char* x )
{
    return amrex::atof(x);
}
```

WeiqunZhang:
This works very well for gcc.  https://godbolt.org/z/rWcEqY But for clang, the constexpr function is not optimization away.  https://godbolt.org/z/GYjTMP

WeiqunZhang:
https://godbolt.org/z/qvxE1s If we force it with a constexpr variable, it works.  But that's not how we use it.

WeiqunZhang:
This however does not compile. https://godbolt.org/z/3Ejzbn

WeiqunZhang:
https://github.com/intel/llvm/pull/2295  `-mlong-double-64` has been fixed.

WeiqunZhang (approved):

ax3l:
Submitted a Clang/LLVM bug report/RFE as discussed on Slack: https://bugs.llvm.org/show_bug.cgi?id=47213

PR #1209, opened 2020-07-28
Domain precision

atmyers:
Due to issues with pointing point precision, there are points inside the `ProbDomain()` that do not map to cells inside the `Domain()` box when binned using something like the `getParticleCell` function. This PR adds an additional domain, the "RoundoffDomain()", to the Geometry object. Points inside the roundoff domain are always sure to map to cells inside `Domain()`. The bounds are calculated when the Geometry object is constructed using bisection. 

This helps solve a number of issues that can happen when you have particles very close to the domain boundaries.

While I have tested this with Nyx and WarpX, there is a chance that there could be a downstream effect on other codes. If the bisection fails when setting the roundoff domain on some problem, then we may need to adjust some tolerances.


Src/Base/AMReX_Algorithm.H
```diff
@@ -62,6 +63,43 @@ namespace amrex
...
+    T bisect (T lo, T hi, F f, T tol=1e-12, int max_iter=100)
+    {
+        AMREX_ASSERT_WITH_MESSAGE(hi > lo,
+            "Error - calling bisect but lo and hi don't describe a reasonable interval.");
+
+        T flo = f(lo);
+        T fhi = f(hi);
+
+        if (flo == 0.0_rt) return flo;
+        if (fhi == 0.0_rt) return fhi;
```
WeiqunZhang: `T(0)` instead of `0.0_rt`? Should we use enable_if to limit T to floating point?

WeiqunZhang: Do you think `std::lower_bound` would work?

Src/Base/AMReX_Geometry.cpp
```diff
@@ -397,4 +400,55 @@ Geometry::growPeriodicDomain (int ngrow) const noexcept
...
+        int ihi = Domain().bigEnd(idim);
+        Real plo = ProbLo(idim);
+        Real phi = ProbHi(idim);
+        Real deltax = CellSize(idim);
+
+#ifdef AMREX_SINGLE_PRECISION_PARTICLE
+        PReal tolerance = std::max(1.e-8*deltax, 1.e-14*phi);
+#else
+        PReal tolerance = std::max(1.e-4*deltax, 1.e-10*phi);
+#endif
```
WeiqunZhang: Seems that the lines should be switched.

Src/Base/AMReX_Geometry.cpp
```diff
@@ -397,4 +400,55 @@ Geometry::growPeriodicDomain (int ngrow) const noexcept
...
+        // bisect the point at which the cell no longer maps to inside the domain
+        PReal lo = static_cast<PReal>(phi) - 0.5_prt*static_cast<PReal>(deltax);
+        PReal hi = static_cast<PReal>(phi) + 0.5_prt*static_cast<PReal>(deltax);
+
+        PReal mid = bisect(lo, hi,
+                           [=] AMREX_GPU_HOST_DEVICE (PReal x) -> PReal
+                           {
+                               int i = int(Math::floor((x - plo)/deltax)) + ilo;
```
WeiqunZhang: Do we mostly use `InvCellSize`?

Src/Base/AMReX_Geometry.cpp
```diff
@@ -397,4 +400,55 @@ Geometry::growPeriodicDomain (int ngrow) const noexcept
...
+                               bool inside = i >= 0 and i <= ihi;
+                               return static_cast<PReal>(inside) - 0.5_prt;
+                           }, tolerance);
+        roundoff_domain.setHi(idim, mid - tolerance);
+    }
+}
+
+bool
+Geometry::outsideRoundoffDomain (AMREX_D_DECL(Real x, Real y, Real z)) const
```
WeiqunZhang, line 437: Should the types be particle_real?

Src/Base/AMReX_Geometry.H
```diff
@@ -275,12 +276,24 @@ public:
...
 
     // is_periodic and RealBox used to be static
     bool    is_periodic[AMREX_SPACEDIM] = {AMREX_D_DECL(false,false,false)};
     RealBox prob_domain;
+
+    // Due to round-off errors, not all floating point numbers for which plo >= x < phi
+    // will map to a cell that is inside "domain". "roundoff_domain" stores a phi
+    // that is very close to that in prob_domain, and for which all floating point numbers
+    // inside it according to a naive inequality check will map to a cell inside domain.
+    RealBox roundoff_domain;
```
WeiqunZhang, line 314: I wonder if this should be two arrays of partice_reals.

WeiqunZhang: Or should we use `Real` everywhere without using `ParticleReal` at all?  Then when particle codes will call `outsideRoundoffDomain` that takes `Real`, we let implicit conversion happen if the types are different.

Src/Base/AMReX_Algorithm.H
WeiqunZhang: Never mind. We don't have iterators here.

atmyers: I originally had an enable_if for floating point types here, but it wouldn't compile with Cuda 9.1: https://github.com/AMReX-Codes/amrex/runs/920763725

9.2 or greater was fine, so I suspect it's not our bug, but I couldn't find a work-around other than removing the enable_if.

WeiqunZhang: I think instead of `typename std::enable_if<std::is_floating_point<T>{}, int>::type = 0` it should be `typename std::enable_if<std::is_floating_point<T>::value, int>::type = 0`.

Src/Base/AMReX_Geometry.cpp
atmyers: Yes, that was ill-formed. Thanks.

Src/Base/AMReX_Geometry.cpp
atmyers: Yes we do, I will do so here as well.

Src/Base/AMReX_Geometry.H
atmyers: The latter seems to work in all the cases I've been testing, anyway.

Src/Base/AMReX_Geometry.cpp
```diff
@@ -397,4 +400,55 @@ Geometry::growPeriodicDomain (int ngrow) const noexcept
...
+#endif
+        // bisect the point at which the cell no longer maps to inside the domain
+        Real lo = static_cast<Real>(phi) - Real(0.5)*static_cast<Real>(deltax);
+        Real hi = static_cast<Real>(phi) + Real(0.5)*static_cast<Real>(deltax);
+
+        Real mid = bisect(lo, hi,
+                          [=] AMREX_GPU_HOST_DEVICE (Real x) -> Real
+                          {
+                              int i = int(Math::floor((x - plo)*idx)) + ilo;
+                              bool inside = i >= 0 and i <= ihi;
```
WeiqunZhang: `i >= 0` -> `i >= ilo`?

WeiqunZhang (approved):

PR #1211, opened 2020-07-28
Add PR template file

drummerdoc:
## Summary

Following CASTRO's lead, this template for PRs will pre-fill PR messages with suggested text.  Hitting "Preview" in the PR window will format the message nicely.

## Checklist

The proposed changes:

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,10 +1,9 @@
 <!-- Thank you for your pull request to AMReX!
 
   1. Be sure the title above is descriptive
-  2. Be sure the PR targets the development branch
-  3. Summarize the changes in a form suitable for use in case the commit is "squashed"
-  4. Provide any background to help reviewers understand the proposed changes
-  5. Fill out the checklist (using [x] to check the box) 
+  2. Provide a summary of the proposed changes
+  3. Provide any background to help reviewers understand the proposed changes
+  4. Fill out the checklist (using [x] to check the box)
 
 NOTE: All text between <-- and --> will not be formatted in md previewers
 -->
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,23 @@
+<!-- Thank you for your pull request to AMReX!
+
+  1. Be sure the title above is descriptive
+  2. Be sure the PR targets the development branch
+  3. Summarize the changes in a form suitable for use in case the commit is "squashed"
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,23 @@
+<!-- Thank you for your pull request to AMReX!
+
+  1. Be sure the title above is descriptive
+  2. Be sure the PR targets the development branch
```
WeiqunZhang: There is only one branch now.  So this can be removed.

.github/pull_request_template.md
```diff
@@ -0,0 +1,23 @@
...
+  3. Summarize the changes in a form suitable for use in case the commit is "squashed"
```
WeiqunZhang: We always squash merge.  So maybe this can be something like
```
Summarize the changes in a form suitable for use in case the combined commit messages are not suitable.
```
Or maybe we should ask the PR submitter to always provide a message that will be used as the final commit message.

WeiqunZhang: This are browser extensions for this.  https://github.com/zachwhaley/squashed-merge-message

WeiqunZhang:
@drummerdoc Your description didn't follow the template. :)

.github/pull_request_template.md
WeiqunZhang: I tried the Firefox extension.  It works.  So I wonder if we should simply use the description (including the checklist) as the final commit message.

drummerdoc:
> @drummerdoc Your description didn't follow the template. :)

It does now!

WeiqunZhang (approved):

atmyers (approved):

kngott:
@drummerdoc How did you misspell "Summary"? It's a template!!

* Also, I feel like we should simplify the checklist substantially (make it much less verbose). TL;DR syndrome. (e.g: Bug Fix... Add Feature... Will Impact Test Suite Answers... etc.)
* Maybe break it into two checklists: one that marks what type of PR it is (the top 2, currently) and then a list of effects that we should be aware of due to this PR (the bottom 3).
* Also, add "updates documentation" to the checklist. We have lots of those PRs!

drummerdoc:
It wasn't a template before this PR!  The answer is that I'm an idiot. fixed above.

I like the abbreviated text, but how about "updates documentation (if appropriate)"

Not sure splitting the checklist helps.

PR #1216, opened 2020-07-31
Average down faces

WeiqunZhang:
New average_down_faces functions that take coarse Geometry and average down periodically.

## Summary

Add new versions of average_down faces factions that take coarse Geometry as an argument.  The average-down takes periodicity into account.

## Checklist

The proposed changes:


asalmgren:
This test fixes the lack of conservation of the no-subcycling multi-level advected scalar problem -- in a 4-level scalar advection problem we now have conservation with this feature.

asalmgren (approved):

PR #1217, opened 2020-07-31
Add a new 3rd order conservative Gaussian Process Based AMR prolongation scheme. 

stevenireeves:
## Summary
Adds a new 3rd order accurate conservative interpolation scheme utilizing Gaussian Processes for 2D and 3D, with ref ratios 2 and 4, both CPU and GPU supported.  Requires USE_LAPACKE=TRUE. 

Most changes are amrex/Src/AmrCore/

AMReX_GP_${DIM}D.{H, cpp} contain the GP class that constructs the linear weights for the GP interpolation -- 2D and 3D. 

AMReX_Interpolator.{H, cpp} 
contains the new class CellGaussianProcess with name gp_interp

AMReX_Interp_${DIM}D_C.H
contains the GP based interpolation routines for 2D and 3D. 

Added a new tutorial, in AmrLevel  -- Slotted Cylinder. Performs the same morphing as the single vortex, but on a slotted cylindrical profile. 

## Additional background
This implementation follows the algorithm described in _An Application of Gaussian Process Modeling for High-order Accurate Adaptive Mesh Refinement Prolongation_ by Steven I Reeves, et al: https://arxiv.org/abs/2003.08508. 

## Checklist

The proposed changes:


stevenireeves:
@WeiqunZhang @asalmgren 
Please review.

WeiqunZhang:
Could you fix the compilation errors first?

stevenireeves:
@WeiqunZhang I'll fix the compilation errors.

stevenireeves:
@WeiqunZhang These features require --llapacke and #include <lapacke.h> to compute interpolation weights. How can I make sure this inclusion is met, or how would  you like this organized?

WeiqunZhang:
We cannot have required dependency on lapacke.  You will have to add `USE_LAPACKE` to the make system and set it to false by default.  The new interpolater will need to be inside `#ifdef AMREX_USE_LAPACKE`.

drummerdoc:
Regarding the dependency: You should add as Weiqun suggests, but there are several ways to manage the building of that library. If you want to include make targets to build/install the lib somewhere in the user's area, I have some examples of that in Pele (the TPL, or third party library, stuff that PeleLM and PelePhysics now use).  However, most AMReX users with these kinds of dependencies manage the install completely outside of AMReX (with Spack, for example, or other modules) and then set a ``XXXX_LIB_DIR=yyy`` variable in the make system(s).

stevenireeves:
@WeiqunZhang I have put the compiler option in for `AMREX_USE_LAPACKE` . Should be good.

stevenireeves:
Closing PR. Will continue to hold GP functionality in my fork.

PR #1218, opened 2020-07-31
Remove instructions....not needed and clutters Slack notices

drummerdoc:
## Summary 

Although the instructions were commented out in the formatted PR text, they clutter up the Slack notices, and aren't really necessary anyway.

## Checklist

The proposed changes:

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,18 +1,8 @@
-<!-- Thank you for your pull request to AMReX!
-
-  1. Be sure the title above is descriptive
-  2. Provide a summary of the proposed changes
-  3. Provide any background to help reviewers understand the proposed changes
-  4. Fill out the checklist (using [x] to check the box)
-
-NOTE: All text between < and > will not be formatted in md previewers
--->
-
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -1,18 +1,8 @@
...
-
-NOTE: All text between < and > will not be formatted in md previewers
--->
-
 ## Summary
 
 ## Additional background
 
-## Checklist
+## Checklist <-- use [x] to check the box -->
```
WeiqunZhang: What do you think about removing the html comment here too or moving to the top of the template?  The reason for it is I use a firefox extension for this and it only removes html comment if it's at the beginning, because it is not always safe to do it in the middle.

drummerdoc: Ha!  The world's most PR'd 3 lines of text!  I say throw that away too.

WeiqunZhang:
```suggestion
## Checklist
```

WeiqunZhang (approved):

PR #1221, opened 2020-08-02
The average_down_faces calls between AMR levels in the linear solvers need to see periodicity

asalmgren:
The average_down_faces calls between AMR levels need to use the interfaces that sees the periodicity.  This requires adding that functionality to EB_average_down_faces (it has already been added to non-EB average_down_faces.)

## Summary
Recently we added the functionality for an average_down_faces call to average fine data from one side to average not only onto the face under it but also onto the face on the other side of the domain if periodic in that direction.  This PR 1) extends that functionality to the EB_average_down_faces call and 2) uses this new functionality when averaging down face-based coefficients between AMR levels in the cell-centered linear solvers.

NOTE: I have left a "TODO" in amrex/Src/EB/AMReX_EBMultiFabUtil.cpp -- I wasn't sure how to best implement the functionality in that case

## Checklist
 
The proposed changes:
- [X] changes answers in the test suite to more than roundoff level
- [X] are likely to significantly affect the results of downstream AMReX users

### DIFF
--- a/Src/EB/AMReX_EBMultiFabUtil.cpp
+++ b/Src/EB/AMReX_EBMultiFabUtil.cpp
@@ -593,82 +593,25 @@ void EB_average_down_faces (const Array<const MultiFab*,AMREX_SPACEDIM>& fine,
 {
     AMREX_ASSERT(crse[0]->nComp() == fine[0]->nComp());
 
-    int ngcrse = 0;
-
-    int ncomp = crse[0]->nComp();
     if (!(*fine[0]).hasEBFabFactory())
     {
-        amrex::average_down_faces(fine, crse, ratio, ngcrse);
+        amrex::average_down_faces(fine, crse, ratio, crse_geom);
     }
     else
     {
-        Dim3 dratio = ratio.dim3();
-
-        const auto& factory = dynamic_cast<EBFArrayBoxFactory const&>((*fine[0]).Factory());
-        const auto&  aspect = factory.getAreaFrac();
-
-        if (isMFIterSafe(*fine[0], *crse[0]))
+        int ngcrse = 0;
+        int ncomp = crse[0]->nComp();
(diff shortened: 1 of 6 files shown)


Src/EB/AMReX_EBMultiFabUtil.cpp
```diff
@@ -587,6 +587,92 @@ void EB_average_down_faces (const Array<const MultiFab*,AMREX_SPACEDIM>& fine,
...
+                            const IntVect& ratio, const Geometry& crse_geom)
+{
+    AMREX_ASSERT(crse[0]->nComp() == fine[0]->nComp());
+
+    int ngcrse = 0;
+
+    int ncomp = crse[0]->nComp();
+    if (!(*fine[0]).hasEBFabFactory())
+    {
+        amrex::average_down_faces(fine, crse, ratio, ngcrse);
```
WeiqunZhang: It seems that we should call the version with Geometry.

Src/EB/AMReX_EBMultiFabUtil.cpp
```diff
@@ -587,6 +587,92 @@ void EB_average_down_faces (const Array<const MultiFab*,AMREX_SPACEDIM>& fine,
...
+#if (AMREX_SPACEDIM == 3)
+                            AMREX_HOST_DEVICE_FOR_3D(tbx,i,j,k,
+                            {
+                                eb_avgdown_face_z(i,j,k,fa,0,ca,0,ap,dratio,ncomp);
+                            });
+#endif
+                        }
+                    }
+                }
+                // TODO: NEED TO ENFORCE PERIODICITY IN THIS CASE
```
WeiqunZhang: I think we could remove this entire block of `if (isMFIterSafe(*fine[0], *crse[0]))`.  If we need to enforce periodicity, we have to do the thing in the else block below.

WeiqunZhang (approved):

PR #1222, opened 2020-08-03
Feature extrapolator gpu

esclapez:
## Summary
Move the Amr/Extrapolater used in IAMR/PeleLM to C++/GPU compliant.

## Additional background
The Extrapolater allows to fill ghost cells with first order extrapolation from interior cells.

## Checklist
The C++ version does not introduce any changes to the results and passed CI compilation tests.

The proposed changes:


drummerdoc (approved):

Src/Amr/AMReX_extrapolater_2D_K.H
```diff
@@ -0,0 +1,143 @@
...
+
+   const int finecell = 1;
+   const int crsecell = 0;
+
+   const auto lo = amrex::lbound(bx);
+   const auto hi = amrex::ubound(bx);
+   int idlo[3] = {D_DECL(lo.x,lo.y,lo.z)};            // base box bounds
+   int idhi[3] = {D_DECL(hi.x,hi.y,hi.z)};
+   int idlo_e[3] = {D_DECL(lo.x,lo.y,lo.z)};          // extended box bounds
+   int idhi_e[3] = {D_DECL(hi.x,hi.y,hi.z)};
```
WeiqunZhang: Have to use `AMREX_D_DECL`.

esclapez: Indeed. I'll have to go over PeleLM to change that too ...

Src/Amr/AMReX_Extrapolater.cpp
```diff
@@ -1,53 +1,42 @@
...
-	    const Box& databox = datafab.box();
-
-	    amrex_first_order_extrap(datafab.dataPtr(), databox.loVect(), databox.hiVect(), N, 
-			       maskfab.dataPtr(), maskbox.loVect(), maskbox.hiVect(),
-			       bx.loVect(), bx.hiVect(), scomp, ncomp);
-	}
+        for (MFIter mfi(mf); mfi.isValid(); ++mfi)
+        {
+            const Box& bx      = mfi.validbox();
+            auto const& mask_arr = mask.array(mfi);
```
WeiqunZhang: We can do `auto const& mask_arr = mask.const_array(mfi);` so that the compiler does not have to do conversion from `Array4<int>` to `Array4<int const>`.

Src/Amr/AMReX_extrapolater_2D_K.H
drummerdoc: Hmm, maybe should declare as int[AMREX_SPACEDIM], just to be more explicit?

Src/Amr/AMReX_Extrapolater.cpp
```diff
@@ -1,53 +1,42 @@
...
+            auto const& data_arr = mf.array(mfi,scomp);
+
+            amrex::launch(bx, [mask_arr,data_arr,ncomp]
```
WeiqunZhang: Can we use `AMREX_LAUNCH_HOST_DEVICE_LAMBDA` so that the gpu launch region is respected.
```
AMREX_LAUNCH_HOST_DEVICE_LAMBDA(bx, tbx, { amrex_first_order_extrap(tbx,ncomp,mask_arr,data_arr); });
```

Src/Amr/AMReX_extrapolater_1D_K.H
```diff
@@ -0,0 +1,33 @@
...
+
+AMREX_GPU_HOST_DEVICE
+AMREX_FORCE_INLINE
+void
+amrex_first_order_extrap(amrex::Box const& bx,
+                         int               nComp,
+                         amrex::Array4<const int>   const& mask,
+                         amrex::Array4<amrex::Real> const& data) noexcept
+{
+   const int crsecell = 0;
```
WeiqunZhang: We can use `constexpr` instead of `const`.  Do the same for 2d and 3d too.

Src/Amr/AMReX_extrapolater_2D_K.H
esclapez: At some point I had 1D/2D and 3D in the same file but Kevin suggested splitting the files to be more compliant with AMReX style and it's a remnant of that. I'll actually clean that up.

Src/Amr/AMReX_extrapolater_2D_K.H
```diff
@@ -0,0 +1,143 @@
...
+                         amrex::Array4<amrex::Real> const& data) noexcept
+{
+   using namespace amrex::literals;
+
+   const int finecell = 1;
+   const int crsecell = 0;
+
+   const auto lo = amrex::lbound(bx);
+   const auto hi = amrex::ubound(bx);
+   int idlo[3] = {AMREX_D_DECL(lo.x,lo.y,lo.z)};            // base box bounds
```
WeiqunZhang: Can we use lo and hi directly without having `idlo` and `idhi`?

Src/Amr/AMReX_extrapolater_2D_K.H
```diff
@@ -0,0 +1,143 @@
...
+   }
+
+   int k = idlo[2];
+
+   for (int n = 0; n < nComp; n++) {
+
+      // set all crse cells to zero first
+      for (int j = idlo_e[1]; j <= idhi_e[1]; ++j) {
+         for (int i = idlo_e[0]; i <= idhi_e[0]; ++i) {
+            if (mask(i,j,k) == crsecell) data(i,j,k,n) = 0.0_rt;
```
WeiqunZhang, line 29: In the GPU case, `bx` has only a single cell and a thread works on a single cell.  This line will cause race conditions.

WeiqunZhang (changes requested):

WeiqunZhang:
It might be easier to have separate host and device implementations.  For the GPU version, a kernel can be launched with `amrex::ParallelFor(amrex::grow(vbx,1),[=] ...)`.  In the GPU kernel, we only work on the cell if `msk(i,j,k) .eq. crsecell`, and a thread does not write to any cells other than (i,j,k).

Src/Amr/AMReX_extrapolater_2D_K.H
esclapez: Can I just use two kernels ?

```
amrex::ParallelFor(amrex::grow(bx,1), ncomp,
[=] AMREX_GPU_DEVICE (int i, int j, int k, int n) noexcept
{   
   if (mask_arr(i,j,k) == crsebnd) data_arr(i,j,k,n) = 0.0;
}); 

AMREX_LAUNCH_HOST_DEVICE_LAMBDA(bx, tbx,
{   
    amrex_first_order_extrap(tbx, ncomp, mask_arr, data_arr);
}); 
```
and not have the first nested loops in the `amrex_first_order_extrap` or we actually have the problem throughout the entire function ? Note that I might just be lucky, but I did not ran into troubles running the code in its current version ... (on both Garuda and Ascent)
Since we actually only care about the ghost cells here, maybe I should go with something like what you did `GpuBndryFuncFab<F>::ccdoit` ?

WeiqunZhang:
My suggestion is to have separate implementations for cpu and gpu.  The cpu version can be a direct translation from Fortran to C++.  The gpu version will work on a single cell (which is either valid cell or ghost cell).  There is no need for two kernel launches.

You didn't notice any problems is because multiple threads write the same value to the same memory location.  But there are no reasons for a thread to do extra work that will be wasted.  A thread should not write to cell (i,j,k), not its neighbours.

WeiqunZhang:
You don't need any loops in the gpu version, just a number of if tests.

Src/Amr/AMReX_extrapolater_2D_K.H
```diff
@@ -0,0 +1,234 @@
...
+            }
+         }
+      }
+      // Edges
+      // xlo, y-valid
+      {
+         int i = lo.x-1;
+         for (int j = lo.y; j <= hi.y; ++j) {
+            if ( mask(i,j,k) == crsecell ) {
+               data(i,j,k,n) = ( mask(i,j-1,k) * data(i,j-1,k,n) + mask(i+1,j,k) + mask(i,j+1,k) * data(i,j+1,k,n) )
```
WeiqunZhang: This should be `data(i,j,k,n) = ( mask(i,j-1,k) * data(i,j-1,k,n) + data(i+1,j,k) + mask(i,j+1,k) * data(i,j+1,k,n) )`.

Src/Amr/AMReX_extrapolater_2D_K.H
```diff
@@ -0,0 +1,234 @@
...
+                               / ( mask(i,j-1,k) + 1 + mask(i,j+1,k) );
+            }
+         }
+      }
+      // xhi, y-valid
+      {
+         int i = hi.x+1;
+         for (int j = lo.y; j <= hi.y; ++j) {
+            if ( mask(i,j,k) == crsecell ) {
+               data(i,j,k,n) = ( mask(i,j-1,k) * data(i,j-1,k,n) + mask(i-1,j,k) + mask(i,j+1,k) * data(i,j+1,k,n) )
```
WeiqunZhang: Similar typo here.

Src/Amr/AMReX_extrapolater_2D_K.H
```diff
@@ -0,0 +1,234 @@
...
+   // set all crse cells to zero first
+   if (mask(i,j,k) == crsecell) data(i,j,k,n) = 0.0_rt;
+
+   // Corners
+   // xlo, ylo
+   if ( (i == lo.x-1) && (j == lo.y-1) ) {
+      if ( mask(i,j,k) == crsecell ) {
+         if ( ( mask(i+1,j,k) == finecell ) ||
+              ( mask(i,j+1,k) == finecell ) ) {
+            data(i,j,k,n) = ( mask(i+1,j,k) * data(i+1,j,k,n) + mask(i,j+1,k) * data(i,j+1,k,n) )
```
WeiqunZhang, line 156: We may have to write this as
```
if ((mask(i+1,j,k) == finecell) && (mask(i,j+1,k) == finecell)) {
} else if (mask(i+1,j,k) == finecell) {
} else if (mask(i,j+1,k) == finecell) {
} else {
}
```
The reason for that is data(i+1,j,k) or data(i,j+1,k,n) may contains nan.  In the CPU version, we have a loop in the beginning to set all coarse cells to zero.  But here we do not have that.  We should also remove the lines
```
// set all crse cells to zero first
   if (mask(i,j,k) == crsecell) data(i,j,k,n) = 0.0_rt;
```
because it won't be able to serve the purpose like in the CPU case.

WeiqunZhang: Maybe it's easier to launch two kernels as you suggested earlier, otherwise the 3d version is going to have so many branches.

Src/Amr/AMReX_extrapolater_2D_K.H
WeiqunZhang: Note that the typo is in GPU version too.

WeiqunZhang (changes requested):
It looks good to me.  Some comments.

(1) There are a few typos, please correct them.

(2) You are right.  We should launch two kernels.  Both kernels will be launched on the grown box.

(3) For the GPU versions, I suggest the following style so that not every tests have to be done and threads can exit earlier especially those assigned to valid cells.
```
if (mask(i,j,k) == crse(i,j,k))
{
   if ((i==lo.x-1) && (j == lo.y-1) && (k == lo.z-1)) {
   } else if (...) {
   } else if (...) {
   .......
   } else {
   } 
}
```

drummerdoc:
I thought threads don't "exit early"...I thought they all did all the branches of the if-tests and masked out results.

WeiqunZhang:
Maybe all 32 threads in a warp are assigned to valid cells.

Maybe 31 threads are assigned to valid cells and the other thread is the x-lo && y-lo && z-corner.

Src/Amr/AMReX_extrapolater_2D_K.H
esclapez: My eyes clearly had trouble focusing there ... too much i,j,k, +1, -1 ... I'll fix those and go over it again.

WeiqunZhang:
Marc, I guess I didn't answer you directly.  On recent Nvidia GPUs, threads within a warp can run asynchronously because they have their own program counter.  Even on GPUs that threads in a warp are synchronous, if a thread only needs to go as far as the third test and the rest only need to go as far as the second test, they do not need the run the rest of the tests in the kernel, because none of them need to.  If and only if one thread has to do the last test, the other threads in that warp cannot exit earlier.

drummerdoc:
ahh, thanks for that explanation!

WeiqunZhang (approved):

PR #1224, opened 2020-08-03
Optimization of FPinfo and complementIn

WeiqunZhang:
## Summary

Use simplified BoxArray in FPinfo and parallelize FPinfo.  Add parallel version of
BoxList::parallelComplementIn and use it in AmrMesh::MakeNewGrids.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_BoxArray.cpp
+++ b/Src/Base/AMReX_BoxArray.cpp
@@ -1275,99 +1275,90 @@ BoxArray::complementIn (BoxList& bl, const Box& bx) const
     bl.set(bx.ixType());
     bl.push_back(bx);
 
-    if (!empty()) 
-    {
-	BARef::HashType& BoxHashMap = getHashMap();
+    if (empty()) return;
 
-	BL_ASSERT(bx.ixType() == ixType());
+    BARef::HashType& BoxHashMap = getHashMap();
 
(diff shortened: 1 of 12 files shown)


WeiqunZhang:
This PR has passed AMReX and incflo regression tests on battra.

Src/Base/AMReX_BoxArray.cpp
```diff
@@ -1275,99 +1275,90 @@ BoxArray::complementIn (BoxList& bl, const Box& bx) const
...
-                            }
-                            bl.swap(newbl);
-                        }
+                }
+            }
+        });
+    } else {
+        AMREX_LOOP_3D(cbx, i, j, k,
+        {
+            amrex::ignore_unused(j,k);
```
atmyers, line 1339: could this be incorporated into the `AMREX_LOOP_3D` macro somehow?

WeiqunZhang: We certainly could.

atmyers (approved):

Src/Base/AMReX_BoxArray.cpp
WeiqunZhang: I will do this in a new PR.

PR #1227, opened 2020-08-04
Optimization of the construction of SFC

WeiqunZhang:
## Summary

The Morton numbers are explicitly computed and stored in SFCTokens.  This
speeds up the sorting in making space filling curve significantly.

## Checklist

The proposed changes:


WeiqunZhang:
I wrote a test code to compare the old and new approaches by sorting a vector of 1e8 tokens.  The results agree with each other.  https://github.com/WeiqunZhang/random/tree/main/sfc

atmyers (approved):

PR #1228, opened 2020-08-04
Small volfrac

drangara:
## Summary

When extend_domain_face is set to true we do the following: If we set a cell that lies on one of the 6 faces of the domain as covered because its volfrac is less than the small_volfrac , we are forcing the corresponding ghost cells on that face to also be covered.

## Checklist

The proposed changes:

### DIFF
--- a/Src/EB/AMReX_EB2_2D_C.cpp
+++ b/Src/EB/AMReX_EB2_2D_C.cpp
@@ -277,28 +277,44 @@ void build_cells (Box const& bx, Array4<EBCellFlag> const& cell,
         }
     });
 
+    // set cells in the extended region to covered if the
+    // corresponding cell on the domain face is covered
     if(extend_domain_face) {
        AMREX_HOST_DEVICE_FOR_3D ( bxg1, i, j, k,
        {
            const auto & dlo = geom.Domain().loVect();
            const auto & dhi = geom.Domain().hiVect();
 
-           if(not cell(i,j,k).isCovered()) {
+           // find the cell(ii,jj,kk) on the corr. domain face
+           // this would have already been set to correct value
+           int ii = i;
+           int jj = j;
+           int kk = k;
+           if(i < dlo[0]) {
+               ii = dlo[0];
+           }
+           else if(i > dhi[0]) {
+               ii = dhi[0];
+           }
 
-              if(not geom.isPeriodic(0)) {
-                 if( (i < dlo[0] and cell(dlo[0],j,k).isCovered()) or 
-                     (i > dhi[0] and cell(dhi[0],j,k).isCovered()) ) 
-                 {
-                     set_covered(i,j,cell,vfrac,vcent,barea,bcent,bnorm);
-                 }
+           if(j < dlo[1]) {
+               jj = dlo[1];
+           }
+           else if(j > dhi[1]) {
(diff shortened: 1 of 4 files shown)


drangara:
When `extend_domain_face` is set to true we do the following: If we set a cell that lies on one of the 6 faces of the domain as `covered` because its volfrac is less than the small_volfrac , we force the corresponding ghost cells on that face to also be `covered`. 

Tested various parameters of a `titled cylinder` test case to make sure the initial projection converges.

Src/EB/AMReX_EB2_3D_C.cpp
```diff
@@ -646,18 +671,48 @@ void build_cells (Box const& bx, Array4<EBCellFlag> const& cell,
...
+                   const auto & dlo = domain.loVect();
+                   const auto & dhi = domain.hiVect();
+                   const auto & lo = bxg1.loVect();
+                   const auto & hi = bxg1.hiVect();
+
+                   // if cell lies on any of the 6 domain faces, force
+                   // all layers of corresponding ghost cells to be covered
+                   if(i == dlo[0]) {
+                     for(int ii = lo[0]; ii < dlo[0]; ++ii) {
+                        set_covered(ii,j,k,cell,vfrac,vcent,barea,bcent,bnorm);
```
WeiqunZhang: The problem is there is a race condition on GPU.  There is another thread that has been assigned to work on cell (ii,j,k).  If both threads are trying to write the data of cell(ii,j,k), it's undefined which thread will win.  I think you will have to fix the extended face after this kernel.  You can launch a kernel on `bxg1`.  In the kernel, you test on if this cell in extended region.  If yes, you test on its original source to see if that has been covered.  If that is covered, you can then set this extended cell to covered.  Also note that you need to make sure periodic boundary has been taken into account.  We do not extend at periodic boundaries.

WeiqunZhang (changes requested):

Src/EB/AMReX_EB2_2D_C.cpp
```diff
@@ -252,19 +272,55 @@ void build_cells (Box const& bx, Array4<EBCellFlag> const& cell,
...
-                bnorm(i,j,0,1) = 0.0;
-                cell(i,j,0).setCovered();
+               set_covered(i,j,cell,vfrac,vcent,barea,bcent,bnorm);
             }
         }
     });
 
+    // set cells in the extended region to covered if the
+    // corresponding cell on the domain face is covered
+    if(extend_domain_face) {
```
WeiqunZhang, line 282: We don't need to launch a kernel if the box is not at the domain boundary.  Moreever, `Geometry` cannot be used on device.  How about something like this here?
```
Box gdomain = geom.Domain();
    if (geom.isPeriodic(idim)) {
        gdomain.setSmall(idim,std::min(gdomain.smallEnd(idim),bxg1.smallEnd(idim)));
        gdomain.setBig(idim,std::max(gdomain.bigEnd(idim),bxg1.bigEnd(idim)));
    }
}
if (not gdomain.contains(bxg1)) {
    AMREX_HOST_DEVICE_FOR_3D ( bxg1, i, j, k,
   {
       // No need to use geom
       // use gdomain
   });
}
```

WeiqunZhang (approved):

PR #1232, opened 2020-08-05
Oversetmask

WeiqunZhang:
## Summary

Switch the overset mask convention to that 1 means unknown and 0 means
known.  The new convention is more convenient for AMR-Wind.

## Checklist

The proposed changes:


WeiqunZhang:
@sayerhs

asalmgren (approved):

PR #1235, opened 2020-08-06
Changing computation at inflow/outflow for tensor solve

OscarAntepara:
## Summary
Computation at inflow boundaries for the MLEBTensor and inflow/outflow for MLTensor has been changed.

## Additional background
At outflow the tensor solve uses extrapolated values from the interior cell to the outflow face. At inflow, the tensor solve uses boundary data, evaluated at the face, that has been stored in the ghost cell next to the inflow face.

## Checklist

The proposed changes:


Src/LinearSolvers/MLMG/AMReX_MLEBTensorOp.cpp
```diff
@@ -479,21 +479,30 @@ MLEBTensorOp::compCrossTerms(int amrlev, int mglev, MultiFab const& mf) const
 	  AMREX_D_TERM(Array4<Real const> const kapxfab = kapmf[0].const_array(mfi);,
 		       Array4<Real const> const kapyfab = kapmf[1].const_array(mfi);,
 		       Array4<Real const> const kapzfab = kapmf[2].const_array(mfi););
-	  
+
+          const auto & bdcv = bcondloc.bndryConds(mfi);
+          Array2D<BoundCond,0,2*AMREX_SPACEDIM-1,0,AMREX_SPACEDIM-1> bct;
+          for (OrientationIter face; face; ++face) {
+              Orientation ori = face();
+              for (int icomp = 0; icomp < AMREX_SPACEDIM; ++icomp) {
```
WeiqunZhang: It will be slightly more efficient to do
```
for (int icomp ...)
    for (OrientationIter ...)
```
Could you switch the looping order here at similar other places?

Src/LinearSolvers/MLMG/AMReX_MLTensorOp.cpp
```diff
@@ -211,6 +211,11 @@ MLTensorOp::apply (int amrlev, int mglev, MultiFab& out, MultiFab& in, BCMode bc
     applyBCTensor(amrlev, mglev, in, bc_mode, s_mode, bndry );
 
     const auto dxinv = m_geom[amrlev][mglev].InvCellSizeArray();
+    const Geometry& geom = m_geom[amrlev][mglev];
+    const Box& domain = amrex::enclosedCells(geom.Domain());
```
WeiqunZhang: The call to `enclosedCells` here and other places is actually unnecessary because `Geometry::Domain()` is always cell-centered.

WeiqunZhang (changes requested):
Looks good to me.  Just two minor comments.

Src/LinearSolvers/MLMG/AMReX_MLEBTensorOp.cpp
OscarAntepara: Thanks for the review!! I did the changes as suggested.

WeiqunZhang (approved):

PR #1241, opened 2020-08-08
Fix bug in FPinfo for EB

WeiqunZhang:
## Summary

PR #1224 introduced a bug that affects FillPatch on nodal grids for EB.
When there is only one ghost cell, the fine patch grids might be degenerate
in the sense that some boxes are only 1 node wide.  In that case, we cannot
build cell-centered EBCellFlags unless there are ghost cells.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_FabArrayBase.cpp
+++ b/Src/Base/AMReX_FabArrayBase.cpp
@@ -1271,11 +1271,12 @@ FabArrayBase::FPinfo::FPinfo (const FabArrayBase& srcfa,
                                            ba_crse_patch,
                                            dm_patch,
                                            {0,0,0}, EBSupport::basic);
+        int ng = boxtype.cellCentered() ? 0 : 1; // to avoid dengerate box
         fact_fine_patch = makeEBFabFactory(index_space,
                                            index_space->getGeometry(fdomain),
                                            ba_fine_patch,
                                            dm_patch,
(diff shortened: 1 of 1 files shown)


cgilet (approved):

Src/Base/AMReX_FabArrayBase.cpp
```diff
@@ -1271,11 +1271,12 @@ FabArrayBase::FPinfo::FPinfo (const FabArrayBase& srcfa,
                                            ba_crse_patch,
                                            dm_patch,
                                            {0,0,0}, EBSupport::basic);
+        int ng = boxtype.cellCentered() ? 0 : 1; // to avoid dengerate box
         fact_fine_patch = makeEBFabFactory(index_space,
                                            index_space->getGeometry(fdomain),
                                            ba_fine_patch,
                                            dm_patch,
-                                           {0,0,0}, EBSupport::basic);
+                                           {ng,ng,ng}, EBSupport::basic);
```
drummerdoc, line 1279: Does this need a AMREX_D_DECL?

WeiqunZhang: No, it's not for directions.  It's for ebflag, volume and face types.

drummerdoc (approved):

PR #1243, opened 2020-08-09
oneAPI Beta 8

WeiqunZhang:
## Summary

In beta8, C++11 user defined literals are broken because of the use of long
double (https://github.com/intel/llvm/issues/2187). 
We have implemented a workaround by overloading operator "" _rt
with `const char*` argument and our own `atof`.  But, because Clang does not
evaluate the new version of the operator at compile time, we now avoid using
`_rt` in AMReX.  In the long term, we may have to remove it entirely if the
long double issue is not resolved in DPC++.

With the workaround, we can enable the DPC++ CI test again.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_REAL.H
+++ b/Src/Base/AMReX_REAL.H
@@ -68,6 +68,63 @@ namespace amrex {
     using Real = amrex_real;
     using ParticleReal = amrex_particle_real;
 
+#ifdef AMREX_USE_DPCPP
+namespace detail {
+    constexpr double atof (const char* p)
+    {
+        while (*p == ' ' || *p == '\t') ++p;
+
+        double sign = (*p == '-') ? -1.0 : 1.0;
+        if (*p == '-' || *p == '+') ++p;
+
+        double r = 0.;
+        while (*p >= '0' && *p <= '9') {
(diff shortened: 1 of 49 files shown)


kngott (approved):
The atof function looks like it will accept a lot of different things, e.g. --0.2efe.233. There aren't any errors for unexpected patterns.

I don't think it'll matter, given it's DPC++ only and temporary, but it's probably good to be aware in case there's any weird output.

atmyers (approved):

ax3l:
>  In the long term, we may have to remove it entirely if the long double issue is not resolved in DPC++.

With https://github.com/intel/llvm/pull/2295 merged, we can just use `-mlong-double-64` in (a potential) beta09 release, I think (unless the underlying issue is fixed directly, too): #1208

Src/Base/AMReX_REAL.H
```diff
@@ -68,6 +68,63 @@ namespace amrex {
     using Real = amrex_real;
     using ParticleReal = amrex_particle_real;
 
+#ifdef AMREX_USE_DPCPP
```
ax3l, line 71: wow.
:point_down:

PR #1252, opened 2020-08-10
CMake: re-factor third party libraries setup

mic84:
## Summary
Centralize setup of all parallel backends dependencies.

## Checklist

The proposed changes:


ax3l (approved):
LGTM, thank you!

PR #1254, opened 2020-08-11
Correcting tensor cross terms computation for periodic bcs

OscarAntepara:
## Summary
Correcting tensor cross terms computation for periodic bcs

## Additional background
Fixing a bug introduced in PR #1235. Now it computes in the correct way the cross terms in the tensor solve for problems with periodic bcs. 

## Checklist

The proposed changes:
- [x] fix a bug or incorrect behavior in AMReX
- [ ] add new capabilities to AMReX
- [x] changes answers in the test suite to more than roundoff level
- [x] are likely to significantly affect the results of downstream AMReX users
- [ ] are described in the proposed changes to the AMReX documentation, if appropriate

PR #1262, opened 2020-08-13
Tensor solver and extension of EB domain face

WeiqunZhang:
## Summary

ParmParse parameter eb2.extend_domain_face is now true by default.  The
reason for the change is that for most applications it's hard to imagine one
would want it to be false.  It will also solve the EB tensor solver issue
when there is a cut cell just outside the domain abutting a covered valid
cell.

Recent changes to the tensor solvers are incorrect and are reverted here.
They are incorrect because the ghost cells in the modified functions
actually have values at the cell centered.  Although the users put domain
face values in the ghost cells, that is not what the solver does.  The
solver stores the boundary values in boundary registers.  Before the stencil
is being applied, bc function is called to fill the ghost cells with
properly interpolated or extrapolated values so that the stencil operations
are just like working on normal interior cells.

Revert "Correcting tensor cross terms computation for periodic bcs (#1254)"
This reverts commit 1197adc5be6144bae9e362c114ca16b879006180.

Revert "Changing computation at inflow/outflow for tensor solve (#1235)"
This reverts commit b391885e918d63fa1741f954d2792e31426a6204.

Revert "Changing computation at outflow for the EBTensor (#1187)"
This reverts commit 9cd538802ea5b3ab7e6db10d6046933e937aa449.

## Checklist

The proposed changes:

### DIFF
--- a/Src/EB/AMReX_EB2.cpp
+++ b/Src/EB/AMReX_EB2.cpp
@@ -18,7 +18,7 @@ namespace amrex { namespace EB2 {
 Vector<std::unique_ptr<IndexSpace> > IndexSpace::m_instance;
 
 int max_grid_size = 64;
-bool extend_domain_face = false;
+bool extend_domain_face = true;
 
 void Initialize ()
 {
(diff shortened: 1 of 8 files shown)


WeiqunZhang:
@OscarAntepara Could you try mfix with tilted cylinder to see if the EB tensor solver is OK?

OscarAntepara:
it looks ok to me. Thanks!

OscarAntepara (approved):

asalmgren (approved):

PR #1265, opened 2020-08-16
updated Make.machines to include new frontier-coe machines and change…

mxxw:
…d 'hostname -f' command to 'hostname'

## Summary
updated Make.machines:
1. added poplar and redwood to frontier-coe machines;
2. changed name string matching for frontier-coe to only
    yield exact matches (filter); hence, !="" means exact match;
3. changed `hostname -f` command (which includes the domain) to `hostname` to
    avoid false negatives, e.g., not recognizing `hostname -f`==tulip.cm.cluster as tulip;

## Checklist

The proposed changes:

### DIFF
--- a/Tools/GNUMake/Make.machines
+++ b/Tools/GNUMake/Make.machines
@@ -9,7 +9,8 @@ ifdef $(HOSTNAME)
 else ifdef $(HOST)
   host_name := $(strip $(HOST))
 else
-  host_name := $(shell hostname)
+  host_name := $(shell hostname -f)
+  host_name_short := $(shell hostname -s)
 endif
 
 # MACHINES supported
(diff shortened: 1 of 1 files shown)


Tools/GNUMake/Make.machines
```diff
@@ -9,7 +9,7 @@ ifdef $(HOSTNAME)
 else ifdef $(HOST)
   host_name := $(strip $(HOST))
 else
-  host_name := $(shell hostname -f)
+  host_name := $(shell hostname)
```
WeiqunZhang: This might break on other machines that include `Make.$(host_name)`.  Maybe we can keep the way `host_name` is and add an extra variable `short_host_name := $(shell hostname -s)`.

WeiqunZhang (changes requested):

mxxw:
Your suggested approach of
     short_host_name := $(shell hostname -s)
looks good to me.
I intend to implement that and
create another pull request,
hopefully later today.

thanks,
marcus


Marcus Wagner, Ph.D.
Application Engineer
Hewlett Packard Enterprise
Work: (909) 248-0597
Cell: (310) 902-0676
marcus.wagner@hpe.com

WeiqunZhang (approved):

PR #1268, opened 2020-08-16
DPC++: Missing synchronize()

WeiqunZhang:
## Summary

Missing Gpu::synnhronize() call after host to device memcpy.

## Checklist

The proposed changes:


atmyers:
Is this redundant with the one at line 229?

WeiqunZhang:
The one at line 229 is certainly redundant for dpc++.   We do need the newly added one for DPC++ because the source of the host memory is on the local stack of that function.  For CUDA, I don't think the sync at line 229 is needed, but I am not sure about HIP and I am not sure I fully understand what's said about these things in the CUDA programming guide.  So I am playing safe here.

atmyers (approved):

PR #1269, opened 2020-08-16
fixed bug in matching of COMP_VERSION to correctly treat CCE > 9

mxxw:
## Summary
changed Tools/GNUMake/comps/cray.mak to fix bugs in the matching of $(COMP_VERSION) in order to correctly treat CCE version > 9

## Additional background
Starting with version 9, the Cray CCE C and C++ compilers changed to being clang/LLVM based
which makes it necessary to change some flags compared to CCE versions <=8.
(This change does not apply to the Cray Fortran compiler versions.)
However, the original version check of
    COMP_VERSION = $(shell echo $(CRAY_CC_VERSION) | cut -f 1 -d .)
in combination with the tests of  "ifeq ($(COMP_VERSION),9)"
differentiated only between V9 and non-V9, which it assumed to be V8.
Howerver, now, we have V10 while V11 is being tested, all of which would
be treated like V8, which is incorrect.
To address this issue the compiler version classification is now done with
     ifeq ($(shell expr $(COMP_VERSION) \>= 9), 1)
            CCE_GE_V9 := TRUE
     else
            CCE_GE_V9 := FALSE
     endif
and it is tested with
     ifeq ($(CCE_GE_V9),TRUE)

## Checklist

The proposed changes:

### DIFF
--- a/Tools/GNUMake/comps/cray.mak
+++ b/Tools/GNUMake/comps/cray.mak
@@ -19,11 +19,16 @@ F90FLAGS =
 # still based on CCE 8 and so it has the same options as before.
 COMP_VERSION = $(shell echo $(CRAY_CC_VERSION) | cut -f 1 -d .)
 
+ifeq ($(shell expr $(COMP_VERSION) \>= 9), 1)
+  CCE_GE_V9 := TRUE
+else
+  CCE_GE_V9 := FALSE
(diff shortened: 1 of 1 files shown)


WeiqunZhang (approved):
Thanks!

PR #1286, opened 2020-08-20
Physical locations in plotfiles

WeiqunZhang:
## Summary

Normally, the integer index of a Geometry's Domain starts with 0.  But if
not, the physical location of boxes stored in plotfiles is incorrect,
because the RealBox constructor we use takes the physical location of index
0.  So we need to shift the box before using RealBox.

## Checklist

The proposed changes:


atmyers (approved):

RevathiJambunathan:
Thank you @WeiqunZhang ! :+1:

PR #1302, opened 2020-08-24
Move declaration of Array4 object

drummerdoc:
## Summary

The autogen function for refinement tagging on physical box doesn't make use of a field, but the controller was trying to make an Array4 using the field that doesn't exist in this case.  This fix moves the declaration down into a place where the data is guaranteed to exist

The proposed changes:

### DIFF
--- a/Src/AmrCore/AMReX_ErrorList.cpp
+++ b/Src/AmrCore/AMReX_ErrorList.cpp
@@ -383,32 +383,35 @@ operator << (std::ostream&    os,
         for (MFIter mfi(tba,TilingIfNotGPU()); mfi.isValid(); ++mfi)
         {
           const auto& bx    = mfi.tilebox();
-          auto const& dat   = mf->array(mfi);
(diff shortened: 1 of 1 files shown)


Src/AmrCore/AMReX_ErrorList.cpp
```diff
@@ -383,32 +383,35 @@ operator << (std::ostream&    os,
...
+            }
+            else if (m_test == GREATER)
+            {
+              AMRErrorTag_GREATER(bx, dat, tag, m_value, tagval);
+            }
+            else if (m_test == VORT)
+            {
+              AMRErrorTag_VORT(bx, dat, tag, level, m_value, tagval);
+            }
+            {
```
WeiqunZhang, line 413: missing `else`?

drummerdoc: Yup!  Fixing...

WeiqunZhang (approved):

PR #1303, opened 2020-08-25
Modify the use of HOST and HOSTNAME in make system

WeiqunZhang:
## Summary

PR #1275 fixed a bug in the use of `HOST` and `HOSTNAME` in the make system.
However, the fix broke on Summit because `HOSTNAME` on Summit login nodes
is set to something like `login1`.  We now set `host_name` to `hostname -f` and
`host_name_short` to `hostname -s` command, which is the behavior before the fix
in #1275.  We add a new make variable `host_name_env` that is set to
`HOSTNAME` and `HOST` from the environment and use it for tulip.

## Checklist

The proposed changes:

### DIFF
--- a/Tools/GNUMake/Make.machines
+++ b/Tools/GNUMake/Make.machines
@@ -4,6 +4,14 @@
 which_site := unknown
 which_computer := unknown
 
+ifdef HOSTNAME
+  host_name_env := $(strip $(HOSTNAME))
+else ifdef HOST
+  host_name_env := $(strip $(HOST))
+else
+  host_name_env := unknown
+endif
+
 host_name := $(shell hostname -f)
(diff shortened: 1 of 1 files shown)


drummerdoc (approved):

maximumcats (approved):

WeiqunZhang:
Unfortunately, it breaks on the compute node of tulip.  The only way there is to use $HOSTNAME.

WeiqunZhang:
I think we can add `hostname_env` in addition to what we have and use the new variable for tulip's login node.

maximumcats:
Does `uname -n` work there?

WeiqunZhang:
No.  It does not.  It returns `node001`.

PR #1307, opened 2020-08-25
Add get_pmap method to amrex_distromap (Fortran)

kweide:
## Summary
Add a new method to the `amrex_distromap` type in F_Interfaces. `get_pmap` fills a caller-owned array of PEs.

Changes to make this more elegant are welcome.

## Additional background
We need something like this in order to actually look into a distribution mapping computed by AMReX from Fortran code.
(Currently this functionality is desired for Implementing a parallel restart capability from FLASH checkpoint files in FLASH.)

## Checklist

The proposed changes:

### DIFF
--- a/Src/F_Interfaces/Base/AMReX_distromap_fi.cpp
+++ b/Src/F_Interfaces/Base/AMReX_distromap_fi.cpp
@@ -33,7 +33,7 @@ extern "C" {
 	Long dmsize = dm->size();
 	AMREX_ASSERT(plen >= dmsize);
 	for (int i = 0; i < dmsize; ++i)
-            pmap[i] = dm[i];
+            pmap[i] = (*dm)[i];
     }
 
     void amrex_fi_print_distromap (const DistributionMapping* dm)
--- a/Src/F_Interfaces/Base/AMReX_distromap_fi.cpp
+++ b/Src/F_Interfaces/Base/AMReX_distromap_fi.cpp
@@ -28,6 +28,14 @@ extern "C" {
 	dmo = new DistributionMapping(*dmi);
     }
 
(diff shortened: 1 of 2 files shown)


Src/F_Interfaces/Base/AMReX_distromap_fi.cpp
```diff
@@ -28,6 +28,14 @@ extern "C" {
 	dmo = new DistributionMapping(*dmi);
     }
 
+    void amrex_fi_distromap_get_pmap (DistributionMapping*& dm, int* pmap, const int plen)
```
WeiqunZhang: Could we change `DistributionMapping*&` to `const DistributionMapping*` and make another change in Fortran to match this?

Src/F_Interfaces/Base/AMReX_distromap_mod.F90
```diff
@@ -71,6 +72,14 @@ subroutine amrex_fi_distromap_maxsize (dm,n) bind(c)
        integer(c_int), value :: n
      end subroutine amrex_fi_distromap_maxsize
 
+     subroutine amrex_fi_distromap_get_pmap (dm,pmap,plen) bind(c)
+       import
+       implicit none
+       type(c_ptr) :: dm
```
WeiqunZhang: Add `value` here to match `const DistributionMapping*` in C++.

WeiqunZhang (changes requested):

Src/F_Interfaces/Base/AMReX_distromap_fi.cpp
kweide: Sure! I'll make those changes.

WeiqunZhang (approved):

PR #1312, opened 2020-08-25
HIP: printf works now on device

WeiqunZhang:
## Checklist

The proposed changes:


kngott (approved):
Anonymous namespace... ew.

WeiqunZhang:
`::` means global namespace.  It works without it, but just want to be explicit.

PR #1315, opened 2020-08-26
Extend the number of unique particles per cpu we can have at once.

atmyers:
Currently, we use two signed integers to store id numbers for each particle as well as the rank it was generated on. This allows unique combinations of 'id', and 'cpu' numbers to be generated without any communication between ranks. However, this does waste some space, since it's unlikely that 2**31-1 MPI ranks will be used any time soon, while the same limit for the id has actually been overflowed in real-world WarpX simulations. To address this, in this PR, we still use 64 bits to represent the combination of (id, cpu), but we devote 40 bits to the id and only 24 to the cpu. This allows ~0.5 trillion unique particles on each of 16.7 million MPI ranks, which should be good enough for the foreseeable future.

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_Particle.H
+++ b/Src/Particle/AMReX_Particle.H
@@ -85,7 +85,7 @@ struct ParticleCPUWrapper
         // zero out the first 24 bits, which are used to store the cpu number
         m_idata &= (~ 0x00FFFFFF);
 
-        AMREX_ASSERT(cpu > 0);
+        AMREX_ASSERT(cpu >= 0);
         AMREX_ASSERT(cpu <= 16777215);  // 2**24-1, the max representable number
 
         m_idata |= cpu;
--- a/Src/Particle/AMReX_Particle.H
+++ b/Src/Particle/AMReX_Particle.H
@@ -123,10 +123,10 @@ struct ConstParticleIDWrapper
 
 struct ConstParticleCPUWrapper
 {
-    const int64_t& m_idata;
+    const uint64_t& m_idata;
 
     AMREX_GPU_HOST_DEVICE
-    ConstParticleCPUWrapper (const int64_t& idata) noexcept
+    ConstParticleCPUWrapper (const uint64_t& idata) noexcept
         : m_idata(idata)
     {}
 
--- a/Src/Particle/AMReX_Particle.H
+++ b/Src/Particle/AMReX_Particle.H
@@ -22,6 +22,108 @@ namespace
     constexpr int NoSplitParticleID  = std::numeric_limits<int>::max()-4;
 }
 
+struct ParticleIDWrapper
+{
+    uint64_t& m_idata;
+
+    ParticleIDWrapper (uint64_t& idata)
+        : m_idata(idata)
+    {}
+
+    ParticleIDWrapper& operator= (const Long id)
+    {
+        // zero out the 40 leftmost bits, which store the sign and the abs of the id;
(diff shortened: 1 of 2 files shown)


MaxThevenet:
On the test I am running, running on this branch (the standard output contains `AMReX (20.08-97-ge99860c8e34c) initialized`), I still get the same error at the same time step
```
STEP 13749 starts ...
amrex::Abort::1::ERROR: overflow on particle id numbers !!!
SIGABRT
```
See the [Backtrace](https://github.com/AMReX-Codes/amrex/files/5134627/Backtrace.txt).

Currently this test crashes after 30 min on 2 V100 GPUs. I can make a reproducer that crashes faster.

MaxThevenet:
@atmyers [This input file](https://github.com/AMReX-Codes/amrex/files/5134884/inputs.txt) is a single-GPU WarpX reproducer where the issue comes after 1802 iterations (2 minutes on 1 V100). Here the number of particles injected should be roughly `256 * 256 * 32 * 1802 = 3.7 billions`.

If you want to make it even faster, you can increase the number of ppc for `plasma_e`. If you encounter memory issues, you can decrease the number of cells longitudinally AND decrease the physical size of the domain in the longitudinal direction accordingly (so dz remains small, so you still inject ~1 cell per time step).

atmyers:
Yes, changes are need on the WarpX side as well, which still uses `int` for the `pid`. This just removes the restriction from AMReX.

atmyers:
I believe this PR to WarpX should do it: https://github.com/ECP-WarpX/WarpX/pull/1266

Src/Particle/AMReX_Particle.H
```diff
@@ -15,13 +15,125 @@ namespace amrex {
...
+        : m_idata(idata)
+    {}
+
+    AMREX_GPU_HOST_DEVICE
+    ParticleCPUWrapper& operator= (const int cpu) noexcept
+    {
+        // zero out the first 24 bits, which are used to store the cpu number
+        m_idata &= (~ 0x00FFFFFF);
+
+        AMREX_ASSERT(cpu > 0);
```
WeiqunZhang: `>= 0`

MaxThevenet:
Oh yes, of course, I just resubmitted the test on both branches. Thanks!

Src/Particle/AMReX_Particle.H
```diff
@@ -15,13 +15,125 @@ namespace amrex {
...
+            val = -id;
+        }
+
+        m_idata |= (sign << 63);  // put the sign in the leftmost bit
+        m_idata |= (val << 24);   // put the val in the next 39
+        return *this;
+    }
+
+    AMREX_GPU_HOST_DEVICE
+    operator long () noexcept
```
WeiqunZhang: This could be const function.

Src/Particle/AMReX_Particle.H
```diff
@@ -15,13 +15,125 @@ namespace amrex {
...
+
+        AMREX_ASSERT(cpu > 0);
+        AMREX_ASSERT(cpu <= 16777215);  // 2**24-1, the max representable number
+
+        m_idata |= cpu;
+        return *this;
+    }
+
+    AMREX_GPU_HOST_DEVICE
+    operator int () noexcept
```
WeiqunZhang, line 97: Could be a const function.

MaxThevenet:
Yes, this fixes the issue for me. Thanks!

WeiqunZhang (approved):

PR #1325, opened 2020-09-01
Add parameter that stores OpenMP support version

kweide:
## Summary

Add a parameter to `AMReX_omp_mod` that can be used by Fortran code to check whether AMReX was built with OpenMP support, and, if yes, what the version number (_OPENMP) was at the time AMReX was compiled.

## Additional background
The name of the parameter is `amrex_omp_support`, the type is (default kind) integer.

## Checklist

The proposed changes:
- [x] add new **(mini)**-capabilities to AMReX

### DIFF
--- a/Src/Base/AMReX_omp_mod.F90
+++ b/Src/Base/AMReX_omp_mod.F90
@@ -4,6 +4,12 @@ module amrex_omp_module
 
   implicit none
 
+#ifdef _OPENMP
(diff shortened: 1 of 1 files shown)


kweide:
Would a function be preferable for some reason? A parameter seems sufficient.

WeiqunZhang:
Sound good.

WeiqunZhang (approved):

PR #1329, opened 2020-09-02
Fix warning with default copy ctor + AMREX_USE_GPU

atmyers:
Otherwise we get `warning #3057-D: __device__ annotation is ignored on a function("ParticleCPUWrapper") that is explicitly defaulted on its first declaration`, etc...

### DIFF
--- a/Src/Particle/AMReX_Particle.H
+++ b/Src/Particle/AMReX_Particle.H
@@ -88,7 +88,6 @@ struct ParticleCPUWrapper
         : m_idata(idata)
     {}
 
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
```
AMREX_GPU_HOST_DEVICE
ParticleCPUWrapper (const ParticleCPUWrapper& rhs) = default;
```
needs update too.

atmyers:
Going to do this a different way: https://github.com/AMReX-Codes/amrex/pull/1342

PR #1331, opened 2020-09-02
Add support for extended particle range to the Fortran interfaces

atmyers:
The proposed changes:


Src/F_Interfaces/Particle/AMReX_particlecontainer_mod.F90
```diff
@@ -56,15 +58,43 @@ end subroutine amrex_fi_delete_particlecontainer
      subroutine amrex_fi_get_next_particle_id (id) bind(c)
        import
        implicit none
-       integer(c_int) id
+       integer(c_long) id
```
WeiqunZhang: We should use `amrex_long` from the `amrex_fort_module`.

WeiqunZhang (approved):

PR #1332, opened 2020-09-03
More on Gpu kernel fusing

WeiqunZhang:
## Summary

* Add Gpu::KernelInfo argument to ParallelFor to allow the user to indicate
  whether the kernel is an candidate for fusing.

* For MFIter, if the local size is less or equal to 3, the fuse region is
  turned on and small kernels marked fusable will be fused.

* Add launch macros for fusing.

* Add fusing to a number of functions used by linear solvers.  Note that
  there are a lot more amrex functions need to be updated for fusing.

* Optimize reduction for bottom solve.

* Consolidate memcpy in communication functions.

* Option to use device memory in communication kernels for packing and
  unpacking buffers.  But it's currently turned off because the performance
  was not improved in testing.  In fact, it was worse than using pinned
  memory.  But this might change in the future.  So the option is kept.

## Checklist

The proposed changes:


WeiqunZhang:
This is still a draft.  I submitted this to run CI and I like the diff provided by github.

I have done scaling test using `amrex/Tutorials/LinearSolvers/ABecLaplacian_C` on summit with 256^3 cells per node recently.  The results are shown below.

```
| Branch      | # of node |   Time |
|-------------+-----------+--------|
| development |         1 | 0.2557 |
| fusing      |         1 | 0.2204 |
|-------------+-----------+--------|
| development |         8 | 0.3232 |
| fusing      |         8 | 0.2820 |
|-------------+-----------+--------|
| development |        64 | 0.3725 |
| fusing      |        64 | 0.3284 |
|-------------+-----------+--------|
| development |       512 | 0.4703 |
| fusing      |       512 | 0.4063 |
|-------------+-----------+--------|
| development |      4096 | 0.6459 |
| fusing      |      4096 | 0.5769 |
```

WeiqunZhang:
This PR is ready for review now.  I have done regression tests on AMReX and various application codes on both GPU and CPU.  It has passed all tests except for a few CPU tests using cell-centered linear solvers.  There are roundoff errors in these tests due to a change in `MultiFab::Dot`.  See  https://ccse.lbl.gov/pub/RegressionTesting/IAMR/2020-09-07-001/index.html for example.  In a different branch, `fusing_regtest`, I reverted the CPU path in that function to the current way in `development`.  The "failed" tests passes with the `fusing_regtest`.  The difference of the PR branch `fusing` and `fusing_regtest` can be seen here, https://github.com/WeiqunZhang/amrex/compare/fusing...WeiqunZhang:fusing_regtest.

Src/Base/AMReX_GpuLaunchFunctsG.H
```diff
@@ -94,7 +94,7 @@ void launch (T const& n, L&& f) noexcept
 }
 
 template <typename T, typename L, typename M=amrex::EnableIf_t<std::is_integral<T>::value> >
-void ParallelFor (T n, L&& f) noexcept
+void ParallelFor (Gpu::KernelInfo const& info, T n, L&& f) noexcept
 {
```
kngott, line 98: Preemptively comment out "info" in these?

Src/Base/AMReX_GpuLaunch.H
```diff
@@ -215,6 +231,22 @@ namespace Gpu {
...
 #define AMREX_HOST_DEVICE_PARALLEL_FOR_4D(...) AMREX_GPU_HOST_DEVICE_PARALLEL_FOR_4D(__VA_ARGS__)
 
+#ifdef AMREX_USE_CUDA
+#define AMREX_HOST_DEVICE_PARALLEL_FOR_1D_FUSABLE(...) AMREX_GPU_HOST_DEVICE_FOR_1D_FUSABLE(__VA_ARGS__)
+#define AMREX_HOST_DEVICE_PARALLEL_FOR_3D_FUSABLE(...) AMREX_GPU_HOST_DEVICE_FOR_3D_FUSABLE(__VA_ARGS__)
+#define AMREX_HOST_DEVICE_PARALLEL_FOR_4D_FUSABLE(...) AMREX_GPU_HOST_DEVICE_FOR_4D_FUSABLE(__VA_ARGS__)
+#define AMREX_HOST_DEVICE_FOR_1D_FUSABLE(...) AMREX_GPU_HOST_DEVICE_FOR_1D_FUSABLE(__VA_ARGS__)
+#define AMREX_HOST_DEVICE_FOR_3D_FUSABLE(...) AMREX_GPU_HOST_DEVICE_FOR_3D_FUSABLE(__VA_ARGS__)
+#define AMREX_HOST_DEVICE_FOR_4D_FUSABLE(...) AMREX_GPU_HOST_DEVICE_FOR_4D_FUSABLE(__VA_ARGS__)
+#else
```
kngott, line 241: Does this mean the "FUSABLE" macros exist to work around HIP and DPCPP and they will go away? (I'm a little concerned about the amount of complexity that's building in this.)

Src/Base/AMReX_GpuLaunch.H
```diff
@@ -215,6 +231,22 @@ namespace Gpu {
 #define AMREX_HOST_DEVICE_PARALLEL_FOR_3D(...) AMREX_GPU_HOST_DEVICE_PARALLEL_FOR_3D(__VA_ARGS__)
 #define AMREX_HOST_DEVICE_PARALLEL_FOR_4D(...) AMREX_GPU_HOST_DEVICE_PARALLEL_FOR_4D(__VA_ARGS__)
 
+#ifdef AMREX_USE_CUDA
+#define AMREX_HOST_DEVICE_PARALLEL_FOR_1D_FUSABLE(...) AMREX_GPU_HOST_DEVICE_FOR_1D_FUSABLE(__VA_ARGS__)
```
kngott: I think this spelling is a bit confusing. This brings to mind the pronunciation "fussy" instead of "fuse".

 I'd lean towards "FUSEABLE" to be very clear, or, apparently the proper English seems to be "FUSIBLE": https://wikidiff.com/fusable/fusible.

Src/Base/AMReX_GpuFuse.cpp
```diff
@@ -86,12 +89,32 @@ void Fuser::Launch ()
         constexpr int nwarps_per_block = nthreads/Gpu::Device::warp_size;
         int nblocks = (ntotwarps + nwarps_per_block-1) / nwarps_per_block;
 
+        bool is_reduction = s_in_fuse_reduction_region;
+
```
kngott, line 93: Again, a little worried about complexity of the implementation: is FuseReductionSafeGuard meant to be hidden from the users, or phased out later?

Src/Base/AMReX_FBI.H
```diff
@@ -54,19 +54,26 @@ ParallelFor (Vector<Array4BoxTag<T> > const& tags, int ncomp, F && f)
...
 
-    constexpr int nthreads = 128;
+    std::memcpy(h_buffer, tags.data(), sizeof_tags);
+    std::memcpy(h_buffer+offset_nwarps, nwarps.data(), sizeof_nwarps);
+    Gpu::htod_memcpy_async(d_buffer, h_buffer, total_buf_size);
+
+    auto d_tags = reinterpret_cast<TagType*>(d_buffer);
+    auto d_nwarps = reinterpret_cast<int*>(d_buffer+offset_nwarps);
+
+    constexpr int nthreads = 256;
```
kngott, line 72: Was the increase in nthreads tested? Or just changing to be consistent with other launches?

Should this be based on MAX_NUM_THREADS or another more universal variable, so it can be easily adjusted for each architecture type without being forgotten?

Src/Base/AMReX_GpuLaunchFunctsG.H
WeiqunZhang: Sure

Src/Base/AMReX_GpuLaunch.H
WeiqunZhang: It's just a hint.  Whether or not HIP and DPCPP support it, we can keep them.

Src/Base/AMReX_GpuLaunch.H
WeiqunZhang: OK.  Then let's use the proper English work.

Src/Base/AMReX_GpuFuse.cpp
WeiqunZhang: This is not meant for the users.  Unfortunately, it is complicated for reduction.

Src/Base/AMReX_FBI.H
WeiqunZhang: Yes, it was tested.  But I didn't really see much difference.

kngott (approved):

PR #1333, opened 2020-09-03
CMake: re-organize tutorials

mic84:
## Checklist

The proposed changes:


.github/workflows/windows.yml
```diff
@@ -14,4 +14,4 @@ jobs:
         mkdir build
         cd build
         cmake .. -DCMAKE_BUILD_TYPE=Debug -DCMAKE_VERBOSE_MAKEFILE=ON -DENABLE_TUTORIALS=ON -DENABLE_FORTRAN=OFF -DENABLE_MPI=OFF
-        cmake --build . --config Debug --target tutorials
+        cmake --build . --config Debug --target all
```
ax3l: Try this:
```suggestion
        cmake --build . --config Debug
```

Tools/CMake/AMReX_Options.cmake
```diff
@@ -144,7 +144,8 @@ endif ()
 #
 # AMReX components selection  ================================================
 #
-option( ENABLE_EB "Build EB Code" OFF )
+cmake_dependent_option( ENABLE_EB "Build EB code" OFF
```
ax3l: We could improve the doc string while changing this:
```suggestion
cmake_dependent_option( ENABLE_EB "Build with Embedded Boundary support" OFF
```

Tutorials/Amr/Advection_AmrLevel/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-if ( (DIM EQUAL 1) OR NOT ENABLE_FORTRAN )
+if ( (DIM EQUAL 1) OR NOT CMAKE_Fortran_COMPILER_LOADE )
```
ax3l: this looks like a typo
```suggestion
if ( (DIM EQUAL 1) OR NOT CMAKE_Fortran_COMPILER_LOADED )
```

Tutorials/Basic/HeatEquation_EX1_CF/CMakeLists.txt
```diff
@@ -1,68 +1,47 @@
...
 
-if (ENABLE_CUDA)
-   set_cpp_sources_to_cuda_language( ${EXENAME} )
-endif ()
 
-#
-# Copy input files to corresponding build dir
-# First copy operation is run when builddir is setup
-set (INPUT_FILES ${EXE_DIR}/inputs_${DIM}d )
+# if ( (NOT ENABLE_CUDA) AND (NOT ENABLE_DPCPP) )
```
ax3l, line 24: Do you want to keep this as an example and maybe add a small comment above?

Tutorials/CMakeLists.txt
```diff
@@ -1,46 +1,119 @@
...
+endif ()
+
+list(TRANSFORM AMREX_TUTORIALS_SUBDIRS PREPEND "${CMAKE_CURRENT_LIST_DIR}/")
 
 #
-# Here we define a macro to accumulate all the tutorial targets
+# Macro to setup the tutorials
 #
-set ( TUTORIAL_TARGETS )
+macro (setup_tutorial _sources  _input_files)
```
ax3l: You could potentially save some `set/unset` if you make this a CMake function

Tutorials/Particles/ElectromagneticPIC/CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
+#
+# Contains 3 tutorials
+#
+#  * CUDA
+#  * OpenMP
+#  * OpenACC (NOT supported anymore)
+#
+if ( (NOT ENABLE_CUDA AND NOT ENABLE_OMP) OR (NOT DIM EQUAL 3) )
+   return()
```
ax3l, line 9: I wonder if we should write a little message such as
```suggestion
   message(STATUS "Tutorials - ElectromagneticPIC: Skipped (requires 3D with CUDA or OMP)")
   return()
```
instead of silently returning.

ax3l (commented):
This looks really great, thank you for modernizing and fixing all these tutorials. I like it a lot. :rocket: :sparkles: 

I left a few inline Q's and ideas inline as suggestions.

Tutorials/CMakeLists.txt
mic84: Good point! Done!

Tutorials/Particles/ElectromagneticPIC/CMakeLists.txt
mic84: I was thinking the same. The problem is that most of the Tutorials have some conditionality attached to them so I will need to modify most of the relevant CMakeLists.txt and include a status message that depends on the name of the Tutorial itself. If the Tutorial name/path changes, the status message again will need to be modified accordingly. I am not sure this is worth the effort. What about just adding a disclaimer in the Docs?

ax3l: Docs is a good idea as well, agreed.

ax3l (approved):
Thank you for modernizing this! :rocket: :sparkles:

ax3l (approved):
Thanks for the docs! :+1:

PR #1351, opened 2020-09-07
Async IO: Check Runtime Threading

ax3l:
##  Summary

Even if `AMREX_MPI_THREAD_MULTIPLE` is defined at compile time, applications might provide their own MPI init routines and/or MPI implementations, runtime arguments or environment arguments can influence the actually provided MPI thread safety level at runtime.

Therefore, we need to check the MPI thread safety level again at runtime where we require a certain minimum.

## Additional background

WarpX PR: https://github.com/ECP-WarpX/WarpX/pull/1298

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_AsyncOut.cpp
+++ b/Src/Base/AMReX_AsyncOut.cpp
@@ -41,12 +41,20 @@ void Initialize ()
     if (s_asyncout and s_noutfiles < nprocs)
     {
 #ifdef AMREX_MPI_THREAD_MULTIPLE
+        int provided = -1;
+        MPI_Query_thread(&provided);
+        if (provided < MPI_THREAD_MULTIPLE)
+            amrex::Abort("AsyncOut with " + std::to_string(s_noutfiles) + " and "
+                         +std::to_string(nprocs) + " processes requires "
+                         +"MPI_THREAD_MULTIPLE at runtime");
+
         int myproc = ParallelDescriptor::MyProc();
         s_info = GetWriteInfo(myproc);
         MPI_Comm_split(ParallelDescriptor::Communicator(), s_info.ifile, myproc, &s_comm);
 #else
         amrex::Abort("AsyncOut with " + std::to_string(s_noutfiles) + " and "
-                     +std::to_string(nprocs) + " processes requires MPI_THREAD_MULTIPLE");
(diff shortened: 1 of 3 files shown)


Src/Base/AMReX_AsyncOut.cpp
```diff
@@ -41,12 +41,20 @@ void Initialize ()
...
+        MPI_Query_thread(&provided);
+        if (provided < MPI_THREAD_MULTIPLE)
+            amrex::Abort("AsyncOut with " + std::to_string(s_noutfiles) + " and "
+                         +std::to_string(nprocs) + " processes requires "
+                         +"MPI_THREAD_MULTIPLE at runtime");
+
         int myproc = ParallelDescriptor::MyProc();
         s_info = GetWriteInfo(myproc);
         MPI_Comm_split(ParallelDescriptor::Communicator(), s_info.ifile, myproc, &s_comm);
 #else
```
ax3l: Looking at it now, I would propose we can also remove the compile-time part here at all, if you like.

WeiqunZhang: Yes.

Src/Base/AMReX_AsyncOut.cpp
```diff
@@ -38,17 +38,21 @@ void Initialize ()
...
     if (s_asyncout and s_noutfiles < nprocs)
     {
-#ifdef AMREX_MPI_THREAD_MULTIPLE
+        int provided = -1;
+        MPI_Query_thread(&provided);
+        if (provided < MPI_THREAD_MULTIPLE)
+            amrex::Abort("AsyncOut with " + std::to_string(s_noutfiles) + " and "
+                         +std::to_string(nprocs) + " processes requires "
+                         +"MPI_THREAD_MULTIPLE at runtime");
+
```
kngott, line 51: I would add the provided thread safety level to the output message.

For example, Cori has an environment variable that must be set to allow THREAD_MULTIPLE support. Seeing if the support is 1 or 3 (SINGLE or SERIALIZED) could help diagnose that quickly.

ax3l: yes, will for that have to move your lambda to a more general location. I just implemented the same lookup in WarpX xD

WeiqunZhang (approved):

PR #1356, opened 2020-09-08
Also use atomics in the particle_dx != dx case

atmyers:
The proposed changes:


asalmgren (approved):

asalmgren:
This changes fixes the Nyx regression test that was previous failing.

PR #1363, opened 2020-09-12
Random Number Generation on Device

WeiqunZhang:
## Summary

In order to support RNG on device with DPC++, we have to change the API
because DPC++ does not support global device variables.  We introduce a new
`ParallelForRNG` function that takes a device lambda with an extra
`RandomEngine` argument.  Random numbers can be generated by calling
`amrex::Random(RandomEngine const&)`.  To maintain backward compatibility in
the short term, the old way of calling `amrex::Random()` on device is kept.
But it will be removed soon because no locking is needed in the new way
unlike in `Random()`.  Note that only the device version of
`amrex::Random()` is deprecated.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_Random.H
+++ b/Src/Base/AMReX_Random.H
@@ -95,6 +95,7 @@ namespace amrex
 #elif defined(__HIP_DEVICE_COMPILE__)
         return hiprand_poisson(random_engine.rand_state, lambda);
 #elif defined (__SYCL_DEVICE_ONLY__)
+        amrex::Abort("RandomPossion not supported in DPC++ device code");
         return 0; // xxxxx DPCPP todo: Poisson distribution
 #else
         amrex::ignore_unused(random_engine);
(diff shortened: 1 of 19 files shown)


WeiqunZhang:
Looks like in the CI the dpc++ compiler is beta8 but mkl is beta9.  The result is the dpcpp compiler cannot find mkl headers.

WeiqunZhang:
I have temporarily removed dpc++ from the list of required checks.  I did test the PR with both dpcpp beta8 and beta9 on DevCloud.

Src/Base/AMReX_Random.H
```diff
@@ -37,14 +87,20 @@ namespace amrex
...
-    AMREX_GPU_DEVICE void free_state (int tid);
+    AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+    unsigned int RandomPoisson (Real lambda, RandomEngine const& random_engine)
+    {
+#if defined(__CUDA_ARCH__)
+        return curand_poisson(random_engine.rand_state, lambda);
+#elif defined(__HIP_DEVICE_COMPILE__)
+        return hiprand_poisson(random_engine.rand_state, lambda);
+#elif defined (__SYCL_DEVICE_ONLY__)
+        return 0; // xxxxx DPCPP todo: Poisson distribution
```
atmyers, line 99: maybe these should explicitly fail instead of returning 0

atmyers (approved):

PR #1364, opened 2020-09-13
Revert "DPCPP Beta09: Package Broken (#1361)"

WeiqunZhang:
## Summary

This reverts commit 198a4958cc2cbfd5a7e1d639d5ff750489104df6.

The workaround does not work with #1363 because of mkl.  We have now
temporarily removed dpc++ from the list of required checks.  So we can
revert this change now.  Once the packages are fixed, we will make the check
required again.

## Checklist

The proposed changes:

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -155,8 +155,6 @@ jobs:
             -DCUDA_ARCH=6.0
         make -j 2 tutorials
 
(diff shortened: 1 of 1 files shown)


ax3l (approved):

ax3l:
Sounds good, they are usually broken now once a month because intel manages to render their oneAPI apt repos into inconsistent state during releases (~24-48hrs).

PR #1367, opened 2020-09-15
Remove direct access of particle m_rdata

atmyers:
The proposed changes:


WeiqunZhang:
Looks like the change here is already in #1368.

PR #1383, opened 2020-09-20
Fix tensor solver with periodic boundaries

WeiqunZhang:
## Summary

Remove erroneous calls to EnforcePeriodicity.  When coarse/fine boundary
meets periodic boundary with fine grids form an L-shape, the uncovered fine
ghost cell at the corner is actually "multi-valued". That is the cells with
the same indices has different values in different boxes.  That's why
calling EnforcePeriodicity is erroneous.  For CPU runs, this bug introduces
deterministic errors, but the solver still converges.  For GPU runs, this
causes convergence issues too because EnforcePeriodicity is not
deterministic when the values inside domain are not consistent.

## Checklist

The proposed changes:

### DIFF
--- a/Src/LinearSolvers/MLMG/AMReX_MLEBTensorOp.cpp
+++ b/Src/LinearSolvers/MLMG/AMReX_MLEBTensorOp.cpp
@@ -402,14 +402,12 @@ MLEBTensorOp::applyBCTensor (int amrlev, int mglev, MultiFab& vel,
         }
     }
 
-    vel.EnforcePeriodicity(0, AMREX_SPACEDIM, IntVect(1),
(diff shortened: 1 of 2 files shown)


asalmgren (commented):
I wonder if we should put a comment where the call was (but shouldn't be) so we don't decide at some later time that we should be enforcing periodicity there?

Otherwise -- since it obviously works the proof is in the pudding, so to speak

WeiqunZhang:
I have updated the PR to add a comment.

drummerdoc (approved):
Ahh, the ole double-valued grow cell at an L-shaped grid issue!

PR #1391, opened 2020-09-22
Intall intel-oneapi-mkl-devel in CI

WeiqunZhang:
## Summary

The package is needed for mkl random number generator header.

## Checklist

The proposed changes:


ax3l (approved):
:rocket:

PR #1399, opened 2020-09-23
Update dpcpp flags

WeiqunZhang:
## Summary

* -fsycl-unnamed-lambda is enabled by default now.  So we do not need to
   pass the flag anymore.

* Early optimizations are enabled by default in beta09.  This makes a lot of
  tests crash.  So we disable it with -fno-sycl-early-optimizations.

## Checklist

The proposed changes:


mic84 (approved):

ax3l:
@WeiqunZhang thank you! :+1:

PR #1402, opened 2020-09-23
CMake: enforce C++17 as minimum C++ standard for ENABLE_DPCPP=ON

mic84:
## Summary
Set minimum C++ standard to 17 when dpc++ is used. Reason:  MKL is using C++17.

Meta-reason: SYCL 2020 is C++17 or newer only, too.

## Checklist

The proposed changes:

### DIFF
--- a/Tools/CMake/AMReX_Config.cmake
+++ b/Tools/CMake/AMReX_Config.cmake
@@ -48,8 +48,12 @@ function (configure_amrex)
    # Moreover, it will also enforce such standard on all the consuming targets
    #
    set_target_properties(amrex PROPERTIES CXX_EXTENSIONS OFF)
-   # minimum: C++11 on Linux, C++17 on Windows
-   target_compile_features(amrex PUBLIC $<IF:$<STREQUAL:$<PLATFORM_ID>,Windows>,cxx_std_17,cxx_std_11>)
+   # minimum: C++11 on Linux, C++17 on Windows, C++17 for dpc++
+   if (ENABLE_DPCPP)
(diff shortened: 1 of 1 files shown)


Tools/CMake/AMReX_Config.cmake
```diff
@@ -48,8 +48,12 @@ function (configure_amrex)
    # Moreover, it will also enforce such standard on all the consuming targets
    #
    set_target_properties(amrex PROPERTIES CXX_EXTENSIONS OFF)
-   # minimum: C++11 on Linux, C++17 on Windows
-   target_compile_features(amrex PUBLIC $<IF:$<STREQUAL:$<PLATFORM_ID>,Windows>,cxx_std_17,cxx_std_11>)
+   # minimum: C++11 on Linux, C++17 on Windows, C++17 for dpc++
+   if (ENABLE_DPCPP)
+      target_compile_features(amrex PUBLIC cxx_std_17)
+   else ()
+      target_compile_features(amrex PUBLIC $<IF:$<STREQUAL:$<PLATFORM_ID>,Windows>,cxx_std_17,cxx_std_11>)
```
ax3l, line 55: Just as an idea, since we don't really need a generator expression here:
```suggestion
   if (ENABLE_DPCPP)
      target_compile_features(amrex PUBLIC cxx_std_17)
   elseif (WIN32)
      target_compile_features(amrex PUBLIC cxx_std_17)
   else ()
      target_compile_features(amrex PUBLIC cxx_std_11)
```

ax3l (approved):

PR #1404, opened 2020-09-23
DPCPP: use C++17 and disable mkl warnings

WeiqunZhang:
* Explicitly use C++17 because MKL contains C++17 codes.

* Disable warnings caused by MKL RNG header.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_RandomEngine.H
+++ b/Src/Base/AMReX_RandomEngine.H
@@ -12,7 +12,7 @@
 #elif defined(AMREX_USE_DPCPP)
 #include <CL/sycl.hpp>
 namespace sycl = cl::sycl;
-// Disable warnings in mkl's rng header.  Too many noises.
+// Disable warnings in mkl's rng header.  Too much noise.
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wmissing-braces"
 #pragma clang diagnostic ignored "-Wunused-variable"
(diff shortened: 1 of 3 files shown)


Src/Base/AMReX_RandomEngine.H
```diff
@@ -12,7 +12,13 @@
 #elif defined(AMREX_USE_DPCPP)
 #include <CL/sycl.hpp>
 namespace sycl = cl::sycl;
+// Disable warnings in mkl's rng header.  Too many noises.
```
ax3l:
```suggestion
// Disable warnings in mkl's rng header.  Too much noise.
```

Tools/GNUMake/comps/dpcpp.mak
```diff
@@ -57,11 +57,7 @@ endif
 
 ########################################################################
 
-ifdef CXXSTD
-  CXXFLAGS += -std=$(strip $(CXXSTD))
-endif
-
-CXXFLAGS += -Wno-error=sycl-strict -fsycl
+CXXFLAGS += -std=c++17 -Wno-error=sycl-strict -fsycl
```
ax3l: We could also keep C++20 (and C++2b) open:
```suggestion
ifdef CXXSTD
  CXXFLAGS += -std=$(strip $(CXXSTD))
else
  CXXFLAGS += -std=c++17
endif

CXXFLAGS += -Wno-error=sycl-strict -fsycl
```

And we just add an `if ($(USE_DPCPP,TRUE))` in WarpX around its setting for `CXXSTD`.

WeiqunZhang: Sure, we could do that.

ax3l (approved):

PR #1421, opened 2020-09-29
Doc: update documentation on CMake+CUDA

mic84:
## Checklist

The proposed changes:

### DIFF
--- a/Docs/sphinx_documentation/source/GPU.rst
+++ b/Docs/sphinx_documentation/source/GPU.rst
@@ -235,8 +235,8 @@ check the :ref:`table <tab:cmakecudavar>` below.
 
 The target architecture to build for can be specified via the configuration option
 ``-DCUDA_ARCH=<target-architecture>``, where ``<target-architecture>`` can be either
-the name of the NVIDIA GPU, i.e. ``Turing``, ``Volta``, ``...`` , or its
-version number, i.e. ``10.0``, ``9.0``,  ``...`` .
+the name of the NVIDIA GPU generation, i.e. ``Turing``, ``Volta``, ``Ampere``, ``...`` , or its
+`compute capability <https://developer.nvidia.com/cuda-gpus>`_, i.e. ``10.0``, ``9.0``,  ``...`` .
 For example, on Cori GPUs you can specify the architecture as follows:
 
 .. highlight:: console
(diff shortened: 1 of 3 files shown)


Docs/sphinx_documentation/source/GPU.rst
```diff
@@ -182,15 +182,61 @@ can run it and that will generate results like:
...
+.. raw:: latex
+
+   \end{center}
+
+The target architecture to build for can be specified via the configuration option
 ``-DCUDA_ARCH=<target-architecture>``, where ``<target-architecture>`` can be either
-the name of the NVIDIA GPU, i.e. ``Turing``, ``Volta``, ``Pascal``, ``...`` , or its
-version number, i.e. ``10.0``, ``9.0``, ``8.0``, ``...`` .
+the name of the NVIDIA GPU, i.e. ``Turing``, ``Volta``, ``...`` , or its
+version number, i.e. ``10.0``, ``9.0``,  ``...`` .
```
```suggestion
the name of the NVIDIA GPU generation, i.e. ``Turing``, ``Volta``, ``Ampere``, ``...`` , or its
`compute capability <https://developer.nvidia.com/cuda-gpus>`_, i.e. ``10.0``, ``9.0``,  ``...`` .
```

Docs/sphinx_documentation/source/GPU.rst
```diff
@@ -200,28 +246,60 @@ For example, on Cori GPUs you can specify the architecture as follows:
    cmake [options] -DENABLE_CUDA=yes -DCUDA_ARCH=Volta /path/to/amrex/source
 
 
-Note that AMReX only supports GPU architectures with version number ``6.0`` or higher.
+If no architecture is specified, CMake will try to determine which GPU architecture is
+supported by the system.
+If more than one is found, CMake will build for all of them.
+This will generally results in a larger library and longer build times.
+If autodetection fails, a set of "common" architectures is assumed.
+**Note that AMReX only supports GPU architectures with version number ``6.0`` or higher**.
```
ax3l: I think we also just support CUDA 9.0 and newer. Let's add this as well.

ax3l (commented):
Thanks for the documentation update :sparkles: 
I added a few notes inline :)

Docs/sphinx_documentation/source/GPU.rst
mic84: Actually, it is still the case.

ax3l: Potentially better wording here as well (right terms):
```suggestion
**Note that AMReX supports NVIDIA GPU architectures with compute capability ``6.0`` or higher**.
```

> Actually, it is still the case.

Sorry, cannot follow. You mean "yes, we support CUDA 9.0+ and we should document that?" :)

maximumcats: NVIDIA

ax3l (approved):

PR #1434, opened 2020-10-02
Fix bug in EB extdir slopes

drangara:
## Checklist

The proposed changes:


drummerdoc (approved):
Looks like a bug alright.  Thanks!

PR #1436, opened 2020-10-03
CMake: Fix missing compile time definitions for HYPRE/PETSc

sayerhs:
When building application codes with an _external_ AMReX install, the `AMReX_Config.H` file that is used does not include the required `AMREX_USE_HYPRE` and `AMREX_USE_PETSC` compile time definitions enabled, even if AMReX was compiled with these options. This PR adds the necessary entries in `CMake/AMReX_Config.H.in` to add these if these were enabled during the CMake build.


mic84:
Thank you very much for catching this!

mic84 (approved):

PR #1441, opened 2020-10-07
Allow tagging value to vary by level

maximumcats:
## Summary

The new error tagging scheme from #1166 is modified to allow the threshold value to vary by level. Using the example there, if we did:

```
amr.refinement_indicators = flame_tracer lo_temp
amr.refine.flame_tracer.max_level = 3
amr.refine.flame_tracer.value_greater = 1.e-6 1.e-5
amr.refine.flame_tracer.field_name = Y(H)

amr.refine.lo_temp.max_level = 2
amr.refine.lo_temp.value_less = 1000.
amr.refine.lo_temp.field_name = temp
```

Then level 0 would be tagged for refinement for Y(H) >= 1.e-6, while level 1 and all higher levels would be tagged for refinement for Y(H) >= 1.e-5. (If a value for every level is not provided, we assume the last value holds for all remaining levels, similar to how we treat quantities like amr.n_error_buf.)

## Additional background

This will help for cases where a field varies substantially on the domain and you want to have better control over refinement so that you don't have too many zones.

## Checklist

The proposed changes:

### DIFF
--- a/Src/AmrCore/AMReX_ErrorList.H
+++ b/Src/AmrCore/AMReX_ErrorList.H
@@ -378,7 +378,7 @@ std::ostream& operator << (std::ostream& os, const ErrorList& elst);
 
   struct AMRErrorTagInfo
   {
-    int m_max_level = 100000;
+    int m_max_level = 1000;
     Real m_min_time = std::numeric_limits<Real>::lowest();
     Real m_max_time = std::numeric_limits<Real>::max();
     RealBox m_realbox;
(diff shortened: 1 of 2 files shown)


Src/AmrCore/AMReX_ErrorList.H
```diff
@@ -378,7 +378,7 @@ std::ostream& operator << (std::ostream& os, const ErrorList& elst);
 
   struct AMRErrorTagInfo
   {
-    int m_max_level = 100000;
+    int m_max_level = 1000;
```
maximumcats, line 381: Note: I changed this just because we're now storing a vector of length m_max_level and I wanted to avoid it being gigantic, and no one will realistically have more than 1000 levels.

drummerdoc (approved):
Awesome!  I like it.

PR #1444, opened 2020-10-08
NodalProjector: use volume-weighted average down functions

mic84:
## Summary
Makes NodalProjector class work in R-Z coordinates

## Checklist

The proposed changes:


cgilet (approved):
Works for IAMR; it seems to do okay with the bubble problem.

cgilet (approved):

asalmgren (approved):

PR #1446, opened 2020-10-09
Allow users to set a Geometry object for ParticleContainer independently of the AmrCore / AmrLevel object.

atmyers:
The proposed changes:

### DIFF
--- a/Src/AmrCore/AMReX_AmrParGDB.H
+++ b/Src/AmrCore/AMReX_AmrParGDB.H
@@ -13,25 +13,30 @@ public:
 
     explicit AmrParGDB (AmrCore* amr) noexcept
 	: m_amrcore(amr),
+          m_geom(amr->maxLevel()+1),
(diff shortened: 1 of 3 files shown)


Src/AmrCore/AMReX_AmrParGDB.H
```diff
@@ -41,20 +46,34 @@ public:
...
     Vector<BoxArray>            m_ba;
 };
 
-inline 
-const Geometry& 
+inline
+const Geometry&
+AmrParGDB::ParticleGeom (int level) const
+{
+    if (m_has_geom[level]) {
```
WeiqunZhang: I think the logic here is wrong.

atmyers: whoops, see update.

WeiqunZhang (approved):

MaxThevenet:
Looks good, thanks for this PR! I'll be happy to use this feature.

PR #1458, opened 2020-10-16
Fix potential out of bounds access in #1441

maximumcats:
## Summary

The user may have given us a vector that has too many values, so we just ignore the remainder.

## Checklist

The proposed changes:

### DIFF
--- a/Src/AmrCore/AMReX_ErrorList.H
+++ b/Src/AmrCore/AMReX_ErrorList.H
@@ -428,7 +428,7 @@ std::ostream& operator << (std::ostream& os, const ErrorList& elst);
       : m_test(test), m_field(field), m_info(info)
       {
           m_value.resize(info.m_max_level);
-          for (int i = 0; i < info.m_max_level; ++i) {
+          for (int i = 0; i < m_value.size(); ++i) {
               m_value[i] = value;
           }
           m_ngrow = SetNGrow();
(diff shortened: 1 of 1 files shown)


Src/AmrCore/AMReX_ErrorList.H
```diff
@@ -442,7 +442,7 @@ std::ostream& operator << (std::ostream& os, const ErrorList& elst);
       {
           AMREX_ASSERT(value.size() > 0);
           m_value.resize(info.m_max_level);
-          for (int i = 0; i < value.size(); ++i) {
+          for (int i = 0; i < m_value.size() && i < value.size(); ++i) {
```
drummerdoc, line 445: An overabundance of caution?  Can't be wrong, but curious what brought that on... -M

drummerdoc (approved):

Src/AmrCore/AMReX_ErrorList.H
maximumcats: In Castro I accidentally gave it a vector with max_level+1 values :-)

jmsexton03: Having the vector of values also gives errors if I'd set max_level=-1, meaning I don't want to use that constraint in that particular run. But max_level=0 works for that purpose as well, since it's strictly less than

maximumcats: Yeah that's right, usually with AMR/tagging related things you just set it to max_level = 0 when you want to turn it off.

PR #1465, opened 2020-10-18
Address warnings in CUDA builds

sayerhs:


Src/Base/AMReX.cpp
```diff
@@ -148,15 +148,13 @@ amrex::write_to_stderr_without_buffering (const char* str)
     }
 }
 
+#if !AMREX_DEVICE_COMPILE
 namespace {
 //  Having both host and device versions to avoid compiler warning
 AMREX_GPU_HOST_DEVICE
```
WeiqunZhang: Seems that we can remove these 2 lines now.  It was added for HIP.  But with the new way, it's not needed anymore.

sayerhs: @WeiqunZhang removed the two lines in 649cf41

WeiqunZhang (approved):

PR #1466, opened 2020-10-18
Use rocrand from HIP APT repository

sayerhs:


WeiqunZhang (approved):

ax3l:
Good to know that package is shipped, @sayerhs. Thx!

PR #1468, opened 2020-10-19
CMake: improve HDF5 support

### LINKED ISSUE #1463: Build Question - Lib folders
Hello there, I'm trying to build AMReX with the following libraries and versions:
```
1) gcc/9.1.1
4) cmake/3.16.7
2) mpich/3.3.1
3) petsc/3.11.3
5) hdf5/1.10.5
```
I'm on tag `20.10`, created a `build` folder and using the following options with cmake:
```
-DCMAKE_INSTALL_PREFIX=/opt/apps/amrex/20.10-mpich-3.3.1-petsc-3.11.3-hdf5-1.10.5
-DDIM=3 
-DENABLE_FORTRAN=yes 
-DENABLE_FORTRAN_INTERFACES=yes 
-DENABLE_MPI=yes 
-DENABLE_HYPRE=yes 
-DENABLE_HDF5=yes 
-DENABLE_PETSC=yes
```

How can I set the paths for those libraries I want to use? I coudn't find in the docs flags to set the path of the libraries ...
None of my libs are instaled in the system path, allways in `/opt/apps`.

mic84:
## Summary

Fix #1463

## Checklist

The proposed changes:


Tools/CMake/AMReXThirdPartyLibraries.cmake
```diff
@@ -3,13 +3,13 @@
 #
 if (ENABLE_HDF5)
     set(HDF5_PREFER_PARALLEL TRUE)
-    find_package(HDF5 1.10.4 REQUIRED COMPONENTS CXX)
+    find_package(HDF5 1.10.4 REQUIRED)
```
ax3l, line 6: I would go with and explicitly request what we need:
```suggestion
    find_package(HDF5 1.10.4 REQUIRED COMPONENTS C)
```

Tools/CMake/AMReXThirdPartyLibraries.cmake
```diff
@@ -3,13 +3,13 @@
...
     if (NOT HDF5_IS_PARALLEL)
         message(FATAL_ERROR "\nHDF5 library does not support parallel I/O")
     endif ()
-    target_include_directories(amrex PUBLIC ${HDF5_CXX_INCLUDE_DIRS})
-    target_compile_definitions(amrex PUBLIC ${HDF5_CXX_DEFINES})
-    target_link_libraries(amrex PUBLIC ${HDF5_CXX_LIBRARIES})
+    target_include_directories(amrex PUBLIC ${HDF5_C_INCLUDE_DIRS})
+    target_compile_definitions(amrex PUBLIC ${HDF5_C_DEFINES})
+    target_link_libraries(amrex PUBLIC ${HDF5_C_LIBRARIES})
```
ax3l: Ok, but this does as well and is what most people use (so it might be better tested in the end):
```suggestion
    target_include_directories(amrex PUBLIC ${HDF5_INCLUDE_DIRS})
    target_compile_definitions(amrex PUBLIC ${HDF5_DEFINITIONS})
    target_link_libraries(amrex PUBLIC ${HDF5_LIBRARIES})
```

Also note the typo in `HDF5_C_DEFINITIONS`/`HDF5_DEFINITIONS`.

Tools/CMake/AMReXThirdPartyLibraries.cmake
mic84: From the [FindHDF5 doc](https://cmake.org/cmake/help/v3.14/module/FindHDF5.html): ```the module will optionally accept the COMPONENTS argument. If no COMPONENTS are specified, then the find module will default to finding only the HDF5 C library. ```

Tools/CMake/AMReXThirdPartyLibraries.cmake
mic84: There is no indication in the doc that one form is preferred over the other, therefore I am inclined to keep the current form with the the language specified in the variable name.

Tools/CMake/AMReXThirdPartyLibraries.cmake
ax3l: Sure, unless this behavior changes in the future and/or a superbuild has an unexpected environment set.

Tools/CMake/AMReXThirdPartyLibraries.cmake
ax3l: In that case I recommend to request `C` explicitly, because you use a component on which you rely with default behavior.

The review I provide here stems from experience to make this just more robust, e.g. if internal dependencies or build variants in a found package change in newer or development versions; for instance, they might introduce publicly undeclared internal side-dependencies that  this variant would nicely cover up without us interacting.

ax3l (approved):
:+1: Thank you! :sparkles:

PR #1474, opened 2020-10-22
Remove compilation Warnings

rporcu:
## Summary
Class `PolynomialIF` has a member called `m_sp` which is of type `std::shared_ptr<PolyTerm>`, which is not used and causes warnings when compiling with GPU enabled.

## Checklist

The proposed changes:


WeiqunZhang:
Unfortunately, the shared_ptr is needed so that `m_dp` can be deleted.  Otherwise we would have a memory leak.  It takes the ownership from `m_dp` and stores in shared_ptr so that it's safe to copy-construct PolynominalIF, which is needed for kernel launch.

rporcu:
Oh, I see. I'll close the pull request then. Thanks for explaining that.

PR #1475, opened 2020-10-22
Including the cell itself in the non-conservative computation

OscarAntepara:
## Summary
In the eb_redistribution algorithm, the computation of the non-conservative divergence should include the cell itself.

## Checklist

The proposed changes:


OscarAntepara:
I will close this PR because we have decided to continue with the current version of redistribution.

PR #1480, opened 2020-10-23
CMake 3.18+: CUDA Arch Policy (OLD)

ax3l:
## Summary

Keep the old policy for now to avoid setting the code generation flags twice and silence a warning.

https://cmake.org/cmake/help/latest/policy/CMP0104.html

We should be able to transition this well, but I need to first find out how device LTO generation flags are handled here, if at all, and how to restore the default-detection of the local GPU architecture, e.g. on Summits head-nodes.

Note: also needs to be set in the a downstream (super-)project early enough when fetching AMReX, e.g.: https://github.com/ECP-WarpX/WarpX/pull/1459

## Checklist

The proposed changes:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -93,6 +93,12 @@ endif ()
 # Enable CUDA if requested
 #
 if (ENABLE_CUDA)
+    # CMake 3.18+: CMAKE_CUDA_ARCHITECTURES
+    # https://cmake.org/cmake/help/latest/policy/CMP0104.html
+    if(POLICY CMP0104)
+        cmake_policy(SET CMP0104 OLD)
+    endif()
+
     enable_language(CUDA)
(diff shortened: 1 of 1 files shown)


Tools/CMake/AMReX_SetupCUDA.cmake
```diff
@@ -53,6 +53,12 @@ endif ()
 #
 message(STATUS "Enabled CUDA options:")
 
+# CMake 3.18+: CMAKE_CUDA_ARCHITECTURES
+# https://cmake.org/cmake/help/latest/policy/CMP0104.html
+if(POLICY CMP0104)
```
ax3l: Oh, can I only set this in the calling project? Hm...
Setting this here does not work, has to be fore `enable_language(CUDA)` in the downstream project.

mic84 (approved):

PR #1481, opened 2020-10-23
Fix warning in FabSet for single precision

### LINKED ISSUE #1479: FabSet Warning in SP
Seen with Intel `icpc` from oneAPI beta09 (single-precision compile) and `-Werror`:
```
Src/Boundary/AMReX_FabSet.cpp(184): error #264: floating-point value does not fit in required floating-point type
          AMREX_HOST_DEVICE_PARALLEL_FOR_4D ( bx, ncomp, i, j, k, n,
          ^

Src/Boundary/AMReX_FabSet.cpp(184): error #264: floating-point value does not fit in required floating-point type
          AMREX_HOST_DEVICE_PARALLEL_FOR_4D ( bx, ncomp, i, j, k, n,
          ^
```

WeiqunZhang:
## Checklist

The proposed changes:

### DIFF
--- a/Src/Boundary/AMReX_FabSet.cpp
+++ b/Src/Boundary/AMReX_FabSet.cpp
@@ -181,10 +181,15 @@ FabSet::linComb (Real a, const MultiFab& mfa, int a_comp,
         const Box& bx = mfi.validbox();
         auto afab = bdrya.array(mfi);
         auto bfab = bdryb.array(mfi);
+#ifdef AMREX_USE_FLOAT
+        Real huge = 1.e30f;
+#else
+        Real huge = 1.e200;
+#endif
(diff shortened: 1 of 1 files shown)


Src/Boundary/AMReX_FabSet.cpp
```diff
@@ -181,10 +181,15 @@ FabSet::linComb (Real a, const MultiFab& mfa, int a_comp,
         const Box& bx = mfi.validbox();
         auto afab = bdrya.array(mfi);
         auto bfab = bdryb.array(mfi);
+#ifdef AMREX_USE_FLOAT
+        Real huge = 1.e30f;
+#else
+        Real huge = 1.e200;
```
ax3l:
```suggestion
        const Real huge = 1.e30f;
#else
        const Real huge = 1.e200;
```

ax3l (approved):
Thank you! :sparkles:

PR #1483, opened 2020-10-27
Divergence Free Stencil and API

kngott:
## Summary

API for multi-dimensional, face-based, coarse-to-fine interpolation (InterpFromCoarse and FillPatchTwoLevels). Includes a variation of Vanella et. al. stencil (doi:10.1016/j.jcp.2010.05.003, section 3.2), but instead of interpolating from a far coarse cell to solve interior closure problem (highly biased result dependent on chosen coarse cell), this uses least squares with initial guess equal to the average of fine face values across the cell. (Mathematica notebooks of solution are available).

~~Currently only includes 2D implementation of the stencil. Will add 3D next (Only the 3D interpolation functions should need to be re-written), but wanted to submit PR to get feedback on API.~~ 2D and 3D complete.
Includes simplified test that was performed on dogora and Summit with MPI and CUDA+MPI. Could use more detailed testing on more complicated geometries.

## Additional background

Discussion points:
* ~~Stencil includes two different implementations, each with different advantages. Would like everyone's thoughts on which to use (this can change number of temporary multifabs, performance and function signature required, so potentially a big deal).~~ Method has been chosen.
* Don't believe degenerate boxes are a problem here. Am I missing anything?
* Do these new function implementations work for general cases (e.g. MFix's potential use case)?
* ~~Should RemoveOverlap be used on the intersected face BoxArrays? (Any known performance info?)~~ Switched to mask.
* Can change the names of the interpolator before merging. Any suggestions? (Bell2020?)

* ~~From earlier discussion with Weiqun: some additional function parameters in InterpFromCoarse and FillPatchTwoLevels should be made const. Which? (I can update in both this and single MF functions).~~ Done.
* Fusible?

## Checklist

The proposed changes:


Src/AmrCore/AMReX_FillPatchUtil.H
```diff
@@ -83,6 +85,40 @@ namespace amrex
                         const PreInterpHook& pre_interp = {},
                         const PostInterpHook& post_interp = {});
 
+    template <typename MF, typename BC, typename Interp,
+              typename PreInterpHook=NullInterpHook<typename MF::FABType::value_type>,
+              typename PostInterpHook=NullInterpHook<typename MF::FABType::value_type> >
+    EnableIf_t<IsFabArray<MF>::value> 
+    FillPatchTwoLevels (Array<MF*, AMREX_SPACEDIM>& mf, IntVect const& nghost, Real time,
```
WeiqunZhang: I think I was talking about `Array<MF*, AMREX_SPACEDIM>&` -> `Array<MF*, AMREX_SPACEDIM> const&`.

Src/AmrCore/AMReX_FillPatchUtil_I.H
```diff
@@ -247,8 +338,8 @@ namespace {
                                                                       cgeom,
                                                                       index_space);
 
-            if ( ! fpc.ba_crse_patch.empty())
-            {
+	    if ( ! fpc.ba_crse_patch.empty())
+	    {
```
WeiqunZhang: Can we revert this change?  It introduced tabs.

WeiqunZhang:
We have two types of launch macros, `AMREX_LAUNCH_*_LAMBDA*` and `AMREX_*_FOR_*`.  The former takes a block of code that works on a `Box` (which is one cell in gpu case), whereas the latter works on `(i,j,k)`.  The historical reason for that is almost all our kernels used to work on a Box.  The newly introduced macros have `LAMBDA` in the name, but work on `(i,j,k)`.  Maybe we should keep the existing convention?

kngott:
I'll swap out the macro names and keep the convention. I just mismatched how this falls under the convention.

I was basing it on input to the macro: the user only gives a box in these and doesn't pass `(i, j, k)` directly, although the macro is based on receiving an `(i,j,k)` lambda. Now that you've pointed it out it makes sense, as it tells the user what's expected, not what we need to give it.

Src/Base/AMReX_GpuLaunch.H
```diff
@@ -205,14 +205,17 @@ namespace Gpu {
...
-#define AMREX_LAUNCH_HOST_DEVICE_LAMBDA_DIM(a1,a2,a3,b1,b2,b3,c1,c2,c3) AMREX_GPU_LAUNCH_HOST_DEVICE_LAMBDA_RANGE(a1,a2,a3)
+#define AMREX_LAUNCH_DEVICE_LAMBDA_DIM(a1,a2,a3,b1,b2,b3,c1,c2,c3)      AMREX_GPU_LAUNCH_DEVICE_LAMBDA_RANGE (a1,a2,a2)
+#define AMREX_LAUNCH_HOST_DEVICE_LAMBDA_DIM(a1,a2,a3,b1,b2,b3,c1,c2,c3) AMREX_GPU_LAUNCH_HOST_DEVICE_LAMBDA_RANGE (a1,a2,a3)
+#define AMREX_LAUNCH_HOST_DEVICE_PARALLEL_FOR_DIM_FLAG(fl,a1,a2,a3,b1,b2,b3,c1,c2,c3) AMREX_GPU_LAUNCH_HOST_DEVICE_PARALLEL_FOR_RANGE_FLAG(fl,a1,a2,a3)
 #elif (AMREX_SPACEDIM == 2)
-#define AMREX_LAUNCH_DEVICE_LAMBDA_DIM(a1,a2,a3,b1,b2,b3,c1,c2,c3)      AMREX_GPU_LAUNCH_DEVICE_LAMBDA_RANGE_2     (a1,a2,a3,b1,b2,b3)
-#define AMREX_LAUNCH_HOST_DEVICE_LAMBDA_DIM(a1,a2,a3,b1,b2,b3,c1,c2,c3) AMREX_GPU_LAUNCH_HOST_DEVICE_LAMBDA_RANGE_2(a1,a2,a3,b1,b2,b3)
+#define AMREX_LAUNCH_DEVICE_LAMBDA_DIM(a1,a2,a3,b1,b2,b3,c1,c2,c3) AMREX_GPU_LAUNCH_DEVICE_LAMBDA_RANGE_2 (a1,a2,a3,b1,b2,b3)
+#define AMREX_LAUNCH_HOST_DEVICE_LAMBDA_DIM(a1,a2,a3,b1,b2,b3,c1,c2,c3) AMREX_GPU_LAUNCH_HOST_DEVICE_PARALLEL_FOR_RANGE_2 (a1,a2,a3,b1,b2,b3)
```
WeiqunZhang: This change in the line above should be reverted.

Src/Base/AMReX_GpuLaunch.H
```diff
@@ -339,6 +342,39 @@ namespace Gpu {
         block; \
     }
 
+#define AMREX_GPU_LAUNCH_HOST_DEVICE_PARALLEL_FOR_RANGE_FLAG(where_to_run,bx1,nc1,block1) \
```
WeiqunZhang: In the existing macros, the `i`, `j`, `k`, `n` are macro arguments.  But the new ones are not, and therefore the users will not be able to change them.  If the enclosing scope has a local variable named `i` or `n`.  This will be confusing.  Also, maybe the name can have `4D` in it to be more explicit as we are currently doing.

Src/AmrCore/AMReX_FillPatchUtil_I.H
```diff
@@ -505,6 +814,146 @@ InterpFromCoarseLevel (MF& mf, IntVect const& nghost, Real time,
...
+    Box fdomain_g(fdomain);
+    for (int d = 0; d < AMREX_SPACEDIM; ++d) {
+        if (fgeom.isPeriodic(d)) {
+            fdomain_g.grow(d,nghost[d]);
+        }
+    }
+
+    BoxArray ba_crse_patch(ba.size());
+    BoxArray ba_fine_patch(ba.size());
+    BoxArray ba_target_patch(ba.size());
```
WeiqunZhang: This is not used.

WeiqunZhang: Actually, it should be used later.

Src/AmrCore/AMReX_FillPatchUtil_I.H
```diff
@@ -505,6 +814,146 @@ InterpFromCoarseLevel (MF& mf, IntVect const& nghost, Real time,
...
+
+            post_interp(dfab, dbx_cc, dcomp, ncomp);
+        }
+    }
+
+    Array<MF, AMREX_SPACEDIM> mf_target_patch;
+    for (int d = 0; d<AMREX_SPACEDIM; ++d)
+    {
+        IndexType typ = mf[d]->boxArray().ixType();
+        BoxArray ba_target_idxed = amrex::convert(ba_fine_patch, typ);
```
WeiqunZhang: I assume ba_target_patch is supposed to be used here.

Src/AmrCore/AMReX_FillPatchUtil_I.H
```diff
@@ -505,6 +814,146 @@ InterpFromCoarseLevel (MF& mf, IntVect const& nghost, Real time,
...
+
+#ifdef AMREX_USE_EB
+        auto target_factory = makeEBFabFactory(fgeom, ba_target_idxed, dm, {0,0,0}, EBSupport::basic);
+        mf_target_patch[d].define(ba_target_idxed, dm, ncomp, 0, MFInfo(), *target_factory);
+#else
+        mf_target_patch[d].define(ba_target_idxed, dm, ncomp, 0);
+#endif
+        amrex::Copy(mf_target_patch[d], mf_fine_patch[d], 0, 0, ncomp, 0);
```
WeiqunZhang: Then this might be a problem.

WeiqunZhang:
For InterpFromCoarseLevel, you might be able to do is this.  If the `nghost` is odd, you add 1 to make it even.  In that case, you build a temporary MultiFab withe the even number of ghost cells.  Then you don't have to worry about fine_region vs. target region.

WeiqunZhang:
What's in my mind is this
```
EnableIf_t<IsFabArray<MF>::value>
InterpFromCoarseLevel (Array<MF*, AMREX_SPACEDIM> const& a_mf, IntVect const& a_nghost,
                       ..., int a_dcomp,
                       ....)
{
    IntVect nghost = a_nghost;
        if (nghost[idim] % 2 == 1) nghost[idim] += 1;
    }

    Array<MF*, AMREX_SPACEDIM> mf = a_mf;
    int dcomp = a_dcomp;
    Array<std::unique_ptr<MF>,AMREX_SPACEDIM> mf_raii;
    if (nghost != a_nghost) {
            mf_raii[idim].reset(new MF(...., ncomp, nghost));
            mf[idim] = mf_raii[idim].get();
        }
        dcomp = 0;
    }

    // Now we work mf that always have even number of ghost cells.                                                                                                                                         

    for (int d = 0; d < AMREX_SPACEDIM; ++d) {
        mf[d]->ParallelCopy(....);
        if (mf[d] != a_mf[d]) {
            amrex::Copy(*a_mf[d], *mf[d], 0, a_dcomp, ncomp, a_nghost);
        }
        fbc[d](*a_mf[d], dcomp, ncomp, nghost, time, fbccomp);
    }
}
```

kngott:
That's a slicker way to do it for sure. I like it. Although I would definitely go with a `ghost % ratio` test to keep the InterpFromCoarse function generalizable for future face-based interpolaters. e.g. `ratio` is tested inside the FaceDivFree interp functions, as `ratio == 2` is a specific requirement for the new interpolater.

I'll make an attempt at this soon. I think this is tricker than first glance because of the possibility that the region of `mf` that is calculated can be altered by the passed domain. From the current creation of the various BoxArrays:
```
    BoxArray ba_crse_patch(ba.size());
    BoxArray ba_fine_patch(ba.size());
    BoxArray ba_target_patch(ba.size());
    {  // TODO: later we might want to cache this
        for (int i = 0, N = ba.size(); i < N; ++i)
        {
            Box bx = amrex::convert(amrex::grow(ba[i], nghost), IntVect::TheZeroVector());
            bx &= fdomain_g;       <====== ((((DOMAIN, WITH GHOST CELLS, MAY REDUCE CALC'ED REGION))))
            ba_target_patch.set(i, bx);
            ba_fine_patch.set(i, amrex::refine(amrex::coarsen(bx, ratio), ratio));
            ba_crse_patch.set(i, coarsener.doit(bx));
        }
    }
```

So, four cases need to be covered correctly (first two are covered by your idea):
* `ghost % ratio = 0` and `domain >= mf`.   (simple case, can just write into `mf`)
* `ghost % ratio != 0` and `domain >= mf`.  (temporary Multifab with extra ghost cells)
* `ghost % ratio = 0` and `domain < mf`. (Can put into `mf` but need to setup a mask, so still need to calc `target` ba?). 
* `ghost % ratio != 0` and `domain <mf`. (Needs both masking and temporary with extra ghost cells. So, similar to current `fine` + `target` implementation, except we're exchanging `target` for masking fab?)

May take some thought to figure out how to incorporate all 4 cases cleanly with your proposed optimizations.

WeiqunZhang:
I don't think you need to worry about the domain size.

Src/AmrCore/AMReX_FillPatchUtil_I.H
```diff
@@ -848,19 +848,31 @@ InterpFromCoarseLevel (Array<MF*, AMREX_SPACEDIM> const& mf, IntVect const& ngho
         }
     }
 
-    const bool nested = ( AMREX_D_TERM(nghost[0] % ratio[0] == 0,
-                                    && nghost[1] % ratio[1] == 0,
-                                    && nghost[2] % ratio[2] == 0));
+    // If needed, adjust to fully overlap the coarse cells.
+    IntVect nghost_adj = nghost;
+    for (int d=0; d<AMREX_SPACEDIM; ++d) {
+        nghost_adj[d] += nghost[d] % ratio[d];
```
WeiqunZhang: I think this should be `if ((nghost[d] % ratio[d]) != 0) nghost_adj[d] += ratio[d] - nghost[d] % ratio[d];`

kngott: Agreed.

WeiqunZhang (approved):

PR #1484, opened 2020-10-27
CI: -Werror for host compilers

ax3l:
## Summary

Add more warning coverage in CI for CMake builds.

## Checklist

The proposed changes:

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -7,6 +7,7 @@ jobs:
   library:
     name: GNU@7.5 C++17 Release [lib]
     runs-on: ubuntu-latest
+    env: {CXXFLAGS: "-Werror"}
     steps:
     - uses: actions/checkout@v2
     - name: Dependencies
@@ -25,6 +26,7 @@ jobs:
   library_clang:
     name: Clang@6.0 C++14 SP NOMPI Debug [lib]
     runs-on: ubuntu-latest
+    env: {CXXFLAGS: "-Werror"}
     steps:
     - uses: actions/checkout@v2
     - name: Dependencies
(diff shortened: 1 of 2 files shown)


ax3l:
@mic84 I think about adding something like this block, guarded by being the root project in a super build, to our CMake logic to be on par with the warning coverage in downstream projects (i.e. WarpX) and GNUmake. What do you think?

https://github.com/ECP-WarpX/WarpX/blob/51f5af9f1950209948c0e91ff9e2882d9b9c5fed/cmake/WarpXFunctions.cmake#L82-L113

Src/Base/AMReX_GpuAsyncArray.H
```diff
@@ -22,6 +22,19 @@ extern "C" {
 #endif
 
 namespace amrex {
+
+namespace type_traits {
+#if __cplusplus < 202002L
```
ax3l: Hah, GCC 10.1 identifies as C++17 if compiled with `-std=c++2a` but already adds the C++20 deprecation notice ;)

mic84:
> @mic84 I think about adding something like this block, guarded by being the root project in a super build, to our CMake logic to be on par with the warning coverage in downstream projects (i.e. WarpX) and GNUmake. What do you think?
> 
> https://github.com/ECP-WarpX/WarpX/blob/51f5af9f1950209948c0e91ff9e2882d9b9c5fed/cmake/WarpXFunctions.cmake#L82-L113

@ax3l I am in favour of adding extra warnings but I would prefer not to do it via ``CMAKE_CXX_FLAGS``

ax3l:
@mic84 sounds good, let's to this in a follow-up PR.

mic84 (approved):

WeiqunZhang (approved):

Issue #1450, opened 2020-10-13
setting eb2.small_volfrac can introduce inconsistencies in EB data


darylbond:
In my implementation, setting `eb2.small_volfrac` triggered a situation where the boundary normal provided by `ebfactory->getBndryNormal()` did not correspond with the associated face area fractions given by `ebfactory->getAreaFrac()`. This lead to non-conservative flux calculation when the area fraction was used to calculate the cell divergence while the boundary flux was calculated using the provided boundary normal. This situation can obviously be avoided by calculating the boundary normal from the area fractions.

asalmgren:
Thank you for pointing this out!

PR #1451 should fix this -- if you have a chance could you check that it fixes the case you have observed?

darylbond:
I have found another inconsistency, this time between volume fraction and `isCovered()`. I was able to fix this by performing a check on volfrac after every call to `set_eb_data` such that the covered status is consistent with whatever was just calculated. However, with further testing, it appears this approach leads to floating-point errors in `EB2::coarsen_from_fine()` so not such a good idea after all.

Edit: removed a bunch of code that didn't work.

PR #1487, opened 2020-10-29
Fix GPU race condition for EB

WeiqunZhang:
## Summary
A GPU race condition was introduced in #1451 trying to fix faces and cells
next to covered cells that were converted from small cut cells.  It is fixed
by launching a separate kernel for each direction.

## Checklist

The proposed changes:


asalmgren (commented):
Does this also address issue #1450?   Or will that need an additional fix?

WeiqunZhang:
Maybe, but I am not sure.

asalmgren (approved):

PR #1505, opened 2020-11-04
Add option for device sync at beginning and end of TinyProfiler region

maximumcats:
## Summary

Now when setting `tiny_profiler.device_synchronize_on_region_end  = 1` in the inputs file, we will synchronize before calling nvtxRangePop() and nvtxRangePush(), which means that TINY_PROFILE regions will include the full kernel time, rather than just the kernel launch time.

## Checklist

The proposed changes:


WeiqunZhang:
I wonder if we should sync before the timer starts.

maximumcats:
Probably a good idea also. Should I add a separate option or rename it and then have it do both?

WeiqunZhang:
I prefer one parameter to rule them all.

maximumcats:
Updated.

WeiqunZhang (approved):

PR #1490, opened 2020-10-30
CMake: overhaul options

mic84:
## Summary

- [x] wait for WarpX CI to be ready for breaking changes: https://github.com/ECP-WarpX/WarpX/pull/1492
- [x] wait for `<unreleased-WarpX-code>` to be ready
- [x] wait for incflo to be ready for breaking changes
- [x] wait for MFIX to be ready for breaking changes
- [x] wait for AMR-Wind to be ready for breaking changes
- [x] wait for PeleC to be ready for breaking changes

## Checklist

The proposed changes:
- [ ] fix a bug or incorrect behavior in AMReX
- [ ] add new capabilities to AMReX
- [ ] changes answers in the test suite to more than roundoff level
- [ ] are likely to significantly affect the results of downstream AMReX users
- [ ] are described in the proposed changes to the AMReX documentation, if appropriate

PR #1507, opened 2020-11-04
GNU Make: AMREX_CUDA_ARCH

WeiqunZhang:
## Summary
Option to use AMREX_CUDA_ARCH in addition to CUDA_ARCH to set CUDA arch.  If
AMREX_CUDA is set, it will be used over CUDA_ARCH.

## Additional background
The CMake system is also planning to support AMREX_CUDA_ARCH.

## Checklist

The proposed changes:

### DIFF
--- a/Docs/sphinx_documentation/source/BuildingAMReX.rst
+++ b/Docs/sphinx_documentation/source/BuildingAMReX.rst
@@ -28,37 +28,40 @@ list of important variables.
 
 .. table:: Important make variables
 
-   +------------+-------------------------------------+--------------------+
-   | Variable   | Value                               | Default            |
-   +============+=====================================+====================+
-   | AMREX_HOME | Path to amrex                       | environment        |
-   +------------+-------------------------------------+--------------------+
-   | COMP       | gnu, cray, ibm, intel, llvm, or pgi | none               |
-   +------------+-------------------------------------+--------------------+
-   | CXXSTD     | C++ standard (``c++11``, ``c++14``, | compiler default,  |
-   |            | ``c++17``, ``c++20``)               | at least ``c++11`` |
-   +------------+-------------------------------------+--------------------+
-   | DEBUG      | TRUE or FALSE                       | FALSE              |
-   +------------+-------------------------------------+--------------------+
-   | DIM        | 1 or 2 or 3                         | 3                  |
-   +------------+-------------------------------------+--------------------+
-   | PRECISION  | DOUBLE or FLOAT                     | DOUBLE             |
-   +------------+-------------------------------------+--------------------+
-   | USE_MPI    | TRUE or FALSE                       | FALSE              |
-   +------------+-------------------------------------+--------------------+
-   | USE_OMP    | TRUE or FALSE                       | FALSE              |
-   +------------+-------------------------------------+--------------------+
-   | USE_CUDA   | TRUE or FALSE                       | FALSE              |
-   +------------+-------------------------------------+--------------------+
-   | USE_HIP    | TRUE or FALSE                       | FALSE              |
-   +------------+-------------------------------------+--------------------+
-   | USE_DPCPP  | TRUE or FALSE                       | FALSE              |
-   +------------+-------------------------------------+--------------------+
-   | USE_RPATH  | TRUE or FALSE                       | FALSE              |
-   +------------+-------------------------------------+--------------------+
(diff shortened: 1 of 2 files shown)


maximumcats (approved):

mic84 (approved):

ax3l:
@WeiqunZhang ! Didn't we also want to add `AMReX_CUDA_ARCH` as environment hint here? :)
This will allow us to pre-define something in `warpx.profile` files for systems without having to tell the user details about a system+queue for the compile commmand.

WeiqunZhang:
I didn't notice that the CMake system was using mixed case.  It thought it was all capitals.  @mic84 said that it's a CMake custom to use the package name (in this case `AMReX`) instead of all capitals.  But doesn't CMake itself uses `CMAKE_*`?

ax3l:
Yes, CMake variables (`-D`/`set(SomeVar ...)`) are [case sensitive](https://cmake.org/cmake/help/latest/manual/cmake-language.7.html#variables).
Variables starting with `CMAKE_` are for set for the whole build, that includes all sub-projects that this build might pull in on the fly (via `add_subdirectories()`). Those are therefore special, as they do not refer to a project. They are essentially [reserved globals](https://cmake.org/cmake/help/latest/manual/cmake-variables.7.html) for a build and often used to initialize defaults on all to-be-built targets.

In order to fine-control sub-projects, e.g. AMReX from WarpX, we prefix WarpX and AMReX variables accordingly. Most projects use their preferred spelling for both CMake `project()` identifier (kind of a namespace) and mirror that in the variables. So we do `-DWarpX_...` and `-DAMReX_...`. Other projects use e.g. `Boost_...`, `HDF5_...`, etc. - some older projects use CAPS for their name, very few projects do all-lowercase.

If you build CMake with CMake, then CMake (the to-be-built project) project built options are controlled as well with [`CMake_...` variables](https://gitlab.kitware.com/cmake/cmake/-/blob/v3.19.0-rc3/CMakeLists.txt#L120), too. General options for the tool CMake are passed again with `CMAKE_...`.

Orthogonal to that, we introduced a default for the `AMReX_CUDA_ARCH` variable via an _environment variable_ in #1490.
I picked the spelling `AMReX_CUDA_ARCH` for symmetry without thinking much, but if its easier to unify with GNUmake we can also adopt the convention to make all AMReX _environment variables_ all-caps, i.e. `AMREX_CUDA_ARCH`. It's probably the first of its kind, so the break is minimal. Would you prefer that? I think it might be more common in the unix world to caps env vars.

WeiqunZhang:
My preference is the environment variables set in the shell should be all caps.  @maxpkatz concurred on slack.

ax3l:
Sounds good, will change it asap: #1522

Docs/sphinx_documentation/source/BuildingAMReX.rst
```diff
@@ -28,37 +28,40 @@ list of important variables.
...
+   | USE_HIP         | TRUE or FALSE                       | FALSE              |
+   +-----------------+-------------------------------------+--------------------+
+   | USE_DPCPP       | TRUE or FALSE                       | FALSE              |
+   +-----------------+-------------------------------------+--------------------+
+   | USE_RPATH       | TRUE or FALSE                       | FALSE              |
+   +-----------------+-------------------------------------+--------------------+
+   | WARN_ALL        | TRUE or FALSE                       | TRUE for DEBUG     |
+   |                 |                                     | FALSE otherwise    |
+   +-----------------+-------------------------------------+--------------------+
+   | AMREX_CUDA_ARCH | CUDA arch such as 70                | 70 if not set      |
```
ax3l, line 62: @WeiqunZhang looking at this PR again, I actually meant to also introduce an **environment** variable `AMREX_CUDA_ARCH` that sets the default for this GNUmake option.

Maybe I don't get the implementation, but at the moment it looks like this PR added a GNUmake option that is defaulted to `70` but cannot be defaulted with an env var. Please correct me if I overlook something :)

I like to introduce env-var control in WarpX, because it simplifies modules and `warpx.profile` files for users with sensible defaults for a queue/system.

WeiqunZhang: If the environment defines `AMREX_CUDA_ARCH`.  `ifdef AMREX_CUDA_ARCH` will be true.  So it looks OK to me.

ax3l: Perfect, I was confused if that is picked up. Thanks! :)

PR #1510, opened 2020-11-05
Fix some build issues at LLNL

maximumcats:
## Checklist

The proposed changes:

### DIFF
--- a/Tools/GNUMake/sites/Make.llnl
+++ b/Tools/GNUMake/sites/Make.llnl
@@ -12,7 +12,7 @@ ifeq ($(which_computer),$(filter $(which_computer),ray rzmanta))
 
   XLC_ROOT = /usr/tce/packages/xl/xl-beta-2018.06.01/xlC/16.1.1/
   XLF_ROOT = /usr/tce/packages/xl/xl-beta-2018.06.01/xlf/16.1.1/
-  MPIROOT=/usr/tce/packages/spectrum-mpi/ibm/spectrum-mpi-2018.06.01
+  MPIROOT=/usr/tce/packages/spectrum-mpi/ibm/spectrum-mpi-rolling-release
 
   ifeq ($(USE_MPI),TRUE)
     CC  = mpicc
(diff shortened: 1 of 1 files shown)


Tools/GNUMake/sites/Make.llnl
```diff
@@ -12,15 +12,20 @@ ifeq ($(which_computer),$(filter $(which_computer),ray rzmanta))
...
+  MPIROOT=/usr/tce/packages/spectrum-mpi/ibm/spectrum-mpi-rolling-release
 
   ifeq ($(USE_MPI),TRUE)
     CC  = mpicc
     CXX = mpicxx
     FC  = mpifort
     F90 = mpifort
 
-    LIBRARIES += -lmpi_ibm_mpifh
+    ifeq ($(USE_GPU_PRAGMA),TRUE)
```
jmsexton03, line 23: Shouldn't this test on AMREX_USE_CUDA_FORTRAN or BL_NO_FORT? Or is there something specific about how USE_GPU_PRAGMA builds that requires this library?

maximumcats: USE_CUDA_FORTRAN isn't a GNU make variable. Really this should technically be LINK_WITH_FORTRAN_COMPILER, but in any case we only link with the Fortran compiler because of USE_GPU_PRAGMA.

maximumcats: I could try to get this right if someone particularly cares, but I was just putting this in as a quick fix, my hope is to rip a lot of this infrastructure out as soon as Castro completes its cutover to gcc-only builds (coming soon!).

jmsexton03: I don't anticipate this change affecting me personally, so a quick fix is fine by me. Thanks for explaining the underlying logic.
My concern was mostly about whether things like amrex/Tutorials/GPU/Launch still compile correctly on that system with `make COMP=ibm USE_MPI=TRUE` or `make COMP=pgi`, since when I test those compilations on other systems it links with mpif90 or pgfortran, not nvcc.

WeiqunZhang (approved):

PR #1512, opened 2020-11-05
Support CUDA arch < 6.0

WeiqunZhang:
## Summary
* Implement atomicAdd for CUDA arch < 6.0.

* Remove the CUDA arch check in CMake.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_BaseFab.H
+++ b/Src/Base/AMReX_BaseFab.H
@@ -1645,10 +1645,12 @@ BaseFab<T>::prefetchToHost () const noexcept
         // auto& q = Gpu::Device::streamQueue();
         // q.submit([&] (sycl::handler& h) { h.prefetch(this->dptr, s); });
 #elif defined(AMREX_USE_CUDA)
-        std::size_t s = sizeof(T)*this->nvar*this->domain.numPts();
-        AMREX_CUDA_SAFE_CALL(cudaMemPrefetchAsync(this->dptr, s,
-                                                  cudaCpuDeviceId,
-                                                  Gpu::gpuStream()));
+        if (Gpu::Device::devicePropMajor() >= 6) {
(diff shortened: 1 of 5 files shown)


WeiqunZhang:
I am able to compile AMReX for CUDA arch 3.7, but I don't have an old GPU like K80 for testing.  @maxpkatz Do you expect AMReX would work on K80?

maximumcats:
> Do you expect AMReX would work on K80?

I'm just worried about atomics on double precision. Do you have an implementation of e.g. atomicAdd based on atomicCAS for compute capability < 6.0?

WeiqunZhang:
Yes, https://github.com/WeiqunZhang/amrex/blob/cuda_arch_lt_600/Src/Base/AMReX_GpuAtomic.H#L48

maximumcats:
OK, then here are my thoughts.

1. We should limit this to sm_35 or above, there's no point in going older than that. The check on compute capability in AMReX_GpuDevice.cpp which currently enforces >= sm_60 at runtime will need to be changed.

2. Unified Memory on pre-Pascal GPUs does not support things like cudaMemPrefetchAsync, so BaseFab prefetchToHost and prefetchToDevice should probably be made no-ops for these architectures. Also, users will need to be careful that they are not violating [the rules for Unified Memory on these GPUs](https://developer.nvidia.com/blog/unified-memory-in-cuda-6/). If I avoid this issue and the previous one locally, I can get the CNS and ParallelReduce tutorials to work on an sm_35 GPU.

3. sm_35, sm_37, and sm_50 are deprecated in CUDA 11, so consider that when deciding whether or how long you want to support them.

WeiqunZhang:
Thanks!  Forgot about we have a check of compute capability.

maximumcats (approved):

PR #1514, opened 2020-11-05
CMake: fix options name in Test directory

mic84:
## Checklist

The proposed changes:


ax3l (approved):
:+1:

PR #1515, opened 2020-11-05
Ascent: Check Bounds Particle Reals

ax3l:
## Summary

Access of particle variable names with a bound-check, so users definitely pass the right amount of names.

cc @cyrush @mclarsen 

## Checklist

The proposed changes:

### DIFF
--- a/Src/Extern/Conduit/AMReX_Conduit_Blueprint_ParticlesI.H
+++ b/Src/Extern/Conduit/AMReX_Conduit_Blueprint_ParticlesI.H
@@ -100,7 +100,7 @@ ParticleTileToBlueprint(const ParticleTile<NStructReal,
     for (int i = 0; i < NStructReal; i++)
     {
         ParticleReal* val = reinterpret_cast<ParticleReal*>(pbuf); pbuf += sizeof(ParticleReal);
-        conduit::Node &n_f = n_fields[real_comp_names[vname_real_idx]];
+        conduit::Node &n_f = n_fields[real_comp_names.at(vname_real_idx)];
         n_f["topology"] = topology_name;
         n_f["association"] = "element";
         n_f["values"].set_external(val,
@@ -146,7 +146,7 @@ ParticleTileToBlueprint(const ParticleTile<NStructReal,
     for (int i = 0; i < NStructInt; i++)
     {
         int* val = reinterpret_cast<int*>(pbuf); pbuf += sizeof(int);
-        conduit::Node &n_f = n_fields[int_comp_names[vname_int_idx]];
+        conduit::Node &n_f = n_fields[int_comp_names.at(vname_int_idx)];
         n_f["topology"] = topology_name;
         n_f["association"] = "element";
         n_f["values"].set_external(val,
@@ -168,7 +168,7 @@ ParticleTileToBlueprint(const ParticleTile<NStructReal,
     // array real fields
     for (int i = 0; i < NArrayReal; i++)
     {
-        conduit::Node &n_f = n_fields[real_comp_names[vname_real_idx]];
+        conduit::Node &n_f = n_fields[real_comp_names.at(vname_real_idx)];
         n_f["topology"] = topology_name;
         n_f["association"] = "element";
         n_f["values"].set_external(const_cast<ParticleReal*>(&soa.GetRealData(i)[0]),
@@ -180,7 +180,7 @@ ParticleTileToBlueprint(const ParticleTile<NStructReal,
     // array int fields
     for (int i = 0; i < NArrayInt; i++)
     {
-        conduit::Node &n_f = n_fields[int_comp_names[vname_int_idx]];
+        conduit::Node &n_f = n_fields[int_comp_names.at(vname_int_idx)];
         n_f["topology"] = topology_name;
         n_f["association"] = "element";
         n_f["values"].set_external(const_cast<int*>(&soa.GetIntData(i)[0]),


Src/Extern/Conduit/AMReX_Conduit_Blueprint_ParticlesI.H
```diff
@@ -166,9 +166,10 @@ ParticleTileToBlueprint(const ParticleTile<NStructReal,
     // since these are contiguous arrays
 
     // array real fields
+    vname_real_idx = 0;
     for (int i = 0; i < NArrayReal; i++)
     {
-        conduit::Node &n_f = n_fields[real_comp_names[vname_real_idx]];
+        conduit::Node &n_f = n_fields[real_comp_names.at(vname_real_idx)];
```
ax3l, line 172: Not sure if `conduit::Node` supports something like `::insert()` to catch double-assignments?

ax3l, line 171: Maybe that's even safer? @cyrush 
```suggestion
        conduit::Node &n_f = n_fields.at(real_comp_names.at(vname_real_idx));
```

Src/Extern/Conduit/AMReX_Conduit_Blueprint_ParticlesI.H
```diff
@@ -178,9 +179,10 @@ ParticleTileToBlueprint(const ParticleTile<NStructReal,
     }
 
     // array int fields
+    vname_int_idx = 0;
     for (int i = 0; i < NArrayInt; i++)
     {
-        conduit::Node &n_f = n_fields[int_comp_names[vname_int_idx]];
+        conduit::Node &n_f = n_fields[int_comp_names.at(vname_int_idx)];
```
ax3l, line 183: As above:
```suggestion
        conduit::Node &n_f = n_fields.at(int_comp_names.at(vname_int_idx));
```

cyrush:
good catch!

conduit doesn't have an `.at`, 

it has `fetch_child` that will fail if the child doesn't exist (not same use case)

`has_child` can be used to check if its already there.

cyrush:
not 100% sure about this, i think `real_comp_names` and  `int_comp_names` should be lists of all of the fields AoS or SoA, in order.  So resetting to zero might not be the right thing?

ax3l:
Oh, you are right - the first iteration is over `NStructReal`.

In WarpX, we do not pass the names of the AoS names into `real_comp_names`, but only the SoA names atm:
- https://github.com/ECP-WarpX/WarpX/blob/5d1b88eaae68353897382423cb87ee0eb7266fa9/Source/Diagnostics/FlushFormats/FlushFormatAscent.cpp#L68-L80
- https://github.com/ECP-WarpX/WarpX/blob/e7f475ae73a84d6d7f155f12d4be2ebd79e708a6/Source/Particles/WarpXParticleContainer.H#L301
- https://github.com/ECP-WarpX/WarpX/blob/e7f475ae73a84d6d7f155f12d4be2ebd79e708a6/Source/Particles/WarpXParticleContainer.cpp#L48-L55

Hm, but WarpX does not add additional AoS attributes, so the first loop is empty in our case anyway:
https://github.com/ECP-WarpX/WarpX/blob/e7f475ae73a84d6d7f155f12d4be2ebd79e708a6/Source/Particles/WarpXParticleContainer.H#L131-L132

ax3l:
> it has `fetch_child` that will fail if the child doesn't exist (not same use case)

and does `Node::operator[]` do a key check or does it create a new empty value in-place if missing?

ax3l:
Hm, but we do not add additional AoS attributes, so the first loop is empty in our case anyway:
https://github.com/ECP-WarpX/WarpX/blob/e7f475ae73a84d6d7f155f12d4be2ebd79e708a6/Source/Particles/WarpXParticleContainer.H#L131-L132

cyrush:
the semantics of non-const `fetch` (and operator[const std::string&]) work like this:

- if the node at the path doesn't exist, it will be created and the reference is returned
- if the node at the path does exist, the reference is returned

if you use `fetch_existing` (or const fetch, const operator[const std::string&]):
- if the node at the path doesn't exist, an error is thrown 
- if the node at the path does exist, the reference is returned

Src/Extern/Conduit/AMReX_Conduit_Blueprint_ParticlesI.H
```diff
@@ -100,7 +100,7 @@ ParticleTileToBlueprint(const ParticleTile<NStructReal,
     for (int i = 0; i < NStructReal; i++)
     {
         ParticleReal* val = reinterpret_cast<ParticleReal*>(pbuf); pbuf += sizeof(ParticleReal);
-        conduit::Node &n_f = n_fields[real_comp_names[vname_real_idx]];
+        conduit::Node &n_f = n_fields[real_comp_names.at(vname_real_idx)];
```
ax3l, line 103: @cyrush thank you for the details.
So the logic part here: is this the place where we create the key in `n_fields` or do we expect the key to already exist at this point?

From what I currently understand, it looks to me like we create the key here for the first time, so I would, contrary to what I thought earlier, not replace it with `fetch_existing`.

cyrush: correct -- this is where we create the key. 
We are using the name from real_comp_names to create a new node in our tree, then populating the children of that node with the rest of the info.

atmyers:
This looks good to me - is it still WIP or is it ready to be merged?

ax3l:
This is good to go now: the updated PR only adds a bounds check.

The issue that we were searching for was downstream: https://github.com/ECP-WarpX/WarpX/pull/1508

atmyers (approved):

PR #1529, opened 2020-11-09
Add anonymous namespace back to parameters in AMReX_Amr.cpp

WeiqunZhang:


maximumcats:
I've actually been relying on this change in Castro:

https://github.com/AMReX-Astro/Castro/blob/main/Source/driver/Castro.cpp#L146-L148

https://github.com/AMReX-Astro/Castro/blob/main/Source/driver/Castro.cpp#L472-L476

Is there a way I can continue to do this?

WeiqunZhang:
We can add `compute_new_dt_on_regrid` as a static member of `Amr`.

maximumcats:
Thank you! Can you also add a static setter function? e.g. a public
```
static void setComputeNewDtOnRegrid (int compute_new_dt) { compute_new_dt_on_regrid = compute_new_dt; }
```
I need that since read_params() is a static function too.

WeiqunZhang:
Sure

maximumcats (approved):

PR #1540, opened 2020-11-14
nvcc --Werror cross-execution-space-call

WeiqunZhang:
## Summary

* GNU make: Turn on the flag by default.  Use `GPU_ERROR_CROSS_EXECUTION_SPACE_CALL=FALSE` to turn it off.

* CMake: Add AMReX_ERROR_CROSS_EXECUTION_SPACE_CALL, which if off by default (just like AMReX_ERROR_CAPTURE_THIS).

## Checklist

The proposed changes:


maximumcats (approved):

mic84:
@WeiqunZhang looks good to me. Could you just update the CMake doc to include this new option? Thanks.

mic84 (approved):

PR #1541, opened 2020-11-14
GPU kernel timer

mrowan137:
## Summary
This PR adds a kernel timer device object that can measure number of elapsed GPU cycled summed thread-wise. This measurement tool can be used to measure relative compute work associated with a kernel. CUDA and HIP are currently supported.  To compile with the kernel timer use the compile flag `USE_KERNEL_TIMER=TRUE`.

## Checklist

The proposed changes:


WeiqunZhang:
There is `__syncthreads()` in the dtor of the timer.  That means it cannot be used in ParalleFor like in the tutorial example.  The reason for that is the threads in the same block may not call the user provided function the same number of times.

Another issue is in the ctor of the timer, there is a line to initialize the shared memory to zero by all threads.  That could cause a race condition.  A thread's result might be zeroed out by another thread in the same block.

Also the tutorial relies on that the memory allocated by `new` being accessible on device.  That's not the case for x86.

mrowan137:
Thank you for catching the many issues here.

> Another issue is in the ctor of the timer, there is a line to initialize the shared memory to zero by all threads. That could cause a race condition. A thread's result might be zeroed out by another thread in the same block.

Would it be problematic (for the same reason you mentioned before) here to add `__syncthreads()`?

> There is `__syncthreads()` in the dtor of the timer. That means it cannot be used in ParalleFor like in the tutorial example. The reason for that is the threads in the same block may not call the user provided function the same number of times.

Since it's not guaranteed that threads in the same block call the user-provided function the same number of times, would you always recommend against having a `__syncthreads()` in the user-provided function run in a ParallelFor?  Do you have any suggested workaround (maybe it will be easiest to forego any used of shared memory and just do atomic add in the dtor; I had tried this before but it was much slower than first storing in shared memory, but at least this way should avoid the problems here).


> Also the tutorial relies on that the memory allocated by `new` being accessible on device. That's not the case for x86.

Sorry I have misunderstood, can you expand a tiny bit to help me understand?  I am confused what numbers are printed out when I ran this on Summit and why their ratio looks reasonable, should the memory allocated by `new` not be modifiable through the `atomicAdd`s?

maximumcats:
> should the memory allocated by new not be modifiable through the atomicAdds?

Right, that is what @WeiqunZhang was saying. It will work on Summit due to a custom CPU-GPU configuration. A more portable solution is to use pinned host memory (the kind you get with e.g. `cudaMallocHost`). This could be allocated from the AMReX pinned arena if you wanted, to avoid explicitly using the CUDA API.

maximumcats:
> Do you have any suggested workaround (maybe it will be easiest to forego any used of shared memory and just do atomic add in the dtor; I had tried this before but it was much slower than first storing in shared memory, but at least this way should avoid the problems here).

Part of the slowness will be because you were writing directly to host memory. It would be higher performance if you wrote to something in device memory. To get that, you could do

```
#include <AMReX_Arena.H>

amrex::Real* cost = The_Managed_Arena()->alloc(sizeof(amrex::Real));
*cost = 0.;
...
The_Managed_Arena()->free(cost);
```

and then use it in the kernel. This pointer will be accessible on both host and device. (With the caveat that it won't be optimally high performance since the pointer needs to migrate to the device; this could be mitigated by putting the initialization to zero in the timer constructor.) Can you try just using atomic adds with this scheme and see if it's acceptable for this use case?

mrowan137:
@maxpkatz @WeiqunZhang 
> Can you try just using atomic adds with this scheme and see if it's acceptable for this use case?

It looks like pinned host memory works great (i.e. small overhead when I instrument WarpX test kernels with the timer) for my test case, which allows to fix the issues.  Thanks!

WeiqunZhang:
I have some concerns.

How much overhead does it have? I have done some testing on A100.  Without the timer, the whole block of MFIter loop takes 0.015 seconds.  With the timer, it takes 3.04 seconds.  I also implemented this in a different way like below.  It takes 0.046 seconds.
```
    {
        BL_PROFILE("reduce_sum");
        ReduceOps<ReduceOpSum> reduce_op;
        ReduceData<Real> reduce_data(reduce_op);
        using ReduceTuple = typename decltype(reduce_data)::Type;
        for (MFIter mfi(mf); mfi.isValid(); ++mfi)
        {
            const Box& bx = mfi.tilebox();
            Array4<Real> const& fab = mf.array(mfi);
            reduce_op.eval(bx, reduce_data,
            [=] AMREX_GPU_DEVICE (int i, int j, int k) -> ReduceTuple
            {
                auto t0 = clock64();
                fab(i,j,k) += 1.;   
                auto t1 = clock64();
                return {amrex::Real(t1-t0)};
            });
        }
        ReduceTuple hv = reduce_data.value();
        amrex::Print() << "result of reduce_sum is " << amrex::get<0>(hv) << std::endl;
    }
```

How accurate is it?  The number of cycles measured by the timer in the PR is about 3.6e14, whereas it's 3.7e12 in the code above.  It's 100x difference.

mrowan137:
> How accurate is it? The number of cycles measured by the timer in the PR is about 3.6e14, whereas it's 3.7e12 in the code above. It's 100x difference.

Is it possible that both are accurate, but there is indeed more cycles when `atomicAdd` is there?  I tried your example with a `KernelTimer` between the clocks, but in one case just comment out the atomic add, and compared to your original example.  I measured (on V100) about 3.7e10 (your original example), 3.7e10 (example below with commented `atomicAdd` in `KernelTimer`'s dtr), 2.1e12 (example below with uncommented `atomicAdd`).  Since your example has relatively low overhead, and also gives more reasonable ratios (I get a ratio of cycles ~ 63 changing from box size 256-->1024 with your example but ~53 with the `atomicAdd`s), I think it will be a huge improvement to do the timing as you suggest.
```
    amrex::Real* cost = (amrex::Real*) The_Managed_Arena()->alloc(sizeof(amrex::Real));
    *cost = 0.;
    ReduceOps<ReduceOpSum> reduce_op;
    ReduceData<Real> reduce_data(reduce_op);
    using ReduceTuple = typename decltype(reduce_data)::Type;
    for (MFIter mfi(mf); mfi.isValid(); ++mfi)
    {
        const Box& bx = mfi.tilebox();
        Array4<Real> const& fab = mf.array(mfi);
        reduce_op.eval(bx, reduce_data,
                       [=] AMREX_GPU_DEVICE (int i, int j, int k) -> ReduceTuple
                       {
                           auto t0 = clock64();
			   amrex::KernelTimer KnlTimer(true, cost);
                           fab(i,j,k) += 1.;   
                           auto t1 = clock64();
                           return {amrex::Real(t1-t0)};
                       });
    }
    ReduceTuple hv = reduce_data.value();
    amrex::Print() << "result of reduce_sum is " << amrex::get<0>(hv) << std::endl;
    The_Managed_Arena()->free(cost);
```

WeiqunZhang:
I believe neither is accurate unless the actual kernel is order of magnitude more expensive than the timer related cost, because of the way how GPU schedules its work.  @maxpkatz will correct me.  Conceptually one can think that a warp starts to work, and after it finishes all its work, another warp will take its place to run.  But in reality, both warps are running.  It might run one instruction from warp 1 and then switch to run two instructions from warp 2, and then back to run warp 1's instructions again.  Then maybe warp 3 will run some instructions.  At that point, all 3 have done some work, but neither has finished.

```
{
   amrex::KernelTimer KnlTimer(true, cost);  // that has a call to clock64
   {
         // actual kernel
   }
   // compiler inserts ~KernelTimer, which calls clock64 first and then store the difference between two times
}
```
The code above is supposed to measure the actual kernel time.   But I I don't think it can.

mrowan137:
Certainly, as you have said, due to the context switching (does @maxpkatz have more insight?), what is measured by the difference of `clock64` can only be an upper bound for the number of active thread cycles.  Timer is probably a bad choice for this object's name (maybe `DeviceCycleCounter` is more accurate?), but I have still found this metric useful in comparing relative cost of different MFIter loops (the usage I had in mind for this 'timer' is providing some way to estimate the relative cost of kernel work, rather than precision measurement of the actual time).  I am encouraged that this is a useful metric by the agreement I have in realistic use cases in WarpX, where I see excellent agreement in the measured costs (normalized) when measuring either with CUPTI, in-gpu `clock64`, or measuring the number of particles:
Screen Shot 2020-11-16 at 3 40 42 PM

If documentation is added to caution that this should only be used in the case you said (kernel is order of magnitude more expensive than the timer related cost), I think this metric can still be a useful addition.

On the other hand, if this metric is not general enough and could cause confusion for other AMReX-based codes, maybe it makes sense to keep this on the WarpX side to prevent potential confusion, let me know what you think.

(from NVidia documentation:
> Sampling this counter at the beginning and at the end of a kernel, taking the difference of the two samples, and recording the result per thread provides a measure for each thread of the number of clock cycles taken by the device to completely execute the thread, but not of the number of clock cycles the device actually spent executing thread instructions. The former number is greater than the latter since threads are time sliced.

)

maximumcats:
@mrowan137, before we get into too much detail about this, I wonder if the use case in question isn't already covered by [the existing TinyProfiler functionality](https://amrex-codes.github.io/amrex/docs_html/AMReX_Profiling_Tools_Chapter.html)? If you do, e.g.,

```
{
  BL_PROFILE("myRegion");

  for (MFIter(mfi); mfi.isValid(); ++mfi) {
    // do work here
  }
}
```

then this will give you a host-side timer that shows you how much time is spent in this MFIter loop. (There is a device synchronize in the MFIter constructor.)

If you're interested in individual loops with an MFIter iteration, you can do, e.g.,

```
for (MFIter(mfi); mfi.isValid(); ++mfi) {
  {
    BL_PROFILE("myRegion1");
    // do work here
  }
  {
    BL_PROFILE("myRegion2");
    // do more work here
  }
}
```
and then, in order to accurately capture the kernel time, you can use the recently added option `tiny_profiler.device_synchronize_around_region=1` to make those timing regions include the full kernel time.

It's very likely you already know about/are using this functionality, I would just like to understand what it doesn't already solve that having the device-side timer would solve.

mrowan137:
Thanks @maxpkatz , I think the use-case I have in mind is more to have on-the-fly measurement of kernel work, without necessarily enforcing a synchronize at each iteration of the MFIter.  But I will need to do some testing, if there is synchronize at the end of each iteration we can just measure with normal CPU timer and get on-the-fly timing.  If the synchronization happens at relatively large intervals, it is probably OK in my use case to use the synchronization and CPU timer.  But if the synchronization happens frequently, I am wondering if the performance will be better to use device-side timing.

WeiqunZhang:
Maybe it's simply the true cost is roughly proportional to the number of cells (or particles), which is the number of timers, and the measured cost (dominated by the timer) is also roughly proportional to the number of timers.  Note that most of our kernels are memory bandwidth bounded.  So the number of cells (or particles) is probably a pretty good measure of the cost.

WeiqunZhang:
It's hard for me to imagine that device side timing involving reduction could be faster than host side timing with synchronization.

mrowan137:
> Maybe it's simply the true cost is roughly proportional to the number of cells (or particles), which is the number of timers,

That's a good point.  It could be tested (correct me if I am wrong) with a uniform box but adding a velocity profile (so that some particles deposit over many cells, and others completely at rest).  in this case, the cost should look uniform if your suggestion is true.

maximumcats:
> But if the synchronization happens frequently, I am wondering if the performance will be better to use device-side timing.

That's a fair concern. If your kernels are relatively long running (say 1 ms or longer) then regular synchronizations will be negligible. It matters more for shorter kernels, but at any rate I imagined you'd only use the `tiny_profiler.device_synchronize_around_region=1` option for development work and then turn it off for production runs.

WeiqunZhang:
Is the difference between `clock64`'s return values from two different kernels meaningful?  If so, one might be able to use two single thread kernels around a kernel to do the measurement without synchronization.  But it's not clear if this would work well with multiple streams.

maximumcats:
It is meaningful (though it's a 64-bit counter and when it wraps, it wraps around silently). There's also the [%globaltimer PTX register](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#special-registers-globaltimer) if you want an actual wall-time (which NVIDIA does not really advertise for end-user consumption, hence the warning there).

mrowan137:
> Maybe it's simply the true cost is roughly proportional to the number of cells (or particles), which is the number of timers, and the measured cost (dominated by the timer) is also roughly proportional to the number of timers.

I did some quick test to check that if there is a step function for velocity but uniform density, the difference in `clock64`-measured difference between the two sides of the box is at about the level expected when comparing the time spent in current deposition for a uniform box with at-rest particles, and a uniform box with moving particles; the `clock64` timing is sensitive enough to capture the difference (though even with quite extreme choice for the velocity, c~0.99 this difference is admittedly small, at the level of 12%; the difference measured with the GPU clock timer is around 5%, but it seems to be more than collinear with the number of particles, as the 5% measurement is repeatable).  This suggests to me that the GPU clock timer is more than a simple proxy for the number of particles.

Screen Shot 2020-11-16 at 6 49 49 PM

mrowan137:
Here are some tests to compare the overhead of different cost-assessment strategies, it seems that synchronization is typically best.  It is simplest and also gives meaningful times, so for other codes that would want to get box times this may be easiest.

Screen Shot 2020-12-03 at 11 01 15 AM

To reiterate, these tests have a simple ParallelFor:
```
            amrex::ParallelFor(bx,
                               [=] AMREX_GPU_DEVICE (int i, int j, int k)
                               {
                                   fab(i,j,k) += 1.;
                               });
```

PR #1542, opened 2020-11-14
fcompare: fix logic error when comparing both rel/abs tolerances

sayerhs:
PR #1537 introduced the option to allow tolerance checks on both absolute and
relative tolerances. However, the check used `and` instead of `or` to allow
tests to pass when either absolute or relative error was below user-specified
tolerance.

## Checklist

The proposed changes:


WeiqunZhang:
Then the next test on relative error is no longer needed.

WeiqunZhang (approved):

PR #1544, opened 2020-11-15
Non-local boundary conditions

WeiqunZhang:
## Summary

* Rotate90 fills the lo-x and lo-y boundary regions by rotating the data
  around (x=0,y=0) by 90 degrees in either direction.  It also fills the
  corner of lo-x and lo-y boundary region by rotating the data by 180
  degrees.

* Rotate180 fills the lo-x boundary by rotating the data around
  (x=0,y=L_y/2) by 180 degrees.

* Fill the polar boundaries of the spherical coordinates (theta, phi, r).
  The lo-x boundary is filled with `f(-x,y) = f(x,mod(y+pi,2*pi))`, and
  the hi-x boundary is filled with `f(pi+x,y) = f(pi-x,mod(y+pi,2*pi))`.

## Additional background

#1538 

## Checklist

The proposed changes:


atmyers:
What do you think about adding some CI tests for this that do things like:

1. Rotate by 90 clockwise, rotate by 90 counter-clockwise, assert unchanged. 
2. Rotate by 90 clockwise twice, rotate by 180, assert the same. 

etc...

WeiqunZhang:
Sure.  In fact, I do have similar tests at https://github.com/WeiqunZhang/nonlocalbc-tests.  I can put them into amrex.  But the functions do not rotate valid cells.  It only touches ghost cells.  So what you are proposing cannot be done exactly.

atmyers:
Yeah, I realized after I wrote that that those exact tests wouldn't work. But there ought to be some symmetry tests that could be done. 

Would you rather merge this and add the tests or add them to this PR?

WeiqunZhang:
Sure, we can do that in a later PR.  There are quite some mesh data tests we need to put into CI.

Note that I have done extensive regression tests on this PR on CPU and GPU using various AMReX codes.

atmyers (approved):

PR #1545, opened 2020-11-17
CMake: generate_buildinfo target

ax3l:
## Summary

This commit generates a static library `buildInfo::<target>` for the passed user target. This target is then automatically linked. Also, the generated `.cpp` file is now scoped properly, avoiding collisions.

This solves the following corner cases:
- workflows with OBJECT libraries, that cannot be passed as   targets since they have no `PRE_BUILD` custom command support
- workflows with multiple targets that are not linked together   (those call `generate_buildinfo` multiple times)
- workflows with multiple targets that are linked together (those call `generate_buildinfo` once and re-use the `buildInfo::<target>`)

Thx to @mic84 for brain-storming & support.

## Checklist

The proposed changes:

### DIFF
--- a/Tools/CMake/AMReXBuildInfo.cmake
+++ b/Tools/CMake/AMReXBuildInfo.cmake
@@ -187,28 +187,34 @@ function (generate_buildinfo _target _git_dir)
 
    # Generate AMReX_buildInfo.cpp
    configure_file( ${AMREX_BUILDINFO_IFILE}
-      ${CMAKE_CURRENT_BINARY_DIR}/AMReX_buildInfo.cpp @ONLY)
+      ${PROJECT_BINARY_DIR}/${_target}/AMReX_buildInfo.cpp @ONLY)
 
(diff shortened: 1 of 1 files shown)


Tools/CMake/AMReXBuildInfo.cmake
```diff
@@ -187,28 +187,34 @@ function (generate_buildinfo _target _git_dir)
 
    # Generate AMReX_buildInfo.cpp
    configure_file( ${AMREX_BUILDINFO_IFILE}
-      ${CMAKE_CURRENT_BINARY_DIR}/AMReX_buildInfo.cpp @ONLY)
+      ${PROJECT_BINARY_DIR}/${_target}/AMReX_buildInfo.cpp @ONLY)
 
-   target_sources( ${_target}
+   # add a new target
```
ax3l:
```suggestion
   # add a re-usable target
```

mic84 (approved):

PR #1546, opened 2020-11-17
Fix BaseFab::maxabs

WeiqunZhang:
## Summary

For the host version, call host function LoopOnCpu instead of host device
function Loop because the function we pass to it is a host function.

## Checklist

The proposed changes:


zingale:
this fixed MAESTROeX compilation

zingale (approved):

PR #1547, opened 2020-11-17
Revert "CMake: generate_buildinfo target"

ax3l:
Reverts AMReX-Codes/amrex#1545

causes regressions in mfix and amr-wind


ax3l:
@sayerhs on Slack:
> I think I’ve figured out what the issue is… so I think I can handle this at my end. :slightly_smiling_face:
> [Revert] No don’t do that… I will fix this at my end.

@mic84 takes on the regression in mfix :bow:

PR #1549, opened 2020-11-17
Enable FFTW Compilation on corigpu

ajnonaka:
## Summary
add CXXFLAGS += $(FFTW) to Make.nersc to enable fftw compilation on cori


WeiqunZhang:
Is `$FFTW` something defined by you or a NERSC module?  I think it's better you add `XTRA_CXXFLAGS=$(FFTW)` in the makefile of your code.

ajnonaka:
I think adding this line in harmless in general, but you still have to 'module include fftw'.  If I can get this to work with a local makefile modification I'll close this PR.

PR #1550, opened 2020-11-17
CMake: Fix Single-Precision Switch

ax3l:
## Summary

Since the latest refactoring, single-recision builds were not activated. This fixes it.

## Additional background

Regression to #1490

Issue reported by @SeverinDiederichs and @MaxThevenet :tada: 

## Checklist

The proposed changes:


mic84:
That's weird. It does work for me without the fix...

ax3l:
It would work for us if we wouldn't use particles.

Anyway, technically, strings must be quoted in CMake.

ax3l:
The reason why it works for fields is that `AMReX_Config.H.in` has a line
```
#cmakedefine AMREX_USE_FLOAT
```

Should this file also contain an `AMREX_SINGLE_PRECISION_PARTICLES` entry?

mic84:
Yes, ``AMReX_Config.H.in`` should contain an ``AMREX_SINGLE_PRECISION_PARTICLES`` entry.
Could you go ahead and add it? Thanks!

ax3l:
I looked at the files
* `Tools/CMake/AMReXInstallHelpers.cmake`
* `Tools/CMake/AMReX_Config.H.in`

and I don't understand how the line
```cmake
#cmakedefine AMREX_USE_FLOAT
```
currently works: I think we don't have a _CMake_ variable `AMREX_USE_FLOAT` that could control it at the moment.

mic84:
``generate_amrex_config_header()`` retrieves the amrex definitions and defines CMake variables with the same name, so that  the ``#cmakedefine`` entries in the configure file can be properly expanded.

An an example, consider ``AMREX_USE_FLOAT``. When ``AMReX_PRECISION=SINGLE``, a macro ``AMREX_USE_FLOAT`` is added to the amrex compile definitions. ``generate_amrex_config_header()`` will then retrieve those compile definitions, and do a ``set(<macro-name> " ")`` for each macro in the compile definition set. That means it will also do a ``set(AMREX_USE_FLOAT " ")``. Since the variable ``AMREX_USE_FLOAT`` is now defined, ``configure_file()`` will replace ``#cmakedefine AMREX_USE_FLOAT`` with ``#define AMREX_USE_FLOAT``.

So the real issue here is not the absence of quotes around a string, but the absence of ``#cmakedefine  AMREX_SINGLE_PRECISION_PARTICLES`` in ``AMReX_Config.H.in``

ax3l:
Thanks for the details - intricate :)

The config header is now updated as well.

mic84 (approved):

PR #1555, opened 2020-11-18
CMake: AMReX_MPI options is dependent on SYCL, not HIP

mic84:
## Summary
Option AMReX_MPI should be available only if AMReX_GPU_BACKEND is not set to SYCL.

## Checklist

The proposed changes:


ax3l (approved):
LGTM, although I would allow AMReX_MPI with SYCL already - we could throw a warning if it us not yet supported/tested.

PR #1557, opened 2020-11-18
allow MLEBABecLap to solve for multiple components

rporcu:
## Summary
MLLinOp::getNComp()  is returning 1 and right now MLEBABecLap is working only with MultiFabs having 1 component.

## Additional background
This PR has the goal to generalize MLEBABecLap solver for a generic N number of components.

## Checklist

The proposed changes:

### DIFF
--- a/Src/LinearSolvers/MLMG/AMReX_MLEBABecLap.H
+++ b/Src/LinearSolvers/MLMG/AMReX_MLEBABecLap.H
@@ -20,7 +20,8 @@ public:
                  const Vector<BoxArray>& a_grids,
                  const Vector<DistributionMapping>& a_dmap,
                  const LPInfo& a_info,
-                 const Vector<EBFArrayBoxFactory const*>& a_factory);
+                 const Vector<EBFArrayBoxFactory const*>& a_factory,
+                 const int ncomp = 1);
 
     virtual ~MLEBABecLap ();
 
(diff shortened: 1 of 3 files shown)


Src/LinearSolvers/MLMG/AMReX_MLLinOp.H
```diff
@@ -178,7 +178,8 @@ public:
     bool getEnforceSingularSolvable () const noexcept { return enforceSingularSolvable; }
 
     virtual BottomSolver getDefaultBottomSolver () const { return BottomSolver::bicgstab; }
-    virtual int getNComp () const { return 1; }
+    virtual void setNComp (int ncomp) noexcept { m_ncomp = ncomp; }
+    virtual int getNComp () const { return m_ncomp; }
```
WeiqunZhang: There is no need to add `int m_ncomp` because this is a virtual function.  This allows MLEBABecLap to override without changing MLLinOp.

rporcu: thank you, Ill revert these changes and modify only MLEBABecLap

rporcu (commented):
The suggested changes have been applied

Src/LinearSolvers/MLMG/AMReX_MLEBABecLap.H
```diff
@@ -59,6 +60,8 @@ public:
     void setEBHomogDirichlet (int amrlev,                      Real beta);
     void setEBHomogDirichlet (int amrlev,                      Vector<Real> const& beta);
 
+    virtual int getNComp () const { return m_ncomp; }
```
WeiqunZhang:
```suggestion
    virtual int getNComp () const override { return m_ncomp; }
```

WeiqunZhang (approved):

PR #1558, opened 2020-11-18
[WIP] CMake: unify on-node acceleration backend under AMReX_COMPUTE

mic84:
## Checklist

The proposed changes:

### DIFF
--- a/Tools/CMake/AMReXOptions.cmake
+++ b/Tools/CMake/AMReXOptions.cmake
@@ -96,48 +96,62 @@ message( STATUS "   AMReX_PRECISION = ${AMReX_PRECISION}")
 
 
 #
-# GPU backends    =============================================================
+# On-node parallel backends    =============================================================
 #
-set(AMReX_GPU_BACKEND_VALUES NONE SYCL CUDA HIP)
-set(AMReX_GPU_BACKEND NONE CACHE STRING "On-node, accelerated GPU backend: <NONE,SYCL,CUDA,HIP>")
-set_property(CACHE AMReX_GPU_BACKEND PROPERTY STRINGS ${AMReX_GPU_BACKEND_VALUES})
-if (NOT AMReX_GPU_BACKEND IN_LIST AMReX_GPU_BACKEND_VALUES)
-   message(FATAL_ERROR "AMReX_GPU_BACKEND=${AMReX_GPU_BACKEND} is not allowed."
-      " Must be one of ${AMReX_GPU_BACKEND_VALUES}")
+set(AMReX_COMPUTE_VALUES NONE OMP SYCL CUDA HIP)
+set(AMReX_COMPUTE NONE CACHE STRING "On-node, accelerated computing backend: <NONE,OMP,SYCL,CUDA,HIP>")
+set_property(CACHE AMReX_COMPUTE PROPERTY STRINGS ${AMReX_COMPUTE_VALUES})
+if (NOT AMReX_COMPUTE IN_LIST AMReX_COMPUTE_VALUES)
+   message(FATAL_ERROR "AMReX_COMPUTE=${AMReX_COMPUTE} is not allowed."
+      " Must be one of ${AMReX_COMPUTE_VALUES}")
 endif ()
 
-if (NOT AMReX_GPU_BACKEND STREQUAL NONE)
-   message( STATUS "   AMReX_GPU_BACKEND = ${AMReX_GPU_BACKEND}")
+if (NOT AMReX_COMPUTE STREQUAL NONE)
+   message( STATUS "   AMReX_COMPUTE = ${AMReX_COMPUTE}")
 endif ()
 
-if (AMReX_GPU_BACKEND STREQUAL SYCL)
-   set(AMReX_CUDA  OFF CACHE INTERNAL "")
-   set(AMReX_DPCPP ON  CACHE INTERNAL "")
-   set(AMReX_HIP   OFF CACHE INTERNAL "")
-elseif (AMReX_GPU_BACKEND STREQUAL CUDA)
-   set(AMReX_CUDA  ON  CACHE INTERNAL "")
-   set(AMReX_DPCPP OFF CACHE INTERNAL "")
-   set(AMReX_HIP   OFF CACHE INTERNAL "")
-elseif (AMReX_GPU_BACKEND STREQUAL HIP)
-   set(AMReX_CUDA  OFF CACHE INTERNAL "")
-   set(AMReX_DPCPP OFF CACHE INTERNAL "")
-   set(AMReX_HIP   ON  CACHE INTERNAL "")
+# Set legacy variables for internal use only
+if (AMReX_COMPUTE STREQUAL "OMP")
+   set(AMReX_OMP   ON )
+   set(AMReX_DPCPP OFF)
+   set(AMReX_CUDA  OFF)
+   set(AMReX_HIP   OFF)
+elseif (AMReX_COMPUTE STREQUAL "SYCL")
+   set(AMReX_OMP   OFF)
+   set(AMReX_DPCPP ON )
+   set(AMReX_CUDA  OFF)
+   set(AMReX_HIP   OFF)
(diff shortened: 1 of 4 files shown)


Tools/CMake/AMReXOptions.cmake
```diff
@@ -96,48 +96,62 @@ message( STATUS "   AMReX_PRECISION = ${AMReX_PRECISION}")
...
+   set(AMReX_DPCPP OFF)
+   set(AMReX_CUDA  OFF)
+   set(AMReX_HIP   OFF)
+elseif (AMReX_COMPUTE STREQUAL "SYCL")
+   set(AMReX_OMP   OFF)
+   set(AMReX_DPCPP ON )
+   set(AMReX_CUDA  OFF)
+   set(AMReX_HIP   OFF)
+elseif (AMReX_COMPUTE STREQUAL "CUDA")
+   set(AMReX_OMP   OFF)
```
sayerhs, line 125: Is there a reason to disable OpenMP when CUDA is active? I thought OpenMP and CUDA could coexist.

WeiqunZhang: Yes, they could coexist.

ax3l: I think there are a few orthogonal aspects that we could declutter:
- the acceleration backend, i.e. MFIter loops & ParallelFor/Reduce, that are "OpenMP" accelerated
- tiling, which is currently only enabled with OpenMP
- general usage of OpenMP elsewhere in the code

Currently, we often mix those aspects. For example, we should replace `#ifdef _OPENMP` usage before MFIter loops (at least in WarpX), with something like `#ifdef AMREX_OMP`. We could add an additional variable that controls tiling that is default "on" only for OMP.

That way, we can use OpenMP in codes in auxiliary functionality (and dependencies like linear algebra/FFT/IO) with CUDA accelerated AMReX primitives, can enable tiling explicitly with GPU when an app supports it, and can disable the latter where it is not supported (i.e. WarpX does not support tiling with CUDA atm.).

What do you think and what's your current and planned usage?

maximumcats: Agreed with Weiqun -- the main reason for the status quo is likely just to prevent people from shooting themselves in the foot (tiling and OpenMP threading of kernel launches are usually not a good idea for MFIter loops in launch regions).

sayerhs: @ax3l To make sure I understand the points you make 

```C++
#ifdef _OPENMP
#pragma omp parallel if (amrex::Gpu::notInLaunchRegion())
#endif
for (amrex::MFIter mfi(mfab, amrex::TilingIfNotGPU()); mfi.isValid(); ++mfi) {
    const auto& bx = mfi.tilebox();
    amrex::ParallelFor(bx, AMREX_GPU_DEVICE[=](int i, int j, int k) { /* ops */ }
}
```

It would be great if we had something like `AMREX_MFITER_LOOP(mfi, mfab)` that handled the boilerplate 

```C++
#ifdef _OPENMP
#pragma omp parallel if (amrex::Gpu::notInLaunchRegion())
#endif
for (amrex::MFIter mfi(mfab, amrex::TilingIfNotGPU()); mfi.isValid(); ++mfi)
```

The other things are already handled by `amrex::TilingIfNotGPU()` correct? 

We will want `-DAMReX_COMPUTE=OMP;CUDA` so that users can select both OMP and CUDA if needed.

WeiqunZhang: Yes, we should give the choice to the users.  In AMReX, we have used `omp parallel if (amrex::Gpu::notInLaunchRegion())` to make sure OMP and CUDA can coexist.  If an application code does not need to have both, it can disable OMP in their own build system when CUDA is on.  If an application likes to have both, it's then their responsibility to make sure to use `amrex::Gpu::notInLaunchRegion()` to disable OMP when necessary.

Currently, `mfi(mf,true)` will not start tiling if GPU is on.  To turn on tiling with GPU, one would need to use `mfi(mf, IntVect_as_tiling_size)`.

Tiling and OMP are orthogonal.  Even without OMP, tiling is usually a good idea on CPU because of cache blocking.

WeiqunZhang: > It would be great if we had something like `AMREX_MFITER_LOOP(mfi, mfab)` that handled the boilerplate
> 
> ```c++
> #ifdef _OPENMP
> #pragma omp parallel if (amrex::Gpu::notInLaunchRegion())
> #endif
> for (amrex::MFIter mfi(mfab, amrex::TilingIfNotGPU()); mfi.isValid(); ++mfi)
> ```

Adding a macro such as `AMREX_OMP_PARALLEL_IF_NOT_GPU` for the first three lines is fine.  But I don't think we should let a macro to expand to `for (amrex::MFIter mfi(mfab, amrex::TilingIfNotGPU()); mfi.isValid(); ++mfi)` that turns on tiling by default for a non-GPU build.

ax3l: What I mean for the sample above is, that we should define our own macro:
```C++
#ifdef AMREX_OMP
#pragma omp parallel if (amrex::Gpu::notInLaunchRegion())
#endif
```
so this code region is independent of me pulling OpenMP for some other, orthogonal functionality in the software stack. (We currently have the legacy `BL_USE_OMP` that could be used for the same purpose).


The runtime check inside does already serve the purpose of safe-guarding for GPU, but what can happen is he following: a user selects "serial" execution and a dependency publicly pulls OpenMP (say, a parallel, header-only compression library or so) - we then will still OpenMP parallelize AMReX loops, which can be confusing.

ax3l: Update: dang, I just saw we define `-DAMREX_USE_OMP`, all good :)

WeiqunZhang: But I agree with you that inside AMReX we should use `#ifdef AMREX_USE_OMP` instead of `#ifdef _OPENMP`.  It doesn't make any difference for a super build.  But if AMReX is built into a library, using `_OPENMP` might result in codes in `AMReX_*.cpp` are not OMP parallelized, but those in `AMReX_*.H` are. Or the other way around.  That could be confusing.

maximumcats: I have independently had this thought several times that we should use `#ifdef AMREX_USE_OMP` instead of `#ifdef _OPENMP`. Big +1 from me.

ax3l: Ok, I'll push a PR to both WarpX and AMReX on this.

Back to the topic of this PR: we probably leave it as currently is for now then? More fine grained with multiple vars gives more control then and we do have some form of tiling control already via `amrex::TilingIfNotGPU()` that can be quickly changed by developers that explore it.

WeiqunZhang: I think we should allow OMP on when GPU is on.

As for using `AMREX_USE_OMP`,  I wonder if we should add two macros
```
#if defined(AMREX_USE_OMP)
#define AMREX_OMP_PARALLEL _Pragma("omp parallel)")
#else
#define AMREX_OMP_PARALLEL
#endif

#if defined(AMREX_USE_OMP)
#define AMREX_OMP_PARALLEL_IF_NOT_GPU _Pragma("omp parallel if (amrex::Gpu::notInLaunchRegion())")
#else
#define AMREX_OMP_PARALLEL_IF_NOT_GPU
#endif
```


What do people think?

ax3l: No strong opinion on it. Pro: nicely compact (savings of 300%!) - Con: more macro indirections.

ax3l: After doing a self-review on changes in WarpX, I think such a macro might indeed help to make intent more clear :D

maximumcats: @drummerdoc had thoughts on this right? Personally I would not use such a macro and would prefer for it not to be used in AMReX source, but of course downstream codes can do what they like and if they want to do this I don't see why they couldn't.

PR #1559, opened 2020-11-18
Remove SUNDIALS support

mic84:
## Summary
SUNDIALS support in AMReX consisted of only few Fortran interfaces to SUNDIALS functions for the convenience of app codes. Since there are no AMReX functionalities depending on SUNDIALS, support for SUNDIALS will be removed.

## Checklist

The proposed changes:


asalmgren (approved):

WeiqunZhang:
There are still some SUNDIALS and CVODE stuffs left.  Could you do a `git grep -i -E "cvode|sundials"`.

mic84:
@WeiqunZhang the only occurrences of "sundials" or "cvode" I can see are from the Doc. We still need to decide what to do with it, thus I haven't removed those parts yet.

WeiqunZhang:
That's right.  I noticed that the tutorials are gone, but the doc referring to the tutorials are still there.  That's why I commented.

Also I think `AMReX_SUNDIALS` appeared twice in `BuildingAMReX.rst`, one was removed, the other was kept.

jmsexton03 (approved):

WeiqunZhang (approved):

PR #1560, opened 2020-11-18
Define: _OPENMP -> AMREX_USE_OMP

ax3l:
## Summary

Replace the define `_OPENMP` with `AMREX_USE_OMP` for all parallel "backend" implementations and control of MFIter loops.

This avoids accidentially enabling OpenMP for the parallel compute components, i.e. when a users explicitly uses OpenMP in auxiliary or implicit dependent functionality, but does not request to also OpenMP-parallelize those sections.

- [x] search & replace
- [x] self-review and review

## Additional background

Ref.: https://github.com/AMReX-Codes/amrex/pull/1558#discussion_r526466768

## Checklist

The proposed changes:

### DIFF
--- a/Docs/sphinx_documentation/source/Basics.rst
+++ b/Docs/sphinx_documentation/source/Basics.rst
@@ -1786,7 +1786,7 @@ parallel region:
   // Dynamic tiling, one box per OpenMP thread.
   // No further tiling details,
   //   so each thread works on a single tilebox.
-  #ifdef _OPENMP
+  #ifdef AMREX_USE_OMP
   #pragma omp parallel
   #endif
       for (MFIter mfi(mf,MFItInfo().SetDynamic(true)); mfi.isValid(); ++mfi)
@@ -1803,7 +1803,7 @@ Dynamic tiling also allows explicit definition of a tile size:
 
   // Dynamic tiling, one box per OpenMP thread.
   // No tiling in x-direction. Tile size is 16 for y and 32 for z.
-  #ifdef _OPENMP
+  #ifdef AMREX_USE_OMP
   #pragma omp parallel
   #endif
       for (MFIter mfi(mf,MFItInfo().SetDynamic(true).EnableTiling(1024000,16,32)); mfi.isValid(); ++mfi)
@@ -2232,7 +2232,7 @@ like below,
 
 ::
 
-  #ifdef _OPENMP
+  #ifdef AMREX_USE_OMP
   #pragma omp parallel if (Gpu::notInLaunchRegion())
   #endif
     for (MFIter mfi(mfa,TilingIfNotGPU()); mfi.isValid(); ++mfi)
--- a/Docs/sphinx_documentation/source/GPU.rst
+++ b/Docs/sphinx_documentation/source/GPU.rst
@@ -1240,7 +1240,7 @@ off using the conditional pragma and :cpp:`Gpu::notInLaunchRegion()`, as shown b
 
 ::
 
-    #ifdef _OPENMP
+    #ifdef AMREX_USE_OMP
     #pragma omp parallel if (Gpu::notInLaunchRegion())
     #endif
 
@@ -1262,7 +1262,7 @@ of common AMReX patterns, such as the one below:
 ::
 
    // Given MultiFab uin and uout
-   #ifdef _OPENMP
+   #ifdef AMREX_USE_OMP
    #pragma omp parallel
    #endif
    {
@@ -1309,7 +1309,7 @@ portable way.
 ::
 
    // Given MultiFab uin and uout
-   #ifdef _OPENMP
+   #ifdef AMREX_USE_OMP
    #pragma omp parallel if (Gpu::notInLaunchRegion())
    #endif
    {
--- a/Src/Base/AMReX.cpp
+++ b/Src/Base/AMReX.cpp
@@ -47,7 +47,7 @@
 #include <AMReX_MemProfiler.H>
 #endif
 
-#ifdef _OPENMP
+#ifdef AMREX_USE_OMP
 #include <omp.h>
 #endif
 
@@ -209,7 +209,7 @@ amrex::Error_host (const char * msg)
     } else {
         write_lib_id("Error");
         write_to_stderr_without_buffering(msg);
-#ifdef _OPENMP
+#ifdef AMREX_USE_OMP
 #pragma omp critical (amrex_abort_omp_critical)
 #endif
         ParallelDescriptor::Abort();
@@ -257,7 +257,7 @@ amrex::Abort_host (const char * msg)
     } else {
        write_lib_id("Abort");
        write_to_stderr_without_buffering(msg);
-#ifdef _OPENMP
+#ifdef AMREX_USE_OMP
 #pragma omp critical (amrex_abort_omp_critical)
 #endif
        ParallelDescriptor::Abort();
@@ -306,7 +306,7 @@ amrex::Assert_host (const char* EX, const char* file, int line, const char* msg)
         throw RuntimeError(buf);
     } else {
        write_to_stderr_without_buffering(buf);
-#ifdef _OPENMP
+#ifdef AMREX_USE_OMP
 #pragma omp critical (amrex_abort_omp_critical)
 #endif
        ParallelDescriptor::Abort();
@@ -592,7 +592,7 @@ amrex::Initialize (int& argc, char**& argv, bool build_parm_parse,
         amrex::Print() << "MPI initialized with thread support level " << provided << std::endl;
 #endif
         
-#ifdef _OPENMP
+#ifdef AMREX_USE_OMP
 //    static_assert(_OPENMP >= 201107, "OpenMP >= 3.1 is required.");
         amrex::Print() << "OMP initialized with "
                        << omp_get_max_threads()
@@ -660,7 +660,7 @@ amrex::Finalize (amrex::AMReX* pamrex)
 	if (ParallelDescriptor::NProcs() == 1) {
 	    if (mp_tot > 0) {
                 amrex::Print() << "MemPool: " 
-#ifdef _OPENMP
+#ifdef AMREX_USE_OMP
                                << "min used in a thread: " << mp_min << " MB, "
                                << "max used in a thread: " << mp_max << " MB, "
 #endif
--- a/Src/Base/AMReX_Extension.H
+++ b/Src/Base/AMReX_Extension.H
@@ -49,7 +49,7 @@
 #elif defined(__HIP_DEVICE_COMPILE__)
 #define AMREX_PRAGMA_SIMD 
 
-//#elif defined(_OPENMP) && (_OPENMP >= 201307) && !defined(__PGI)
+//#elif defined(AMREX_USE_OMP) && defined(_OPENMP) && (_OPENMP >= 201307) && !defined(__PGI)
 //#define AMREX_PRAGMA_SIMD _Pragma("omp simd")
 
 #elif defined(__INTEL_COMPILER)
--- a/Src/Base/AMReX_FabArrayUtility.H
+++ b/Src/Base/AMReX_FabArrayUtility.H
@@ -1808,7 +1808,7 @@ indexFromValue (FabArray<FAB> const& mf, int comp, IntVect const& nghost,
 
             if (priv_loc.allGT(IntVect::TheMinVector())) {
                 bool old;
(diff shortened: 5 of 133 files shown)


Src/Base/AMReX.cpp
```diff
@@ -556,8 +556,8 @@ amrex::Initialize (int& argc, char**& argv, bool build_parm_parse,
         amrex::Print() << "MPI initialized with thread support level " << provided << std::endl;
 #endif
         
-#ifdef _OPENMP
-//    static_assert(_OPENMP >= 201107, "OpenMP >= 3.1 is required.");
+#ifdef AMREX_USE_OMP
+//    static_assert(AMREX_USE_OMP >= 201107, "OpenMP >= 3.1 is required.");
```
WeiqunZhang: This should use `_OPENMP`.

Src/Base/AMReX_Extension.H
```diff
@@ -48,7 +48,7 @@
 #elif defined(__HIP_DEVICE_COMPILE__)
 #define AMREX_PRAGMA_SIMD 
 
-//#elif defined(_OPENMP) && (_OPENMP >= 201307) && !defined(__PGI)
+//#elif defined(AMREX_USE_OMP) && (AMREX_USE_OMP >= 201307) && !defined(__PGI)
```
WeiqunZhang:
```suggestion
//#elif defined(AMREX_USE_OMP) && defined(_OPENMP) && (_OPENMP >= 201307) && !defined(__PGI)
```
Although this is commented out.

Src/Base/AMReX_FabArrayUtility.H
```diff
@@ -1806,11 +1806,11 @@ indexFromValue (FabArray<FAB> const& mf, int comp, IntVect const& nghost,
 
             if (priv_loc.allGT(IntVect::TheMinVector())) {
                 bool old;
-// we should be able to test on _OPENMP < 201107 for capture (version 3.1)
+// we should be able to test on AMREX_USE_OMP < 201107 for capture (version 3.1)
 // but we must work around a bug in gcc < 4.9
-#if defined(_OPENMP) && _OPENMP < 201307 // OpenMP 4.0
+#if defined(AMREX_USE_OMP) && AMREX_USE_OMP < 201307 // OpenMP 4.0
```
WeiqunZhang:
```suggestion
#if defined(AMREX_USE_OMP) && defined(_OPENMP) && _OPENMP < 201307 // OpenMP 4.0
```

Src/Base/AMReX_OpenMP.H
```diff
@@ -1,7 +1,7 @@
-#ifndef AMREX_OPENMP_H_
-#define AMREX_OPENMP_H_
+#ifndef AMREXAMREX_USE_OMP_H_
+#define AMREXAMREX_USE_OMP_H_
```
WeiqunZhang: I guess, it's fine because it's consistent.

Src/Base/AMReX_omp_mod.F90
```diff
@@ -4,8 +4,8 @@ module amrex_omp_module
 
   implicit none
 
-#ifdef _OPENMP
-  integer, parameter :: amrex_omp_support = (_OPENMP)
+#ifdef AMREX_USE_OMP
+  integer, parameter :: amrex_omp_support = (AMREX_USE_OMP)
```
WeiqunZhang:
```suggestion
  integer, parameter :: amrex_omp_support = (_OPENMP)
```

We need the value of `_OPENMP`.

Src/Particle/AMReX_Particle.H
```diff
@@ -760,11 +760,11 @@ Long
 Particle<NReal, NInt>::NextID ()
 {
     Long next;
-// we should be able to test on _OPENMP < 201107 for capture (version 3.1)
+// we should be able to test on AMREX_USE_OMP < 201107 for capture (version 3.1)
 // but we must work around a bug in gcc < 4.9
-#if defined(_OPENMP) && _OPENMP < 201307
+#if defined(AMREX_USE_OMP) && AMREX_USE_OMP < 201307
```
WeiqunZhang:
```suggestion
#if defined(AMREX_USE_OMP) && defined(_OPENMP) && _OPENMP < 201307
```

Tools/libamrex/mkconfig.py
```diff
@@ -59,11 +59,11 @@ def doit(defines, undefines, comp, allow_diff_comp, use_omp):
         print("#endif")
 
     if use_omp == "TRUE":
-        print("#ifndef _OPENMP")
+        print("#ifndef _OPENMP")  # TODO: should this be AMREX_USE_OMP ?
```
WeiqunZhang: Maybe we can remove the checks???

Docs/sphinx_documentation/source/GPU.rst
```diff
@@ -1269,7 +1269,7 @@ of common AMReX patterns, such as the one below:
 ::
 
    // Given MultiFab uin and uout
-   #ifdef _OPENMP
+   #ifdef AMREX_USE_OMP
```
ax3l, line 1272: Should potentially stay `_OPENMP`?

Docs/sphinx_documentation/source/Basics.rst
```diff
@@ -1803,7 +1803,7 @@ Dynamic tiling also allows explicit definition of a tile size:
 
   // Dynamic tiling, one box per OpenMP thread.
   // No tiling in x-direction. Tile size is 16 for y and 32 for z.
-  #ifdef _OPENMP
+  #ifdef AMREX_USE_OMP
```
ax3l, line 1806: Potentially `_OPENMP` at this point (2x)?

Docs/sphinx_documentation/source/Basics.rst
```diff
@@ -1786,7 +1786,7 @@ parallel region:
   // Dynamic tiling, one box per OpenMP thread.
   // No further tiling details,
   //   so each thread works on a single tilebox.
-  #ifdef _OPENMP
+  #ifdef AMREX_USE_OMP
```
ax3l, line 1789: Potentially `_OPENMP` at this point (2x)?

Src/Base/AMReX_MFIter.H
```diff
@@ -11,7 +11,7 @@
 
 #include <AMReX_Gpu.H>
 
-#ifdef _OPENMP
+#ifdef AMREX_USE_OMP
```
ax3l, line 15: Just to make usage in complex user code easier, we might want to write
```suggestion
#if defined(AMREX_USE_OMP) || defined(_OPENMP)
```
?

Src/Base/AMReX.cpp
ax3l:
```suggestion
//    static_assert(_OPENMP >= 201107, "OpenMP >= 3.1 is required.");
```

Src/Base/AMReX_OpenMP.H
ax3l: Haha, but it's unnecessary :) Will keep like the file name.

ax3l:
```suggestion
#ifndef AMREX_OPENMP_H_
#define AMREX_OPENMP_H_
```

Src/Base/AMReX_omp_mod.F90
```diff
@@ -4,7 +4,7 @@ module amrex_omp_module
 
   implicit none
 
-#ifdef _OPENMP
+#ifdef AMREX_USE_OMP
   integer, parameter :: amrex_omp_support = (_OPENMP)
```
ax3l, line 8: Curious crash - seems `_OPENMP` is not defined even with `BL_USE_OMP`/`AMREX_USE_OMP` being defined.

I think we forget to pass OpenMP flags to the Fortran compiler in CMake:
```
$ gfortran ... -DAMREX_USE_OMP ... -DBL_USE_OMP -I/...  -O3 -DNDEBUG -J../mod_files -pthread -fimplicit-none \
  -c /home/runner/work/amrex/amrex/Src/Base/AMReX_omp_mod.F90 -o CMakeFiles/amrex.dir/Base/AMReX_omp_mod.F90.o
Src/Base/AMReX_omp_mod.F90:8:45:
   integer, parameter :: amrex_omp_support = (_OPENMP)
                                             1
Error: Syntax error in expression at (1)
```

WeiqunZhang: maybe `-fopenmp` is missing.

ax3l: Yes, it definitely is missing for the Fortran part, let's add it: #1563

WeiqunZhang:
@ax3l Do you need help finishing this?

ax3l:
@WeiqunZhang thank you, this PR slipped me a bit end of the year.
I rebased it now, yes review & help to finalize it is very welcome.

Src/Base/AMReX_FabArrayUtility.H
```diff
@@ -1808,11 +1808,11 @@ indexFromValue (FabArray<FAB> const& mf, int comp, IntVect const& nghost,
 
             if (priv_loc.allGT(IntVect::TheMinVector())) {
                 bool old;
-// we should be able to test on _OPENMP < 201107 for capture (version 3.1)
+// we should be able to test on AMREX_USE_OMP < 201107 for capture (version 3.1)
```
WeiqunZhang:
```suggestion
// we should be able to test on _OPENMP < 201107 for capture (version 3.1)
```

Src/Base/AMReX_MFIter.H
WeiqunZhang: `AMReX_Config.H` contains
```
#if defined(AMREX_USE_OMP) && !defined(_OPENMP)                                                                                                                                                            
#error "libamrex was built with OpenMP, so the downstream project must activate it, too"                                                                                                                   
#endif 
```
So we do not allow the downstream project to build without OpenMP if AMReX is built with OpenMP.

Src/Base/AMReX_NonLocalBCImpl.H
```diff
@@ -163,7 +163,7 @@ local_copy_cpu (FabArray<FAB>& mf, int scomp, int ncomp, FabArrayBase::CommMetaD
     auto const& LocTags = *(cmd.m_LocTags);
     int N_locs = LocTags.size();
     if (N_locs == 0) return;
-#ifdef _OPENMP
+#ifdef _OPENMP // Note: intentionally not AMREX_USE_OMP
```
WeiqunZhang: Why is `_OPENMP` used here?  I think the only complicated situation we can support is AMReX is built without OpenMP whereas the downstream projects is built with OpenMP.  In that case, I don't think we should enable OpenMP here because it's part of AMReX.  If we do and this template is instantiated in AMReX itself (although currently that's not the case for this template), we might end up with this template is instantiated in two different ways.  That might be a problem.

Src/Base/AMReX_Reduce.H
```diff
@@ -842,7 +842,7 @@ template <typename T, typename N, typename F,
 T Sum (N n, F&& f, T init_val = 0)
 {
     T r = init_val;
-#ifdef _OPENMP
+#ifdef _OPENMP // Note: intentionally not AMREX_USE_OMP
```
WeiqunZhang: The same argument for using `AMREX_USE_OMP` here as in `AMReX_NonLocalBCImpl.H`.

Src/Particle/AMReX_Particle.H
```diff
@@ -761,11 +761,11 @@ Long
 Particle<NReal, NInt>::NextID ()
 {
     Long next;
-// we should be able to test on _OPENMP < 201107 for capture (version 3.1)
+// we should be able to test on AMREX_USE_OMP < 201107 for capture (version 3.1)
```
WeiqunZhang:
```suggestion
// we should be able to test on _OPENMP < 201107 for capture (version 3.1)
```

Src/Base/AMReX_NonLocalBCImpl.H
ax3l: I think that sounds reasonable, yes.

I was thinking about a similar pattern downstream: in this case this is an auxiliary function/loop. That works well with e.g. CUDA being on. Will change to avoid confusion.

```suggestion
#ifdef AMREX_USE_OMP
```

Src/Base/AMReX_Reduce.H
ax3l:
```suggestion
#ifdef AMREX_USE_OMP
```

WeiqunZhang (approved):

PR #1562, opened 2020-11-19
CI: MSVC + OpenMP

ax3l:
## Summary

Try if we can compile with MSVC's OpenMP support on Windows.

## Additional background

Wow, OpenMP 2.0 support (`_OPENMP` is `200203`) but exposes itself as `2019` with `/openmp:experimental`, which in turn just adds some SIMD parts: https://docs.microsoft.com/en-us/cpp/build/reference/openmp-enable-openmp-2-0-support?view=msvc-160

## Checklist

The proposed changes:

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -15,3 +15,16 @@ jobs:
         cd build
         cmake .. -DCMAKE_BUILD_TYPE=Debug -DCMAKE_VERBOSE_MAKEFILE=ON -DAMReX_BUILD_TUTORIALS=ON -DAMReX_FORTRAN=OFF -DAMReX_MPI=OFF
         cmake --build . --config Debug
+
+  # Build libamrex and all tutorials
+  tutorials_omp:
+    name: MSVC C++17 w/ OMP w/o MPI w/o Fortran
+    runs-on: windows-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Build & Install
+      run: |
+        mkdir build
+        cd build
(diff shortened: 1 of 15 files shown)


ax3l:
lol, MSVC: no `omp atomic read` and `omp atomic` requires: https://docs.microsoft.com/en-us/cpp/error-messages/compiler-errors-2/compiler-error-c3048?view=msvc-160
Update: `&=` is also not supported.

ax3l:
Ah gosh, looks like they don't support OpenMP 3.1 over at MSVC:
```
D:\a\amrex\amrex\Src\Base\AMReX_FabArrayUtility.H(356,9): error C3036: 'min': invalid operator token in OpenMP 'reduction' clause [D:\a\amrex\amrex\build\Src\amrex.vcxproj]
```
https://docs.microsoft.com/en-us/cpp/parallel/openmp/openmp-in-visual-cpp?view=msvc-160
https://docs.microsoft.com/en-us/cpp/build/reference/openmp-enable-openmp-2-0-support?view=msvc-160

And `_OPENMP` is defined to `2019` for `/openmp:experimental` :rofl:

ax3l:
Although this works, I don't like to permanently support OpenMP 2.0 from 2002 in a modern code base.

I just learned that we can get OpenMP 4.5 support for free with Clang on Windows: https://clang.llvm.org/docs/MSVCCompatibility.html (also with ICC).
Clang serves the need of where I want to use this for: conda-forge continuous delivery to Windows users.

PR #1563, opened 2020-11-19
CMake: OpenMP for Fortran

ax3l:
## Summary

Fix missing activation.

## Additional background

Isolated from #1560

## Checklist

The proposed changes:

### DIFF
--- a/Tools/CMake/AMReXParallelBackends.cmake
+++ b/Tools/CMake/AMReXParallelBackends.cmake
@@ -31,7 +31,7 @@ endif ()
 #
 if (AMReX_OMP)
    set(_omp_comps CXX)
-   if (AMReX_FORTRAN_INTERFACES)
+   if (AMReX_FORTRAN_INTERFACES OR AMReX_FORTRAN)
       list(APPEND _omp_comps Fortran)
    endif ()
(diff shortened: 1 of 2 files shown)


Tools/CMake/AMReXParallelBackends.cmake
```diff
@@ -31,7 +31,7 @@ endif ()
 #
 if (AMReX_OMP)
    set(_omp_comps CXX)
-   if (AMReX_FORTRAN_INTERFACES)
+   if (AMReX_FORTRAN_INTERFACES OR AMReX_FORTRAN)
```
mic84:
```suggestion
   if (AMReX_FORTRAN)
```
Testing for ``AMReX_FORTRAN_INTERFACES`` is redundant if we also test for ``AMReX_FORTRAN``

mic84:
Can you also update ``AMReXConfig.cmake.in`` accordingly? Thanks!

ax3l:
Good catch - and I just see `set( _mpi_components C CXX )` needs the same patch.

mic84:
I am not sure it does. U

ax3l:
@mic84 did you accidentally close this instead of merging it?

mic84 (approved):

PR #1565, opened 2020-11-19
CMake: MPI for Fortran

ax3l:
## Summary

Fix missing activation. Same as #1563 but for MPI

## Checklist

The proposed changes:


ax3l:
@mic84 on slack:
> MPI for Fortran is needed ONLY if `AMReX_FORTRAN_INTERFACES` are enabled

PR #1570, opened 2020-11-25
Fused FillBoundary

WeiqunZhang:
## Summary

Add a new FillBoundary function for a Vector of FabArrays.  There are two
implementations in this PR.  The one being used simply calls
`FillBoundary_nowait` and `FillBoundary_finish`.  The other version
aggregates the MPI messages from multiple FabArrays into one.  Tests on
summit shows the latter is slower.  But I don't want to lose the code, so it
is kept for future investigation.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_Array4.H
+++ b/Src/Base/AMReX_Array4.H
@@ -59,6 +59,21 @@ namespace amrex {
               ncomp(rhs.ncomp-start_comp)
         {}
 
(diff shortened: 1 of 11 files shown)


kngott (approved):

kngott:
Agreed on saving the code. A lot of good MPI cleanup in here.

PR #1571, opened 2020-11-25
Fix some warnings related to double-->float conversions

lucafedeli88:
## Summary
I tried to compile  `WarpX` in single precision and I noticed several warnings related to casts from `double` to `float` in `AMReX`.
This PR should fix some of these warnings. I plan to continue in the next days with other PRs, if you agree.

I don't know if you have a strong preference between C-style `Real(1.0)` and the C++-style `static_cast<Real>(1.0)`

## Additional background

I compiled `WarpX` with `make DIM=2 QED=TRUE PRECISION=FLOAT USE_SINGLE_PRECISION_PARTICLES=TRUE`
using the compiler `g++ (Ubuntu 10.2.0-13ubuntu1) 10.2.0` 

## Checklist

The proposed changes:

### DIFF
--- a/Src/AmrCore/AMReX_AmrMesh.H
+++ b/Src/AmrCore/AMReX_AmrMesh.H
@@ -25,7 +25,7 @@ struct AmrInfo {
     // Buffer cells around each tagged cell.
     Vector<IntVect> n_error_buf {{IntVect(1)}};
     // Grid efficiency.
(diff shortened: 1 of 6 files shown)


WeiqunZhang:
Thanks!  As for the style, if the line is long, `Real()` might be better because it's shorter.

WeiqunZhang (approved):

PR #1580, opened 2020-11-30
eval_genex correctly replaces the matching compiler version

robertmaynard:
## Summary

Previously it would place the resulting output in the variable
`1` instead of `_in`. Additionally it would try to add an extra `>`.

## Checklist

The proposed changes:


robertmaynard:
Found when building on windows and the `/experimental:preprocessor` or `/Zc:preprocessor` selection was wrong in the CXX and CUDA state information that AMReX prints out during configuration.

mic84 (approved):

PR #1592, opened 2020-12-02
Eb from stl

hsitaram:
## Summary
This is a work-in-progress PR that adds a preliminary set of tools and helper functions for using triangulated STL files in amrex based applications. Coupling with AMReX's EB framework will be added in later pull requests. This PR is done early so as to provide a software framework for incremental changes. 

As of now, I have added one helper function that provides a nodal multifab with a marker value indicating inside or outside the STL geometry. I guess it can be used for particle-wall collisions. 

I have also added a test case in the EB folder that generates a nodal multifab with the blanked values for an airfoil.

The proposed changes:


Src/EB/AMReX_EB_STL_utils.H
```diff
@@ -0,0 +1,42 @@
+#ifndef AMREX_EB_STL_UTILS_H_
+#define AMREX_EB_STL_UTILS_H_
```
WeiqunZhang, line 2: Could you add `#include <AMReX_Config.H>` after this line and do similar thing for the other new header too?  We just switched to a new style of including `AMReX_Config.H` in every header file, so that the macros defined during the generation of libamrex can be reliably used later.

Src/EB/AMReX_EB_STL_utils.H
```diff
@@ -0,0 +1,42 @@
...
+#include <AMReX.H>
+#include <AMReX_ParmParse.H>
+#include <AMReX_MultiFab.H>
+#include<AMReX_REAL.H>
+#include<AMReX_Vector.H>
+#include <AMReX_Geometry.H>
+#include <AMReX_FArrayBox.H>
+#include <AMReX_Box.H>
+
+using namespace amrex;
```
WeiqunZhang: Could you put everything into `namespace amrex {}` instead of `using namespace amrex` to avoid polluting the global namespace?

Src/EB/AMReX_EB_STL_utils.cpp
```diff
@@ -0,0 +1,151 @@
...
+{
+    //local variables for lambda capture
+    int data_stride   = m_ndata_per_tri;
+    int num_triangles = m_num_tri;
+    Real outvalue     = m_outside;
+    Real invalue      = m_inside;
+
+    const auto plo   = geom.ProbLoArray();
+    const auto dx    = geom.CellSizeArray();
+    Real outp[]      = {point_outside[0],point_outside[1],point_outside[2]};
```
WeiqunZhang: Can we make this `GpuArray<Real,3> outp`?

Src/EB/AMReX_EB_triGeomOps_K.H
```diff
@@ -0,0 +1,303 @@
...
+#define AMREX_EB_TRIGEOMOPS_K_H_
+
+#include <AMReX.H>
+
+using namespace amrex;
+
+namespace tri_geom_ops
+{
+    //================================================================================
+    AMREX_GPU_HOST_DEVICE AMREX_INLINE Real Distance2(Real P1[3],Real P2[3])
```
WeiqunZhang: Can we use `AMREX_FORCE_INLINE` instead of `AMREX_INLINE`?

Src/EB/AMReX_EB_triGeomOps_K.H
```diff
@@ -0,0 +1,303 @@
...
+    AMREX_GPU_HOST_DEVICE AMREX_INLINE void getunitvec(Real v[3],Real vu[3])
+    {
+        Real vmag;
+        vmag=sqrt(v[0]*v[0]+v[1]*v[1]+v[2]*v[2]);
+        vu[0]=v[0]/vmag;
+        vu[1]=v[1]/vmag;
+        vu[2]=v[2]/vmag;
+    }
+    //================================================================================
+    AMREX_GPU_HOST_DEVICE AMREX_INLINE void CrossProd(Real v1[3],Real v2[3],Real v[3])
```
WeiqunZhang: We have `amrex::RealVect`, which has some of the operations like dot product and cross product defined.  I wonder if that can be used.  Note that `RealVect` has only two elements in 2D.  Anyway, this is not a high priority.

Src/EB/AMReX_EB_triGeomOps_K.H
```diff
@@ -0,0 +1,303 @@
...
+        getvec(P1,P3,v2);
+        CrossProd(v1,v2,n);
+
+
+        centr[0]=0.333333*(P1[0]+P2[0]+P3[0]);
+        centr[1]=0.333333*(P1[1]+P2[1]+P3[1]);
+        centr[2]=0.333333*(P1[2]+P2[2]+P3[2]);
+
+        getvec(centr,testp,c_tp_vec);
+        magn=sqrt(n[0]*n[0]+n[1]*n[1]+n[2]*n[2]);
```
WeiqunZhang: Can we add  `std::` to the math functions like `sqrt`?

Src/EB/AMReX_EB_triGeomOps_K.H
```diff
@@ -0,0 +1,303 @@
...
+
+        if(DotProd(c_tp_vec,n) < 0.0)
```
WeiqunZhang: Can we make the constants in the PR something like `Real(0.0)` so that the compile does not complain for single precision builds?

WeiqunZhang: Actually, don't worry about this.  The EB codes won't work with single precision anyway because there are so many other places that need to be fixed.

Src/EB/AMReX_EB_STL_utils.cpp
```diff
@@ -0,0 +1,151 @@
+#include<AMReX_EB_STL_utils.H>
+#include<AMReX_EB_triGeomOps_K.H>
+
+//================================================================================
+void STLtools::read_ascii_stl_file(std::string fname)
+{
+    std::string tmpline,tmp1,tmp2;
+    int nlines=0;
+
+    std::ifstream infile(fname.c_str());
```
WeiqunZhang: All processes reading from disk will not scale well on supercomputers.  I suggest we use `ParallelDescriptor::ReadAndBcastFile` and then read from stringstream.  For an example of using this, see `AMReX_VisMF.cpp`.

Src/EB/AMReX_EB_STL_utils.cpp
```diff
@@ -0,0 +1,151 @@
...
+    Print()<<"STL file name:"<<fname<<"\n";
```
WeiqunZhang: Can we put all Prints under `if (amrex::Verbose())`?  xSDK requires us to be able to turn off the print message.

Src/EB/AMReX_EB_STL_utils.H
hsitaram: this is done

Src/EB/AMReX_EB_STL_utils.H
hsitaram: everything is in amrex namespace now

Src/EB/AMReX_EB_STL_utils.cpp
hsitaram: I have changed this

Src/EB/AMReX_EB_triGeomOps_K.H
hsitaram: using FORCE_INLINE for all functions now

Src/EB/AMReX_EB_triGeomOps_K.H
hsitaram: I am not doing this now. Is it ok if I do it in a future PR?

Src/EB/AMReX_EB_triGeomOps_K.H
hsitaram: now using std namespace

Src/EB/AMReX_EB_triGeomOps_K.H
hsitaram: not worrying about this now!

Src/EB/AMReX_EB_STL_utils.cpp
hsitaram: This is a neat trick! I have added it to the stl reading function

Src/EB/AMReX_EB_STL_utils.cpp
hsitaram: Yes, I have done this now

Src/EB/AMReX_EB_triGeomOps_K.H
WeiqunZhang: Yes.

WeiqunZhang:
Two more things in AMReX_EB_triGeomOps_K.H.  Can you put the functions into amrex namespace and remove the using line?  Can you replace `fabs` with `amrex::Math::abs`?

hsitaram:
> Two more things in AMReX_EB_triGeomOps_K.H. Can you put the functions into amrex namespace and remove the using line? Can you replace `fabs` with `amrex::Math::abs`?

This is done now. Sorry I missed this earlier. Do you think it is better to put the test case I added in the tests folder or in the tutorials folder?

WeiqunZhang (approved):

PR #1593, opened 2020-12-02
Fix: Windows and/or includes (iso646)

ax3l:
## Summary

Add missing includes to `AMReX_Extension.H`, which pulls the `<iso646.h>` include on Windows for support for `and`/`or`.

## Additional background

Seen with Clang 11.0.0 on x64 Windows with Visual Studio 2017 on conda-forge during a WarpX build.

See also #947

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_Extension.H
+++ b/Src/Base/AMReX_Extension.H
@@ -137,7 +137,7 @@
 #endif
 
 #if defined(__cplusplus) && defined(_WIN32)
-#include <iso646.h>
+#include <ciso646>
 #endif
 
 #if (__cplusplus >= 201703L)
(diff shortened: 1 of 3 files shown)


ax3l:
Hm, patch looks needed but the original issue I saw with VS2017 and Clang11 on CF is not fixed yet.

@WeiqunZhang It looks like even other major libs gave up on `and`/`or`/`not` keywords: https://github.com/nlohmann/json/pull/2259

Background:
- Clang-Cl does not work well with the iso646 header in C++ mode: https://github.com/nlohmann/json/issues/1818#issuecomment-636316654 (weirdly, works with VS2019 + Clang 10 in CI)
- And even if it does, some Windows headers use these keywords so one cannot define them always: http://lists.llvm.org/pipermail/cfe-commits/Week-of-Mon-20170522/193579.html

Maybe VS2019 just fixed the need for this header... https://docs.microsoft.com/en-us/cpp/standard-library/ciso646?view=msvc-160&viewFallbackFrom=vs-2019

WeiqunZhang (approved):

PR #1602, opened 2020-12-04
CI: Disable DPC++ oneAPI 2020-11

ax3l:
## Summary

Someone patched in a dedicated CMake compiler ID for Intel DPC++
but did not add full compiler support.

Release: 2021.1.1

Ref.: https://gitlab.kitware.com/cmake/cmake/-/issues/21551

## Checklist

The proposed changes:

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -162,7 +162,10 @@ jobs:
             -DAMReX_CUDA_ARCH=6.0
         make -j 2
 
+# disabled since Intel patched a new CMake Compiler ID in with incomplete support
+# for dpcpp
   tutorials-dpcpp:
+    if: false
     name: DPCPP@PubBeta GFortran@7.5 C++17 [tutorials]
     runs-on: ubuntu-latest
     steps:


sayerhs:
@ax3l Given that this is probably not the last time this is going to happen, I wonder if we should pull DPC++ workflow into its own workflow so that we can [enable/disable workflow](https://docs.github.com/en/free-pro-team@latest/actions/managing-workflow-runs/disabling-and-enabling-a-workflow) without having to introduce commits into the codebase?

sayerhs (approved):

WeiqunZhang:
The dpcpp check has not been a required check for several days.  So we could keep it.

ax3l:
Yes, it will just be grey now instead of red, which avoids questions from developers I get on slack every other hour :)

ax3l:
> Given that this is probably not the last time this is going to happen, I wonder if we should pull DPC++ workflow into its own workflow so that we can enable/disable workflow without having to introduce commits into the codebase?

Oh interesting, that's neat! Easier to switch, did not know that functionality.
I can separate it out once I re-enable it. It's likely there will be more changes before I can do.

mic84 (approved):

PR #1603, opened 2020-12-04
Migrate building and deploying the documentation to Github CI.

atmyers:
Travis CI has put in a cap on the number of hours free accounts can use, so we are migrating this task to Github actions.

The proposed changes:


.github/workflows/docs.yml
```diff
@@ -0,0 +1,30 @@
...
+        run: |
+          ./build_docs.sh
+
+      - name: Deploy
+        if: github.event_name == 'push' && github.repository == 'AMReX-Codes/amrex' && github.ref == 'refs/heads/development'
+        uses: JamesIves/github-pages-deploy-action@3.7.1
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          ACCESS_TOKEN: ${{ secrets.DEPLOY_DOCS }}
+          REPOSITORY_NAME: atmyers/atmyers.github.io
```
WeiqunZhang: Is this the right repo?

atmyers: Er, no. I was testing it on mine and forgot to update.

WeiqunZhang (approved):

PR #1625, opened 2020-12-09
Fuse nodal jacobi smoother for GPU

WeiqunZhang:
## Summary
Fuse nodal jacobi smoother kernels by switch the order of the sweep and MFIter loops.  For Amr-Wind's ABL test with `amr.n_cell = 256 256 64` and `amr.max_grid_size = 32` on one V100, the nodal project is about 2x faster.

## Checklist

The proposed changes:

### DIFF
--- a/Src/LinearSolvers/MLMG/AMReX_MLNodeLap_1D_K.H
+++ b/Src/LinearSolvers/MLMG/AMReX_MLNodeLap_1D_K.H
@@ -45,16 +45,16 @@ inline void mlndlap_bc_doit (Box const& vbx, Array4<T> const& a, Box const& doma
 {}
 
 AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
-void mlndlap_adotx_ha (int /*i*/, int /*j*/, int /*k*/, Array4<Real> const& y, Array4<Real const> const& x,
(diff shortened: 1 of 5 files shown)


maximumcats:
The overall nodal projection performance increase I observe trying this on a single GPU on Summit (with the same setup, 256x256x64 zones with max_grid_size=32) is more like 20% than 100%, but the change looks good to me and the smoother looks exactly like we want in the profile.

maximumcats (approved):

WeiqunZhang:
That's interesting.  I did it on Cori and it went from
```
NodalProjector::project                                                14       6.55       6.55       6.55  36.08%
```
to
```
NodalProjector::project                                                14      2.969      2.969      2.969  28.40%
```

Note that I only ran for 10 steps and with `time.plot_interval=-1` and `time.checkpoint_interva=-1`.  I will try this on summit tomorrow.

WeiqunZhang:
I ran the test on summit and the difference was only 20% as what you saw.  If I reran the executable I built yesterday, I could reproduce the big difference.  But if I rebuild the executable using the same old amrex version, then the old version is much faster.  I don't understand why.  Maybe I used a wrong cmake flag yesterday.

PR #1627, opened 2020-12-10
Convert AMReX_eb_to_pvd.F90 to cpp

drangara:
## Summary
Convert the fortran file that writes the EB surface for paraview visualization to cpp

## Additional background
Tested with the visualization files produced by `mfix/benchamrk/07-hopper` and `mfix/tutorials/clr/prototype`. Paraview visualization seems to look "good". When comparing the actual data from the files I do see some differences b/w the fortran and cpp version, but I believe the differences can be attributed to the precision of datatypes used in the fortran vs cpp code. These changes should not affect the solution results, just affects visualization.

## Checklist

The proposed changes:


asalmgren:
It doesn't look like you finished the changes in Make.package.

Also -- I believe this routine is only relevant in 3D?   Do we have a way of keeping it from being included in a 2D build?

Src/EB/AMReX_EBToPVD.H
```diff
@@ -0,0 +1,60 @@
+#ifndef AMReX_EBToPVD_H_
+#define AMReX_EBToPVD_H_
+#include <AMReX_REAL.H>
```
WeiqunZhang:
```suggestion
#include <AMReX_Config.H>
#include <AMReX_REAL.H>
```

Src/EB/AMReX_EBToPVD.cpp
```diff
@@ -0,0 +1,495 @@
...
+
+            if(flag(i,j,k).isSingleValued()) {
+               Real axm = apx(i  ,j  ,k  );
+               Real axp = apx(i+1,j  ,k  );
+               Real aym = apy(i  ,j  ,k  );
+               Real ayp = apy(i  ,j+1,k  );
+               Real azm = apz(i  ,j  ,k  );
+               Real azp = apz(i  ,j  ,k+1);
+
+               Real apnorm = sqrt((axm-axp)*(axm-axp) + (aym-ayp)*(aym-ayp) + (azm-azp)*(azm-azp));
```
WeiqunZhang:
```suggestion
               Real apnorm = std::sqrt((axm-axp)*(axm-axp) + (aym-ayp)*(aym-ayp) + (azm-azp)*(azm-azp));
```

Src/EB/AMReX_EBToPVD.cpp
```diff
@@ -0,0 +1,495 @@
...
+      myfile.close();
+   }
+}
+
+void EBToPVD::WritePVTP(const int nProcs) const
+{
+   std::ofstream myfile("eb.pvtp");
+
+   if(myfile.is_open()) {
+      myfile << "<?xml version=\"1.0\"?>" << std::endl;
```
WeiqunZhang: It will be a little faster if `std::endl` here and many other places are replaced with `"\n"`, because `std::endl` flushes the buffer.

WeiqunZhang (approved):

PR #1633, opened 2020-12-12
AssignDensity and AssignCellDensitySingleLevel copy -> ParallelCopy

jmsexton03:
## Summary

"ParallelCopy" should be preferred over "copy" because the name is more explicit that it works even if the ba and dm are different.

## Checklist

The proposed changes:


WeiqunZhang:
`copy` is `ParallelCopy`.  `Copy` is serial.

WeiqunZhang:
We used to have `copy` vs. `Copy`.  That was very confusing.  So we added `ParallelCopy` and kept `copy` for backward compatibility.  Anyway, although it's not a bug, it's still better to use `ParallelCopy` for clarity.

atmyers:
Even though these copies are already parallel, let's merge this because `ParallelCopy` makes that more explicit than `copy`.

atmyers (approved):

PR #1638, opened 2020-12-14
Write a plotfile to ASCII

ajnonaka:
## Summary

This utility reads in a single-level plotfile, copies it to 
a MultiFab with a single box, then writes out all the data in
'i j k comp <value>' format to the screen
The user can modify this cpp file to write out on certain components,
coordinates, row/column formatting, etc.

Usage:
./WritePlotfileToASCII2d.gnu.MPI.ex infile=inputFileName


asalmgren:
Could you add something about this functionality into the AMReX docs?   We should have a section on what tools are available to post-process plotfiles ...

ajnonaka:
I started a new chapter in the docs about post-processing and described the usage of this utility.  PR is re-opened.

WeiqunZhang (approved):

PR #1642, opened 2020-12-15
Fix NodalProjector::getGradPhi for multilevel case

cgilet:
## Summary
    Two bug fixes for multilevel in NodalProjector:
    
    1. Ensure NodalProjector::getGradPhi() returns the correct result. Previously could return \sum{sigma grad phi} / (2**dim sigma_coarse).
    2. Average down the complete projected velocity at the end. For alpha!=1, this ensures the projected velocity is averaged down correctly.

This could change the results of multilevel regression tests. If the application uses the affected outputs from NodalProjector as is, then changes are expected. If the application subsequently averages down the outputs, then changes greater than roundoff level are not expected.   

## Checklist

The proposed changes:

### DIFF
--- a/Src/LinearSolvers/Projections/AMReX_NodalProjector.cpp
+++ b/Src/LinearSolvers/Projections/AMReX_NodalProjector.cpp
@@ -272,12 +272,6 @@ NodalProjector::project ( Real a_rtol, Real a_atol )
     // Get fluxes -- fluxes = - sigma * grad(phi)
     m_mlmg -> getFluxes( GetVecOfPtrs(m_fluxes) );
 
-    // At this time, the fluxes are "correct" only on regions not covered by finer grids.
-    // We average the fluxes down so that they are "correct" everywhere in each level.
-    // This is necessary because the caller can access grad(phi) and may use it for
-    // computations involving the whole level.
-    averageDown(GetVecOfPtrs(m_fluxes));
-
     // Compute sync residual BEFORE performing projection
     computeSyncResidual();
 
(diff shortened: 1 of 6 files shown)


Src/LinearSolvers/Projections/AMReX_NodalProjector.cpp
```diff
@@ -296,31 +290,20 @@ NodalProjector::project ( Real a_rtol, Real a_atol )
...
-        }
-
+	m_linop->compGrad(lev,m_fluxes[lev],m_phi[lev]);
     }
 
+    //
+    // At this time, results are "correct" only on regions not covered by finer grids.
+    // We average them down so that they are "correct" everywhere in each level.
+    //
+    averageDown(GetVecOfPtrs(m_fluxes));
```
WeiqunZhang, line 303: Is this necessary for non-EB?  When `LinOp::compGrad` is called, the nodal phi has been averaged down.  So I don't think this is necessary at least for non-EB.

Src/LinearSolvers/Projections/AMReX_NodalProjector.cpp
```diff
@@ -296,31 +290,20 @@ NodalProjector::project ( Real a_rtol, Real a_atol )
...
+    averageDown(m_vel);
```
WeiqunZhang, line 304: Maybe we can add `if (m_has_alpha)` here.

Src/LinearSolvers/Projections/AMReX_NodalProjector.cpp
cgilet: Looking at the numbers says we do really need to average down here. ... For nodal, doesn't average down just copy the overlying fine node? Whereas gradPhi is cell centered, so it does a more traditional averaging, which would result in non-overlying nodes getting included.

WeiqunZhang: Then something is not quite right.

cgilet: Yeah. I was just thinking the same thing

WeiqunZhang: Actually you are right.  They are supposed to be different.

WeiqunZhang (approved):
@mic84 Could you take a look at this too?

mic84:
@WeiqunZhang it looks good to me.

mic84 (approved):

PR #1643, opened 2020-12-15
PolymorphicArray4

WeiqunZhang:
## Summary

PolymorphicArray4 is a thin layer around Array4.  It allows us to access
both AoS and SoA data with operator().  For example,

    constexpr int ncomp = 2;
    Box box(...);
    BaseFab<Real> soa_fab(box, ncomp);
    auto soa_array = makePolymorphic(soa_fab.array());
    int i = ...; int j = ...; int k = ...;
    soa_array(i,j,k,0) = 0;
    soa_array(i,j,k,1) = 1;

    BaseFab<GpuArray<Real,ncomp>> aos_fab(box,1);
    auto aos_array = makePolymorphic(aos_fab.array());
    aos_array(i,j,k,0) = 0;
    aos_array(i,j,k,1) = 1;

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_Array.H
+++ b/Src/Base/AMReX_Array.H
@@ -33,6 +33,9 @@ namespace amrex {
     template <class T, std::size_t N>
     struct GpuArray
     {
+        using value_type = T;
+        using reference_type = T&;
+
         AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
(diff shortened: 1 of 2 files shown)


maximumcats:
Is it possible to do this directly to Array4 without adding a new class? Or does that break existing uses.

WeiqunZhang:
The problem of that is it will break the existing code.

atmyers (approved):

PR #1646, opened 2020-12-15
DPCPP MPI wrapper

WeiqunZhang:
## Summary

Replace mpiicpx with mpiicpc because mpiicpc is shipped by Intel, whereas
mpiicpx is a script that is only available on JLSE testbeds.

## Checklist

The proposed changes:

### DIFF
--- a/Tools/CMake/AMReXOptions.cmake
+++ b/Tools/CMake/AMReXOptions.cmake
@@ -126,10 +126,12 @@ endif ()
 
 # --- SYCL ---
 if (AMReX_DPCPP)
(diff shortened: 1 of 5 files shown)


sayerhs:
@WeiqunZhang We should also update CMake message

https://github.com/AMReX-Codes/amrex/blob/cd2e06e47cb586d0b8e2ef3eb0c534548634ad97/Tools/CMake/AMReXOptions.cmake#L128-L133

WeiqunZhang:
Yes, I have pinged @mic84 on slack.

mic84 (approved):

PR #1653, opened 2020-12-16
CI: CUDA 9.1 & 11.0

ax3l:
## Summary

Add a recent CUDA 11 release for cutting-edge coverage, too.

## Checklist

The proposed changes:


.github/workflows/linux.yml
```diff
@@ -6,7 +6,7 @@ jobs:
   # Build and install libamrex as AMReX CMake project
   library:
     name: GNU@7.5 C++17 Release [lib]
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
```
ax3l, line 9: I am just changing this, since `-latest` will default to `20.04` very soon.

.github/workflows/linux.yml
```diff
@@ -163,6 +163,35 @@ jobs:
...
+    runs-on: ubuntu-20.04
+    env: {CXXFLAGS: "-fno-operator-names"}
+    steps:
+    - uses: actions/checkout@v2
+    - name: Dependencies
+      run: .github/workflows/dependencies/dependencies_nvcc11.sh
+    - name: Build & Install
+      run: |
+        export PATH=/usr/local/nvidia/bin:/usr/local/cuda/bin:${PATH}
+        export LD_LIBRARY_PATH=/usr/local/nvidia/lib:/usr/local/nvidia/lib64:${LD_LIBRARY_PATH}
```
maximumcats: I think you want `/usr/local/cuda/lib64` here too.

ax3l: Already in there (2nd added argument) :)

ax3l: Oh wait, it's not. Will add!
(Curiously, the current syntax works for WarpX.)

maximumcats: It's not always necessary since nvcc statically links the CUDA runtime library libcudart.so by default -- it is more relevant when you're using things like cuBLAS. All of the other driver libraries like libcuda.so and libnvidia-ml.so go in system standard places like /usr/lib64.

ax3l: Thanks for the details, will modernize WarpX' CI then as well: https://github.com/ECP-WarpX/WarpX/pull/1581

ax3l:
@maxpkatz I changed the nvcc 11.2 build to use C++14 for now. We can fix the C++17 NVCC build independently and then increase the CI coverage in a follow-up.

WeiqunZhang (approved):

maximumcats:
> We can fix the C++17 NVCC build independently and then increase the CI coverage in a follow-up.

Good by me. Thanks for finishing this up.

ax3l:
Cross-link: NVCC C++17 CI ensured via #1665

PR #1684, opened 2020-12-23
unsigned long long iMultiFab sum.

kngott:
## Summary
Changes iMultiFab sum to use hardware supported unsigned long long.

## Additional background
Using ParallelReduce, this changes sum time on Cori (default CUDA, 10.1) from 0.07546 to 0.001185, 63x increase.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_iMultiFab.cpp
+++ b/Src/Base/AMReX_iMultiFab.cpp
@@ -325,10 +325,14 @@ iMultiFab::sum (int comp, int nghost, bool local) const
     Long sm = 0;
 
 #ifdef AMREX_USE_GPU
+    // If on GPUs, cast to unsigned long long to take advantage of hardware support.
     if (Gpu::inLaunchRegion())
     {
+        unsigned long long ulsm = 0;
+        unsigned long long points = 0;
+        const unsigned long long imax = static_cast<unsigned long long>(INT_MAX);
         ReduceOps<ReduceOpSum> reduce_op;
-        ReduceData<Long> reduce_data(reduce_op);
+        ReduceData<unsigned long long> reduce_data(reduce_op);
         using ReduceTuple = typename decltype(reduce_data)::Type;
 
         for (MFIter mfi(*this); mfi.isValid(); ++mfi)
@@ -338,12 +342,14 @@ iMultiFab::sum (int comp, int nghost, bool local) const
             reduce_op.eval(bx, reduce_data,
             [=] AMREX_GPU_DEVICE (int i, int j, int k) -> ReduceTuple
             {
-                return { static_cast<Long>(arr(i,j,k,comp)) };
+                return { static_cast<unsigned long long>(arr(i,j,k,comp))+imax };
             });
+            points += bx.numPts();
         }
 
         ReduceTuple hv = reduce_data.value();
-        sm = amrex::get<0>(hv);
+        ulsm = amrex::get<0>(hv) - points*imax;
+        sm = static_cast<Long>(ulsm);
     }
     else
 #endif
--- a/Src/Base/AMReX_iMultiFab.cpp
+++ b/Src/Base/AMReX_iMultiFab.cpp
@@ -328,9 +328,7 @@ iMultiFab::sum (int comp, int nghost, bool local) const
     // If on GPUs, cast to unsigned long long to take advantage of hardware support.
     if (Gpu::inLaunchRegion())
     {
-        unsigned long long ulsm = 0;
-        unsigned long long points = 0;
-        const unsigned long long imax = static_cast<unsigned long long>(INT_MAX);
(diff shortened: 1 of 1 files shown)


Src/Base/AMReX_iMultiFab.cpp
```diff
@@ -338,12 +342,14 @@ iMultiFab::sum (int comp, int nghost, bool local) const
             reduce_op.eval(bx, reduce_data,
             [=] AMREX_GPU_DEVICE (int i, int j, int k) -> ReduceTuple
             {
-                return { static_cast<Long>(arr(i,j,k,comp)) };
+                return { static_cast<unsigned long long>(arr(i,j,k,comp))+imax };
```
WeiqunZhang: Why not using the macro directly here instead of capturing it?  Also it should be `-INT_MIN` instead because `INT_MIN` is usually `-INT_MAX-1`

Src/Base/AMReX_iMultiFab.cpp
```diff
@@ -338,12 +342,14 @@ iMultiFab::sum (int comp, int nghost, bool local) const
...
             });
+            points += bx.numPts();
         }
 
         ReduceTuple hv = reduce_data.value();
-        sm = amrex::get<0>(hv);
+        ulsm = amrex::get<0>(hv) - points*imax;
+        sm = static_cast<Long>(ulsm);
```
WeiqunZhang: I don't think this is correct. Because `ulsm` is unsigned, it can never be negative after the cast.

WeiqunZhang: You have to cast `amrex::get<0>(hv)` and `points*imax` to `long long` first.

Src/Base/AMReX_iMultiFab.cpp
kngott: Are there any warnings/issues/error messages that may interact with type_casting a negative number to an unsigned long long that has a minimum of 0? Or, are the compilers known to be smart enough to do the math in the correct order and never overflow?

This seems to yield an identical time and give the correct answer, baring any overflow concerns:
```               
return { static_cast<unsigned long long>(arr(i,j,k,comp))-
         static_cast<unsigned long long>(INT_MIN) };
```

Note: static_cast<unsigned long long>(-INT_MIN) is outside the int bounds, due to the -INT_MAX-1, so we'd need to do (-(INT_MIN+1)) to suppress the warning, if we wanted to avoid potential issues manually:
```
/ccs/home/kngott/amrex/Src/Base/AMReX_iMultiFab.cpp(346): warning: integer operation result is out of range
```

WeiqunZhang: I think it should be `static_cast<unsigned long long>(static_cast<long long>(arr(i,j,k,comp)) - static_cast<long long>(INT_MIN))`.

Src/Base/AMReX_iMultiFab.cpp
kngott: Good catch. I'm currently testing:
```
        long long ulsm = static_cast<long long>(amrex::get<0>(hv)) 
                       - static_cast<long long>(INT_MAX)*points;
```
I think this'll work, given we expect the final answer to be a `long`.

Src/Base/AMReX_iMultiFab.cpp
WeiqunZhang: Actually `return { static_cast<unsigned long long>(arr(i,j,k,comp)) - static_cast<unsigned long long>(INT_MIN) };` is fine because it's always well defined with the modulo behavior.

Src/Base/AMReX_iMultiFab.cpp
WeiqunZhang: Assuming you are using `{ static_cast<unsigned long long>(arr(i,j,k,comp)) - static_cast<unsigned long long>(INT_MIN) }` in the kernel, I think this should be `static_cast<long long>(amrex::get<0>(hv)) + static_cast<long long>(INT_MIN)*points` and you want `points` to be `long long`, not unsigned.

WeiqunZhang (approved):

Src/Base/AMReX_iMultiFab.cpp
```diff
@@ -338,12 +340,15 @@ iMultiFab::sum (int comp, int nghost, bool local) const
             reduce_op.eval(bx, reduce_data,
             [=] AMREX_GPU_DEVICE (int i, int j, int k) -> ReduceTuple
             {
-                return { static_cast<Long>(arr(i,j,k,comp)) };
+                return { static_cast<unsigned long long>(arr(i,j,k,comp)) -
+                         static_cast<unsigned long long>(INT_MIN)) };
```
WeiqunZhang: extra `)`

PR #1685, opened 2020-12-23
Sanitize the input of mkconfig

### LINKED ISSUE #1682: mkconfig.py creates a bad AMReX_Config.H when the user has defined CPPFLAGS to include -I flags
I have `CPPFLAGS` set to include `-I/usr/local/...` which trips up the compiler when using GNU make:
```
tmp_build_dir/s/3d.gnu.MPI.EXE/AMReX_Config.H:3:9: error: macro name must be an identifier
#ifndef -I/usr/local/opt/binutils/include -I/usr/local/opt/llvm/include -I/usr/local/opt/binutils/include -I/usr/local/opt/llvm/include -I/usr/local/opt/llvm/include -I/usr/local/opt/llvm/include -I/usr/local/opt/binutils/include -I/usr/local/opt/llvm/include -I/usr/local/opt/llvm/include
```
which is due to a bad `AMReX_Config.H`:
```
#ifndef AMREX_HAVE_NO_CONFIG_H
#define AMREX_HAVE_NO_CONFIG_H
#ifndef -I/usr/local/opt/binutils/include -I/usr/local/opt/llvm/include -I/usr/local/opt/binutils/include -I/usr/local/opt/llvm/include -I/usr/local/opt/llvm/include -I/usr/local/opt/llvm/include -I/usr/local/opt/binutils/include -I/usr/local/opt/llvm/include -I/usr/local/opt/llvm/include
#define -I/usr/local/opt/binutils/include -I/usr/local/opt/llvm/include -I/usr/local/opt/binutils/include -I/usr/local/opt/llvm/include -I/usr/local/opt/llvm/include -I/usr/local/opt/llvm/include -I/usr/local/opt/binutils/include -I/usr/local/opt/llvm/include -I/usr/local/opt/llvm/include 1
#endif
```
After consulting Michele, I found that `mkconfig.py` does not expect `-I` flags in the `--defines`:
```
$(srcTempDir)/AMReX_Config.H:
        @echo Generating AMReX_Config.H ...
        @if [ ! -d $(srcTempDir) ]; then mkdir -p $(srcTempDir); fi
        @ $(MKCONFIG) --defines="$(CPPFLAGS)" --undefines="" \
                      --comp="$(lowercase_comp)" \
                      --allow-different-compiler="$(ALLOW_DIFFERENT_COMP)" \
                      > $@
```
I suggest sanitizing the input to `--defines` when the user specifies include directories (or better still, non `-D` flags) in `CPPFLAGS`

@WeiqunZhang tag?

@mic84 wants me to make sure that I mention that CMake doesn't run into this problem (my guess would be that CMake sanitizes that inputs into `mkconfig.py` byt GNU Make does not).

WeiqunZhang:
## Summary

The mkconfig.py script takes CPPFLAGS and generates AMReX_Config.H.  If
CPPFLAGS contains -I flags (as advised by homebrew) and the input is not
santized, it would generate ill-formed code.

## Checklist

The proposed changes:

### DIFF
--- a/Tools/libamrex/mkconfig.py
+++ b/Tools/libamrex/mkconfig.py
@@ -2,7 +2,7 @@
 
 from __future__ import print_function
 
-import sys
+import sys, re
 
 if sys.version_info < (2, 7):
     sys.exit("ERROR: need python 2.7 or later for mkconfig.py")
@@ -12,6 +12,9 @@ import argparse
 def doit(defines, undefines, comp, allow_diff_comp):
     print("#ifndef AMREX_HAVE_NO_CONFIG_H")
     print("#define AMREX_HAVE_NO_CONFIG_H")
+
+    # Remove -I from input
+    defines = re.sub(r'-I.*?(-D|$)', r'\1', defines)
 
     defs = defines.split("-D")
     for d in defs:


JBlaschke:
Just tested -- it fixes #11682 on my setup.

mic84 (approved):

PR #1690, opened 2020-12-24
Optimization of MLCellLinOp::applyBC

WeiqunZhang:
## Summary

Launch one kernel per box to work on all faces and all components.  If the
number of boxes exceeds the kernel fusing threshold, the kernels from all
boxes will be fused.

## Checklist

The proposed changes:

### DIFF
--- a/Src/LinearSolvers/MLMG/AMReX_MLCellLinOp.cpp
+++ b/Src/LinearSolvers/MLMG/AMReX_MLCellLinOp.cpp
@@ -504,7 +504,10 @@ MLCellLinOp::applyBC (int amrlev, int mglev, MultiFab& in, BCMode bc_mode, State
                     bvhi[idim] = (bndry != nullptr) ? bndry->bndryValues(ohi).array(mfi) : foo;
                 }
                 const auto len = vbx.length3d();
-                const int nthreads = amrex::max(len[0]*len[1],len[0]*len[2],len[1]*len[2]);
+                const int nthreads
+                    = AMREX_D_PICK(1;,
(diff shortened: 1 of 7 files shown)


Src/LinearSolvers/MLMG/AMReX_MLCellLinOp.cpp
```diff
@@ -791,101 +865,250 @@ MLCellLinOp::prepareForSolve ()
...
-                                mllinop_comp_interp_coef0_z_eb
-                                    (0, tboxlo, blen, flo, mlo, ap, bctlo, bcllo,
-                                     imaxorder, dzi, icomp);
-                                },
-                                bhi, tboxhi, {
-                                mllinop_comp_interp_coef0_z_eb
-                                    (1, tboxhi, blen, fhi, mhi, ap, bcthi, bclhi,
-                                     imaxorder, dzi, icomp);
-                                });
+                    if (fabtyp == Fabtyp::singlevalued) {
```
maximumcats: I think this is a typo?

maximumcats:
This change looks like it works out -- on the 256x256x64 single-box ABL case I was looking at in amr-wind, this brings the kernel time in prepareForSolve and applyBC down by a factor of 4.

Src/LinearSolvers/MLMG/AMReX_MLCellLinOp.cpp
WeiqunZhang: Yes, thanks!

maximumcats (approved):

PR #1696, opened 2021-01-04
Support when CMake 3.18 policy 105 is set to new

robertmaynard:
## Summary
When AMReX is using a newer version of CMake (3.18+) it sets policy `105` to new, causing `target_link_options` rules to be applied to CUDA device linking. When that happens we need to make sure that host only linking flags are excluded.
 
## Checklist

The proposed changes:

### DIFF
--- a/Tools/CMake/AMReX_Config.cmake
+++ b/Tools/CMake/AMReX_Config.cmake
@@ -141,11 +141,22 @@ function (configure_amrex)
 
    if ( BUILD_SHARED_LIBS OR AMReX_CUDA )
       if(APPLE)
-         target_link_options(amrex PUBLIC -Wl,-undefined,warning)
+        target_link_options(amrex PUBLIC "LINKER:-undefined,warning")
       elseif(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC" OR
              CMAKE_CXX_SIMULATE_ID STREQUAL "MSVC")
+
       else()
-         target_link_options(amrex PUBLIC -Wl,--warn-unresolved-symbols)
+
+        set(host_link_supported OLD)
+        if(CMP0105)
+          cmake_policy(GET CMP0105 host_link_supported)
+        endif()
+
+        if( host_link_supported STREQUAL "NEW" )
+          target_link_options(amrex PUBLIC "$<HOST_LINK:LINKER:--warn-unresolved-symbols>")
+        else()
+          target_link_options(amrex PUBLIC "LINKER:--warn-unresolved-symbols")
+        endif()
       endif()
    endif()
 


ax3l (approved):
Thank you for the fix! :tada:

ax3l:
@robertmaynard can you please rebase against `development`? I just fixed the macOS CI issue.

robertmaynard:
> @robertmaynard can you please rebase against `development`? I just fixed the macOS CI issue.

Rebased

ax3l:
This might have introduced a regression in mfix and AMR-Wind for CUDA builds, showing up since today:
```
unrecognized command line option '--warn-unresolved-symbols'
```

robertmaynard:
> This might have introduced a regression in mfix for CUDA builds:
> 
> ```
> unrecognized command line option '--warn-unresolved-symbols'
> ```

It sounds like the `LINKER:` is getting resolved to an empty string instead of the correct linker flag. 

Was this with NVCC?

Tools/CMake/AMReX_Config.cmake
```diff
@@ -141,11 +141,22 @@ function (configure_amrex)
...
+
+        set(host_link_supported OLD)
+        if(CMP0105)
+          cmake_policy(GET CMP0105 host_link_supported)
+        endif()
+
+        if( host_link_supported STREQUAL "NEW" )
+          target_link_options(amrex PUBLIC "$<HOST_LINK:LINKER:--warn-unresolved-symbols>")
+        else()
+          target_link_options(amrex PUBLIC "LINKER:--warn-unresolved-symbols")
```
ax3l, line 158: Maybe rolling this line back to
```
target_link_options(amrex PUBLIC "-Wl,--warn-unresolved-symbols")
```
could solve the regression? @robertmaynard

robertmaynard: That will break CMake 3.18+ as the `--warn-unresolved-symbols` will be propagated to the device link step for nvcc.

ax3l:
> It sounds like the LINKER: is getting resolved to an empty string instead of the correct linker flag.
>
> Was this with NVCC?

The issue occurs only with pre-CMake 3.18 (seen with 3.14.1 and 3.15.2) when using NVCC on Linux (e.g. cmake/3.15.2 + cuda/10.1.243).

Tools/CMake/AMReX_Config.cmake
ax3l: @mic84 is it possible that we patch-out the `LINKER:`-expression when stripping the generator expressions?

ax3l: > That will break CMake 3.18+ ...

Not sure, since this line (158) is not taken for CMake 3.18+ (it uses 156 instead, no?)

robertmaynard: Sorry, I misread your first comment @ax3l. I thought you meant roll back both the `if` and `else`. I think rolling back `LINKER:` would be sufficient.

robertmaynard:
> > It sounds like the LINKER: is getting resolved to an empty string instead of the correct linker flag.
> > Was this with NVCC?
> 
> The issue occurs only with pre-CMake 3.18 (seen with 3.14.1 and 3.15.2) when using NVCC on Linux (e.g. cmake/3.15.2 + cuda/10.1.243).

Found the issue. CMake < 3.18 didn't establish the proper information to evaluate `LINKER` for nvcc.
So yes we need to roll back line 158

PR #1699, opened 2021-01-05
fix -Werror=sign-compare

mwm126:
## Summary

Use size_t for index variables to avoid sign-compare warnings and casting to size_t

## Checklist

The proposed changes:


atmyers:
Thank you for this pull request. Would you mind fixing this slightly differently? I'd prefer leaving the template parameters as type `int`, but casting them to `size_t` as needed when we need to compare them to an unsigned type. That would be more consistent with the way we handle this in other parts of AMReX.

mwm126:
Ah, yes I see now that changing the template parameter type wasn't necessary.  Sure!  done.

atmyers (approved):

atmyers:
Thanks!

PR #1700, opened 2021-01-05
Update write plot file to ascii tool

emotheau:
## Summary

A flag (fast = 1) has been added to the WritePlotfileToAscii tool to propose to export data via Amrex::Print() instead of the 4 i,j,k,n imbricated loops. Indeed, on large plotfiles, exporting data to an ascii file is extremely slow with the procedure originally implemented in the tool.

## Checklist

The proposed changes:


WeiqunZhang:
@ajnonaka Since this tool is recently added, maybe we simply remove the old way?

emotheau:
I agree, I hesitated to remove the old way, but as the format of the output data changes (all components are now in a unique row), it may break the post-processing pipeline of @ajnonaka

Tools/Postprocessing/C_Src/WritePlotfileToASCII.cpp
```diff
@@ -131,16 +134,20 @@ main (int   argc,
...
-            
+        if (fast == 1){
+          // This way is much faster than the loops below
+          amrex::Print() << mf_onegrid[mfi];
+        }else{
+          for (auto n=0; n<ncomp; ++n) {
+            for (auto k = lo.z; k <= hi.z; ++k) {
+              for (auto j = lo.y; j <= hi.y; ++j) {
+                for (auto i = lo.x; i <= hi.x; ++i) {
+                  Print() << i << " " << j << " " << k << " " << n << " " << mfdata(i,j,k,n) << std::endl;
```
WeiqunZhang: @emotheau Could you test to see if the change below makes it as fast as the "fast" option?
```suggestion
                  std::cout << i << " " << j << " " << k << " " << n << " " << mfdata(i,j,k,n) << "\n";
```

emotheau: I tested and it does not really help, it is as slow as the `std::endl` version. I think the loop over the n components is making things bad.

WeiqunZhang: Sorry, can you test it again?  It should use `std::cout` instead of `Print`.  The later has flush internally.

emotheau: Ah ! This time it is almost as fast as the "fast" option ! Do we keep this suggestion and I remove all my previous stuff ?

WeiqunZhang: Maybe you can keep both options, but choose a different name for the argument to show the difference in the order of the printed data.

I also think you should use `std::cout` instead of `Print()` too for the new option you added.

emotheau: Ok I updated my PR to reflect your suggestions. Both methods give different representation of data but with the same (roughly) computational time. I think we are good now?

ajnonaka (approved):
Perhaps leave a note in PrintUsage about what comp_in_line does?

Tools/Postprocessing/C_Src/WritePlotfileToASCII.cpp
```diff
@@ -20,7 +20,7 @@ PrintUsage (const char* progName)
             << "coordinates, row/column formatting, etc." << std::endl << std::endl;
     
     Print() << "Usage:" << '\n';
-    Print() << progName << " infile=inputFileName" << '\n' << '\n';
+    Print() << progName << " infile=inputFileName (optional comp_in_line = 1)" << '\n' << '\n';
```
WeiqunZhang:
```suggestion
    Print() << progName << " infile=inputFileName [comp_in_line = 1]" << '\n' << '\n';
```
Square brackets are commonly used to indicate optional parameters. https://en.wikipedia.org/wiki/Usage_message

Tools/Postprocessing/C_Src/WritePlotfileToASCII.cpp
```diff
@@ -131,16 +134,20 @@ main (int   argc,
...
-       for (auto j = lo.y; j <= hi.y; ++j) {
-       for (auto i = lo.x; i <= hi.x; ++i) {
-            Print() << i << " " << j << " " << k << " " << n << " " << mfdata(i,j,k,n) << std::endl;
-       }
-       }
-       }
-       }
-            
+        if (comp_in_line == 1){
+          // This way is much faster than the loops below
```
WeiqunZhang: Can we remove this line of comment now?

emotheau: Done

WeiqunZhang (approved):

PR #1701, opened 2021-01-05
CMake<3.18: Fix Linker Regression

ax3l:
## Summary

Roll-back to explicit `-Wl,` linker flags for CMake<3.18

## Additional background

Regression of #1696

cc @robertmaynard @mic84 

## Checklist

The proposed changes:

### DIFF
--- a/Tools/CMake/AMReX_Config.cmake
+++ b/Tools/CMake/AMReX_Config.cmake
@@ -140,22 +140,25 @@ function (configure_amrex)
    endif ()
 
    if ( BUILD_SHARED_LIBS OR AMReX_CUDA )
+      set(host_link_supported OLD)
+      if(CMP0105)
+        cmake_policy(GET CMP0105 host_link_supported)
+      endif()
+
       if(APPLE)
-        target_link_options(amrex PUBLIC "LINKER:-undefined,warning")
+        if( host_link_supported STREQUAL "NEW" ) # CMake 3.18+
+          target_link_options(amrex PUBLIC "LINKER:-undefined,warning")
+        else()
+          target_link_options(amrex PUBLIC "-Wl,-undefined,warning")
+        endif()
       elseif(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC" OR
              CMAKE_CXX_SIMULATE_ID STREQUAL "MSVC")
-
+        # nothing
       else()
-
-        set(host_link_supported OLD)
-        if(CMP0105)
-          cmake_policy(GET CMP0105 host_link_supported)
-        endif()
-
-        if( host_link_supported STREQUAL "NEW" )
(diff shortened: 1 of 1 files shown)


Tools/CMake/AMReX_Config.cmake
```diff
@@ -155,7 +155,7 @@ function (configure_amrex)
         if( host_link_supported STREQUAL "NEW" )
           target_link_options(amrex PUBLIC "$<HOST_LINK:LINKER:--warn-unresolved-symbols>")
         else()
-          target_link_options(amrex PUBLIC "LINKER:--warn-unresolved-symbols")
+          target_link_options(amrex PUBLIC "-Wl,--warn-unresolved-symbols")
```
ax3l, line 158: @robertmaynard in line 144: should we roll back or make the `LINKER:` conditional as well?
We don't have CUDA these days on macOS anymore, but if `LINKER:` is a CMake 3.18+ feature we might better not place it in the `if(APPLE)` path unconditionally, yet.

Tools/CMake/AMReX_Config.cmake
```diff
@@ -140,22 +140,25 @@ function (configure_amrex)
    endif ()
 
    if ( BUILD_SHARED_LIBS OR AMReX_CUDA )
+      set(host_link_supported OLD)
+      if(CMP0105)
+        cmake_policy(GET CMP0105 host_link_supported)
+      endif()
+
       if(APPLE)
```
ax3l, line 148: @robertmaynard in this line: should we roll back or make the `LINKER:` conditional as well?
We don't have CUDA these days on macOS anymore, but if `LINKER:` is a CMake 3.18+ feature we might better not place it in the `if(APPLE)` path unconditionally, yet.

robertmaynard (commented):
`LINKER:` was added at the same time as `target_link_options`, we just missed the `NVIDIA-CUDA` language/compiler configuration and so it didn't have a proper mapping.

Tools/CMake/AMReX_Config.cmake
robertmaynard: `LINKER:` was added at the same time as `target_link_options`, we just missed the `NVIDIA-CUDA` language/compiler configuration and so it didn't have a proper mapping.

ax3l:
Thanks for the info!
I'll leave it with the same logic anyway, not sure if someone uses an old CUDA on macOS.
We can drop both paths to the new logic once we upgrade to 3.18+.

Does the current patch look good to you?

Tools/CMake/AMReX_Config.cmake
ax3l: (answer below)

robertmaynard:
It all looks good to me

mic84 (approved):

PR #1703, opened 2021-01-06
ParallelDescriptor::ReduceReal wrappers

WeiqunZhang:
## Summary

Make ParallelDescriptor::ReduceReal wrappers template on the floating point
data type.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_ParallelDescriptor.H
+++ b/Src/Base/AMReX_ParallelDescriptor.H
@@ -422,31 +422,100 @@ while ( false )
     void ReduceBoolOr  (bool& rvar, int cpu);
 
     //! Real sum reduction.
-    void ReduceRealSum (Real& rvar);
-    void ReduceRealSum (Real* rvar, int cnt);
+    template <typename T>
+    typename std::enable_if<std::is_floating_point<T>::value>::type
+    ReduceRealSum (T& rvar);
+
+    template <typename T>
+    typename std::enable_if<std::is_floating_point<T>::value>::type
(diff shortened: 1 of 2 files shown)


Src/Base/AMReX_ParallelDescriptor.H
```diff
@@ -1120,6 +1189,333 @@ Scatter(T* /*t*/, size_t /*n*/, const T1* /*t1*/, size_t /*n1*/, int /*root*/)
...
+					  MyTeam().get_lead_comm()) );
+	}
+	BL_MPI_REQUIRE( MPI_Bcast(recv.dataPtr(), cnt, Mpi_typemap<T>::type(),
+				  0, MyTeam().get_team_comm()) );
+    }
+    else
+#endif
+    {
+	BL_MPI_REQUIRE( MPI_Allreduce(r,
+				      recv.dataPtr(),
```
sayerhs: @WeiqunZhang I am wondering if there is a reason why AMReX doesn't use `MPI_IN_PLACE` here?

WeiqunZhang: Thanks for pointing this out.  We should `MPI_IN_PLACE` here.  Will update it.

atmyers (approved):

atmyers:
Do think the test failure is related to `MPI_IN_PLACE`?

WeiqunZhang:
Yes, I think it's fixed now.

WeiqunZhang:
For `MPI_Reduce`, `MPI_IN_PLACE` can only be used on the root process.

PR #1707, opened 2021-01-07
Add ability to calculate slopes on face centroids, and apply the expl…

drangara:
…icit stencil when computing the laplacian

## Summary
These changes allow the computation of EB viscous laplacian operator when phi is defined on cell centroids by using the Least squares method for calculating slopes on face centroids. Currently only works when Dirichlet BC is set.

## Checklist

The proposed changes:


Src/EB/AMReX_EB_LeastSquares_2D_K.H
```diff
@@ -0,0 +1,813 @@
+#ifndef AMREX_MLEB_LEASTSQUARES_K_H_
+#define AMREX_MLEB_LEASTSQUARES_K_H_
+
```
WeiqunZhang:
```suggestion
#include <AMReX_Config.H>
```

Src/EB/AMReX_EB_LeastSquares_3D_K.H
```diff
@@ -0,0 +1,1184 @@
+#ifndef AMREX_MLEB_LEASTSQUARES_K_H_
+#define AMREX_MLEB_LEASTSQUARES_K_H_
+
```
WeiqunZhang:
```suggestion
#include <AMReX_Config.H>
```

Src/LinearSolvers/MLMG/AMReX_MLEBABecLap.cpp
```diff
@@ -185,14 +185,22 @@ MLEBABecLap::setBCoeffs (int amrlev, Vector<Real> const& beta)
 void
 MLEBABecLap::setEBDirichlet (int amrlev, const MultiFab& phi, const MultiFab& beta)
 {
+    bool  phi_on_centroid = (m_phi_loc  == Location::CellCentroid);
+
+    if (phi_on_centroid) {
+      // If we call setEBDirichlet instead of setEBHomogDirichlet then we assume posisbly inhomogeneous
```
WeiqunZhang: Don't think this is needed.  `m_is_eb_inhomog` is a `mutable` that is set in `applyBC`.     Then in `Fapply` and `getEBFluxes`, the value of `m_is_eb_inhomog` is used.  It is actaully false (homogeneous) for inhomogeneous Drichlet EB when the multigrid solver is in correct form.

Src/LinearSolvers/MLMG/AMReX_MLEBABecLap.cpp
```diff
@@ -252,16 +260,29 @@ MLEBABecLap::setEBDirichlet (int amrlev, const MultiFab& phi, const MultiFab& be
...
 }
 
 void
 MLEBABecLap::setEBDirichlet (int amrlev, const MultiFab& phi, Real beta)
 {
+    bool  phi_on_centroid = (m_phi_loc  == Location::CellCentroid);
+
+    if (phi_on_centroid) {
+       // If we call setEBDirichlet instead of setEBHomogDirichlet then we assume posisbly inhomogeneous
+       m_is_eb_inhomog = true;
```
WeiqunZhang: See the comment above.

Src/LinearSolvers/MLMG/AMReX_MLEBABecLap.cpp
```diff
@@ -366,18 +400,30 @@ MLEBABecLap::setEBDirichlet (int amrlev, const MultiFab& phi, Vector<Real> const
...
+    }
 }
 
 void
 MLEBABecLap::setEBHomogDirichlet (int amrlev, const MultiFab& beta)
 {
+    bool  phi_on_centroid = (m_phi_loc  == Location::CellCentroid);
+
+    if (phi_on_centroid) {
+      m_is_eb_inhomog = false;
```
WeiqunZhang: See the comment above.

Src/LinearSolvers/MLMG/AMReX_MLEBABecLap.cpp
```diff
@@ -435,16 +481,28 @@ MLEBABecLap::setEBHomogDirichlet (int amrlev, const MultiFab& beta)
...
+    }
 }
 
 void
 MLEBABecLap::setEBHomogDirichlet (int amrlev, Real beta)
 {
+    bool  phi_on_centroid = (m_phi_loc  == Location::CellCentroid);
+
+    if (phi_on_centroid) {
+      m_is_eb_inhomog = false;
```
WeiqunZhang: See the comment above.

Src/LinearSolvers/MLMG/AMReX_MLEBABecLap.cpp
```diff
@@ -490,16 +548,28 @@ MLEBABecLap::setEBHomogDirichlet (int amrlev, Real beta)
...
+    }
 }
 
 void
 MLEBABecLap::setEBHomogDirichlet (int amrlev, Vector<Real> const& hv_beta)
 {
+    bool  phi_on_centroid = (m_phi_loc  == Location::CellCentroid);
+
+    if (phi_on_centroid) {
+      m_is_eb_inhomog = false;
```
WeiqunZhang: See the comment above.

Src/LinearSolvers/MLMG/AMReX_MLEBABecLap.cpp
```diff
@@ -1345,7 +1440,7 @@ MLEBABecLap::applyBC (int amrlev, int mglev, MultiFab& in, BCMode bc_mode, State
 
     int m_is_inhomog = bc_mode == BCMode::Inhomogeneous;
     int flagbc = m_is_inhomog;
-    m_is_eb_inhomog = s_mode == StateMode::Solution;
+    if (!phi_on_centroid) m_is_eb_inhomog = s_mode == StateMode::Solution;
```
WeiqunZhang: I see why you made changes to `m_is_eb_inhomog` in `setEBDirichlet`.   Have to think about it.  Seems wrong to me.  If the multigrid solver is in correct form, the original phi on EB should not be used, I think.

asalmgren: If by "correct form" you mean "residual-correction form", then wouldn't s_mode == StateMode::Correction if that were true?

WeiqunZhang: Yes, I meant correction form (i.e., residual-correction form).  `StateMode` has only two possible values, `Solution` and `Correction`.  So `!= Solution` is equivalent to ` == Correction`.

Src/LinearSolvers/MLMG/AMReX_MLEBABecLap.cpp
drangara: thanks for pointing this out. currently the unit test calls [MLMG::apply()](https://github.com/drangara/amrex/blob/eb_viscous_explicit/Tests/LinearSolvers/LeastSquares/MyTest.cpp#L322)
which seems to call the `MLCellLinOp::solutionResidual()` by setting `inhomogeneous=true`. When `MLEBABecLap::applyBC()` is called, the inhomog flag is always true. I think removing the statements which set `m_is_eb_inhomog=false` inside `setEBHomogDirichlet` is therefore causing these tests to fail. Not sure if perhaps the test setup itself needs to change to use the "correction" form?

WeiqunZhang: Maybe there is a bug in the development branch or your PR branch.  But I think the logic in the development branch is not wrong.  Function `setEBHomogDirichlet` sets `phi` on EB to zero.  So setting `m_is_eb_inhomog=true` is not wrong for homogeneous Dirichlet because inhomogeneous boundary with a Dirichlet value of zero is homogeneous Dirichlet.

drangara: ah i see. let me check if there is a bug on my branch that is causing the failure when the flag is true. sounds like it should be able to handle the case since phi on EB should be set to 0. thank you.

drangara: moved the handling of `inhomog=true` into the slopes methods and removed setting it in `setEBHomogDirichlet` functions.

WeiqunZhang:
I have run AMReX regression tests.  Four GPU tests crashed because it ran out of memory.  Looks like it's because the new functions use a lot more memory because of local variables like `Array2D<Real,0,35,0,9> Amatrix`.  So maybe we should keep the old `mlebabeclap_adotx` and simply add a new version of the whole adotx function for the phi-on-centroid case.  The test on whether phi is on centroid is then done on CPU and different kernels are launched depending on the location of phi.  I think in that case at least the currently default phi-on-cell-center case will not use that much memory.

@maxpkatz I run the same test on one computer with 6 GB memory and cuda 9.1, and another with 8 GB memory and cuda 11.0.  It's fine on the one with 6 GB, but it runs out of memory on the computer with 8 GB.  Kind of strange.

WeiqunZhang:
https://ccse.lbl.gov/pub/GpuRegressionTesting/AMReX/2021-01-19-001/index.html  Here are the regression tests.

WeiqunZhang (approved):

PR #1721, opened 2021-01-11
RoundRobin: option to not sort processes

WeiqunZhang:
## Summary

Add an optional parameter `sort` to RoundRobin constructors.  This makes it
consistent with other DistributionMapping strategies.  And this fixes an
incorrect behavior when RoundRobin is used by KnapSack that has been given
`sort=false`.  This should be able to fix the hanging issue observed in
WarpX simulations with load balancing enabled and more MPI processes than
the number of boxes.

## Additional background

https://github.com/ECP-WarpX/WarpX/issues/1614

## Checklist

The proposed changes:


ax3l:
cc @mrowan137 @guj

ax3l (approved):
Thank you for this!
I did a small local test with this and can confirm this fixes the issue for me :sparkles:

PR #1726, opened 2021-01-14
Update Make.nrel to accomodate MPT MPI with CUDA enabled

jrood-nrel:
## Summary

For our main computer at NREL, Eagle, we mostly use HPE's MPT for MPI. Until these changes, CUDA did not work with MPT when using the GNU makefiles.

## Checklist

The proposed changes:

### DIFF
--- a/Tools/GNUMake/sites/Make.nrel
+++ b/Tools/GNUMake/sites/Make.nrel
@@ -40,7 +40,7 @@ else ifeq ($(which_computer), rhodes)
   endif
 endif
 
-# Account for Intel-MPI, MPICH, OpenMPI, and MPT
+# Account for Intel-MPI, MPICH, OpenMPI, and HPE MPT
 ifeq ($(USE_MPI),TRUE)
   ifeq ($(COMP), intel)
     CXX := mpiicpc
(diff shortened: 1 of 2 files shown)


sayerhs (approved):

Tools/GNUMake/sites/Make.nrel
```diff
@@ -58,6 +58,14 @@ ifeq ($(USE_MPI),TRUE)
...
       mpif90_link_flags := $(shell $(F90) -showme:link)
       LIBRARIES += $(mpif90_link_flags)
+    else
+      # MPT case (no option available to query link flags)
+      LIBRARIES += -lmpi
+      ifeq ($(USE_CUDA),TRUE)
+        export MPICXX_CXX := nvcc
+        export MPICC_CC   := nvcc
+        export MPIF90_F90 := $(F90)
+      endif
```
WeiqunZhang: Maybe we can do this in `Tools/GNUMake/Make.defs` at line 678.  We are already setting `OMPI_CXX`, `MPICH_CXX`, etc.

jrood-nrel: I noticed that, but didn't want to do it if it was too NREL-specific. I can make the change and update the PR.

WeiqunZhang: I assume that's Cray's way.  So adding it to `Make.defs` might help others.

WeiqunZhang (approved):

PR #1739, opened 2021-01-20
HIP indirect function and kernel fusing

WeiqunZhang:
Add a GNU Make flag, HIP_INDIRECT_FUNCTION=[TRUE|FALSE], and modify the Gpu
kernel fusing test.

Currently HIP does not support indirect function calls via device function
pointers, which our GPU fusing implementation depends on.

### DIFF
--- a/Tests/GPU/Fuse/main.cpp
+++ b/Tests/GPU/Fuse/main.cpp
@@ -82,7 +82,7 @@ int main(int argc, char* argv[])
         mfc.setVal(-1);
 
         {
-            BL_PROFILE("fuesd-test1");
+            BL_PROFILE("fused-test1");
             fused_test(mfa,mfb,mfc);
         }
         r = r && verify(mfc);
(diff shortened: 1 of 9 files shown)


Tests/GPU/Fuse/main.cpp
```diff
@@ -43,56 +43,75 @@ void unfused_test (MultiFab const& mfa, MultiFab const& mfb, MultiFab& mfc)
...
-        mfc.setVal(-1.0);
+        iMultiFab mfa(ba,dm,1,0);
+        iMultiFab mfb(ba,dm,1,0);
+        iMultiFab mfc(ba,dm,2,0);
+        mfa.setVal(1);
+        mfb.setVal(3);
+        mfc.setVal(-1);
 
         {
             BL_PROFILE("fuesd-test1");
```
kngott: Misspelled: should be "fused-test1".

kngott: Since we're sharing this...

WeiqunZhang: fixed.

WeiqunZhang:
Results on v100.
```
Test passes.
Unfused kernels take 0.00373884 seconds.
Fused   kernels take 0.000714362 seconds.
```

kngott (approved):

PR #1741, opened 2021-01-21
Template copy/add particles based on src type.

atmyers:
This is needed so that that they can used between PCs that have different memory allocators, for example from device to pinned memory particle containers.

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_ParticleContainerI.H
+++ b/Src/Particle/AMReX_ParticleContainerI.H
@@ -987,9 +987,10 @@ clearParticles ()
 
 template <int NStructReal, int NStructInt, int NArrayReal, int NArrayInt,
           template<class> class Allocator>
+template <class PCType, amrex::EnableIf_t<IsParticleContainer<PCType>::value, int> foo>
 void
 ParticleContainer<NStructReal, NStructInt, NArrayReal, NArrayInt, Allocator>::
-copyParticles (const ParticleContainerType& other, bool local)
+copyParticles (const PCType& other, bool local)
 {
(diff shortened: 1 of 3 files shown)


Src/Particle/AMReX_ParticleContainerI.H
```diff
@@ -987,31 +987,34 @@ clearParticles ()
 
 template <int NStructReal, int NStructInt, int NArrayReal, int NArrayInt,
           template<class> class Allocator>
+template <class PCType, amrex::EnableIf_t<IsParticleContainer<PCType>::value, int> foo>
```
WeiqunZhang, line 990: I thought you need `... foo = 0` to make it an optional parameter.

WeiqunZhang: Never mind. It has been declared elsewhere.

Src/Particle/AMReX_ParticleContainerI.H
```diff
@@ -987,31 +987,34 @@ clearParticles ()
...
     addParticles(other, [=] AMREX_GPU_HOST_DEVICE (const PData& /*data*/, int /*i*/) { return 1; }, local);
 }
 
 template <int NStructReal, int NStructInt, int NArrayReal, int NArrayInt,
           template<class> class Allocator>
-template <class F,
-          amrex::EnableIf_t<! std::is_integral<F>::value, int> foo>
+template <class F, class PCType,
+          amrex::EnableIf_t<IsParticleContainer<PCType>::value, int> foo,
+          amrex::EnableIf_t<! std::is_integral<F>::value, int> bar>
```
WeiqunZhang, line 1014: You could merge the two into one.

WeiqunZhang (approved):

PR #1742, opened 2021-01-21
HIP: New Warp Size Macro

ax3l:
## Summary

Use new the HIP macro for the warp size on various AMD devices.

## Additional background

Credits to @psychocoderHPC and the mallocMC developers (https://github.com/alpaka-group/mallocMC/pull/204)

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_GpuDevice.H
+++ b/Src/Base/AMReX_GpuDevice.H
@@ -123,7 +123,15 @@ public:
     static std::size_t freeMemAvailable ();
 
 #ifdef AMREX_USE_GPU
+
+#   if defined(AMREX_USE_HIP) && (HIP_VERSION_MAJOR >= 4)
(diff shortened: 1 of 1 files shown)


WeiqunZhang (approved):

Src/Base/AMReX_GpuDevice.H
```diff
@@ -123,7 +123,15 @@ public:
     static std::size_t freeMemAvailable ();
 
 #ifdef AMREX_USE_GPU
+
+#   if defined(AMREX_USE_HIP) && (HIP_VERSION_MAJOR >= 4)
```
psychocoderHPC, line 127: Note: You could also write `#if defined(AMREX_USE_HIP) && defined(__AMDGCN_WAVEFRONT_SIZE)`
But I decided to use exactly what you implemented in this PR because this will guarantee that we use the new macro together with the latest ROCM versions and not silently switch to the old implementation.

ax3l: Thanks for the note!

PR #1746, opened 2021-01-23
Add additional tests when interpolating from cell centroids to face c…

asalmgren:
…entroids --

when the face centroid is exactly at the center of the face we need an additional
test to know which way we can look transversely and not hit covered cells.

## Checklist

The proposed changes:
- [X ] fix a bug or incorrect behavior in AMReX
- [X ] changes answers in the test suite to more than roundoff level

### DIFF
--- a/Src/EB/AMReX_EBMultiFabUtil_2D_C.H
+++ b/Src/EB/AMReX_EBMultiFabUtil_2D_C.H
@@ -415,7 +415,15 @@ void eb_interp_centroid2facecent_x (Box const& ubx,
       else 
       {
           int ii = i - 1;
-          int jj = (fcx(i,j,k) < 0.0) ? j - 1 : j + 1;
+          // We must add an additional test to avoid the case where fcx is very close to zero, but (j-1) or (j+1)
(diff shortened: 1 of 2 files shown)


Src/EB/AMReX_EBMultiFabUtil_2D_C.H
```diff
@@ -415,7 +415,15 @@ void eb_interp_centroid2facecent_x (Box const& ubx,
       else 
       {
           int ii = i - 1;
-          int jj = (fcx(i,j,k) < 0.0) ? j - 1 : j + 1;
+          // We must add an additional test to avoid the case where fcx is very close to zero, but (j-1) or (j+1)
+          //    might be covered cells -- this can happen when the EB is exactly aligned with the horizontal grid lines
+          int jj;
+          if (std::abs(fcx(i,j,k)) > 1.e-8)
```
WeiqunZhang:
```suggestion
          if (amrex::Math::abs(fcx(i,j,k)) > 1.e-8)
```

WeiqunZhang (approved):

PR #1753, opened 2021-01-26
Fix out of bound error in AsyncWrite

jmsexton03:
## Summary
Fix out of bound error when the local number of fabs is zero.

## Checklist

The proposed changes:


Src/Base/AMReX_VisMF.cpp
```diff
@@ -2236,7 +2236,7 @@ VisMF::AsyncWriteDoit (const FabArray<FArrayBox>& mf, const std::string& mf_name
     bool strip_ghost = valid_cells_only && mf.nGrowVect() != 0;
 
     int64_t total_bytes = 0;
-    auto pld = (char*)(&(localdata[1]));
+    auto pld = (char*)(&(localdata[0]));
```
WeiqunZhang: The fix should be
```suggestion
    char* pld = (localdata.size() > 1) ? (char*)(&(localdata[1])) : nullptr;
```

WeiqunZhang: It's a bug, but it's not a typo.  The index is meant to be 1.

WeiqunZhang (approved):

WeiqunZhang:
@jmsexton03 Can you tes if it works for you now?

PR #1757, opened 2021-01-27
Anyof

kngott:
## Summary

* DPC++ versions of AnyOf. Written to match CUDA/HIP version. Performance not explicitly tested.
* Also, adjustment to AMREX_DEVICE_PRINTF to remove the need for AMREX_DEVICE_COMPILE.
* If desired, I can delete the test.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_Reduce.H
+++ b/Src/Base/AMReX_Reduce.H
@@ -630,7 +630,8 @@ bool AnyOf (N n, T const* v, P&& pred)
     ec.numBlocks.x = std::min(ec.numBlocks.x, Gpu::Device::maxBlocksPerLaunch());
 
 #ifdef AMREX_USE_DPCPP
-    const std::size_t shared_mem_bytes = sizeof(int);
+    const int num_ints = std::max(Gpu::Device::warp_size, int(ec.numThreads.x)/Gpu::Device::warp_size); 
+    const std::size_t shared_mem_bytes = num_ints*sizeof(int);
     amrex::launch(ec.numBlocks.x, ec.numThreads.x, shared_mem_bytes, Gpu::nullStream(),
     [=] AMREX_GPU_DEVICE (Gpu::Handler const& gh) noexcept {
         int* has_any = static_cast<int*>(gh.sharedMemory());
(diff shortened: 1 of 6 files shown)


WeiqunZhang:
There is a bug.  Too little shared memory is allocated for DPC++.  `blockReduce` needs `sizeof(T) * max(warpsize,blocksize/warpsize)` bytes of shared memory.

WeiqunZhang:
Maybe we should add a function to return the minimal requirement for shared memory for DPC++ to make it more friendly.

Src/Base/AMReX_Reduce.H
```diff
@@ -630,7 +630,8 @@ bool AnyOf (N n, T const* v, P&& pred)
     ec.numBlocks.x = std::min(ec.numBlocks.x, Gpu::Device::maxBlocksPerLaunch());
 
 #ifdef AMREX_USE_DPCPP
-    const std::size_t shared_mem_bytes = sizeof(int);
+    const int num_ints = std::max(Gpu::Device::warp_size, int(ec.numThreads.x)/Gpu::Device::warp_size);
```
WeiqunZhang: I think you either have to allocate space for one more int to be used by the `has_any` flag, or add an extra barrier to make sure no threads can touch it before `if (!(*has_any))` finishes.

kngott: My first attempt included an extra shared memory int and pointed (has_any) at the final location in the shared memory space. However, the answer in the test was wrong. I looked like the value of "has_any" was being included in the reduction.

Rather than try to alter the reduction, I opted to reuse the memory space. The blockReduce function performs a pre-emptive sharedBarrier before writing to it the first time. Is that sufficient, or should we add an additional barrier in AnyOf for safety? Immediately before the blockReduce call, maybe?

WeiqunZhang: Having extra space at the final location alter the reduction answer does not make sense.

kngott: You're right. Seems to be working now. Must've been related to the indexing error I fixed in the last push.

WeiqunZhang (approved):

PR #1765, opened 2021-01-30
ParallelCopy_nowait & ParallelCopy_finish

kngott:
## Summary
Most direct implementation of `ParallelCopy_nowait` and `ParallelCopy_finish`, with minimal changes.

Previous version had a max component limit per message, set by the input parameter `fabarray.maxcomp`, 25 by default. This is preserved. If `ncomp > fabarray.maxcomp`, `ParallelCopy_nowait` calls `ParallelCopy_finish` internally for all iterations of the loop except the final iteration. For most apps using GPUs that haven't been manually setting `fabarray.maxcomp`, this will almost certainly mean behavior identical to `FillBoundary_nowait`, as `ncomp > 25` would require very small `MultiFabs` to fit on current GPUs.

Keeps `ParallelCopy`, which simply calls `ParallelCopy_nowait` followed by `ParallelCopy_finish`.

No doubt, there are potential optimizations available: e.g. if `ncomp>fabarray.maxcomp` and `ncomp%fabarray.maxcomp != 0` returning from `nowait` with a message with a full `fabarray.maxcomp` components may allow the largest possible overlap. But, this version gets the feature running while we discuss additional improvements such as this.

Plan to follow up with a scaling study on the effect of message size on `ParallelCopy` timing on Summit by exploiting the `fabarray.maxcomp` feature to see if rewriting with more controlled buffer sizes yields enough benefit to justify a rewrite.

## Additional background
Tested on dogora and Summit up to 64 GPUs and compared to previous one function version. Also tested with `ncomp > fabarray.maxcomp` (e.g 4 > 1). Yielded correct results and consistent timings with the previous one function version, e.g:

```
Name                                        NCalls  Incl. Min  Incl. Avg  Incl. Max   Max %
**** ParallelCopy - 3                            1  1.846e-06    0.01367    0.01502   9.40%
---- ParallelCopy_nowait -- 3                    1  3.058e-06     0.0137    0.01495   9.35%
**** ParallelCopy - 2                            1  2.299e-06    0.01346    0.01474   9.22%
---- ParallelCopy_nowait -- 1                    1  5.748e-06    0.01321    0.01465   9.17%
**** ParallelCopy - 1                            1  2.388e-06    0.01282    0.01465   9.17%
---- ParallelCopy_nowait -- 2                    1  7.208e-06    0.01049    0.01302   8.15%
```

Test included. Can be removed to limit clutter, once unneeded.

## Checklist

The proposed changes:


Src/Base/AMReX_FabArray.H
```diff
@@ -472,6 +472,18 @@ public:
...
         { ParallelCopy(fa,period,op); }
 
+    void ParallelAdd_nowait(const FabArray<FAB>& fa,
+                            const Periodicity& period = Periodicity::NonPeriodic())
+       { ParallelCopy_nowait(fa,period,FabArray::ADD); }
+    void ParallelCopy_nowait (const FabArray<FAB>& fa,
+                              const Periodicity&   period = Periodicity::NonPeriodic(),
+                              CpOp                 op = FabArrayBase::COPY)
+       { ParallelCopy_nowait(fa,0,0,nComp(),0,0,period,op); }
+    void copy_nowait (const FabArray<FAB>& fa,
```
WeiqunZhang: We don't need provide `copy_nowait`.  We only kept `FabArray::copy` for backward compatibility.

Src/Base/AMReX_FabArrayCommI.H
```diff
@@ -408,73 +436,120 @@ FabArray<FAB>::ParallelCopy (const FabArray<FAB>& src,
...
+
+template <class FAB>
+void
+FabArray<FAB>::ParallelCopy_finish ()
+{
+    BL_PROFILE("FabArray::ParallelCopy_finish()");
+
+    if (size() == 0 || pc_src->size() == 0) return;
+
+    if ((pc_src->boxArray().ixType().cellCentered() || pc_op == FabArrayBase::COPY) &&
```
WeiqunZhang: The test here is non-trivial.  So maybe we should add a boolean flag and set it in the `nowait` function for this shortcut test and others that allow an immediate return.

Src/Base/AMReX_FabArray.H
```diff
@@ -864,6 +938,31 @@ public:
...
+    int                  pc_SC, pc_NC, pc_DC;
+
+    //
+    char*                pc_the_recv_data = nullptr;
+    char*                pc_the_send_data = nullptr;
+    Vector<int>          pc_recv_from;
+    Vector<char*>        pc_recv_data;
+    Vector<std::size_t>  pc_recv_size;
+    Vector<MPI_Request>  pc_recv_reqs;
+//    Vector<MPI_Status>  pc_recv_stat;  // why kept in FB? Needed in PC?
```
WeiqunZhang: I think it's because we have `FillBoundary_test`, we must save the MPI_Status for FillBoundary.  Not needed in PC.

Src/Base/AMReX_FabArrayCommI.H
```diff
@@ -275,7 +303,7 @@ FabArray<FAB>::ParallelCopy (const FabArray<FAB>& src,
         // we're doing plus()s on cell-centered data.  Don't do plus()s on
         // non-cell-centered data this simplistic way.
         //
-#ifdef AMREX_USE_OMP
+#ifdef _OPENMP
```
WeiqunZhang: This change should be reverted.  We prefer `AMREX_USE_OMP` to `_OPENMP`.  See a PR early on this subject.

Tests/ParallelCopyBuffer/main.cpp
```diff
@@ -0,0 +1,143 @@
+#include <AMReX.H>
```
WeiqunZhang: Let's remove the test as we have discussed.

WeiqunZhang (changes requested):

WeiqunZhang (approved):

PR #1768, opened 2021-02-02
IntVect::toArray()

ax3l:
## Summary

Add a ::toArray() conversion member to IntVect that can be used to convert back to an array type and cast the result on the way, e.g. to unsigned.

## Additional background

Functions in https://github.com/ECP-WarpX/WarpX/pull/1660

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_IntVect.H
+++ b/Src/Base/AMReX_IntVect.H
@@ -135,6 +135,17 @@ public:
 #endif
     }
 
+#if __cplusplus >= 201402L
+    template< typename T = int >
+    AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+    Array<T, AMREX_SPACEDIM>
+    toArray () const noexcept {
+        return Array<T, AMREX_SPACEDIM>{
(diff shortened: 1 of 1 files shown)


Src/Base/AMReX_IntVect.H
```diff
@@ -135,6 +135,15 @@ public:
 #endif
     }
 
+    template< typename T = int >
+    AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+    Array<T, AMREX_SPACEDIM>
+    toArray () const noexcept {
+        return Array<T, AMREX_SPACEDIM>{
```
WeiqunZhang, line 143: `Array` is `std::array` that does not work on device for C++11.

ax3l: Oh, thanks for the hint!
Would guarding this method with `__cplusplus` for C++14 be sufficient (now updated)? I can alternatively make this a standalone function.

WeiqunZhang (approved):

PR #1770, opened 2021-02-02
Make.unknown MPI Detection Update

kngott:
## Summary
Simplify/robust-ify Make.unknown's MPI detection.

**** Suggest a thorough testing on "unknown" systems to confirm this works across compiler versions and environments.

## Additional background
* Tests for error codes to determine appropriate flag to return wrapper arguments.
* Filters all flags out of the wrapper, based on a leading dash.
* Keeps one specific compiler test (OpenMPI) for a recent flag to remove a warning. (Way around this?)
* Tested on dogora, checked compiler behavior was consistent with this method on Cori and Summit. However, needs thorough testing on other "unknown" systems.

## Checklist

The proposed changes:

### DIFF
--- a/Tools/GNUMake/sites/Make.unknown
+++ b/Tools/GNUMake/sites/Make.unknown
@@ -20,28 +20,35 @@ ifeq ($(USE_MPI),TRUE)
     MPI_OTHER_COMP := mpif90
   endif
 
-  # error codes - find flag that displays the data
+  # error codes - find flag that displays the link line wrapper. 
   code_sml := $(shell $(MPI_OTHER_COMP) -showme:link > /dev/null 2>&1; echo $$?)
   code_li :=  $(shell $(MPI_OTHER_COMP) -link_info   > /dev/null 2>&1; echo $$?)
 
-  # f90: filter all flags (-%)
-  # cxx: filter additional library flags (-l%)
+  # use correct flag to store link line and cxx libraries (-l%).
   ifeq ($(code_sml), 0)
-     mpif90_link_flags := $(filter -%,$(shell $(MPI_OTHER_COMP) -showme:link))
-     mpicxx_link_flags := $(filter -l%,$(shell mpicxx -showme:link))
+     mpi_link_flags  := $(shell $(MPI_OTHER_COMP) -showme:link)
+     mpicxx_link_libs := $(filter -l%,$(shell mpicxx -showme:link))
   else ifeq ($(code_li), 0)
-     mpif90_link_flags := $(filter -%,$(shell $(MPI_OTHER_COMP) -link_info))
-     mpicxx_link_flags := $(filter -l%,$(shell mpicxx -link_info))
+     mpi_link_flags  := $(shell $(MPI_OTHER_COMP) -link_info)
+     mpicxx_link_libs := $(filter -l%,$(shell mpicxx -link_info))
   else
      $(error Unknown mpi wrapper.  You can try setting MPI stuff in amrex/Tools/GNUMake/Make.local and then compile with NO_MPI_CHECKING=TRUE.)
   endif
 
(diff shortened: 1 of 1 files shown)


drummerdoc (approved):
Works for me on OSX+brew

kngott:
Which MPI? (and/or do you have multiple you could try)?

drummerdoc:
On my newer Nrel MacBook, I have a pretty clean install of use openmpi 4.0.5; not real keen on loading up additional ones.  Rood might be a good person to bring in to test.

kngott:
No need to load up additional ones. Just confirming what seems to be working. Given OpenMPI has the extra flag, that's a good sign. :)

Not opposed to extra testing, maybe we follow up with Rood after Weiqun takes a look

Tools/GNUMake/sites/Make.unknown
```diff
@@ -20,101 +20,34 @@ ifeq ($(USE_MPI),TRUE)
...
   else
+     $(error Unknown mpi wrapper.  You can try setting MPI stuff in amrex/Tools/GNUMake/Make.local and then compile with NO_MPI_CHECKING=TRUE.)
+  endif
 
-    $(error Unknown mpi implementation.  You can try setting MPI stuff in amrex/Tools/GNUMake/Make.local and then compile with NO_MPI_CHECKING=TRUE.)
+  LIBRARIES += $(mpif90_link_flags) $(mpicxx_link_flags)
 
+  # OpenMPI specific flag
+  ifneq ($(findstring Open MPI, $(shell $(MPI_OTHER_COMP) -showme:version 2>&1)),)
+     DEFINES += -DOMPI_SKIP_MPICXX
```
WeiqunZhang, line 50: It's probably safe to always define this.

Tools/GNUMake/sites/Make.unknown
```diff
@@ -20,101 +20,34 @@ ifeq ($(USE_MPI),TRUE)
...
+  code_sml := $(shell $(MPI_OTHER_COMP) -showme:link > /dev/null 2>&1; echo $$?)
+  code_li :=  $(shell $(MPI_OTHER_COMP) -link_info   > /dev/null 2>&1; echo $$?)
+
+  # f90: filter all flags (-%)
+  # cxx: filter additional library flags (-l%)
+  ifeq ($(code_sml), 0)
+     mpif90_link_flags := $(filter -%,$(shell $(MPI_OTHER_COMP) -showme:link))
+     mpicxx_link_flags := $(filter -l%,$(shell mpicxx -showme:link))
+  else ifeq ($(code_li), 0)
+     mpif90_link_flags := $(filter -%,$(shell $(MPI_OTHER_COMP) -link_info))
```
WeiqunZhang: I don't think it is safe to filter-in `-`.  We should instead filter out the first word if it doe not start with `-`.  On one machine I am testing `mpif90 -link_info` gives
```
gfortran .... -Xlinker '/glob/.....'.
``` 
Because of the space between `-Xlinker` and `/glob/...`, the latter gets dropped.  This could also happen to `-I xxx` and `-L xxx`.

WeiqunZhang (changes requested):

Tools/GNUMake/sites/Make.unknown
kngott: I was worried about that one. I believe "eliminate the first word if it doesn't start with a dash" should be pretty doable. I'll try writing that instead.

Tools/GNUMake/sites/Make.unknown
kngott: 👍 I'll comment the if-statement out for now (and future reference) and we can test to make sure it doesn't break anything.

WeiqunZhang (approved):

PR #1772, opened 2021-02-03
For FillPatchTwoLevels, BCs should be set on the coarse box, since th…

cgilet:
…ey are only used on the coarse level (when computing slopes).

## Summary
This corrects certain cases where a level of refinement is very close to, but not actually touching an ext_dir or hoextrap boundary.

## Checklist

The proposed changes:


WeiqunZhang:
How many fine cells away is the fine level from the boundary?  It looks we need to change it in other similar places including AmrLevel's FillPatchIteratorHelper::fill.

cgilet:
The exact number depends on the code and how many ghost cells are needed. I saw this in IAMR when the valid fine cell was 5 cells away from the boundary and 3 ghost cells needed to be filled, so that the ghost cell closest to the boundary was 1 cell away from it. So while the fine cell didn't touch the boundary, the coarse cell that does touch the boundary was needed for the interpolation. 

I'll go through the code and correct similar cases, including FillPatchIteratorHelper::fill

WeiqunZhang (approved):

PR #1773, opened 2021-02-03
MSVC: Proper __cplusplus macro

ax3l:
## Summary

Modern MSVC versions finally set the __cplusplus macro correctly, but we need to request this with an additional flag.

## Additional background

- https://docs.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=msvc-160
- https://gitlab.kitware.com/cmake/cmake/-/issues/18837

## Checklist

The proposed changes:


mic84 (approved):

ax3l:
Huh, how does this break the `linux / CUDA@11.2 GNU@9.3.0 C++17 Release` build :thinking:

PR #1774, opened 2021-02-03
Refinement Ratio of 4 Support in Nodal Solver

WeiqunZhang:
This adds refinement ratio of 4 support in non-EB nodal linear solver.
Roundoff errors are expected for multi-level solves with a refinement ratio
of 2.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_Box.H
+++ b/Src/Base/AMReX_Box.H
@@ -259,6 +259,21 @@ public:
 	BL_ASSERT(sameType(b));
 	return b.smallend.allGT(smallend) && b.bigend.allLT(bigend);
     }
+
+    //! Returns true if argument is strictly contained within Box.
+    AMREX_GPU_HOST_DEVICE
+#if (AMREX_SPACEDIM == 1)
+    bool strictly_contains (int i, int, int) const noexcept {
+#elif (AMREX_SPACEDIM == 2)
+    bool strictly_contains (int i, int j, int) const noexcept {
+#else
+    bool strictly_contains (int i, int j, int k) const noexcept {
+#endif
+        return AMREX_D_TERM(i > smallend[0] && i < bigend[0],
+                         && j > smallend[1] && j < bigend[1],
+                         && k > smallend[2] && k < bigend[2]);
+    }
+
     /**
     * \brief Returns true if Boxes have non-null intersections.
     * It is an error if the Boxes have different types.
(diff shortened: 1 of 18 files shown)


WeiqunZhang:
@cgilet Will you be able to test this with IAMR?

asalmgren (approved):

cgilet:
I'm getting an out of bounds error in setBndryValues during the MAC projection. Is that supposed to be working with ref_ratio==4?

Here's part of the backtrace:
```
amrex::InterpBndryData::BndryValuesDoIt(amrex::BndryRegister&, int, amrex::MultiFab const*, int, int, int, amrex::IntVect const&, amrex::BCRec const*, int) at /home/candace/CCSE/IAMR_cleanup/Exec/run3d/../../../amrex_weiqun/amrex/Src/Base/AMReX_Array4.H:96
 (inlined by) ?? at /home/candace/CCSE/IAMR_cleanup/Exec/run3d/../../../amrex_weiqun/amrex/Src/Boundary/AMReX_InterpBndryData_3D_K.H:34
 (inlined by) amrex::InterpBndryData::BndryValuesDoIt(amrex::BndryRegister&, int, amrex::MultiFab const*, int, int, int, amrex::IntVect const&, amrex::BCRec const*, int) at /home/candace/CCSE/IAMR_cleanup/Exec/run3d/../../../amrex_weiqun/amrex/Src/Boundary/AMReX_InterpBndryData.cpp:225

amrex::InterpBndryData::setBndryValues(amrex::BndryRegister&, int, amrex::MultiFab const&, int, int, int, amrex::IntVect const&, amrex::BCRec const&, int) at /home/candace/CCSE/IAMR_cleanup/Exec/run3d/../../../amrex_weiqun/amrex/Src/Boundary/AMReX_InterpBndryData.cpp:144
```

WeiqunZhang:
MAC projection is supposed to work.  Could you push your IAMR change somewhere so that I can reproduce the error you see.

WeiqunZhang:
The problem is in `MacProjector::setDomainBC`.  It should not call `setLevelBC`, which needs to know the amr refinement ratio.  We did it this way to make `MacProjector` more user friendly.  But unfortunately it breaks for refinement ratio of 4.  So we will have to change this, and it will be the user's responsibility to call `setLevelBC`.

WeiqunZhang:
Actually, there is a way to keep the current behavior of calling `setLeveBC` implicitly.

PR #1778, opened 2021-02-04
Overset Solver with Refinement Ratio of 4

WeiqunZhang:
We need to define overset mask for on coarse MG levels of fine AMR levels.
We didn't need to do it for refinement ratio of 2 because there are no
coarse MG levels on fine AMR levels.


sayerhs (approved):
Works for simple overset problem

PR #1783, opened 2021-02-09
Fix precision issue in ParmParse::add

WeiqunZhang:
## Summary

When adding values to ParmParse, we need to call `std::setprecision` so as
not to lose precision.

## Checklist

The proposed changes:


WeiqunZhang:
Thank @dpgrote for reporting this!

atmyers (approved):

PR #1790, opened 2021-02-15
fix table formatting in GPU.rst

asalmgren:
## Checklist

The proposed changes:


sayerhs (approved):

Docs/sphinx_documentation/source/Post_Processing.rst
```diff
@@ -20,7 +20,7 @@ row/column formatting, etc.
 
 
 fextract
-====================
+========
```
sayerhs, line 22: Is there some utf8 character issue here?

asalmgren: rst wants the length of the underline to match the length of the word(s)

PR #1791, opened 2021-02-15
Disable the pdf build of the docs for now.

WeiqunZhang:
Several PRs are stuck at building docs.  It seems to be related to building with pdflatex.


asalmgren (approved):

WeiqunZhang:
Closing this PR, because the pdf is not the issue.  The issue appears to be Sphinx v3.5.0 that is used by the current CI.  It used to be v3.4.3.

PR #1793, opened 2021-02-16
Fortran Interfaces for new FillPatch

kngott:
## Summary
Fortran Interfaces for new FillPatch. Compiles, and I believe the indexing is correct, but didn't test a Fortran implementation.

I welcome adjustments and checking of my indexing.

## Checklist

The proposed changes:

### DIFF
--- a/Src/F_Interfaces/AmrCore/AMReX_fillpatch_mod.F90
+++ b/Src/F_Interfaces/AmrCore/AMReX_fillpatch_mod.F90
@@ -221,7 +221,7 @@ contains
     type(amrex_multifab), intent(in   ) :: mfold_c(amrex_spacedim), mfnew_c(amrex_spacedim)
     type(amrex_multifab), intent(in   ) :: mfold_f(amrex_spacedim), mfnew_f(amrex_spacedim)
     integer, intent(in) :: scomp, dcomp, ncomp, rr, interp
-    integer, dimension(amrex_spacedim,scomp+ncomp-1), target, intent(in) :: lo_bc, hi_bc
+    integer, dimension(amrex_spacedim,scomp+ncomp-1,amrex_spacedim), target, intent(in) :: lo_bc, hi_bc
     real(amrex_real), intent(in) :: told_c, tnew_c, told_f, tnew_f, time
     type(amrex_geometry), intent(in) :: geom_c, geom_f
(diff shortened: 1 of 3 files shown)


WeiqunZhang:
Made a few changes.  Please check if you agree.

Src/F_Interfaces/AmrCore/AMReX_fillpatch_mod.F90
```diff
@@ -184,6 +201,135 @@ subroutine amrex_fillpatch_two (mf, told_c, mfold_c, tnew_c, mfnew_c, geom_c, fi
...
+        &                                   fill_physbc_fz, &
+#endif
+#endif
+        &                               time, scomp, dcomp, ncomp, rr, interp, lo_bc, hi_bc, &
+        &                               pre_interp, post_interp)
+    type(amrex_multifab), intent(inout) :: mf(amrex_spacedim)
+    type(amrex_multifab), intent(in   ) :: mfold_c(amrex_spacedim), mfnew_c(amrex_spacedim)
+    type(amrex_multifab), intent(in   ) :: mfold_f(amrex_spacedim), mfnew_f(amrex_spacedim)
+    integer, intent(in) :: scomp, dcomp, ncomp, rr, interp
+    integer, dimension(amrex_spacedim,scomp+ncomp-1), target, intent(in) :: lo_bc, hi_bc
```
WeiqunZhang: We have an issue here.  This should be a 3D array, `(amrex_spacedim,amrex_spacedim,scomp+ncomp-1)`.  Let's suppose `scomp = 1` and `ncomp = 1`, then we need 9, not 3 ints to specify the lower BC, because there are lo-x, lo-y and lo-z directions for x-, y- and z-faces.  I think we should simplify this and assume that there is only one component and scomp = 1 so that we can drop the component dimension.

WeiqunZhang (changes requested):

Src/F_Interfaces/AmrCore/AMReX_fillpatch_mod.F90
kngott: I agree with the intent changes in the PR. I'm double-checking that the new FillPatch uses 9 BCs correctly as is. Then, we can either do this simplification or take the time to make it robust? Depends on how difficult it is, I suppose.

WeiqunZhang (approved):

PR #1794, opened 2021-02-16
Add FabArray::tags() to return the tags.

WeiqunZhang:


atmyers (approved):

ax3l:
cc @sayerhs this can be useful for Python bindings querying available fabs :)

PR #1800, opened 2021-02-18
add interface between sundials SUNMemory and the Arena

balos1:
## Summary

As @asalmgren @jmsexton03 @mic84 @ajnonaka @dwillcox @gardner48 @cswoodward and I discussed... this PR adds an interface between the AMReX Arena and the SUNDIALS SUNMemory API. This is currently used in Pele and will be used in Nyx also. In general, any AMReX based app that also uses SUNDIALS (and its provided data structures) will likely find this useful when targeting GPUs.

CC @drummerdoc 

## Additional background

In SUNDIALS 5.4.0, the SUNMemory API was introduced to allow users to substitute their own memory management underneath SUNDIALS data structures. In Pele, using the AMReX arena underneath the SUNDIALS data structures results in as much as a 15% speedup.

I have tried to structure the code to follow AMReX conventions. Since we also will be bringing in an interface between the MultiFab and the SUNDIALS N_Vector, I have also tried to structure the code so that it will be easy to add that interface in as well.

## TODO 
- [x] Add optional SUNDIALS support in the AMReX build system (@mic84  can you help with this?)
- [x] Add DPC++ support in the interface
- [x] Add documentation
- [x] Verify/test in PelePhysics

## Checklist

The proposed changes:


Src/Extern/SUNDIALS/AMReX_SUNMemory.H
```diff
@@ -0,0 +1,22 @@
+#include <AMReX.H>
```
WeiqunZhang, line 4: Can we add include guard here?

Src/Extern/SUNDIALS/AMReX_Sundials.H
```diff
@@ -0,0 +1 @@
+#include <AMReX_SUNMemory.H>
```
WeiqunZhang, line 4: Include guard here too?

Tools/CMake/AMReXSetDefines.cmake
```diff
@@ -153,6 +153,12 @@ endif ()
 add_amrex_define(AMREX_USE_HDF5 NO_LEGACY IF AMReX_HDF5)
 add_amrex_define(AMREX_USE_HDF5_ASYNC NO_LEGACY IF AMReX_HDF5_ASYNC)
 
+
+#
+# SUNDIALS
+#
+add_amrex_define(AMREX_USE_SUNDIALS IF AMReX_SUNDIALS)
```
WeiqunZhang: Add `NO_LEGACY`?

Src/Extern/SUNDIALS/AMReX_SUNMemory.cpp
```diff
@@ -0,0 +1,140 @@
...
+    mem->ptr = NULL;
+    mem->own = SUNTRUE;
+    mem->type = mem_type;
+
+    if (mem_type == SUNMEMTYPE_HOST) {
+      mem->ptr = The_Cpu_Arena()->alloc(memsize);
+    } else if (mem_type == SUNMEMTYPE_UVM) {
+      mem->ptr = The_Managed_Arena()->alloc(memsize);
+    } else if (mem_type == SUNMEMTYPE_DEVICE) {
+      mem->ptr = The_Device_Arena()->alloc(memsize);
```
WeiqunZhang: I suggest we use `The_Arena` when it's possible so that the memory is less fragmented.

```suggestion
    } else if (mem_type == SUNMEMTYPE_UVM) {
        if (The_Arena()->isManaged()) {
          mem->ptr = The_Arena_Arena()->alloc(memsize);
        } else if (The_Managed_Arena()->isManaged()) {
          mem->ptr = The_Managed_Arena()->alloc(memsize);
        } else {
          free(mem);
          return -1;
        }
    } else if (mem_type == SUNMEMTYPE_DEVICE) {
      mem->ptr = The_Arena()->alloc(memsize);
```

Src/Extern/SUNDIALS/AMReX_SUNMemory.cpp
```diff
@@ -0,0 +1,140 @@
...
+    if (mem->type == SUNMEMTYPE_HOST) {
+      if (mem->own) The_Cpu_Arena()->free(mem->ptr);
+    } else if (mem->type == SUNMEMTYPE_UVM) {
+    if (mem->own) The_Managed_Arena()->free(mem->ptr);
+    } else if (mem->type == SUNMEMTYPE_DEVICE) {
+      if (mem->own) The_Device_Arena()->free(mem->ptr);
+    } else if (mem->type == SUNMEMTYPE_PINNED) {
+      if (mem->own) The_Pinned_Arena()->free(mem->ptr);
+    } else {
+      return(-1);
```
WeiqunZhang: I assume this means that the caller is responsible for freeing `mem`.

balos1: This should be

```c++
free(mem);
return -1;
```

Src/Extern/SUNDIALS/AMReX_SUNMemory.cpp
```diff
@@ -0,0 +1,148 @@
...
+    } else {
+      free(mem);
+      return -1;
+    }
+
+    *memptr = mem;
+    return 0;
+  }
+
+  int Dealloc(SUNMemoryHelper, SUNMemory mem)
```
WeiqunZhang, line 60: This function `Dealloc` needs to use the same Arena as `Alloc`.  If they do not match, there will be a runtime error.  Maybe we can put the logic in a function say
```
Arena* getArena (SUNMemoryType type)
{
    if (type == ...) {
        return ....;
    } else if (type == ...) {
    } ....
    else {
        return nullptr;
    } 
}
```

Then `Alloc` can use it and this function will be
```
if (mem == nullptr) return 0;
auto arena = getArena(mem->type);
if (arena) {
    arena->free(mem->ptr);
    free(mem);
    return 0;
else {
    free(mem);
    return -1;
}
}
```

Src/Extern/SUNDIALS/AMReX_SUNMemory.cpp
```diff
@@ -0,0 +1,148 @@
...
+}
+
+MemoryHelper::MemoryHelper()
+  : helper(CreateMemoryHelper())
+{}
+
+MemoryHelper::~MemoryHelper()
+{
+  if (helper->ops) free(helper->ops);
+  free(helper);
```
WeiqunZhang, line 135: Because `helper` was allocated by `SUNMemoryHelper_NewEmpty`, could SUNDIALS could provide a `free` function too, `SUNMemoryHelper_Free` instead of `std::free`?  (I am assuming `free` here is `std::free`.)  I think the detail of using `std::free` should be hidden inside SUNDIALS because the user (i.e., AMReX) never made a call to `std::malloc`.  `free(helper->ops)` should also be hidden from the user.

There are similar concerns in `Alloc` and `Dealloc`.

Src/Extern/SUNDIALS/AMReX_SUNMemory.cpp
jmsexton03: I get a runtime error when I use the current code with HIP, because of the Dealloc mismatch with The_Device_Arena and the isManaged(). When I duplicate the logic from Alloc into Dealloc, this runtime error goes away, and the Nyx gpu_feature branch built with GMake works for Cuda, Hip, and Sycl. Having a separate getArena function would be great.

Tools/CMake/AMReXThirdPartyLibraries.cmake
```diff
@@ -73,3 +73,19 @@ if (AMReX_PETSC)
     find_package(PETSc 2.13 REQUIRED)
     target_link_libraries( amrex PUBLIC PETSC )
 endif ()
+
+#
+# SUNDIALS
+#
+if (AMReX_SUNDIALS)
+    find_package(SUNDIALS 5.7.0 REQUIRED)
```
jmsexton03, line 81: This causes a failure in the Nyx superbuild: 
https://github.com/AMReX-Astro/Nyx/runs/2513476050?check_suite_focus=true#step:4:102

If I use an installed version of sundials, I get an error from some of the components (such as IDA not being installed and failing an alias setup in SUNDIALSConfig.cmake

If I comment this line out I can generate and build Nyx, but get a runtime error for Cuda during init.

Src/Extern/SUNDIALS/AMReX_SUNMemory.cpp
balos1: We do not have a `SUNMemoryHelper_Free` in the current version of SUNDIALS, but we will add it in the next version. Can we proceed with how it is for now and then update it later?

WeiqunZhang: Sure.

Src/Extern/SUNDIALS/AMReX_SUNMemory.cpp
balos1: @jmsexton03 Can you update the PR with the fixes you have described?

jmsexton03: @balos1  I submitted a PR to this branch [https://github.com/balos1/amrex/pull/2](https://github.com/balos1/amrex/pull/2) 
It includes the initial commit I referenced earlier, my unsuccessful attempt to tweak cmake to work for a Nyx superbuild, and the final commit for the getArena implementation. I'll go in and fix the spacing and push again, but I don't know yet if the other CI tests will pass.

balos1: @jmsexton03 OK, I merged that PR. Just waiting to see if the CI tests pass here.

jmsexton03: @balos1 Looks like it passed.
I'm testing the Nyx CI with this PR in the Nyx branch `gpu_feature`, which is ready to become Nyx `development` 
https://github.com/AMReX-Astro/Nyx/commit/636371e2df64f13fbe300422e3c0acec9004c143

jmsexton03 (approved):
This has been tested with both PelePhysics and Nyx

balos1:
LGTM.

Src/Extern/SUNDIALS/AMReX_SUNMemory.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+  int Alloc(SUNMemoryHelper, SUNMemory* memptr, size_t memsize, SUNMemoryType mem_type)
+  {
+    SUNMemory mem = SUNMemoryNewEmpty();
+
+    mem->ptr = NULL;
+    mem->own = SUNTRUE;
+    mem->type = mem_type;
+
+    if (mem == nullptr) return 0;
```
gardner48: Move this above setting the `mem` content.

Src/Extern/SUNDIALS/AMReX_SUNMemory.cpp
```diff
@@ -0,0 +1,162 @@
...
+    auto arena = getArena(mem->type);
+    if (arena) {
+      mem->ptr = arena->alloc(memsize);
+      *memptr = mem;
+      return 0;
+    }
+    else {
```
gardner48, line 51: Minor style question. Should this be `} else {` like above?

Src/Extern/SUNDIALS/AMReX_SUNMemory.cpp
```diff
@@ -0,0 +1,162 @@
...
+    auto arena = getArena(mem->type);
+    if (arena) {
+      if(mem->own)
+      {
+        arena->free(mem->ptr);
+        free(mem);
+        return 0;
+      }
+    }
+    else {
```
gardner48, line 73: Minor style question. Should this be `} else {` like above?

gardner48 (changes requested):
One minor change and a style question, otherwise this looks good to me.

Src/Extern/SUNDIALS/AMReX_SUNMemory.cpp
WeiqunZhang:
```suggestion
    if (mem == nullptr) return 0;
 
    mem->ptr = NULL;
    mem->own = SUNTRUE;
    mem->type = mem_type;
```

Src/Extern/SUNDIALS/AMReX_SUNMemory.cpp
```diff
@@ -38,11 +38,11 @@ namespace {
   {
     SUNMemory mem = SUNMemoryNewEmpty();
 
+    if (mem == nullptr) return 0;
```
gardner48: Just noticed this should be `return -1`
```suggestion
    if (mem == nullptr) return -1;
```

Src/Extern/SUNDIALS/AMReX_SUNMemory.cpp
```diff
@@ -0,0 +1,162 @@
...
+      return nullptr;
+    }
+  }
+
+  int Alloc(SUNMemoryHelper, SUNMemory* memptr, size_t memsize, SUNMemoryType mem_type)
+  {
+    SUNMemory mem = SUNMemoryNewEmpty();
+
+    if (mem == nullptr) return -1;
+
```
WeiqunZhang:
```suggestion

```

WeiqunZhang (approved):

PR #1809, opened 2021-02-22
HIP: atomicAddNoRet

WeiqunZhang:
It has been deprecated since ROCm 4.1.0. However, for performance reasons,
we should still use it for versions < 5.

### DIFF
--- a/Src/Base/AMReX_GpuAtomic.H
+++ b/Src/Base/AMReX_GpuAtomic.H
@@ -269,7 +269,7 @@ namespace detail {
 #endif
     }
 
-#if defined(AMREX_USE_HIP)
+#if defined(AMREX_USE_HIP) && defined(HIP_VERSION_MAJOR) && (HIP_VERSION_MAJOR < 5)
     AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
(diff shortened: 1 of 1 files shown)


kngott (approved):

Src/Base/AMReX_GpuAtomic.H
```diff
@@ -152,18 +152,6 @@ namespace detail {
 #endif
     }
 
-#if defined(AMREX_USE_HIP)
```
ax3l, line 272: Do we already want to remove this? So far it's only deprecated.

Did they provide an equally fast replacement yet via `atomicAdd?`

WeiqunZhang: Doesn't seem so.  That's why it's marked "do-not-merge".  4.1 is not formally released yet.

ax3l: Oh got it, sorry I overlooked the label :see_no_evil:

ax3l:
X-ref: #2567 / #2577

Src/Base/AMReX_GpuAtomic.H
WeiqunZhang: ROCm 5.0 has fixed the performance issue.

PR #1824, opened 2021-02-27
Ensure that Particle is 8-byte aligned, even in single precision.

atmyers:
This will allow us to load particles as a bunch of `uint64_t`s into shared memory.

### DIFF
--- a/Src/Particle/AMReX_ArrayOfStructs.H
+++ b/Src/Particle/AMReX_ArrayOfStructs.H
@@ -69,8 +69,8 @@ public:
 
     bool empty () const { return m_data.empty(); }
 
-    const RealType* data () const { return &(m_data[0].m_pos[0]); }
-    RealType* data () { return &(m_data[0].pos(0)); }
+    const RealType* data () const { return (RealType*)&(m_data[0]); }
+    RealType* data () { return (RealType*)&(m_data[0]); }
 
     const RealType* dataPtr () const { return data(); }
     RealType*       dataPtr ()       { return data(); }
--- a/Src/Particle/AMReX_ArrayOfStructs.H
+++ b/Src/Particle/AMReX_ArrayOfStructs.H
@@ -69,8 +69,8 @@ public:
 
     bool empty () const { return m_data.empty(); }
 
-    const RealType* data () const { return (RealType*)&(m_data[0]); }
-    RealType* data () { return (RealType*)&(m_data[0]); }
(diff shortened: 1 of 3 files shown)


WeiqunZhang:
What about `amrex_particle` in `amrex_particlecontainer_module`?  Do we need to change that too?

atmyers:
Yes - but the thing about this is that it will require app code changes too - see this code in FLASH for example:

```
     type, bind(C), public :: pt_amrexParticle_t
     real(amrex_particle_real)    :: pos(AMREX_SPACEDIM) !< Position
     real(amrex_particle_real)    :: vel(AMREX_SPACEDIM) !< Particle velocity
     real(amrex_particle_real)    :: fdata(NPART_PROPS-(2*MDIM+4)) ! attribs except pos[xyz],
                                                                   ! vel[xyz],tag,cpu,blk,proc
     integer(c_int), private      :: id
     integer(c_int), private      :: cpu
  end type pt_amrexParticle_t
```

WeiqunZhang:
OK.  Maybe we should use `alignas` instead.

WeiqunZhang:
But that's probably not portable to Fortran either.

Src/Particle/AMReX_ArrayOfStructs.H
```diff
@@ -69,8 +69,8 @@ public:
 
     bool empty () const { return m_data.empty(); }
 
-    const RealType* data () const { return &(m_data[0].m_pos[0]); }
-    RealType* data () { return &(m_data[0].pos(0)); }
+    const RealType* data () const { return (RealType*)&(m_data[0]); }
+    RealType* data () { return (RealType*)&(m_data[0]); }
```
WeiqunZhang: Maybe `(RealType*)m_data.data()`?

atmyers:
I think we should do it this way and then fix on the application side in FLASH5 and Maui.

drummerdoc:
Does this require explicit changes in Pele as well?

WeiqunZhang (approved):

atmyers:
> Does this require explicit changes in Pele as well?

No, it should only affect codes that still update particles through Fortran, which I believe Pele is not.

drummerdoc:
Fortran?  People still use that dinosaur??  :)
Thanks Andrew.

atmyers:
EDIT - on further thought this is not necessary, so to avoid breakage I will close this PR.

PR #1839, opened 2021-03-04
Use unique_ptr instead of naked pointer

maikel:
## Summary

This commit replaces naked pointer to char with `std::unique_ptr` in `NonLocalBC::CommHandler`.
This will lead to proper cleanup in case of any exceptions. 
Note, that `Comm_nowait` and `Comm_finish` are *not* marked noexcept. 

Furthermore, future PRs will add interpolation and some kind of projection for non-local data transactions.  
Error handling for those is still unclear. Maybe, we want to mark everything noexcept.

This is merely an implementation change without nearly any noticable difference in user code.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_Extension.H
+++ b/Src/Base/AMReX_Extension.H
@@ -151,20 +151,13 @@
 #define AMREX_FALLTHROUGH ((void)0)
 #endif
 
-#ifndef AMREX_NO_NODISCARD
-#  ifdef __has_cpp_attribute
-#    if __has_cpp_attribute(nodiscard)
-#      define AMREX_NODISCARD [[nodiscard]]
-#    elif defined(__GNUC__) || defined(__clang__)
-#      define AMREX_NODISCARD __attribute__ ((warn_unused_result))
-#    else
-#      define AMREX_NODISCARD
-#    endif
-#  elif defined(__GNUC__) || defined(__clang__)
-#    define AMREX_NODISCARD __attribute__ ((warn_unused_result))
-#  else
-#    define AMREX_NODISCARD
-#  endif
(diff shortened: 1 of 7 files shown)


maikel:
If `FAB::nBytes` is really state dependend (particles?) I will revert the changes to static. They are not really needed.

Src/Base/AMReX_Extension.H
```diff
@@ -151,6 +151,22 @@
...
+#    else
+#      define AMREX_NODISCARD
+#    endif
+#  elif defined(__GNUC__) || defined(__clang__)
+#    define AMREX_NODISCARD __attribute__ ((warn_unused_result))
+#  else
+#    define AMREX_NODISCARD
+#  endif
+#endif
+
```
WeiqunZhang: I like this!  It seems that the logic can be simplified with
```
#if !defined(AMREX_NO_NODISCARD) && defined(__has_cpp_attribute) && __has_cpp_attribute(nodiscard)
# 
#elif !defined(AMREX_NO_NODISCARD) && (defined(__GNU__) || defined(__clang__))
# 
#else
# 
#endif
```
I still we need to define `AMREX_NODISCARD` as empty even if `AMREX_NO_NODISCARD` is defined.

WeiqunZhang (changes requested):
Looks good to me.

Src/Base/AMReX_Extension.H
maikel: Of course, you are right. We still need to define an empty macro.
My macro-fu is weak, so I didn't know if the preprocessor can do partial evaluation, i.e. check for the existence of a macro and use it in the and clause. I will try it!

WeiqunZhang (approved):

PR #1840, opened 2021-03-04
CMake: move AMReX_INSTALL option to AMReXOptions.cmake

mic84:
## Summary
Option to control generation of install target must be loaded at the beginning of execution.
Follow-up to #1831 and #1833
## Checklist

The proposed changes:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -143,7 +143,6 @@ endif ()
 #
 # Install amrex  -- Export
 #
-option(AMReX_INSTALL "Generate Install Targets" YES)
 if(AMReX_INSTALL)
     include(AMReXInstallHelpers)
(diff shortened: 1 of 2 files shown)


ax3l:
@mic84 should the `option()` in `CMakeLists.txt` then be removed?

mic84:
Yes, it should be removed. The reason for moving `AMReX_INSTALL` to `AMReXOptions.cmake` is that it needs to be loaded together with all the other options so that it is visible from the very beginning. That was not the case and  as a result `generate_config_header` in `Src/CMakeLists.txt` was never executed.

ax3l (approved):
Thank you for catching this!

ax3l:
@RemiLehe that's the issue you saw a minute ago, will be fixed now :)

PR #1842, opened 2021-03-05
Feature/non local parallel copy

maikel:
## Summary

This PR introduces "Multiblock" capabilities to AMReX.

## Additional background

Multiblock refers to having multiple distinct computational domains, known as blocks, that have some relation to each other.
This PR proposes a generic NonLocalBC::ParallelCopy that performs copies between two FabArrays that might live on different blocks. The ParallelCopy respects

 - An index mapping such as already present in NonLocalBC (DTOS)
 - A projection function that will be called on either sender or receiver side to perform actions like e.g. swapping two components or changing the sign of velocity components
 
The proposed usage is

```cpp
NonLocalBC::ParallelCopy(dest, destbox, src, scomp, dcomp, ncomp, ngrow, index_mapping, fab_projection);
```

## Notes

- In the current implementation, `DTOS` needs to be a map between index spaces where inverse images of boxes are boxes again.
- The name `NonLocalBC::ParallelCopy` is confusing if it does something else than `amrex::ParallelCopy`. Alternatives?
- Parts of the implementation might be shared with other places in AMReX, such as in communication procedures of FabArray. But this is out of scope.
- In addition to `SwapComponents`, a more generic `PermuteComponents` is useful.
- The test is not commented yet.

## Checklist

The proposed changes:

The following tasks need to be done:
- [x] Define CPOs for dispatching packing and unpacking of communication buffers
- [x] Implement the first version without interpolation
- [x] Provide a test case to prove correct index mapping, projection and communication behaviour 
- [ ] Write a tutorial with an example of using at least two blocks in a simple simulation
- [x] Improve type documentation and requirements

**The advection example produces the following result**

<p align="center">

</p>
<p align="center">Figure 1. A test run of the multi block advection test case. It was initiated with 4 MPI Ranks. It shows a surface plot of a scalar component and arrow glyphs of an advected vector.</p>


WeiqunZhang:
It looks good to me.  There are a lot of changes.  So I will go through this a second time tomorrow.  I have one minor comment about the order of parameters of `ParallelCopy`.  Currently it is
```
(FabArray<FAB>& dest, const Box& destbox, const FabArray<FAB>& src, int destcomp,
              int srccomp, int numcomp, ...)
```
Can we swap the order of `srccomp` and `destcomp` so that it's consistent with existing `ParallelCopy` functions?

maikel:
> Can we swap the order of `srccomp` and `destcomp` so that it's consistent with existing `ParallelCopy` functions?

Sure, I remembered the order incorrectly.

WeiqunZhang (approved):

PR #1836, opened 2021-03-04
CI: macOS Shared Lib

ax3l:
## Summary

CI for macOS: Build a shared library. This is the default in Spack.

I see issues with AppleClang 12.0 raising:
```
ld: can't use -undefined warning or suppress with -twolevel_namespace
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

@RemiLehe reported this again here: https://github.com/AMReX-Codes/amrex/issues/425#issuecomment-473153152

- [x] reproduce issue
- [x] add fix

## Additional background

Introduced in 1b6af3cb630 to fix #425

## Checklist

The proposed changes:
- [x] fix a bug or incorrect behavior in AMReX
- [ ] add new capabilities to AMReX
- [ ] changes answers in the test suite to more than roundoff level
- [ ] are likely to significantly affect the results of downstream AMReX users
- [ ] are described in the proposed changes to the AMReX documentation, if appropriate

PR #1847, opened 2021-03-06
Build: Windows with Shared Libs (.dll)

ax3l:
## Summary

Add build support, link support and CI coverage for `.dll` builds on Windows.
Remove linker tinkering for weak symbols on Linux & macOS.

## Additional background

- [x] Apply changes from #1852 (remove weak symbols)
- [x] Remove work-around for weak-symbols on macOS #1836
- [x] rebase after #1852 was merged
- [x] add macro to export globals on Windows [[SO]](https://stackoverflow.com/questions/54560832/cmake-windows-export-all-symbols-does-not-cover-global-variables/54568678#54568678)

## Checklist

The proposed changes:

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -13,12 +13,30 @@ jobs:
       run: |
         cmake -S . -B build   `
               -DCMAKE_BUILD_TYPE=Debug      `
+              -DBUILD_SHARED_LIBS=ON        `
               -DCMAKE_VERBOSE_MAKEFILE=ON   `
               -DAMReX_BUILD_TUTORIALS=ON    `
               -DAMReX_FORTRAN=OFF           `
               -DAMReX_MPI=OFF               `
               -DAMReX_PARTICLES=ON
         cmake --build build --config Debug -j 2
+
+  # Build libamrex and all tutorials (static)
+  tutorials_msvc_static:
+    name: MSVC C++17 w/o Fortran w/o MPI static
+    runs-on: windows-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Build & Install
+      run: |
+        cmake -S . -B build   `
+              -DCMAKE_BUILD_TYPE=RelWithDebInfo `
+              -DCMAKE_VERBOSE_MAKEFILE=ON   `
+              -DAMReX_BUILD_TUTORIALS=ON    `
+              -DAMReX_FORTRAN=OFF           `
+              -DAMReX_MPI=OFF               `
+              -DAMReX_PARTICLES=ON
+        cmake --build build --config RelWithDebInfo -j 2
 
   # Build libamrex and all tutorials
   tutorials_clang:
@@ -34,6 +52,7 @@ jobs:
         cmake -S . -B build   ^
               -T "ClangCl"    ^
               -DCMAKE_BUILD_TYPE=Release    ^
+              -DBUILD_SHARED_LIBS=ON        ^
               -DCMAKE_VERBOSE_MAKEFILE=ON   ^
               -DAMReX_BUILD_TUTORIALS=ON    ^
               -DAMReX_FORTRAN=OFF           ^
--- a/Src/Base/AMReX_MFIter.H
+++ b/Src/Base/AMReX_MFIter.H
@@ -5,6 +5,7 @@
 #include <memory>
 
 #include <AMReX_Arena.H>
+#include <AMReX_Extension.H>
 #include <AMReX_FabArrayBase.H>
 #include <AMReX_IntVect.H>
 #include <AMReX_FArrayBox.H>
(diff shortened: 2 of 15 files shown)


WeiqunZhang:
I merged changes from #1852 for testing.

ax3l:
@WeiqunZhang just a quick note if you haven't seen it yet: there is still an issue with the patches from #1852 applied

ax3l:
Ok, removed the macOS linker flag work-arounds to fix that platform :)

Not sure what we do about the `extern` declared globals like `amrex::ParallelDescriptor::ioProcessor` on Windows yet, but it's also nappying time :)

WeiqunZhang:
Looks like we have a problem with Windows.

https://stackoverflow.com/questions/19373061/what-happens-to-global-and-static-variables-in-a-shared-library-when-it-is-dynam

ax3l:
Hm, but we do export those symbols on Windows with a blanket option developed by CERN. Maybe the logic it uses does not apply to `extern static` defined vars and we have to add export macros ourselves for them.

Very off topic:
> Stack Overflow: And of course, as a necessary final note: global variables should be avoided.

:D I mean that will fall on our feet as well at some point, e.g. if a users tries to import multiple amrex-powered libraries into an executable. I guess you are aware of that, sorry for mumbling out loud. For another time... we can think how to refactor those patterns for better C/C++-level composability :)

ax3l:
Ah yes, the problem is with globals in AMReX: https://stackoverflow.com/a/54568678/2719194
https://cmake.org/cmake/help/latest/prop_tgt/WINDOWS_EXPORT_ALL_SYMBOLS.html

We just need to add another macro for those (and long term avoid global states for better modularity).

.github/workflows/windows.yml
```diff
@@ -13,6 +13,7 @@ jobs:
       run: |
         cmake -S . -B build   `
               -DCMAKE_BUILD_TYPE=Debug      `
+              -DBUILD_SHARED_LIBS=ON        `
```
ax3l, line 16: TODO: Careful - also check if static builds still work on both Windows CI entries.

.github/workflows/macos.yml
```diff
@@ -19,6 +19,7 @@ jobs:
         cmake -S . -B build             \
             -DBUILD_SHARED_LIBS=ON      \
             -DCMAKE_BUILD_TYPE=Debug    \
+            -DBUILD_SHARED_LIBS=ON      \
```
ax3l: TODO: revert me (see two lines above)

ax3l:
Rebased since #1852 landed.
@WeiqunZhang @mic84 @sayerhs ready for review :)

Src/Base/AMReX_MFIter.H
```diff
@@ -204,9 +205,9 @@ protected:
     std::unique_ptr<Gpu::FuseSafeGuard> gpu_fsg;
 #endif
 
-    static int nextDynamicIndex;
-    static int depth;
-    static int allow_multiple_mfiters;
+    static AMREX_EXPORT int nextDynamicIndex;
+    static AMREX_EXPORT int depth;
+    static AMREX_EXPORT int allow_multiple_mfiters;
```
WeiqunZhang, line 210: Is that true we have mark all the static variables and functions in classes with AMREX_EXPORT?

WeiqunZhang: Note that these are not static variables at file scopes.

ax3l: Only `static` (and `export`) variables, since they are encapsulated global states of a library.

`static` functions don't need that, as they do not carry a state.

ax3l: In general, for modern C++ libs that template a lot one would manually expose only the public classes with such a macro and use `-fvisibility=hidden -fvisibility-inlines-hidden`.

Incomplete ramblings with mention of some benefits / Q&A: memo I wrote a while back: https://gist.github.com/ax3l/ba17f4bb1edb5885a6bd01f58de4d542

We don't do this and instead simplify through an export of all symbols by making everything visible by default (classes, functions, enums) - even for Windows; but [the tooling for Windows](https://cmake.org/cmake/help/latest/prop_tgt/WINDOWS_EXPORT_ALL_SYMBOLS.html) does not include those global variables (*data symbols*).

Src/Base/AMReX_MFIter.H
```diff
@@ -174,7 +175,7 @@ public:
 
     const DistributionMapping& DistributionMap () const noexcept { return fabArray.DistributionMap(); }
 
-    static int allowMultipleMFIters (int allow);
+    static AMREX_EXPORT int allowMultipleMFIters (int allow);
```
WeiqunZhang: This is a function.  Does it need `AMREX_EXPORT`?

ax3l: Thanks, that's a mistake. It's already exposed.

WeiqunZhang (approved):

sayerhs (approved):

ax3l:
@mic84 ping :)

mic84 (approved):

PR #1848, opened 2021-03-06
Robin BC in Linear Solver

WeiqunZhang:
## Summary

Add Robin BC to MLABecLaplacian.  There are two steps for setting up Robin
BC.  The first step is to call setDomainBC, and the second step is to call
setLevelBC, which has been modified to take a, b and f in Robin BC, `a*phi +
b*dphi/dn = f`.

## Checklist

The proposed changes:


WeiqunZhang:
Tests at https://github.com/WeiqunZhang/amrex-devtests/tree/main/robin_bc show second order convergence for all orientations.

asalmgren (approved):

PR #1855, opened 2021-03-12
Slopes fix

vzendejasl:
## Additional background
Slope computation on cell centroids using the Least Square method. These tests are based from the LeastSquares Test (#1707) but instead of calculating scalar values on face centroids they are calculated on cell centroids using a linear function rather than a second order polynomial (#1707). 
## Checklist

The proposed changes:


asalmgren (commented):
@vzendejasl -- I see some sections of the code that are commented out (with slashes or if 0) or have an error message (e.g."combustor) -- could you remove those completely from the commit?

vzendejasl:
@asalmgren ahh yes! I'll clean those up and resubmit.

PR #1860, opened 2021-03-13
Remove tabs and trailing white spaces

ccse:
The changes are purely white spaces only. `git diff -w development..HEAD`
shows nothing.  It was done with

```
find . -type d \( -name .git \
                  -o -path ./paper \
               \) -prune -o \
       -type f \( -name "*.H" -o -name "*.h" -o -name "*.hh" -o -name "*.hpp" \
                  -o -name "*.c" -o -name "*.cc" -o -name "*.cpp" -o -name "*.cxx" \
                  -o -name "*.f" -o -name "*.F" -o -name "*.f90" -o -name "*.F90" \
                  -o -name "*.py" \
                  -o -name "*.md" -o -name "*.rst" \
                  -o -name "*.sh" \
                  -o -name "*.tex" \
                  -o -name "*.txt" \
               \) \
    -exec grep -Iq . {} \; \
    -exec sed -i 's/[[:blank:]]\+$//g' {} + \
    -exec sed -i 's/\t/\ \ \ \ \ \ \ \ /g' {} +
```

We will add a CI check as a follow-up.


atmyers:
After this is merged, we should add a style guide with things like this, space before `(` after function names, etc...

atmyers (approved):

PR #1862, opened 2021-03-14
CI for Code Style

WeiqunZhang:
Check for tabs and trailing whitespaces at the end of line in CI.


.github/workflows/style.yml
```diff
@@ -0,0 +1,13 @@
...
+on: [push, pull_request]
+
+jobs:
+  style:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Tabs
+        run: .github/workflows/style/check_tabs.sh
+      - name: Trailing Whitespaces
```
sayerhs, line 17: @WeiqunZhang If the tabs check fails, it will not run trailing whitespaces check. Probably want to split this into two independent jobs?

WeiqunZhang: Sure, we can split this.

sayerhs (approved):

PR #1871, opened 2021-03-16
Fix bug with face-centroid slopes when using multigrid

drangara:
## Summary
When computing the Least Square slopes on face-centroid stencil with multigrid, we need to grow `m_eb_phi` and make additional checks.

## Additional background
This adds to the changes made in https://github.com/AMReX-Codes/amrex/pull/1707

## Checklist

The proposed changes:

### DIFF
--- a/Src/EB/AMReX_EB_LeastSquares_2D_K.H
+++ b/Src/EB/AMReX_EB_LeastSquares_2D_K.H
@@ -498,6 +498,9 @@ Real grad_x_of_phi_on_centroids_extdir(int i,int j,int k,int n,
               continue;
             }
 
+            if (ii == phi.end.x || jj == phi.end.y)
+               continue;
+
             if (!flag(ii,jj,k).isCovered())
             {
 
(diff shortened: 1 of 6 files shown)


asalmgren (approved):

Src/EB/AMReX_EB_LeastSquares_2D_K.H
```diff
@@ -498,6 +498,9 @@ Real grad_x_of_phi_on_centroids_extdir(int i,int j,int k,int n,
               continue;
             }
 
+            if (ii == phi.end.x || jj == phi.end.y)
```
WeiqunZhang: Is this to prevent out-of-bound access to phi?  Why is it only checking the end, but not the begin?

drangara: that's a good point. i guess my test cases are not hitting that but i will update it to also check the lower bound. thanks

WeiqunZhang: Array4 has a member function for testing this
```
bool contains (int i, int j, int k) const
```
For 2d, you can pass 0 as k.

WeiqunZhang (approved):

PR #1873, opened 2021-03-17
Add FabArray::release and Make BaseFab virtual

WeiqunZhang:
Add FabArray::release functions that release the ownership of the FAB at
the given index and return a raw pointer to the FAB.

Make BaseFab virtual so that it's safe to use the new release functions.

Also declare a number of functions `[[nodiscard]]`.

Fix some warnings in EB/CNS example.

## Checklist

The proposed changes:

### DIFF
--- a/Src/AmrCore/AMReX_TagBox.H
+++ b/Src/AmrCore/AMReX_TagBox.H
@@ -46,7 +46,7 @@ public:
 
     TagBox (const TagBox& rhs, MakeType make_type, int scomp, int ncomp);
 
-    ~TagBox () = default;
+    virtual ~TagBox () noexcept override final = default;
 
     TagBox (TagBox&& rhs) noexcept = default;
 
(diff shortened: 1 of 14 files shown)


Src/Base/AMReX_iMultiFab.cpp
```diff
@@ -169,6 +169,7 @@ iMultiFab::iMultiFab (const BoxArray&            bxs,
     :
     FabArray<IArrayBox>(bxs,dm,ncomp,ngrow,info,factory)
 {
+    factory.clone();  // xxxxx this must be removed later
```
WeiqunZhang: @maikel Let's see if this produces a warning.

WeiqunZhang: In the CI tests, some compilers caught it, but some didn't.

WeiqunZhang: Anyway, that's good enough.

atmyers (approved):

PR #1883, opened 2021-03-19
Header file order and iosfwd

WeiqunZhang:
Include amrex headers before standard C++ headers.

Use iosfwd instead of iostream if we can.

### DIFF
--- a/Src/Amr/AMReX_Amr.cpp
+++ b/Src/Amr/AMReX_Amr.cpp
@@ -1,17 +1,3 @@
-#include <algorithm>
-#include <cstdio>
-#include <list>
-#include <iostream>
-#include <iomanip>
-#include <sstream>
-#include <iomanip>
-#include <limits>
-#include <cmath>
-
-#ifdef AMREX_USE_OMP
-#include <omp.h>
-#endif
-
 #include <AMReX_Geometry.H>
 #include <AMReX_TagBox.H>
 #include <AMReX_Array.H>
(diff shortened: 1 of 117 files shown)


Src/Amr/AMReX_Amr.cpp
```diff
@@ -48,6 +34,20 @@
...
+#include <iostream>
+#include <iomanip>
+#include <sstream>
+#include <iomanip>
+#include <limits>
+#include <cmath>
+
+#ifdef AMREX_USE_OMP
+#include <omp.h>
+#endif
```
ax3l: Only off in this location: include of OMP / MPI et al. before stdlib headers.

Src/Base/AMReX_BLProfiler.H
```diff
@@ -15,8 +15,6 @@
 #include <AMReX_Box.H>
 #endif
 
-#include <iostream>
-#include <ostream>
```
ax3l, line 19: you probably still need `ostream` here since it's appearing as an argument in the `operator<<`

WeiqunZhang: Yes, we need to add `isofwd` for this header to be independent.  But, because the `ostream` argument is a reference, not value, I think we are fine without `<iostream>`.

ax3l: Right, `isofwd` should work here. https://en.cppreference.com/w/cpp/header/iosfwd

ax3l (approved):

PR #1884, opened 2021-03-20
Make FabArray::setFab safer

WeiqunZhang:
Instead of taking a raw pointer, it takes std::unique_ptr or rvalue FAB, so
that the ownership is very clear.  It also deletes the old FAB if not null.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_FabArray.H
+++ b/Src/Base/AMReX_FabArray.H
@@ -351,10 +351,18 @@ public:
     Array4<typename FabArray<FAB>::value_type const> const_array (int K, int start_comp) const noexcept;
 
     //! Explicitly set the Kth FAB in the FabArray to point to elem.
-    void setFab (int K, FAB* elem);
+    void setFab (int K, std::unique_ptr<FAB> elem);
+
+    //! Explicitly set the Kth FAB in the FabArray to point to elem.
+    template <class F=FAB, std::enable_if_t<std::is_move_constructible<F>::value,int> = 0>
+    void setFab (int K, FAB&& elem);
 
     //! Explicitly set the FAB associated with mfi in the FabArray to point to elem.
-    void setFab (const MFIter&mfi, FAB* elem, bool assertion=true);
+    void setFab (const MFIter&mfi, std::unique_ptr<FAB> elem);
+
+    //! Explicitly set the FAB associated with mfi in the FabArray to point to elem.
+    template <class F=FAB, std::enable_if_t<std::is_move_constructible<F>::value,int> = 0>
+    void setFab (const MFIter&mfi, FAB&& elem);
 
     //! Release ownership of the FAB. This function is not thread safe.
     AMREX_NODISCARD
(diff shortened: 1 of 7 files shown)


maikel:
When do you use `std::enable_if_t` over `EnableIf_t` ?

WeiqunZhang:
We added `amrex::EnableIf_t` when we had C++11 as the minimal requirement.  But recently we changed to use C++14.  I am planning to replace all uses of `amerx::EnableIf_t` with `std::enable_if_t` so that it's easy for the readers to understand.

Src/Extern/amrdata/AMReX_AmrData.cpp
```diff
@@ -877,7 +877,7 @@ bool AmrData::ReadNonPlotfileData(const string &filename, Amrvis::FileType /*fil
     dataGrids[LevelZero][ComponentZero]->define(fabBoxArray, dMap, NVarZero,
                                                 fabNGrow, Fab_noallocate);
     if(ParallelDescriptor::IOProcessor()) {
```
maikel, line 879: Is AMReX supposed to be expcetion safe? `auto newfab = std::make_unique<FArrayBox>()` and following `newfab.get()` for non-owning accesses would be more c++ idiomatic

WeiqunZhang: We never really care much about exception safety.  The problem is for us it's not clear how we could handle it if the memory allocation throws an exception other than crashing the code.  I am afraid of touching the code in AmrData.  So I just did the minimal change to it.

WeiqunZhang: We also have quite a lot of `unique_ptr::reset(new ...)` because we had been using C++11.  Some day, we need to replace it with `make_unique`.

maikel: Makes sense. LGTM

Src/Base/AMReX_FabArray.H
```diff
@@ -1561,56 +1573,99 @@ FabArray<FAB>::AllocFabs (const FabFactory<FAB>& factory, Arena* ar,
...
+    if (m_fabs_v.size() == 0) {
+        m_fabs_v.resize(indexArray.size(),nullptr);
+    }
+
+    const int li = localindex(boxno);
+    if (m_fabs_v[li]) {
+        m_factory->destroy(m_fabs_v[li]);
+    }
+    m_fabs_v[li] = new FAB(std::move(elem));
+}
```
kngott, line 1627: Is it possible to collapse these down into fewer copies of the same function? e.g. MFIter version calls the int version?

Looks like the problem is mfi.index() and mfi.LocalIndex(). Would it make sense to have a setFab_doit in this case, or too much hassle?

WeiqunZhang: We could get the global index with `MFIter::index()` and call the int version, not the other way around.  But the MFIter version is more efficient than the int version.

kngott: Eww. I so dislike weird edge cases without an elegant solution. Oh well, otherwise, looks good to me.

kngott (approved):

PR #1886, opened 2021-03-20
Remove __attribute__((warn_unused_result)) from AMREX_NODISCARD

atmyers:
This appears to be buggy in our CI, for example: https://github.com/AMReX-Codes/amrex/pull/1876/checks?check_run_id=2146238004

The proposed changes:


WeiqunZhang (approved):

maikel:
The following article might be related https://quuxplusone.github.io/blog/2019/01/25/nodiscard-philosophies/
`__wuc`, is, as it seems, too aggressive.

PR #1887, opened 2021-03-20
[CMake] Fix path to tests inputs file

atmyers:
The inputs file gets copied to the test run directory from a location that sometimes includes a subdirectory, for example `Exec/inputs`. But, then the run command will also include the sub-directory. This makes it so that the directory is stripped out, so that the run command will be `./<exe> inputs` instead of `./<exe> Exec/inputs`.  

Furthermore, in the event that multiple files are passed in as _`inputs`, so that e.g. `probin` can be copied, we should only actually run the test with the first inputs file. 

The proposed changes:

### DIFF
--- a/Tests/CMakeLists.txt
+++ b/Tests/CMakeLists.txt
@@ -80,7 +80,8 @@ function (setup_test _srcs  _inputs)
 
    if (${_inputs})
       file( COPY ${${_inputs}} DESTINATION ${_exe_dir} )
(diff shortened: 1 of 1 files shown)


atmyers:
This has already come in on PR #1876 - @mic84 could you let me know if this change looks good to you?

mic84:
@atmyers Looks good to me

PR #1890, opened 2021-03-21
CMake: IPO/LTO

ax3l:
## Summary

Add optional interprocedural optimization (IPO) aka link-time optimization (LTO) to AMReX.

In my WarpX tests recently on x86_64 and ppc64le, using those options incurred a 2% performance hit (although one would expect a slight improvement). Thus, they are not enabled by default. Also, enabling them with some toolchains needs users to set the linker properly.

Nonetheless, enabling IPO shrinks the binary size a lot. This is very valuable and worth the runtime penalty for our binary deployments, e.g. when shipping generic x86_64, aarch64 and ppc64le binaries on [conda-forge](https://github.com/conda-forge/warpx-feedstock) to users. Those binaries are not tuned to micro-architectures anyway (contrary to Spack builds, which are), and thus already leave at least another 2% of performance on the table (tests done on KNL and Power9/ppc64le again).

WarpX binary size (default build, GCC 9.3.0 on x86_64 with Ubuntu 20.04):
* no IPO: 217MiB
* WarpX IPO, AMReX no IPO: 155MiB
* both WarpX and AMReX with IPO: 127MiB

## Additional background

Related to:
- https://github.com/ECP-WarpX/WarpX/pull/1746 cc @lucafedeli88 
- https://github.com/ECP-WarpX/WarpX/pull/1822

## Checklist

The proposed changes:


sayerhs (approved):

WeiqunZhang (approved):

ax3l:
Hi @mic84, I am going ahead and merge this already. Please don't hesitate to review it later as you will get a ping nonetheless. If anything is off, we will do a follow-up PR :)

ax3l:
Via Slack:
> Looks good to me

PR #1910, opened 2021-03-29
Dont zero out input multifab in amrex::ParticleToMesh

atmyers:
This is allows the function to be used when, e.g. depositing multiple particle species to the same multifab.

The proposed changes:


atmyers:
Note, I have also changed this to always use a tmp multifab, instead of only doing it for the dual grid case. This is because with ghost cells, I think it is actually a bug if there is already some mass in the ghost cells before the adding the particles because it will be double counted in the SumBoundary call. It is cleaner and safer (and probably not that expensive) to just always use a tmp here.

dwillcox (approved):
Thanks for the heads-up!

I think if we promise to call `setVal(0.)` first ourselves that would fix the ghost cell issue too, right?

Src/Particle/AMReX_ParticleMesh.H
```diff
@@ -69,18 +68,13 @@ ParticleToMesh (PC const& pc, MF& mf, int lev, F&& f)
...
 
-    mf_pointer->SumBoundary(pc.Geom(lev).periodicity());
-
-    if (mf_pointer != &mf)
-    {
-        mf.copy(*mf_pointer,0,0,mf_pointer->nComp());
-        delete mf_pointer;
-    }
+    mf_tmp.SumBoundary(pc.Geom(lev).periodicity());
+    mf.copy(mf_tmp,0,0,mf_tmp.nComp());
```
WeiqunZhang: Maybe we can do
```
mf.setVal(0.);
mf.ParallelAdd(mf_tmp, 0, 0, mf_tmp.nComp(), mf_tmp.nGrowVect(), IntVect(0), pc.Geom(lev).periodicity());
```
This will save one parallel communication if the grids of mf and mf_tmp are different.

WeiqunZhang: I am also confused now.  `mf.copy(mf_tmp,0,0,mf_tmp.nComp());` overrides what's in mf, whereas the title of the PR says `Dont zero out input multifab`.

Src/Particle/AMReX_ParticleMesh.H
```diff
@@ -14,11 +14,10 @@ ParticleToMesh (PC const& pc, MF& mf, int lev, F&& f)
...
 
-    MultiFab* mf_pointer = pc.OnSameGrids(lev, mf) ?
-        &mf : new MultiFab(pc.ParticleBoxArray(lev),
-                           pc.ParticleDistributionMap(lev),
-                           mf.nComp(), mf.nGrowVect());
-    mf_pointer->setVal(0.);
+    MultiFab mf_tmp(pc.ParticleBoxArray(lev),
+                    pc.ParticleDistributionMap(lev),
+                    mf.nComp(), mf.nGrowVect());
+    mf_tmp.setVal(0.0);
```
ax3l, line 22: Offline discussed: I was wondering if an additional flag could save us from allocating a temporary multifab if we subsequently pass multiple `pc`s

WeiqunZhang: Not worried about temporary multifabs.  But it would certainly save communications.

atmyers: It seems that if 1) the input multifab has zero ghost cells (it doesn't need any for NGP deposition) and 2) dual grid is off, we could skip the temporary altogether. That would cover the "particle_count" case anyway.

ax3l (approved):
Thank you, @atmyers! :)

WeiqunZhang:
Does it make sense to add an optional argument so that the function is still backward compatible in its behavior?

WeiqunZhang (approved):

PR #1920, opened 2021-04-02
Use std::make_unique instead of reset since we are using C++14

WeiqunZhang:


Src/Particle/AMReX_NeighborParticlesI.H
```diff
@@ -141,7 +141,7 @@ NeighborParticleContainer<NStructReal, NStructInt, NArrayReal, NArrayInt>
         {
             const Geometry& geom = this->Geom(lev);
 
-            mask_ptr[lev].reset(new iMultiFab(ba, dmap, num_mask_comps, m_num_neighbor_cells));
+            mask_ptr[lev] = std::make_unique<iMultiFab>(ba, dmap, int(num_mask_comps), m_num_neighbor_cells);
```
WeiqunZhang, line 144: For reasons I don't completely understand, we have to use `int(num_mask_comps)` here for debug build.  The reason might be this function is not inline in debug mode and `num_mask_comps` is a `static constexpr` member that has no definition.  Thus it cannot be odr-used before C++17.  `std::make_unique` is probably trying to match `num_mask_comps` to `const int&` (an odr-use).

atmyers (approved):

Docs/sphinx_documentation/source/LinearSolvers.rst
```diff
@@ -925,7 +925,7 @@ viscous term `divtau` explicitly:
        {
            BoxArray edge_ba = grids[lev];
            edge_ba.surroundingNodes(dir);
-           b[lev][dir].reset(new MultiFab(edge_ba, dmap[lev], 1, nghost, MFInfo(), *ebfactory[lev]));
+           b[lev][dir] = std::make_unique<MultiFab>(edge_ba, dmap[lev], 1, nghost, MFInfo(), *ebfactory[lev]);
```
ax3l, line 928: @WeiqunZhang There is just one detail here that hit me recently:

if you assign a new `make_unique` data structure that is *large* then there is a short moment in which both the old one and the new one are alive and occupy memory: https://godbolt.org/z/PTr4MsPY8

```C++
#include <iostream>
#include <memory>

struct T {
    T(){ std::cout << "alive\n";}
    ~T(){ std::cout << "dead\n";}
};

int main()
{
    auto p = std::make_unique<T>();
    p = std::make_unique<T>();
}
```
```
alive
alive
dead
dead
```

In case your MultiFabs are large, this might be quite expensive for peak memory.

(The `.reset(new ...);` syntax has the same problem, though. I `= nullptr;` explicitly in such a case.)

WeiqunZhang: That is true.

PR #1928, opened 2021-04-05
Update RNG for latest oneAPI

WeiqunZhang:
MKL now has device support for uniform distribution of unsigned int and
Poisson distribution.

## Checklist

The proposed changes:


kngott:
Approving, but putting this thought here:

Do we also want to add a quick check to the make system to error out if SYCL is under this version, given this is always compiled?

More people are exploring HIP/DPC++ implementations, so it might be helpful for the users and save us some explanation and headache for random errors down the road.

kngott (approved):

PR #1931, opened 2021-04-06
Alias HostVector to PinnedVector and update documentation.

atmyers:
Currently, `amrex::Gpu::HostVector` uses `std::allocator`. However, the normal use case for this class is for vectors that need to be initialized on the host and then copied into a amrex::Gpu::DeviceVector. For this purpose, pinned memory is expected to perform better. Additionally, it is arguably less surprising for Gpu::HostVector to use `cudaHostAlloc` than `std::allocator`. Thus, this PR changes `amrex::Gpu::HostVector` to alias `amrex::Gpu::PinnedVector`. If a vector that uses `std::allocator` by default is desired, people can always use `amrex::Vector`.

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_GpuContainers.H
+++ b/Src/Base/AMReX_GpuContainers.H
@@ -42,7 +42,7 @@ namespace Gpu {
 
     /**
      * \brief A PODVector that uses pinned host memory. Same as PinnedVector.
-     * For a vector class that uses std::allocator, see amrex::Vector.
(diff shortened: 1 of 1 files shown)


maximumcats:
I think this is a reasonable decision. We really should be avoiding pageable memory for any code in the performance critical path and making that just work under the hood makes it easier on the developer.

WeiqunZhang (approved):

PR #1933, opened 2021-04-07
Fix other missing Real()

lucafedeli88:
## Summary
This PR fixes few missing `Real()` in `AMReX_EB2_3D_C.cpp`, which currently prevent the compilation of some methods in single precision.

## Checklist

The proposed changes:


WeiqunZhang:
Please note that the EB code has never been tested for single precision and there are some parameters that are certainly not suitable for single precision.

WeiqunZhang (approved):

PR #1934, opened 2021-04-07
CMake: add support for BOUND_CHECK

mic84:
## Checklist

The proposed changes:

### DIFF
--- a/Tools/CMake/AMReXOptions.cmake
+++ b/Tools/CMake/AMReXOptions.cmake
@@ -283,6 +283,8 @@ endif ()
 
 print_option( AMReX_ASSERTIONS )
 
+option(AMReX_BOUND_CHECK  "Enable bound checking in Array4 class" OFF)
+print_option( AMReX_BOUND_CHECK )
 
 #
 # Profiling options  =========================================================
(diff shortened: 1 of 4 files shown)


WeiqunZhang (approved):

Tools/CMake/AMReXOptions.cmake
```diff
@@ -283,6 +283,8 @@ endif ()
 
 print_option( AMReX_ASSERTIONS )
 
+option(AMReX_BOUND_CHECK  "Enable bound checking in Array4 class" OFF)
```
ax3l, line 286: Oh cool, should we enable this by default with Debug builds?
How is the corresponding GNUmake logic?

WeiqunZhang: It's enabled with debug builds.

WeiqunZhang: My message was probably not very clear.  The `AMREX_BOUND_CHECK` macro is only used like below
```
```

ax3l: Ok perfect, then we can keep the build system logic simple in both GNUmake & CMake.

Issue #1937, opened 2021-04-08
amrex crashes when built with latest version of intel oneapi


lucaparisi91:
Current version of Amrex crashes with Segfault when initialized. 
Amrex library was built with cmake command from build directory 

export CC=icc
export CXX=icpc
export CF=ifort
cmake ../amrex  -DAMReX_OMP=YES -DAMReX_BASE_PROFILE=YES 

Compilation is successfull but crashes when doing

make test_install

cmake version : 3.20.0
intel oneApi version : 2021.2.0

WeiqunZhang:
I can reproduce it with cmake.  It's also reproducible with
```
cd  amrex/Tests/Amr/Advection_AmrCore/Exec
make -j16 COMP=intel USE_MPI=FALSE
./main2d.intel.ex inputs
```

This seems to be a compiler bug, and I think I have found a workaround.  #1938

WeiqunZhang:
Could you verify the issue is gone with the latest amrex?

lucaparisi91:
I confirm that amrex is now building properly. Thanks for the fix.

PR #1938, opened 2021-04-08
Workaround a compiler bug(?) reported in #1937

WeiqunZhang:


kngott:
Do we want to add a

`// todo intel bug`

comment here to track this bug?

WeiqunZhang:
There is nothing we need to change even after this bug is fixed.  Both are legal code.

kngott (approved):

PR #1941, opened 2021-04-11
Avoid AMREX_NO_UNIQUE_ADDRESS

WeiqunZhang:
* Avoid AMREX_NO_UNIQUE_ADDRESS because it causes gcc 9.3 to crash.

* Fix AMREX_NO_UNIQUE_ADDRESS and AMREX_NODISCARD macros.

* Adjust compiler flag in CI to allow for C++17 extension when compiling with C++14.

## Checklist

The proposed changes:


kngott (approved):
Ooo... I wonder where we gonna use "if constexpr". 😄

WeiqunZhang:
Next PR.

PR #1948, opened 2021-04-15
Update HDF5 I/O to use new HDF5 Async VOL APIs

houjun:
## Summary
Update the HDF5 I/O to (optionally) use the latest HDF5 Async VOL EventSet APIs (*_async) when enabled in GNUMakefile, the original HDF5 I/O is not affected if Async VOL is not enabled.

## Checklist

The proposed changes:


Src/Base/AMReX_PlotFileUtil.cpp
```diff
@@ -14,6 +14,7 @@
 
 #ifdef AMREX_USE_HDF5_ASYNC
 #include "h5_vol_external_async_native.h"
+hid_t es_id_g = 0;
```
WeiqunZhang: Can we put this inside `namespace {}`?

Src/Particle/AMReX_Particles.H
```diff
@@ -48,6 +48,10 @@
 
 #ifdef AMREX_USE_HDF5
 #include <hdf5.h>
+#ifdef AMREX_USE_HDF5_ASYNC
+#include "h5_vol_external_async_native.h"
+extern hid_t es_par_g;
```
WeiqunZhang: Could we prefix this global variable with `amrex_` or put it in `namespace amrex`?

houjun: @WeiqunZhang  Just moved them (es_par_g and es_id_g) to amrex namespace.

houjun:
Hi @atmyers, got a chance to review this?

atmyers (approved):

PR #1958, opened 2021-04-19
DPCPP: fix atomicAdd for float

WeiqunZhang:
We recently removed the template specialization for atomicAdd of float in
the DPC++ backend.  That was a mistake because SYCL does not support
atomicAdd for float, and we have to use atomicCAS to implement it.

## Checklist

The proposed changes:


kngott (approved):

ax3l (approved):
Thank you! :)

Looks like we could add also more SP tests to CI in AMReX to cover those code paths.

PR #1962, opened 2021-04-22
Using rocPRIM in HIP Version of Scan

WeiqunZhang:
Reimplement HIP version of scan using rocprim.  This uses a few things in
rocprim::detail.  Hopefully it will not break by a future release of HIP.

Also some minor tweaks to the CUDA version of Scan.

## Checklist

The proposed changes:


kngott (approved):

WeiqunZhang:
This PR also fixes the issue caused by the latest rocm release that broke our own implementation of scan for long int.

PR #1967, opened 2021-04-23
GNU Make: -MMD -MP

WeiqunZhang:
Use -MMD -MP instead of -M to generate dependency unless it is not supported
by the compiler.  In the new approach, the dependencies are generated during
compilation without a separate step.

## Checklist

The proposed changes:


WeiqunZhang:
Ready for review, but more testing is needed.

WeiqunZhang:
This is now ready.

kngott (approved):

PR #1973, opened 2021-04-26
PCData object for non-blocking ParallelCopy

kngott:
## Summary

Adds an object to pass the temporary data needed for non-blocking comms. Should substantially reduce the amount of metadata data stored in the FabArray object, as well as open up additional options for non-blocking comms.

Blocking comms hide this, no changes needed; non-blocking users will need to store the object from `_nowait` and pass it to `_finish`. Current expected use is:
```
auto pcdata = mf.ParallelCopy_nowait(<user's favorite flavor of the call>);
<Do other work>
mf.ParallelCopy_finish(pcdata);
```

Data type of pcdata is `std::unique_ptr<PCData<FAB>>`. Unique_ptr is used to avoid an unwanted call to the PCData destructor at the end of `_nowait`, allowing the destructor to test for correctness (or anything else we may want to do).

Once this is discussed and flushed out, I'll add a `SumBoundary_nowait` and `SumBoundary_finish` using this object (and this methodology can be expanded to any other PC wrapped call, as desired).

## Additional background

- Apparently, non-blocking comms isn't anywhere in the docs? I'll add that once these changes are completed. 
- Once this is flushed out, will go for FillBoundary as well.

## Checklist

The proposed changes:


WeiqunZhang:
Looks good to me.  I have a few minor comments.  We could use `AMREX_NODISCARD` to catch errors at compile time.  It seems that the `nowaited` member variable in PCData is unnecessary.  The `_finish` function's parameter could be value instead of reference.  This does force the user to use an extra move.  However, this will also force the life time of it to end in the finish function.

kngott:
`NO_DISCARD` and `nowaited`: agreed.

I like the `std::move` idea. We'd definitely need to add docs to make it clear we expect users to do a `std::move`, as I doubt we require that in many places. I guess the question is: would that be too restrictive? If we go with a move, the user then needs to be extra careful with the remaining "object" (I'm imagining problems of accidentally trying to access the now-dead-object if it's stored in a cache of vectors.) Do we want to force that on users in one of the most basic AMReX functions?

WeiqunZhang:
I can agree that requiring `std::move` might be too restrictive for users.  But I cannot imagine that anyone would need to access the internal data of PCData, especially after the finish call.  It should be regarded as an opaque object.

WeiqunZhang (approved):

PR #1979, opened 2021-04-27
Use full path to AMReX_Config.H in gnu make target

### LINKED ISSUE #1978: GNUmake dependency issues
Hello, is there anything we need to do with GNUmake in our projects to accommodate #1975 and #1967 ? We are getting:
```
make: *** No rule to make target `AMReX_Config.H', needed by `tmp_build_dir/o/3d.llvm.EXE/prob.o'.  Stop.
```

WeiqunZhang:
## Checklist

The proposed changes:

### DIFF
--- a/Tools/GNUMake/Make.rules
+++ b/Tools/GNUMake/Make.rules
@@ -194,7 +194,7 @@ FORCE:
 #
 # Rules for objects.
 #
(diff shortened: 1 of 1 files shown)


zingale (approved):
Castro builds fine with this change

PR #1990, opened 2021-05-04
Define BL_USE_MPI compilation flag when MPI and Fortran are enabled

### LINKED ISSUE #1989: Initialization fails when MPI and Fortran are enabled
When calling `amrex_init` from Fortran:

I get the following error message:

```
*** The MPI_Comm_f2c() function was called before MPI_INIT was invoked.
*** This is disallowed by the MPI standard.
*** Your MPI job will now abort.
```

This seems like a regression that was introduced in v21.01. I'll submit a pull request with a fix.

zjibben:
## Summary
Resolves #1989 

## Checklist

The proposed changes:

### DIFF
--- a/Tools/CMake/AMReXSetDefines.cmake
+++ b/Tools/CMake/AMReXSetDefines.cmake
@@ -79,6 +79,12 @@ if (AMReX_FORTRAN)
       $<$<COMPILE_LANGUAGE:Fortran>:BL_LANG_FORT AMREX_LANG_FORT>
       )
 
+    if (AMReX_MPI)
+      target_compile_definitions( amrex PRIVATE
+        $<$<COMPILE_LANGUAGE:Fortran>:BL_USE_MPI>
+        )
+    endif()
+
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
I don't think this is the right fix.

WeiqunZhang:
We should only define `BL_USE_MPI` and `AMREX_USE_MPI` when AMReX_MPI is on.

WeiqunZhang:
Other than that, I believe this is a valid fix.

zjibben:
Thanks. You're right, I didn't quite do the right thing here. In fact the problem only arises due to the lack of `BL_USE_MPI`. I'll modify my commit.

zjibben:
I just overwrote my commit. This one will only define `BL_USE_MPI` if `AMReX_MPI` is on.

WeiqunZhang:
Great! Thanks!

mic84 (approved):

PR #1991, opened 2021-05-04
Non-blocking SumBoundary

kngott:
## Summary

Non-blocking SumBoundary, using the stored MF pointer in PCData. 

Could use a double-check that this is the preferred, non-memory leak way to do this. Feels a bit unsafe, as is.

## Additional background

SumBoundary isn't described in the docs anywhere, so we'd have to add SumBoundary before adding this.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_MultiFab.cpp
+++ b/Src/Base/AMReX_MultiFab.cpp
@@ -1238,7 +1238,8 @@ MultiFab::SumBoundary_finish ()
 {
     BL_PROFILE("MultiFab::SumBoundary_finish()");
 
-    if ( n_grow == IntVect::TheZeroVector() && boxArray().ixType().cellCentered()) return;
+    // If pcd doesn't exist, was all local and operation was fully completed in "SumBoundary_nowait".
+    if ( (n_grow == IntVect::TheZeroVector() && boxArray().ixType().cellCentered()) || !(this->pcd) ) return;
 
     FabArray<FArrayBox>* tmp = const_cast<FabArray<FArrayBox>*> (this->pcd->src);
     this->ParallelCopy_finish();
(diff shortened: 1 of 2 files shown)


Src/Base/AMReX_MultiFab.cpp
```diff
@@ -1187,29 +1187,63 @@ MultiFab::negate (const Box& region, int comp, int num_comp, int nghost)
...
+MultiFab::SumBoundary_finish ()
+{
+    BL_PROFILE("MultiFab::SumBoundary_finish()");
+
+    // If pcd doesn't exist, was all local and operation was fully completed in "SumBoundary_nowait".
+    if ( (n_grow == IntVect::TheZeroVector() && boxArray().ixType().cellCentered()) || !(this->pcd) ) return;
+
+    FabArray<FArrayBox>* tmp = const_cast<FabArray<FArrayBox>*> (this->pcd->src);
+    this->ParallelCopy_finish();
+    tmp->clear();
```
WeiqunZhang: There is a memory leak here.  This should be `delete tmp` instead.

Src/Base/AMReX_MultiFab.cpp
```diff
@@ -1187,29 +1187,63 @@ MultiFab::negate (const Box& region, int comp, int num_comp, int nghost)
...
 {
-    SumBoundary(0, n_comp, IntVect(0), period);
+    BL_PROFILE("MultiFab::SumBoundary_nowait()");
+
+    if ( n_grow == IntVect::TheZeroVector() && boxArray().ixType().cellCentered()) return;
+
+    MultiFab* tmp = new MultiFab( boxArray(), DistributionMap(), ncomp, n_grow, MFInfo(), Factory() );
+    MultiFab::Copy(*tmp, *this, scomp, 0, ncomp, n_grow);
+    this->setVal(0.0, scomp, ncomp, nghost);
+    this->ParallelCopy_nowait(*tmp,0,scomp,ncomp,n_grow,nghost,period,FabArrayBase::ADD);
```
WeiqunZhang, line 1233: There is a memory leak here.  Need to test if this->pcd is built.  If not delete tmp.

WeiqunZhang (approved):

PR #1997, opened 2021-05-05
Templated implementation of ostream operator for both Dim3 and XDim3

atmyers:
This makes `amrex::Print()`, etc. work for both data types.

The proposed changes:


Src/Base/AMReX_Dim3.H
```diff
@@ -6,13 +6,21 @@
 #include <AMReX_GpuQualifiers.H>
 
 #include <iosfwd>
```
WeiqunZhang: This is no longer needed.

WeiqunZhang (approved):

PR #1999, opened 2021-05-06
Move Morton code utility functions somewhere they can be reused.

atmyers:
This also marks them host device so they can be used on the GPU. 

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_Morton.H
+++ b/Src/Base/AMReX_Morton.H
@@ -68,6 +68,8 @@ std::uint32_t splitBits10 (std::uint32_t x) noexcept {
     // 0x55555555  : 0101,0101,0101,0101,0101,0101,0101,0101
     // x           : 0g0f,0e0d,0c0b,0a09,0807,0605,0403,0201
     return x;
+#elif (AMREX_SPACEDIM == 1)
+    return x;
 #endif
 }
 
--- a/Src/Base/AMReX_Morton.H
+++ b/Src/Base/AMReX_Morton.H
@@ -1,0 +1,112 @@
+#ifndef AMREX_MORTON_H_
+#define AMREX_MORTON_H_
+
+#include <AMReX_REAL.H>
+#include <AMReX_Dim3.H>
+#include <AMReX_Array.H>
+#include <AMReX_GpuQualifiers.H>
+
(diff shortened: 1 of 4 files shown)


Src/Base/AMReX_Morton.H
```diff
@@ -0,0 +1,114 @@
+#ifndef AMREX_MORTON_H_
+#define AMREX_MORTON_H_
+
```
WeiqunZhang, line 4: Can we add `#include <AMReX_Config.H>` here?

Src/Base/AMReX_Morton.H
```diff
@@ -0,0 +1,114 @@
...
+ * \brief
+ *  On input, the lowest 10 bits of x are filled, the next 22 are 0
+ *  On output, x has been "stretched" to 30 bits, with each input bit
+ *  followed by interleaving bits set to 0. In 2D, there is one
+ *  interleaving bit, and in 3D, there are 2.
+ *
+ * \param x unsigned int holding the input to be split
+ */
+AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+std::uint32_t splitBits10 (std::uint32_t x) noexcept {
```
WeiqunZhang: The name of this function is only true for 3d.  So is the comment.

Src/Base/AMReX_Morton.H
```diff
@@ -0,0 +1,114 @@
...
+ * \param plo the low end of the domain.
+ * \param phi the high end of the domain.
+ */
+AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+std::uint32_t get30BitCode (const XDim3& p,
+                            const GpuArray<amrex::Real,AMREX_SPACEDIM>& plo,
+                            const GpuArray<amrex::Real,AMREX_SPACEDIM>& phi) noexcept {
+    std::uint32_t a = packBits10(p.x, plo[0], phi[0]);
+    std::uint32_t b = packBits10(p.y, plo[1], phi[1]);
+    std::uint32_t c = packBits10(p.z, plo[2], phi[2]);
```
WeiqunZhang: This is out of bound.

WeiqunZhang (approved):

PR #2001, opened 2021-05-07
Cray complier type detection

WeiqunZhang:
There are two types of Cray compilers, the classical and clang based.  The version number
is not a reliable way to determine the type.  For example, cce/9 on Cori has both
versions.  Cray compilers for aarch64 on Ookami are classical.  Thus we switch to grep
"LLVM" in `CC --version`.  This works on both Cori and Ookami.

## Checklist

The proposed changes:

### DIFF
--- a/Tools/GNUMake/comps/cray.mak
+++ b/Tools/GNUMake/comps/cray.mak
@@ -13,26 +13,13 @@ F90FLAGS =
 
 ########################################################################
 
-# CRAY_CC_VERSION is defined by the 'cce' module. Starting with CCE 9, Cray
-# changed the C and C++ compilers to clang/LLVM based, so all of the
-# options/flags for those compilers changed. But the Cray Fortran compiler is
-# still based on CCE 8 and so it has the same options as before.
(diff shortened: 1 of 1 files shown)


Tools/GNUMake/comps/cray.mak
```diff
@@ -13,26 +13,18 @@ F90FLAGS =
...
-  CCE_GE_V9 := FALSE
+  CRAY_IS_CLANG_BASED = FALSE
 endif
 
 # we need to be careful on the A64fx
-ifeq ($(CRAY_CC_VERSION),10.0.1)
-  ifeq ($(CRAY_ARCH_TARGET),aarch64)
-    CCE_GE_V9 := FALSE
-  endif
+ifeq ($(CRAY_ARCH_TARGET),aarch64)
```
atmyers: Why is this needed, in addition to the grep check above? Wouldn't this be false anyway on Ookami?

WeiqunZhang: Oh I forgot to remove the old check.

atmyers (approved):

PR #2007, opened 2021-05-11
Arena initial size

WeiqunZhang:
Add ParmParse parameters to allow the users to change the initial size of
various Arenas. In particular this allows the users to not allocate any
managed memory if they set `amrex.the_arena_is_managed=0
amrex.the_managed_arena_init_size=0`.

## Checklist

The proposed changes:


atmyers:
It looks like we managed to crash the HIP compiler in CI - I guess we should mark that check optional for now?

atmyers (approved):

WeiqunZhang:
That's the known VOP bug.

PR #2025, opened 2021-05-15
Work-Around: oneAPI 2021.3.0 SYCL with oneDPL

ax3l:
Add recommended work-around flags to oneAPI 2021.3.0 SYCL compilers.

https://software.intel.com/content/www/us/en/develop/articles/intel-oneapi-dpcpp-library-release-notes.html#inpage-nav-2-3
```
The use of oneDPL together with the GNU C++ standard library
(libstdc++) version 9 or 10 may lead to compilation errors
(caused by oneTBB API changes). To overcome these issues,
include oneDPL header files before the standard C++ header
files, or disable parallel algorithms support in the standard
library. For more information, please see Intel® oneAPI
Threading Building Blocks (oneTBB) Release Notes.
```
```
An application using Parallel STL algorithms in libstdc++ versions 9
and 10 may fail to compile due to incompatible interface changes
between earlier versions of Threading Building Blocks (TBB) and
oneAPI Threading Building Blocks (oneTBB). Disable support for
Parallel STL algorithms by defining PSTL_USE_PARALLEL_POLICIES
(in libstdc++ 9) or _GLIBCXX_USE_TBB_PAR_BACKEND (in libstdc++ 10)
macro to zero before inclusion of the first standard header file in
each translation unit.
```

## Checklist

The proposed changes:


WeiqunZhang:
Could we add more changes to this PR?  Could we have a new cmake bool flag say `AMReX_ONEDPL` or `AMReX_DPCPP_ONEDPL`?  It will only be used when the dpcpp backend is used and the default is off.  If the flag is on, we will define and export those macros as a workaround.  Additionally we define `-DAMREX_USE_ONEDPL` and add it to `AMReX_Config.H` if the onedpl flag is enabled.

There are two reasons for this.  It's not clear what impact this workaround may have.  Second, I like to use oneDPL for scan, but it seems that the performance is actually not as good as what we already have.

ax3l:
Is the main purpose of `oneDPL` to provide a parallel STL? In that case, we could also make a `AMReX_PSTL` flag and use that to try to enable PSTL features for various backends.

I am wondering about that, because NVC++ would be another backend that could react to it.

WeiqunZhang:
Yes, the main purpose of oneDPL for Intel is to provide parallel STL.  But that's not why I am interested in using oneDPL.  I would like to use something like CUB or rocPRIM.  But oneAPI does not have such a component.   We use CUB and rocPRIM in a number of places including iterator based inclusive and exclusive scan.  For those scans, we may eventually use oneDPL.

I am not sure if we want to add `AMReX_PSTL` right now, because we do not have any immediate plan to use Nvidia's PSTL.

ax3l:
Xref: #2027

ax3l:
Updated - shall I merge in #2027?

WeiqunZhang:
It's fine to be two separate PRs.

ax3l:
Works for me, we also have them cross-linked so they are in relation :)

Tools/CMake/AMReXSYCL.cmake
```diff
@@ -68,6 +68,16 @@ if (AMReX_MPI)
     $<${_cxx_dpcpp}:-fsycl-unnamed-lambda>)
 endif ()
 
+if(AMReX_DPCPP_ONEDPL)
+    # TBB and PSTL are broken in oneAPI 2021.3.0
+    # https://software.intel.com/content/www/us/en/develop/articles/intel-oneapi-dpcpp-library-release-notes.html#inpage-nav-2-3
+    if(DPCPP_VERSION VERSION_GREATER_EQUAL 2021.3.0 AND DPCPP_VERSION VERSION_LESS 2021.4.0)
```
WeiqunZhang: The issue has existed at least since 2021.1.1, and probably won't be fixed until glibc version 10 is gone.  So we probably should drop the version test.

WeiqunZhang: I tested this PR with 2021.2.  It fails.  But if I remove the test, it compiles.

ax3l: Thanks for the test, will update!

Tools/CMake/AMReXSYCL.cmake
```diff
@@ -68,6 +68,16 @@ if (AMReX_MPI)
...
+        target_compile_definitions( SYCL
+            INTERFACE
+            $<${_cxx_dpcpp}:_GLIBCXX_USE_TBB_PAR_BACKEND=0 PSTL_USE_PARALLEL_POLICIES=0>)
+    endif()
```
ax3l:
```suggestion
    # at least since 2021.1.1 and probably won't be fixed until glibc version 10 is gone
    target_compile_definitions( SYCL
        INTERFACE
        $<${_cxx_dpcpp}:_GLIBCXX_USE_TBB_PAR_BACKEND=0 PSTL_USE_PARALLEL_POLICIES=0>)
```

WeiqunZhang (approved):

PR #2033, opened 2021-05-18
nvcc --display-error-number --diag-error errNum

WeiqunZhang:
Since 11.2, nvcc can display a diagnostic number for any warning message and
it can turn specified warnings into errors These two new flags are added to
the GNU make system, and it's now an error to write a `__device__` variable in
a host function.

## Checklist

The proposed changes:

### DIFF
--- a/Tools/CMake/AMReXParallelBackends.cmake
+++ b/Tools/CMake/AMReXParallelBackends.cmake
@@ -143,6 +143,12 @@ if (  AMReX_GPU_BACKEND STREQUAL "CUDA"
       endif ()
    endif ()
 
+   # Flags to make it an error to write a device variable in
+   # a host function.
+   if (CMAKE_CUDA_COMPILER_VERSION VERSION_GREATER_EQUAL)
+      list(APPEND _cuda_flag --display-error-number --diag-error 20092)
+   endif ()
+
    target_compile_options( amrex PUBLIC $<${_genex}:${_cuda_flags}> )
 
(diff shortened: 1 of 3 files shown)


WeiqunZhang:
@mic84 @ax3l Could we support this in cmake too?

kngott (approved):

mic84:
@WeiqunZhang  sure. Just to be clear: `--display-error-number --diag-error errNum` should be added for nvcc >= 11.2, correct?

mic84:
@WeiqunZhang done :-)

Tools/CMake/AMReXParallelBackends.cmake
```diff
@@ -143,6 +143,12 @@ if (  AMReX_GPU_BACKEND STREQUAL "CUDA"
       endif ()
    endif ()
 
+   # Flags to make it an error to write a device variable in
+   # a host function.
+   if (CMAKE_CUDA_COMPILER_VERSION VERSION_GREATER_EQUAL 11.2)
+      list(APPEND _cuda_flag --display-error-number --diag-error 20092)
```
ax3l, line 149: Great idea!
So far we make a similarish warning (capture `this`) a warning by default and only error it in CI with an extra option. The basic idea behind this is that "shipping" any kind of "-Werror" is usually considered bad practice and hurts adoption in unexpected environments (e.g. with other/new/in-development compilers).

PR #2034, opened 2021-05-18
HIP: AMREX_AMD_ARCH Env Variable

ax3l:
## Summary

Adding an AMREX_AMD_ARCH env variable to control the AMD GPU arch from user environments & profiles.

## Additional background

Same as #1507 / #1522, but for HIP.

## Checklist

The proposed changes:

### DIFF
--- a/Docs/sphinx_documentation/source/GPU.rst
+++ b/Docs/sphinx_documentation/source/GPU.rst
@@ -333,6 +333,7 @@ Since CMake does not support autodetection of HIP compilers/target architectures
 yet, ``CMAKE_CXX_COMPILER`` must be set to a valid HIP compiler, i.e. ``hipcc`` or ``nvcc``,
 and ``AMReX_AMD_ARCH`` to the target architecture you are building for.
 Thus **AMReX_AMD_ARCH and CMAKE_CXX_COMPILER are required user-inputs when AMReX_GPU_BACKEND=HIP**.
+We again read also an *environment variable*: ``AMREX_AMD_ARCH`` (note: all caps).
 Below is an example configuration for HIP on Tulip:
 
 .. highlight:: console
--- a/Tools/CMake/AMReXOptions.cmake
+++ b/Tools/CMake/AMReXOptions.cmake
@@ -155,8 +155,14 @@ print_option(  AMReX_DPCPP_SPLIT_KERNEL )
 
 # --- HIP ----
 if (AMReX_HIP)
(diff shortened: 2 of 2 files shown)


Docs/sphinx_documentation/source/GPU.rst
```diff
@@ -333,6 +333,7 @@ Since CMake does not support autodetection of HIP compilers/target architectures
 yet, ``CMAKE_CXX_COMPILER`` must be set to a valid HIP compiler, i.e. ``hipcc`` or ``nvcc``,
 and ``AMReX_AMD_ARCH`` to the target architecture you are building for.
```
ax3l, line 334: At some point we could also re-consider that this variable takes right now a comma-separated list instead of `;`-separated (the latter is the std for CMake, the current & former is the pass-through we use for HIP's compiler option)

Tools/CMake/AMReXOptions.cmake
```diff
@@ -155,8 +155,14 @@ print_option(  AMReX_DPCPP_SPLIT_KERNEL )
 
 # --- HIP ----
 if (AMReX_HIP)
-   set(AMReX_AMD_ARCH "IGNORE" CACHE STRING
+   set(AMReX_AMD_ARCH_DEFAULT "IGNORE")
```
ax3l, line 158: Do we really need to require a user-input here?
@WeiqunZhang mentioned that without a flag, `hipcc` picks `gfx803` these days? (Note that AMD build architectures are not forward compatible as of today.) It seem though like `gfx803` was dropped from rocM 4.0? https://github.com/RadeonOpenCompute/ROCm/issues/1353

WeiqunZhang: I saw this if we don't specify amd gpu target
```2021-05-18T23:13:33.9776231Z /opt/rocm-4.1.1/hip/include/hip/hcc_detail/hip_runtime_api.h:55:41: note: expanded from macro 'DEPRECATED'
2021-05-18T23:13:33.9777132Z #define DEPRECATED(msg) __attribute__ ((deprecated(msg)))
2021-05-18T23:13:33.9777664Z                                         ^
2021-05-18T23:13:35.1172406Z 1 warning generated when compiling for gfx803.
```

ax3l: Since things are not forward compatible, we might continue to require a user-choice for the time being.

mic84 (approved):

PR #2035, opened 2021-05-18
Generalize fextract allowing to specify coords perpendicular to slicedir

drummerdoc:
Generalize fextract.  Allow spec as coordinate + dir, so that 1D slice can be positioned as desired, within constraints of backward compatibility.

## Checklist

The proposed changes:

### DIFF
--- a/Tools/Plotfile/fextract.cpp
+++ b/Tools/Plotfile/fextract.cpp
@@ -148,7 +148,7 @@ void main_main()
         kloc = (hi0.z-lo0.z+1)/2 + lo0.z;
     }
 
-    if (xcoord > -1.e-36 && AMREX_SPACEDIM >= 1) {
+    if (xcoord > -1.e36 && AMREX_SPACEDIM >= 1) {
         // we specified the x value to pass through
         iloc = hi0.x;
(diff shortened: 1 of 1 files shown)


Tools/Plotfile/fextract.cpp
```diff
@@ -155,8 +148,21 @@ void main_main()
         kloc = (hi0.z-lo0.z+1)/2 + lo0.z;
     }
 
+    if (xcoord > -1.e-36 && AMREX_SPACEDIM >= 1) {
```
WeiqunZhang: What's special about `-1.e-36`?  Was it meant to be `-1.e36` or `std::numeric_limits<Real>::lowest()`?

drummerdoc: I dunno, I just copied what was there.  Do a git blame and ask the offender :)

WeiqunZhang: I asked myself and concluded that it was a typo.  :facepalm: Could we change all these to `-1.e36`?

WeiqunZhang (approved):

PR #2036, opened 2021-05-19
Wrappers for builtin clz extensions and tests thereof.

atmyers:
Counting the number of leading zeros in an integer is useful for a number of bit-fiddling tasks. GCC, clang, and CUDA all provide builtin functions to do this. However, the results are platform-dependent, because the size of many integer types can vary, and also the CUDA interface is slightly different. 

This:

1) Wraps the builtin versions to work with unsigned, fixed-width types, which should behave the same anywhere they are present. 
2) Provides a backup implementation for when the extension is not available.
3) Adds a test.

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_Algorithm.H
+++ b/Src/Base/AMReX_Algorithm.H
@@ -11,6 +11,8 @@
 #include <algorithm>
 #include <limits>
 #include <type_traits>
+#include <cstdint>
+#include <climits>
 
 namespace amrex
 {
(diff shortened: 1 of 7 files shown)


Src/Base/AMReX_Algorithm.H
```diff
@@ -134,6 +136,117 @@ namespace amrex
...
+// in gcc and clang, there are three versions of __builtin_clz taking unsigned int,
+// unsigned long, and unsigned long long inputs. Because the sizes of these data types
+// vary on different platforms, we work with fixed-width integer types.
+// these tags and overloads select the smallest version of __builtin_clz that will hold the input type
+struct clzll_tag_host {};
+struct clzl_tag_host : clzll_tag_host {};
+struct clz_tag_host : clzl_tag_host {};
+
+template <typename T, typename = typename std::enable_if<sizeof(T) <= sizeof(unsigned int)>::type>
+AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
```
WeiqunZhang: Does this need to be host and device?

atmyers: No I don't believe they do.

WeiqunZhang (approved):

PR #2031, opened 2021-05-18
CMake: HIP Modernizing & RDC

ax3l:
## Summary

Add `-fgpu-rdc` flags to HIP if requested via `AMReX_GPU_RDC`.

Modernize HIP logic with recommended targets that avoid the flaky `hipcc` compiler scripts: https://rocmdocs.amd.com/en/latest/Installation_Guide/Using-CMake-with-AMD-ROCm.html#using-hip-in-cmake
Add support for AMDs `clang++`/`clang` compiler for HIP instead of using the legacy `hipcc` perl wrapper as C++ compiler.
This also increases support towards Cray compiler wrappers, which also refer to `clang++`/`clang` underneath (Spock/OLCF).

Close #1688

## Additional background

Relocatable-device-code (RDC) flags are needed for `extern` device variable support (for codes that use global variables on device). Also needed when linking with Ascent.

Follow-up to  #2029

With HIP GPU RDC, static libs emitting & linking does get more fancy:
https://github.com/ROCmSoftwarePlatform/rccl/pull/260

## Tests

- [x] Compiled in CI
- [ ] Compile on Tulip (ROCm 4.1.1) with Cray Wrappers:
```bash
module unload cuda11.2
module load craype-accel-amd-gfx908
module load rocm/4.1.1
module load cmake/3.18.2
# default: cray-mvapich2/2.3.5

export CC=$(which cc)
export CXX=$(which CC)
export CXXFLAGS="--rocm-path=${ROCM_PATH}"

cmake -S . -B build -DWarpX_COMPUTE=HIP -DAMReX_AMD_ARCH=gfx908
cmake --build build -j 32
# In linking of final executable:
# error: Illegal instruction detected: VOP* instruction violates constant bus restriction
# renamable $sgpr4_sgpr5 = V_CMP_EQ_U64_e64 $exec, killed $vcc, implicit $exec, debug-location !293636; /opt/rocm-4.1.1/hip/../include/hip/hcc_detail/device_functions.h:1021:12
```
- [x] Compile on Tulip (ROCm 4.1.1) with AMD Clang:
```bash
module unload cuda11.2
module load craype-accel-amd-gfx908
module load rocm/4.1.1
module load cmake/3.18.2
# default: cray-mvapich2/2.3.5 as CUDA awareness in it and needs this at link time
module unload cray-mvapich2
module load cray-mvapich2_nogpu

export CC=$(which clang)
export CXX=$(which clang++)  # legacy works, too: export CXX=$(which hipcc)
[...]

PR #2037, opened 2021-05-19
[DO NOT MERGE] Test libamrex built with hip rdc using gnu make

WeiqunZhang:
TESTING ONLY


WeiqunZhang:
Since we have figured out why the test failed to link in #2031, I will close this PR.  It turns out `hipcc  .... -Lpath -lxxx` works, but `hipcc ... path/libxxx.a` does not, when rdc is on.

PR #2041, opened 2021-05-20
Async Output Docs

kngott:
## Summary
Adding Async Output documentation.

Could we check that the listed functions that use Async Output are the forward-facing functions that we want users to interact with and not lower-level functions? I believe I got them all, but I would appreciate a second set of eyes.

## Additional background

We should probably follow up with the HDF5 crowd and see if they can add how to turn on the HDF5 version (and update any HDF5 part of the I/O docs. (Who would we ask, @asalmgren?)

## Checklist

The proposed changes:


Docs/sphinx_documentation/source/IO.rst
```diff
@@ -91,6 +91,54 @@ We note that AMReX does not overwrite old plotfiles if the new
...
+============
+
+AMReX provides the ability to print MultiFabs, plotfiles and
+particle data asynchronously.  Asynchronous output works by creating
+a copy of the data at the time of the call, which is written to disk
+by a persistent thread created during AMReX's initialization.  This allows
+the calculation to continue immediately, which can drastically reduce
+walltime spent writing to disk.
+
+If writing to more than one file per MPI rank, AMReX's async output requires
```
WeiqunZhang: This does not seem right.

Docs/sphinx_documentation/source/IO.rst
```diff
@@ -91,6 +91,54 @@ We note that AMReX does not overwrite old plotfiles if the new
...
+Async Output is performed.  However, this can introduce additional overhead as
+the thread's MPI operations are scheduled safely around each other.  Therefore,
+AMReX uses a lower level of support, SERIALIZED, by default and applications
+have to turn on THREAD_MULTIPLE support.
+
+To turn on Async Output, use the input flag ``amrex.async_out=1``.  The number
+of output files can also be set, using ``amrex.async_out_nfiles``.  The default
+number of files is ``64``. If the number of ranks is larger than the number of
+files, THREAD_MULTIPLE must be turned on by adding
+``DEFINES+=-DMPI_THREAD_MULTIPLE=TRUE`` to the GNUMakefile. Otherwise, AMReX
```
WeiqunZhang: `MPI_THREAD_MULTIPLE=TRUE` is arguably better.

Docs/sphinx_documentation/source/IO.rst
```diff
@@ -91,6 +91,54 @@ We note that AMReX does not overwrite old plotfiles if the new
...
+* Amr::writeSmallPlotFile()
+* Amr::checkpoint()
+* AmrLevel::writePlotFile()
+* StateData::checkPoint()
+* FabSet::write()
+
+Be aware: when using Async Output, a thread is spawned and exclusively used
+to perform output throughout the runtime.  As such, you may oversubscribe
+resources if you launch an AMReX application that assigns all available
+hardware threads in another way, such as OpenMP threadsin.  If you see any
```
WeiqunZhang: `threadsin`: typo?

WeiqunZhang (changes requested):

WeiqunZhang (approved):

PR #2043, opened 2021-05-22
ParamParse: Find Entries under Prefix

ax3l:
## Summary

Extend the "unused params" checks to find generally parameters under a given inputs prefix.

## Additional background

Demonstrator & test in: https://github.com/AMReX-Codes/amrex-tutorials/pull/7

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_ParmParse.cpp
+++ b/Src/Base/AMReX_ParmParse.cpp
@@ -19,6 +19,8 @@
 #include <cctype>
 #include <vector>
 #include <list>
+#include <set>
+#include <string>
 
 extern "C" void amrex_init_namelist (const char*);
 extern "C" void amrex_finalize_namelist ();
@@ -1148,12 +1150,12 @@ ParmParse::getUnusedInputs (const std::string& prefix)
     return r;
 }
 
-std::vector<std::string>
+std::set<std::string>
 ParmParse::getEntries (const std::string& prefix)
 {
     std::vector<std::string> r;
     get_entries_under_prefix(r, g_table, prefix, false, false);
-    return r;
+    return std::set<std::string>(r.begin(), r.end());
 }
 
 void
--- a/Src/Base/AMReX_ParmParse.cpp
+++ b/Src/Base/AMReX_ParmParse.cpp
@@ -1112,23 +1112,30 @@ ParmParse::hasUnusedInputs (const std::string& prefix)
 
 static
 void
-get_unused_inputs(std::vector<std::string>& unused, const ParmParse::Table& table,
-                  const std::string& prefix)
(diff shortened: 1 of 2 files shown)


Src/Base/AMReX_ParmParse.cpp
```diff
@@ -1112,23 +1112,30 @@ ParmParse::hasUnusedInputs (const std::string& prefix)
 
 static
 void
-get_unused_inputs(std::vector<std::string>& unused, const ParmParse::Table& table,
-                  const std::string& prefix)
+get_entries_under_prefix (std::vector<std::string>& found_enties,
+                         const ParmParse::Table& table,
+                         const std::string& prefix,
+                         const bool only_unused = false,
+                         const bool add_values = false)
```
jmsexton03: Consider found_enties => found_entries
While you're adding flags, did you want to add a flag to leverage the duplicate resolution behavior of ParmParse? It looks like this may give you a list of all matching entries, not all unique or last matching entries. ppindex uses a reversed iterator in order to take the last entry. I'm good either way.

ax3l: Thanks, fixed the typo!

> While you're adding flags, did you want to add a flag to leverage the duplicate resolution behavior of ParmParse?

I think I don't know what that is, but the test I did for my use case in https://github.com/AMReX-Codes/amrex-tutorials/pull/7 works well :)

WeiqunZhang: It is allowed to have multiple entries in a ParmParse input file.  For example
```
a = 2
a = 3
a = 4
```
By default, `ParmParse::get` and `query` returns the last one.  The new `getEntries` function will return a vector with `a` appearing 3 times.  I think this is what @jmsexton03 is talking about.

ax3l: Thanks, got it!

For my current use case, it's not problematic that duplicates are returned because I'll query their values with the usual `get` and `query` calls :)
But if it's clearner, I can add the values to an (`unordered_`)`set` if you like?

jmsexton03: Thanks for clarifying my question @WeiqunZhang, I think the tutorial update mentioned above wouldn't have shown the multiple entries case since it doesn't use an input file.
Using a `set` sounds clearer than a `vector` to me, but I don't have a strong preference.

ax3l: I'll use a `set` then over an `unordered_set`. The former is more expensive, but the sorting comes handy for use cases like sorting for sub-sub options later on.

WeiqunZhang (approved):

PR #2046, opened 2021-05-24
When building neighbor masks, perform setVal on device so it will work without managed memory.

atmyers:
This fixes an issue reported by Chris Lishka from Intel re: mfix. 

The proposed changes:


WeiqunZhang:
I wonder if we should do something like what's in https://github.com/AMReX-Codes/amrex/blob/development/Src/Base/AMReX_MultiFabUtil.cpp#L712 to reduce the number of setVal kernel launches to 1.

atmyers:
See update.

WeiqunZhang (approved):

PR #2056, opened 2021-05-25
Fix to correct CUDA flag and add missing libraries in hypre cmake

drangara:
## Summary
When CUDA and HYPRE are enabled these libraries need to be added to the link line when using CMake similar to how they are currently added in `Tools/GNUMake/packages/Make.hypre`.  Also fixed to using the correct flag for CUDA.

## Checklist

The proposed changes:


Tools/CMake/AMReXThirdPartyLibraries.cmake
```diff
@@ -57,8 +57,11 @@ endif ()
 # HYPRE
 #
 if (AMReX_HYPRE)
-    if(ENABLE_CUDA)
+    if(AMReX_CUDA)
         find_package(HYPRE 2.20.0 REQUIRED)
+        find_library(LIBCUSPARSE cusparse PATHS ${CMAKE_CUDA_IMPLICIT_LINK_DIRECTORIES})
+        find_library(LIBCURAND curand PATHS ${CMAKE_CUDA_IMPLICIT_LINK_DIRECTORIES})
+        target_link_libraries(amrex PUBLIC ${LIBCUSPARSE} ${LIBCURAND})
```
sayerhs: Recommend using the CMake targets syntax

```
find_package(CUDAToolKit REQUIRED)
target_link_libraries(amrex PUBLIC 
    CUDA::cusparse    
    CUDA::curand    
    CUDA::cudart
    CUDA::cublas
    CUDA::nvToolsExt)
```

This will propagate additional include flags to application targets.

drangara: thank you. will make the change. do you think we need the other libraries though i.e. `cudart`, `cublas` and `nvToolsExt`? I was trying to be consistent with [Make.hypre](https://github.com/AMReX-Codes/amrex/blob/development/Tools/GNUMake/packages/Make.hypre) and don't see them there

sayerhs: @drangara I was just being explicit about all the additional libraries hypre links to. It is likely that `CUDA::cusparse` and `CUDA::curand` will be sufficient for the code to build as they will in turn depend on `cudart` and `cublas`. `nvToolsExt` is necessary if you use hypre profiling capabilities.

sayerhs (approved):

PR #2057, opened 2021-05-26
Add roctx ranges to TinyProfiler

jmsexton03:
## Summary
Adds roctx markup similarly to existing nvtx markup. 

## Additional background
New features are controlled by added compile flag USE_ROCTX. This assumes the location of roctracer/libroctracer64 and roctracer/libroctx64 are similar to other ROCm library installations such as rocrand. 

Example run to generate results.json
```
export HIP_PATH=/path/to/rocm/root
cd amrex/Tests/GPU/CNS/Exec/Sod
make USE_HIP=TRUE TINY_PROFILE=TRUE USE_ROCTX=TRUE USE_MPI=FALSE NO_CONFIG_CHECKING=TRUE
srun -n 1 ${HIP_PATH}/bin/rocprof  --hsa-trace --stats --timestamp on --roctx-trace  ./CNS3d.*.TPROF.HIP.ex inputs
```

This results.json file can be viewed with a browser as described in the documentation using chrome://tracing/
https://github.com/ROCm-Developer-Tools/rocprofiler/blob/amd-master/doc/rocprof.md#43rd-party-visualization-tools

TODO:
Add CMake support for the new flag
Minimum ROCm version enforcement should be updated (or removed)

## Checklist

The proposed changes:


Tools/CMake/AMReXParallelBackends.cmake
```diff
@@ -210,6 +210,10 @@ if (AMReX_HIP)
    find_package(rocrand REQUIRED CONFIG)
    find_package(rocprim REQUIRED CONFIG)
    find_package(hiprand REQUIRED CONFIG)
+   if (AMReX_ROCTX)
+   target_include_directories(amrex PUBLIC ${HIP_PATH}/../roctracer/include ${HIP_PATH}/../rocprofiler/include)
```
ax3l: I think this might be a tool we can `find_package()` for: https://github.com/ROCm-Developer-Tools/roctracer

Then we can probably do: `target_link_libraries(roc::roctracer)` or so - that sets includes & libs :)

jmsexton03: I tried something similar to that here: https://github.com/AMReX-Codes/amrex/pull/2057/commits/81230627bc8391f1f0e0c5f373312b569e11cdaa#diff-f817df7657a979c7f7e4013a878563aeaadfbbc09a12d5b60bbf697a04ef4ff4R212-R218

There's no cmake directory in roctracer/libs like there is in rocrand/libs. I couldn't find any obvious place for a cmake config file, and it didn't look like roctracer generates one. How do I find out what alias is appropriate?

Tools/GNUMake/comps/hip.mak
```diff
@@ -19,6 +22,10 @@ else
 endif
 
 #if less than a given version, throw error.
+hipcc_major_lt_4 = $(shell expr $(hipcc_major_version) \< 4)
```
WeiqunZhang: I don't think we need to worry about ROCm version right now.  ROCm is under actively development and our target will be the latest version that works for us until it's more stable.

Src/Base/AMReX_GpuDevice.cpp
```diff
@@ -19,6 +19,16 @@
 #endif
 #endif
 
+#if defined(AMREX_USE_HIP)
+#include <hip/hip_runtime.h>
+#if defined(AMREX_USE_ROCTX)
+#include "roctracer_ext.h"
```
WeiqunZhang: Can we use `<>`?

WeiqunZhang:
On spock, one can also compile with `make USE_HIP=TRUE TINY_PROFILE=TRUE USE_ROCTX=TRUE HIP_PATH=/opt/rocm-4.1.0/hip` without using `NO_CONFIG_CHECKING=TRUE`.

Src/Base/AMReX_GpuDevice.cpp
```diff
@@ -343,6 +353,9 @@ Device::Initialize ()
             amrex::Print() << "HIP initialized.\n";
         }
     }
+#if defined(AMREX_USE_ROCTX)
+    roctracer_start();
```
WeiqunZhang: For some reason, we call `profileStart` twice in `ifdef AMREX_USE_CUDA`.  I assume that's uncessary. (@maxpkatz)  So I suggest we move `roctracer_start` into `profileStart`, remove the two calls to `profileStart` in `#ifdef AMREX_USE_CUDA`, and put a call to `profileStart` at the end of `Device::Initialize`.

Src/Base/AMReX_GpuDevice.cpp
```diff
@@ -358,6 +371,11 @@ Device::Finalize ()
     cudaProfilerStop();
 #endif
 
+#if defined(AMREX_USE_HIP) && defined(AMREX_USE_ROCTX)
+    roctracer_stop();
```
WeiqunZhang: I suggest we move roctracer_stop to `Device::profilerStop`, remove the call to `cudaProfilerStop` above, and call `Device::profileStop()`.

Src/Base/AMReX_TinyProfiler.cpp
```diff
@@ -98,6 +98,12 @@ TinyProfiler::start () noexcept
         }
         nvtxRangePush(fname.c_str());
 #endif
+#if defined(AMREX_USE_HIP) && defined(AMREX_USE_ROCTX)
```
WeiqunZhang: Might be better to use `#elif defined()`

Src/Base/AMReX_TinyProfiler.cpp
```diff
@@ -178,6 +184,12 @@ TinyProfiler::stop () noexcept
                 amrex::Gpu::Device::synchronize();
             }
             nvtxRangePop();
+#endif
+#if defined(AMREX_USE_HIP) && defined(AMREX_USE_ROCTX)
+        if (device_synchronize_around_region) {
```
WeiqunZhang: Also we probably should move
```
        if (device_synchronize_around_region) {
            amrex::Gpu::Device::synchronize();
        }
```
outside the ifdefs so that it's called for dpcpp too.

WeiqunZhang:
Once the rocm issue is resolved on spock, we should remove AMREX_USE_ROCTX to be consistent with how we do things in CUDA.

Tools/CMake/AMReXParallelBackends.cmake
jmsexton03: I get this error when I try to use findPackage on roctracer, with or without setting `roctracer_DIR`:

```CMake Error at subprojects/amrex/Tools/CMake/AMReXParallelBackends.cmake:214 (find_package):
  Could not find a package configuration file provided by "roctracer" with
  any of the following names:

    roctracerConfig.cmake
    roctracer-config.cmake

  Add the installation prefix of "roctracer" to CMAKE_PREFIX_PATH or set
  "roctracer_DIR" to a directory containing one of the above files.  If
  "roctracer" provides a separate development package or SDK, be sure it has
  been installed.```

File: Src/Base/AMReX_GpuDevice.cpp (hunk 4 above)
Comment (maximumcats) [reply Src/Base/AMReX_GpuDevice.cpp]: I don't know why it's there twice but I agree that we only need one of those two., and it makes sense to put it at the end of Initialize.

Comment (kngott) [reply Src/Base/AMReX_GpuDevice.cpp]: I might not be remembering this right, but did we try to put a nvtx region around the Initialization to ID if there was unneeded start-up at some point? Is that why it's in there twice, we tried to move the init back to allow wrapping Init?

File: Tools/CMake/AMReXParallelBackends.cmake (hunk 1 above)
Comment (ax3l) [reply Tools/CMake/AMReXParallelBackends.cmake]: Ah I see, ok I submitted an issue so that we can modernize accordingly in the near future: https://github.com/ROCm-Developer-Tools/roctracer/issues/56

File: Tools/CMake/AMReXParallelBackends.cmake [hunk 8]
```diff
@@ -210,6 +210,10 @@ if (AMReX_HIP)
...
+   target_link_libraries(amrex PUBLIC "-L${HIP_PATH}/../roctracer/lib/ -lroctracer64" "-L${HIP_PATH}/../roctracer/lib -lroctx64")
```
ax3l: Just alignment:
```suggestion
       target_include_directories(amrex PUBLIC ${HIP_PATH}/../roctracer/include ${HIP_PATH}/../rocprofiler/include)
       target_link_libraries(amrex PUBLIC "-L${HIP_PATH}/../roctracer/lib/ -lroctracer64" "-L${HIP_PATH}/../roctracer/lib -lroctx64")
```

ax3l:
Is there already an `apt` package that we can install in CI for `roctracer`?
https://github.com/AMReX-Codes/amrex/blob/db8e47927065fcf283f6aae631f91a07b222876b/.github/workflows/dependencies/dependencies_hip.sh#L39

jmsexton03:
> Is there already an `apt` package that we can install in CI for `roctracer`?
> https://github.com/AMReX-Codes/amrex/blob/db8e47927065fcf283f6aae631f91a07b222876b/.github/workflows/dependencies/dependencies_hip.sh#L39

roctracer-dev should be a dependency of rocm-dev, which we already include

ax3l:
Ok, feel free to add `roctracer-dev` explicitly anyway to make the dependency more visible.

Src/Base/AMReX_TinyProfiler.cpp
```diff
@@ -92,12 +92,18 @@ TinyProfiler::start () noexcept
...
+            }
+#endif
+
 #ifdef AMREX_USE_CUDA
-        if (device_synchronize_around_region) {
-            amrex::Gpu::Device::synchronize();
-        }
         nvtxRangePush(fname.c_str());
 #endif
+#if defined(AMREX_USE_HIP) && defined(AMREX_USE_ROCTX)
```
ax3l: Weiqun's suggestion above:
```suggestion
#elif defined(AMREX_USE_HIP) && defined(AMREX_USE_ROCTX)
```

Tools/CMake/AMReXOptions.cmake
```diff
@@ -321,6 +321,10 @@ cmake_dependent_option(AMReX_PROFPARSER "Enable profile parser" OFF
    "AMReX_BASE_PROFILE;AMReX_TRACE_PROFILE;AMReX_AMRDATA" OFF)
 print_option( AMReX_PROFPARSER )
 
+cmake_dependent_option(AMReX_ROCTX  "Enable roctx markup for HIP with ROCm" OFF
+     "AMReX_GPU_BACKEND STREQUAL HIP" ON)
```
ax3l: This default looks a bit mixed up to me. Ref: https://cmake.org/cmake/help/latest/module/CMakeDependentOption.html

Do you try to turn it on by default if HIP is used (and off otherwise)?
```suggestion
cmake_dependent_option(AMReX_ROCTX  "Enable roctx markup for HIP with ROCm" ON
     "AMReX_GPU_BACKEND STREQUAL HIP" OFF)
```

If you want this to be OFF by default even if HIP is found and _OFF and hidden_ otherwise, then this:
```suggestion
cmake_dependent_option(AMReX_ROCTX  "Enable roctx markup for HIP with ROCm" OFF
     "AMReX_GPU_BACKEND STREQUAL HIP" OFF)
```

Docs/sphinx_documentation/source/BuildingAMReX.rst
```diff
@@ -473,6 +473,8 @@ The list of available options is reported in the :ref:`table <tab:cmakevar>` bel
    +------------------------------+-------------------------------------------------+-------------------------+-----------------------+
    | AMReX_PROFPARSER             |  Build with profile parser support              | NO                      | YES, NO               |
    +------------------------------+-------------------------------------------------+-------------------------+-----------------------+
+   | AMReX_ROCTX                  |  Build with roctx markup profiling support      | NO                      | YES, NO               |
```
ax3l, line 476: Depending on default choice above:
```suggestion
   | AMReX_ROCTX                  |  Build with roctx markup profiling support      | ON (if HIP)                    | YES, NO               |
```

Tools/GNUMake/comps/hip.mak
```diff
@@ -114,6 +115,15 @@ ifeq ($(HIP_COMPILER),clang)
   # rocThrust - Header only
   # SYSTEM_INCLUDE_LOCATIONS += $(ROC_PATH)/rocthrust/include
 
+  ifeq ($(USE_ROCTX),TRUE)
```
ax3l, line 118: We could also turn this on by default with GNUmake builds, as soon as HIP is found.

Tools/GNUMake/comps/hip.mak
```diff
@@ -114,6 +115,15 @@ ifeq ($(HIP_COMPILER),clang)
...
+  # rocTracer
+  CXXFLAGS += -DAMREX_USE_ROCTX
+  HIPCC_FLAGS += -DAMREX_USE_ROCTX
+  SYSTEM_INCLUDE_LOCATIONS += $(ROC_PATH)/roctracer/include $(ROC_PATH)/rocprofiler/include
+  LIBRARY_LOCATIONS += $(ROC_PATH)/roctracer/lib $(ROC_PATH)/rocprofiler/lib
+  LIBRARIES +=  -Wl,--rpath=${ROC_PATH}/roctracer/lib -lroctracer64 -lroctx64
```
ax3l, line 124: Don't we add RPATHS automatically these days from the PR I added in #590 ?
We should probably rely on the `USE_RPATH` GNUmake option for this lib as well, so users can explicitly control RPATHs with a central option alone:
```suggestion
  LIBRARIES +=  -lroctracer64 -lroctx64
```

Tools/CMake/AMReXParallelBackends.cmake
```diff
@@ -210,6 +210,10 @@ if (AMReX_HIP)
    find_package(rocrand REQUIRED CONFIG)
    find_package(rocprim REQUIRED CONFIG)
    find_package(hiprand REQUIRED CONFIG)
+   if (AMReX_ROCTX)
+       target_include_directories(amrex PUBLIC ${HIP_PATH}/../roctracer/include ${HIP_PATH}/../rocprofiler/include)
```
ax3l, line 216: Just to keep taps:
```suggestion
       # To be modernized in the future, please see:
       # https://github.com/ROCm-Developer-Tools/roctracer/issues/56
       target_include_directories(amrex PUBLIC ${HIP_PATH}/../roctracer/include ${HIP_PATH}/../rocprofiler/include)
```

We could use [find_library](https://cmake.org/cmake/help/latest/command/find_library.html) and [find_file](https://cmake.org/cmake/help/latest/command/find_file.html), but I think there is atm no reason to over-engineer this.

Src/Base/AMReX_GpuDevice.cpp
WeiqunZhang: There are only calls to Print() between the two calls to cudaProfilerStart.

Tools/CMake/AMReXOptions.cmake
WeiqunZhang: I believe it should be the second one, which is consistent with the gnu make system.

Docs/sphinx_documentation/source/BuildingAMReX.rst
WeiqunZhang: My view is we should simply remove this option in the future once it's fixed on spock.

WeiqunZhang (approved):

PR #2059, opened 2021-05-26
Perlmutter Make: CUDA_PATH

kngott:
## Summary
Copy CUDA_ROOT/CUDA_PATH logic from cgpu. Required for TINY_PROFILE, among other things.

For now, force users to load a Cuda module (defining a version they're using). Probably good practice anyway, instead of running the default.

## Checklist

The proposed changes:

### DIFF
--- a/Tools/GNUMake/sites/Make.nersc
+++ b/Tools/GNUMake/sites/Make.nersc
@@ -44,7 +44,20 @@ ifeq ($(which_computer),$(filter $(which_computer),perlmutter))
     endif
   endif
 
-  CUDA_ARCH = 80
+  ifeq ($(USE_CUDA),TRUE)
+    CUDA_ARCH = 80
+
+    ifneq ($(CUDA_ROOT),)
+        SYSTEM_CUDA_PATH := $(CUDA_ROOT)
+        COMPILE_CUDA_PATH := $(CUDA_ROOT)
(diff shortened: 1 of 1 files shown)


Tools/GNUMake/sites/Make.nersc
```diff
@@ -44,7 +44,20 @@ ifeq ($(which_computer),$(filter $(which_computer),perlmutter))
...
+    CUDA_ARCH = 80
+
+    ifneq ($(CUDA_ROOT),)
+        SYSTEM_CUDA_PATH := $(CUDA_ROOT)
+        COMPILE_CUDA_PATH := $(CUDA_ROOT)
+    else ifneq ($(CUDA_HOME),)
+        SYSTEM_CUDA_PATH := $(CUDA_HOME)
+        COMPILE_CUDA_PATH := $(CUDA_HOME)
+    else
+        $(error No CUDA_ROOT or CUDA_PATH found. Please load a cuda module.)
```
WeiqunZhang: `CUDA_PATH` or `CUDA_HOME`?

kngott: All three are available in the module as identical paths, so I just added the third as well.

We can tweak later if they change or start pointing to different things.

WeiqunZhang (approved):

PR #2060, opened 2021-05-27
Clean up some unused variables.

kngott:
## Summary

Found during the Perlmutter CMake testing.

- Alternatively, in MultiBlock tests, am I correct that the `ParallelCopyWithItselfIsCorrect` function doesn't do anything if DEBUG is off? Would it be cleaner to wrap the entire function or the call in `#ifdef DEBUG` to make it a no-op?

- Also found:
`/global/homes/k/kngott/amrex/Src/LinearSolvers/MLMG/AMReX_MLMG.cpp(409): warning #177-D: parameter "xs" was declared but never referenced
          detected during instantiation of "std::string amrex::<unnamed>::make_str(Ts...) [with Ts=<const char *, int>]" 
(428): here`

    Do we have anything to quiet an unused parameter pack?

## Checklist

The proposed changes:

### DIFF
--- a/Tests/MultiBlock/IndexType/main.cpp
+++ b/Tests/MultiBlock/IndexType/main.cpp
@@ -56,7 +56,7 @@ bool ParallelCopyWithItselfIsCorrect(amrex::iMultiFab& mf, const amrex::Box& dom
         amrex::LoopOnCpu(section, [&](int i, int j, int k)
         {
             amrex::Dim3 si = dtos(amrex::Dim3{i,j,k});
-            int value = si.x + si.y*nx + si.z*nx*xy;
+            int value = si.x + si.y*nx + si.z*nx*ny;
             fails += (array(i,j,k) != value);
 
             AMREX_ASSERT(fails);  // If DEBUG, crash on first error.
--- a/Tests/MultiBlock/IndexType/main.cpp
+++ b/Tests/MultiBlock/IndexType/main.cpp
@@ -56,7 +56,10 @@ bool ParallelCopyWithItselfIsCorrect(amrex::iMultiFab& mf, const amrex::Box& dom
         amrex::LoopOnCpu(section, [&](int i, int j, int k)
         {
             amrex::Dim3 si = dtos(amrex::Dim3{i,j,k});
-            AMREX_ASSERT(array(i,j,k) == (si.x + si.y*nx + si.z*nx*ny));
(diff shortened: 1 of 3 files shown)


Tests/MultiBlock/IndexType/main.cpp
```diff
@@ -56,8 +56,7 @@ bool ParallelCopyWithItselfIsCorrect(amrex::iMultiFab& mf, const amrex::Box& dom
         amrex::LoopOnCpu(section, [&](int i, int j, int k)
         {
             amrex::Dim3 si = dtos(amrex::Dim3{i,j,k});
-            int value = si.x + si.y*nx + si.z*nx*ny;
-            AMREX_ASSERT(array(i,j,k) == value);
+            AMREX_ASSERT(array(i,j,k) == (si.x + si.y*nx + si.z*nx*ny));
```
WeiqunZhang: I suppose we also need `fails += array(i,j,k) != value` in addition to the assert.  Otherwise `fails` does not change.

WeiqunZhang:
I assume `std::forward` can be used on parameter pack if needed.  However, it looks to me what it should be complaining about is `blp_mgv_down_lev_str` in line 428 of AMReX_MLMG.cpp is never used.  So you can try to merge that line into the next line without creating `blp_mgv_down_lev_str`.

WeiqunZhang:
Note that `make_str` is only used in two places in that file.  So we could remove it entirely and simply to
```
BL_PROFILE_VAR("MLMG::...."+std::to_string(mglev), blp_mgv_up_lev);
```

Tests/MultiBlock/IndexType/main.cpp
```diff
@@ -57,7 +57,9 @@ bool ParallelCopyWithItselfIsCorrect(amrex::iMultiFab& mf, const amrex::Box& dom
         {
             amrex::Dim3 si = dtos(amrex::Dim3{i,j,k});
             int value = si.x + si.y*nx + si.z*nx*ny;
-            AMREX_ASSERT(array(i,j,k) == value);
+            fails += (array(i,j,k) != value);
+
+            AMREX_ASSERT(fails);  // If DEBUG, crash on first error.
```
WeiqunZhang: It should be `AMREX_ASSERT(fails == 0);`

WeiqunZhang (approved):

PR #2061, opened 2021-05-27
GPU RDC Support

WeiqunZhang:
Make GPU RDC configurable, but still on by default.  The user can disable it by `USE_GPU_RDC=FALSE` in
gnu make and `-DAMReX_GPU_RDC=OFF` in cmake.

Remove CUDA specific random number functions that rely on rdc.

## Checklist

The proposed changes:

### DIFF
--- a/Tools/CMake/AMReXOptions.cmake
+++ b/Tools/CMake/AMReXOptions.cmake
@@ -183,7 +183,13 @@ endif ()
 #
 # GPU RDC support
 #
-cmake_dependent_option( AMReX_GPU_RDC "Enable GPU RDC" OFF
+# https://cmake.org/cmake/help/latest/variable/CMAKE_CUDA_SEPARABLE_COMPILATION.html
+set(_GPU_RDC_default OFF)
+if(AMReX_CUDA AND DEFINED CMAKE_CUDA_SEPARABLE_COMPILATION)
+    set(_GPU_RDC_default "${CMAKE_CUDA_SEPARABLE_COMPILATION}")
+endif()
+cmake_dependent_option( AMReX_GPU_RDC "Enable GPU RDC" ${_GPU_RDC_default} )
+unset(_GPU_RDC_default)
    "AMReX_CUDA OR AMReX_HIP" OFF)
 print_option(AMReX_GPU_RDC)
 
--- a/Tools/CMake/AMReXOptions.cmake
+++ b/Tools/CMake/AMReXOptions.cmake
@@ -180,6 +180,12 @@ if (AMReX_CUDA OR AMReX_HIP)
    print_option(GPUS_PER_NODE)
 endif ()
 
+#
+# GPU RDC support
+#
+cmake_dependent_option( AMReX_GPU_RDC "Enable GPU RDC" OFF
+   "AMReX_CUDA OR AMReX_HIP" OFF)
+print_option(AMReX_GPU_RDC)
 
 #
 # Parallel backends    ========================================================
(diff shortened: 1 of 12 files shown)


Tools/CMake/AMReXOptions.cmake
```diff
@@ -180,6 +180,12 @@ if (AMReX_CUDA OR AMReX_HIP)
    print_option(GPUS_PER_NODE)
 endif ()
 
+#
+# GPU RDC support
+#
+cmake_dependent_option( AMReX_GPU_RDC "Enable GPU RDC" OFF
```
ax3l: In order to play well with dependent projects that use the standardized CMake CUDA logic, let's add this logic: https://cmake.org/cmake/help/latest/prop_tgt/CUDA_SEPARABLE_COMPILATION.html#prop_tgt:CUDA_SEPARABLE_COMPILATION
```suggestion
# https://cmake.org/cmake/help/latest/variable/CMAKE_CUDA_SEPARABLE_COMPILATION.html
set(_GPU_RDC_default OFF)
if(AMReX_CUDA AND DEFINED CMAKE_CUDA_SEPARABLE_COMPILATION)
    set(_GPU_RDC_default "${CMAKE_CUDA_SEPARABLE_COMPILATION}")
endif()
cmake_dependent_option( AMReX_GPU_RDC "Enable GPU RDC" ${_GPU_RDC_default} )
unset(_GPU_RDC_default)
```

That way a user can do a super-build and pass `-DCMAKE_CUDA_SEPARABLE_COMPILATION=ON` in a deep software stack - and we will honor that request for our targets, too. (There is no standardization on HIP yet.)

Tools/CMake/AMReXTargetHelpers.cmake
```diff
@@ -114,12 +114,12 @@ endfunction ()
...
-   set_target_properties( ${_target}
-      PROPERTIES
-      CUDA_SEPARABLE_COMPILATION ON      # This adds -dc
-      )
+   if (AMReX_GPU_RDC)
+      set_target_properties( ${_target}
+         PROPERTIES
+         CUDA_SEPARABLE_COMPILATION ON      # This adds -dc
+         )
+   endif ()
```
ax3l: With the default set above, we can do this:
```suggestion
   set_target_properties( ${_target}
      PROPERTIES
      CUDA_SEPARABLE_COMPILATION ${AMReX_GPU_RDC}      # This adds -dc
   )
```

WeiqunZhang:
Actually we have to have rdc always on, otherwise kernel fusing won't work.

ax3l:
Ah, curious. We got WarpX compiled without this yesterday though for Perlmutter.
Probably because we don't use the kernel fusion feature?

WeiqunZhang:
It only happens in some places where there are several small kernels.

WeiqunZhang:
And it's likely to happen in amr regrid.

ax3l:
So still worth an option (but default ON)? :)

WeiqunZhang:
Yes, we can still add the gpu rdc option with the default being on and turn off fusing if it's disabled.  We are experimenting with alternative ways of fusing, and we can probably remove the current fusing approach and make gpu rdc disabled by default in the near future.

maximumcats:
OK, in that case can we simplify this PR to just add the new flag but not change any other behavior? I would be OK with it, given that change.

WeiqunZhang:
Yes, I will make the modification when I have a chance so that the behavior does not change.  I do however want to remove the old cuda specific random number stuff that nobody uses anymore.  In any case, don't want to make big changes when it's so close to the next release day.

atmyers (approved):

WeiqunZhang:
Forgot to update the commit message before merging.  :(

PR #2062, opened 2021-05-27
CMake: HIP Compiler - don't fail

ax3l:
## Summary

Potentially, users might use `CC` (Cray Prg Envs) and later `clang++` (underlying compiler #2031) ) as compilers. Only warn if not met.

ROCm/Cray environments are currently being developed and we want to keep things flexible for codes trying to use the compiler wrapper instead of `hpcc`.

cc @mwm126 @jmsexton03 

## Additional background

That said, the working mode to build for now is to build on Cray with AMD's `clang`/`hipcc` and pass MPI flags manually.
See e.g.: https://warpx.readthedocs.io/en/latest/install/hpc/spock.html

## Checklist

The proposed changes:

### DIFF
--- a/Tools/CMake/AMReXParallelBackends.cmake
+++ b/Tools/CMake/AMReXParallelBackends.cmake
@@ -171,12 +171,13 @@ endif ()
 #
 if (AMReX_HIP)
 
-   set(_valid_hip_compilers hipcc nvcc)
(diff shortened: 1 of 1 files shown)


WeiqunZhang (approved):

jmsexton03 (approved):
Tested with these instructions on Spock using PrgEnv-cray:
https://warpx.readthedocs.io/en/latest/install/hpc/spock.html

PR #2063, opened 2021-05-27
Particle plotfile/checkpoint without managed memory.

atmyers:
Currently, `ParticleContainer::WritePlotFile` and `ParticleContainer::Checkpoint` assume managed memory when compiled for GPUs. This PR extends them to work without managed memory. Operations such as particle filtering is performed on the device if the container uses managed or device memory, and on the host otherwise. 

The AMReX GPU and CPU tests all pass with the PR. 

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_ParticleIO.H
+++ b/Src/Particle/AMReX_ParticleIO.H
@@ -572,7 +572,7 @@ ParticleContainer<NStructReal, NStructInt, NArrayReal, NArrayInt, Allocator>
                   Vector<int>& which, Vector<int>& count, Vector<Long>& where,
                   const Vector<int>& write_real_comp,
                   const Vector<int>& write_int_comp,
-                  const Vector<std::map<std::pair<int, int>, Gpu::DeviceVector<int>>>& particle_io_flags) const
+                  const Vector<std::map<std::pair<int, int>, IntVector>>& particle_io_flags) const
 {
     BL_PROFILE("ParticleContainer::WriteParticles()");
 
@@ -588,11 +588,7 @@ ParticleContainer<NStructReal, NStructInt, NArrayReal, NArrayInt, Allocator>
 
         // Only write out valid particles.
         int cnt = 0;
-        for (int k = 0; k < kv.second.GetArrayOfStructs().numParticles(); ++k)
-        {
-            if (pflags[k]) cnt++;
-        }
-
+        particle_detail::countFlags(pflags, cnt);
         count[grid] += cnt;
     }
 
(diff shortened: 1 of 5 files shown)


jmsexton03 (approved):

Src/Particle/AMReX_ParticleIO.H
```diff
@@ -588,11 +588,7 @@ ParticleContainer<NStructReal, NStructInt, NArrayReal, NArrayInt, Allocator>
 
         // Only write out valid particles.
         int cnt = 0;
-        for (int k = 0; k < kv.second.GetArrayOfStructs().numParticles(); ++k)
-        {
-            if (pflags[k]) cnt++;
-        }
-
+        particle_detail::countFlags(pflags, cnt);
```
WeiqunZhang: Minor comment:  Would it be better to do `int cnt = particle_detail::countFlags(pflags)` instead?

atmyers: Yes, but I was using the return type for `enable_if` for these functions so I stuck with that. I could change it if you have a strong opinion, but there aren't user-facing functions.

Src/Particle/AMReX_WriteBinaryParticleData.H
```diff
@@ -24,6 +27,284 @@ std::size_t PSizeInFile (const Vector<int>& wrc, const Vector<int>& wic)
...
+    const auto write_int_comp_d_ptr = write_int_comp_d.data();
+    const auto write_real_comp_d_ptr = write_real_comp_d.data();
+
+    std::size_t poffset = 0;
+    for (int i = 0; i < tiles.size(); i++) {
+        const auto& ptile = pc.ParticlesAt(lev, grid, tiles[i]);
+        const auto& pflags = particle_io_flags[lev].at(std::make_pair(grid, tiles[i]));
+        int np_tile = ptile.GetArrayOfStructs().numParticles();
+        typename PC::IntVector offsets(np_tile);
+        Gpu::exclusive_scan(pflags.begin(), pflags.end(), offsets.begin());
```
WeiqunZhang: Potentially we could save one memcpy by using ExclusiveSum that will return the total sum.

Src/Particle/AMReX_ParticleIO.H
WeiqunZhang: std::enable_if_t has an optional parameter.

WeiqunZhang (approved):

Src/Particle/AMReX_ParticleIO.H
atmyers: oh, cool.

PR #2065, opened 2021-05-28
Set HIP_LIBRARIES before using

mwm126:
## Summary

HIP_LIBRARIES is empty before finding rocrand,rocprim,hiprand, which causes MFIX build to still fail with HIP.

## Checklist

The proposed changes:


ax3l (approved):
Ouch, thanks!

PR #2067, opened 2021-05-28
CI: Cover Fortran

ax3l:
## Summary

It looks like at some point we changed the `AMReX_Fortran` default from ON to OFF and forgot to adjust CI. This re-enables Fortran compile tests.

We still have at least one entry that makes sure we can compile in pure C/C++ on Linux.

## Checklist

The proposed changes:

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -17,6 +17,7 @@ jobs:
         mkdir build
         cd build
         cmake ..                                  \
+            -DAMReX_FORTRAN=ON                    \
             -DAMReX_PLOTFILE_TOOLS=ON             \
             -DCMAKE_VERBOSE_MAKEFILE=ON           \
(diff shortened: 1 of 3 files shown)


.github/workflows/linux.yml
```diff
@@ -168,8 +174,9 @@ jobs:
         cmake ..                        \
             -DCMAKE_BUILD_TYPE=Debug    \
             -DCMAKE_VERBOSE_MAKEFILE=ON \
-            -DAMReX_ENABLE_TESTS=ON    \
-            -DAMReX_PARTICLES=ON       \
+            -DAMReX_ENABLE_TESTS=ON     \
+            -DAMReX_FORTRAN=OFF         \
```
ax3l, line 178: This is the intentional w/o Fortran test. I am adding the flag just to be explicit here, even though it is the default.

atmyers (approved):

PR #2069, opened 2021-05-28
[WIP] Add support for MPI+HIP+OMP

jmsexton03:
## Summary
This adds flags and code changes to allow compilation with HIP kernels and CPU OpenMP threads
## Additional background
It's unclear whether defining `_OPENMP` is good practice for consistency with Cuda behavior. The hip compiler does not define this flag, although the Cuda compiler does.
The AMReX_Math.H change resolves this error:
```
../../subprojects/amrex//Src/Particle/AMReX_ParticleUtil.H:156:30: error: call to 'floor' is ambiguous
                         int(amrex::Math::floor((p.pos(1)-plo[1])*dxi[1])),
                             ^~~~~~~~~~~~~~~~~~
```
This setup will require changes to user code if compiling with HIP+OMP similar to this Nyx branch: https://github.com/AMReX-Astro/Nyx/tree/test_hip_omp

TODO:
- [ ] Add CMake flags
- [ ] Determine if there's a better namespace choice for Hip math operations 
## Checklist

The proposed changes:


WeiqunZhang:
`hipcc -fopenmp -E -dM - < /dev/null | grep OPENMP` shows `#define _OPENMP 201811`.  It's not clear why `_OPENMP` is undefined when `hipcc -fopenmp` is used to compile.

jmsexton03:
`_OPENMP` being undefined with ROCm is expected behavior:
https://rocmdocs.amd.com/en/latest/Programming_Guides/HIP-FAQ.html#openmp-is-undefined-when-compiling-with-fopenmp

I included a comment example of how to override this behavior with an explicit define, and an assumption about a minimal version

Tools/libamrex/mkconfig.py
```diff
@@ -78,10 +78,15 @@ def doit(defines, undefines, comp, allow_diff_comp):
 
     print("#endif") #  ifdef __cplusplus
 
-    print("#if defined(AMREX_USE_OMP) && !defined(_OPENMP)")
+    print("#if defined(AMREX_USE_OMP) && !defined(_OPENMP) && !defined(AMREX_USE_HIP)")
```
WeiqunZhang, line 81:
```suggestion
    print("#if defined(AMREX_USE_OMP) && !defined(_OPENMP) && !defined(__HIP_DEVICE_COMPILE__)")
```

PR #2073, opened 2021-06-01
ParallelFor for MultiFab/FabArray

WeiqunZhang:
## Summary

Add an experimental feature that can be used to launch a single kernel to
work on MultiFabs or FabArrays.  An example is shown below, where `mf` is a
MultiFab.
```
        auto ma = mf.arrays();
        amrex::ParallelFor(mf,
                           [=] AMREX_GPU_DEVICE (int box_no, int i, int j, int k)
        {
            ma[box_no](i,j,k) = 1.0;
        });
        Gpu::synchronize();
```

It's still unclear how we handle tiling and OpenMP and how it performs on CPU.  However, there is strong evidence that this improves performance for GPU runs when there are small boxes.  The purpose of this PR is to get the feature into the AMReX repo first and do further tuning later.  Since the new function is  not used at all, it should not affect any codes.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_FabArray.H
+++ b/Src/Base/AMReX_FabArray.H
@@ -25,6 +25,11 @@
 #include <AMReX_TypeTraits.H>
 #include <AMReX_LayoutData.H>
 #include <AMReX_BaseFabUtility.H>
+#ifdef AMREX_USE_GPU
+#include <AMReX_MFParallelForG.H>
+#else
+#include <AMReX_MFParallelForC.H>
+#endif
 
 #include <AMReX_Gpu.H>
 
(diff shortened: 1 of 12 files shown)


Src/Base/AMReX_FabArray.H
```diff
@@ -1262,6 +1361,14 @@ FabArray<FAB>::clear ()
         }
     }
     m_fabs_v.clear();
+#ifdef AMREX_USE_GPU
+    The_Pinned_Arena()->free(m_hp_arrays);
```
maximumcats, line 1365: This is relying on the newly changed behavior of CArena that free() is valid even if the pointer wasn't allocated, I take it?

WeiqunZhang: No, it does not.  The change in CArena is just to avoid constructing a mutex is the pointer is null.

maximumcats (approved):

PR #2076, opened 2021-06-02
Fix Plotfile Regression (introduced in #2063)

atmyers:
Fix a regression from #2063

The proposed changes:


ax3l (approved):
Thanks, this fixes the issue I saw with WarpX in https://github.com/ECP-WarpX/WarpX/pull/1998 :tada:

PR #2081, opened 2021-06-03
AMReX_PROBINT: Depends on AMReX_AMRLEVEL

ax3l:
## Summary

CMake: do not enable if `AMReX_AMRLEVEL` is off anyway.

## Checklist

The proposed changes:


WeiqunZhang (approved):

WeiqunZhang:
Note that this may cause issues for the linear solver.

WeiqunZhang:
I could submit a PR to AMReX to disable GPU kernel fusing (which depends on rdc) when gpu rdc is off.  We are working an alternative approach of kernel fusing not using rdc, but it may take a little time.

WeiqunZhang:
Sorry I will remove my comments.  Posted in the wrong place.  They were meant for @ax3l's WarpX PR.

PR #2083, opened 2021-06-03
Disable kernel fusing if gpu rdc is off.

WeiqunZhang:
## Additional background

Because currently the kernel fusing depends on rdc.

## Checklist

The proposed changes:


atmyers (approved):

WeiqunZhang:
Running regression tests now.

PR #2084, opened 2021-06-03
Fix some macro issues in CMake Fortran build

### LINKED ISSUE #2071: CMake: Fortran+MPI Warning
It looks like we are still adding `BL_USE_MPI` to the compiler defines in CMake, but I think we should just rely on the one set in `AMReX_Config.H`:
 
```
... Src/Boundary/AMReX_lo_bctypes_mod.F90 ...
AMReX_Config.H:17:0:

   17 | #define BL_USE_MPI
      | 
Warning: "BL_USE_MPI" redefined
<command-line>:5:0:

note: this is the location of the previous definition
```

cc @kngott

WeiqunZhang:
Include AMReX_Config.H in Fortran files that need preprocessing.  Do not
pass those definitions already defined in AMReX_Config.H to the Fortran
compiler in the cmake build.  This fixes the warnings about redefined macros.

Fix #2071

## Checklist

The proposed changes:

### DIFF
--- a/Src/AmrCore/AMReX_INTERP_2D.F90
+++ b/Src/AmrCore/AMReX_INTERP_2D.F90
@@ -331,7 +331,7 @@ contains
       integer icase
 
       if (MAX(lratiox,lratioy).gt.rMAX) then
-#ifdef AMREX_DEBUG
+#if 0
          print *,'rMAX in INTERP_2D::AMREX_PROTECT_INTERP must be >= ',MAX(lratiox,lratioy)
 #endif
          call bl_abort("rMAX in INTERP_2D")
@@ -551,7 +551,7 @@ contains
                enddo
                enddo
 
-#ifdef AMREX_DEBUG
+#if 0
                if (abs(crseTotnew - crseTot)/cvol .gt. 1.e-8) then
                   print *,' '
                   print *,'BLEW CONSERVATION with ICASE = ',icase
--- a/Src/AmrCore/AMReX_INTERP_3D.F90
+++ b/Src/AmrCore/AMReX_INTERP_3D.F90
@@ -437,7 +437,7 @@ contains
                enddo
                enddo
 
-#ifdef AMREX_DEBUG
+#if 0
                if (abs(sum_fine_new - sum_fine_old) .gt. 1.e-8) then
                   print *,' '
                   print *, &
--- a/Src/Boundary/AMReX_LO_UTIL.F90
+++ b/Src/Boundary/AMReX_LO_UTIL.F90
@@ -37,7 +37,7 @@ contains
             num = num*(xInt - x(i))
             den = den*(x(j) - x(i))
          end do
-#ifdef AMREX_DEBUG
+#if 0
          if (den .eq. zero) STOP 'polyInterpCoeff::invalid data'
 #endif
          c(j) = num/den
--- a/Tools/CMake/AMReXSetDefines.cmake
+++ b/Tools/CMake/AMReXSetDefines.cmake
@@ -70,22 +70,13 @@ add_amrex_define( AMREX_USE_ASSERTION NO_LEGACY IF AMReX_ASSERTIONS )
 # Bound checking
 add_amrex_define( AMREX_BOUND_CHECK NO_LEGACY IF AMReX_BOUND_CHECK )
 
-#
-# Fortran-specific defines: BL_LANG_FORT and AMREX_LANG_FORT
-#
 if (AMReX_FORTRAN)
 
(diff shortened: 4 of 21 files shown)


Tools/CMake/AMReXSetDefines.cmake
```diff
@@ -81,11 +81,35 @@ if (AMReX_FORTRAN)
       $<$<COMPILE_LANGUAGE:Fortran>:BL_LANG_FORT AMREX_LANG_FORT>
       )
 
-    if (AMReX_MPI)
+   if (AMReX_MPI)
       target_compile_definitions( amrex PRIVATE
```
ax3l: Would downstream codes depend on these compiler define flags or include `AMReX_Config.H` in their Fortran files?

If we want to make sure they also have the defines added, then we can make those `PUBLIC` definitions
```suggestion
      target_compile_definitions( amrex PUBLIC
```
(same below)

A user that does `target_link_libraries(mytarget [PRIVATE|PUBLIC] amrex::amrex)` will then inherit the `PUBLIC` flags from AMReX.

ax3l:
Added a question on Slack. I don't understand it yet fully, can't we just include `AMReX_Config.H` everywhere we need it in Fortran files, too?

WeiqunZhang:
Yes, including `AMReX_Config.H` could also work.  That might be simpler.

WeiqunZhang:
If we include `AMReX_Config.H` in Fortran, does CMake know the dependency?

WeiqunZhang:
In the GNU Make, we do have *.F and *.F90 files depend on AMReX_Config.H.  But I am not sure about cmake.  How does cmake generate dependency for Fortran files?  @ax3l

WeiqunZhang:
Testing shows cmake does generate the correct dependency if a header file is included in a Fortran file.

ax3l:
> In the GNU Make, we do have *.F and *.F90 files depend on AMReX_Config.H. But I am not sure about cmake. How does cmake generate dependency for Fortran files? @ax3l

Discussed on Slack: same way as for C sources in that case. Most compilers have a special flag (I think it's `-M` [most of the time](https://stackoverflow.com/a/5835035/2719194)) to output only transitive includes, otherwise they probably filter. I found the source for this once in the CMake repos, but cannot spot it quickly today.

We generate the `AMReX_Config.H` at configure time and the generated Makefile dependencies takes care about recompiles of object files that include it.

Src/AmrCore/AMReX_INTERP_2D.F90
```diff
@@ -331,7 +331,7 @@ subroutine AMREX_PROTECT_INTERP (fine, fine_l1,fine_l2,fine_h1,fine_h2, fblo, fb
       integer icase
 
       if (MAX(lratiox,lratioy).gt.rMAX) then
-#ifdef AMREX_DEBUG
+#if 0
```
ax3l, line 334: Do you want to avoid including `AMReX_Config.H` in this one or is this a leftover?

Src/AmrCore/AMReX_INTERP_3D.F90
```diff
@@ -437,7 +437,7 @@ subroutine AMREX_PROTECT_INTERP (fine, fine_l1,fine_l2,fine_l3,fine_h1,fine_h2,f
                enddo
                enddo
 
-#ifdef AMREX_DEBUG
+#if 0
```
ax3l, line 440: Same question as above :)

Src/Boundary/AMReX_LO_UTIL.F90
```diff
@@ -37,7 +37,7 @@ subroutine polyInterpCoeff(xInt, x, N, c)
             num = num*(xInt - x(i))
             den = den*(x(j) - x(i))
          end do
-#ifdef AMREX_DEBUG
+#if 0
```
ax3l, line 40: Same question as above :)

Src/AmrCore/AMReX_INTERP_2D.F90
WeiqunZhang: We are not including any header in this file.

Src/AmrCore/AMReX_INTERP_3D.F90
WeiqunZhang: Just leftover.

Src/Boundary/AMReX_LO_UTIL.F90
WeiqunZhang: Same answer. :)

ax3l (approved):
:tada:

PR #2085, opened 2021-06-04
 Use device warp size instead of defaulting to 32

jmsexton03:
## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_GpuDevice.cpp
+++ b/Src/Base/AMReX_GpuDevice.cpp
@@ -905,7 +905,7 @@ Device::grid_stride_threads_and_blocks (dim3& numBlocks, dim3& numThreads) noexc
 {
     int num_SMs = device_prop.multiProcessorCount;
 
-    int SM_mult_factor = 32;
+    int SM_mult_factor = device_prop.warp_size;
 
     if (num_SMs > 0) {
 
--- a/Src/Base/AMReX_GpuDevice.cpp
+++ b/Src/Base/AMReX_GpuDevice.cpp
@@ -905,7 +905,7 @@ Device::grid_stride_threads_and_blocks (dim3& numBlocks, dim3& numThreads) noexc
 {
     int num_SMs = device_prop.multiProcessorCount;
 
-    int SM_mult_factor = device_prop.warp_size;
+    int SM_mult_factor = device_prop.warpSize;
 
     if (num_SMs > 0) {
 


Src/Base/AMReX_GpuDevice.cpp
```diff
@@ -905,7 +905,7 @@ Device::grid_stride_threads_and_blocks (dim3& numBlocks, dim3& numThreads) noexc
 {
     int num_SMs = device_prop.multiProcessorCount;
 
-    int SM_mult_factor = 32;
+    int SM_mult_factor = device_prop.warpSize;
```
maximumcats, line 908: This number is not warp size, it is a measurement of how many blocks can fit per SM. It corresponds to [maxBlocksPerMultiProcessor](https://docs.nvidia.com/cuda/cuda-runtime-api/structcudaDeviceProp.html#structcudaDeviceProp_17f337476973ea65db85c277f4646f0b3).

jmsexton03: So this should be a different device property, instead of hardcoded to 32? Will that be 32 for CUDA_ARCH 60, 70, and 80?

maximumcats: Right, maxBlocksPerMultiProcessor has been 32 on all of the recent HPC GPU architectures which is why I didn't think much about the hardcoding at the time.

WeiqunZhang: Is that function still used anywhere?  Can we remove it?

maximumcats: It's still used in the no-argument constructor of `ExecutionConfig()`. I think that is the only place.

jmsexton03: It's called from ExecutionConfig() which isn't called anywhere in amrex:
https://github.com/AMReX-Codes/amrex/blob/1628b84f6eae8c280754683f4c486eddef0402a0/Src/Base/AMReX_GpuLaunch.H#L125-L129

drummerdoc: But apps might use it.  Even I did at one point.

WeiqunZhang: The problem is it may not be a good default for AMD and Intel.

jmsexton03: I was looking at the ExecutionConfig setup because I wanted to see if there was a good way to pass along amrex choices to sundials's ExecPolicy setup.

WeiqunZhang: We need to revisit the execution configuration setup for MI100.  In particular, we might simply get rid of grid strided loop.

maximumcats: I have no allegiance to the grid-stride loop even on NVIDIA GPUs. I added it way back in the early days because it was quite useful for the CUDA Fortran implementation but that is gone now.

jmsexton03: There's no direct analog in hip device properties for `maxBlocksPerMultiProcessor`, the closest is from amd_detail: `hipError_t hipModuleOccupancyMaxPotentialBlockSize (int ∗ gridSize, int ∗ blockSize, hipFunction_t f, size_t
dynSharedMemPerBlk, int blockSizeLimit)` which looks like it's achieved occupancy, rather than a device property.

PR #2091, opened 2021-06-07
Remove VecReduce and FabReduce

WeiqunZhang:
They were used for the old atomics based reduction approach.

## Additional background

Only PeleC uses VecReduce.  The use of  VecReduce there can be replaced with Reduce::Sum.

## Checklist

The proposed changes:


WeiqunZhang:
@jrood-nrel

atmyers (approved):

PR #2092, opened 2021-06-07
Add link to "Building with CMake" page in INSTALL file

wyphan:
## Summary
Add a link to "Building with CMake" page in the AMReX documentation web page into the INSTALL file.

## Additional background
I think it is useful for people who want to build AMReX to have the link handy, and it's more descriptive than simply listing CMake as a possible way to build AMReX.

## Checklist

The proposed changes:

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -17,5 +17,5 @@ There are three ways to use AMReX.
     `Tools/GNUMake/`.  See `Tools/GNUMake/README.md` for more details
     on the make system.
 
-(3) CMake.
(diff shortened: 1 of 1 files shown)


atmyers:
Looks like the `INSTALL` file doesn't render markdown - you could either change it to `INSTALL.md` or just use a raw url.

wyphan:
That's right, my bad. I think the `INSTALL` file is traditionally in plain text, so I'll keep it that way.

atmyers (approved):

atmyers:
Thanks!

PR #2104, opened 2021-06-13
Add more warnings to CI and fix warnings

WeiqunZhang:
Also fixed a bug in device version of `Random_int`.

## Checklist

The proposed changes:

### DIFF
--- a/.github/workflows/hip.yml
+++ b/.github/workflows/hip.yml
@@ -9,7 +9,17 @@ jobs:
   tests-hip:
     name: HIP ROCm GFortran@9.3 C++17 [tests]
     runs-on: ubuntu-20.04
-    env: {CXXFLAGS: "-fno-operator-names"}
+    # Have to have -Wno-deprecated-declarations due to deprecated atomicAddNoRet
+    # Have to have -Wno-gnu-zero-variadic-macro-arguments to avoid
+    #    amrex/Src/Base/AMReX_GpuLaunchGlobal.H:15:5: error: must specify at least one argument for '...' parameter of variadic macro [-Werror,-Wgnu-zero-variadic-macro-arguments]
+    #        __launch_bounds__(amrex_launch_bounds_max_threads)
+    #        ^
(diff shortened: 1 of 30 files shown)


atmyers (approved):

.github/workflows/hip.yml
```diff
@@ -28,7 +38,8 @@ jobs:
             -DAMReX_AMD_ARCH=gfx908                       \
             -DCMAKE_C_COMPILER=$(which clang)             \
             -DCMAKE_CXX_COMPILER=$(which hipcc)           \
-            -DCMAKE_Fortran_COMPILER=$(which gfortran)
+            -DCMAKE_Fortran_COMPILER=$(which gfortran)    \
+            -DCMAKE_CXX_STANDARD=17
```
ax3l, line 42: Since this will only test C++17 now with HIP: Do we bump our HIP requirement to C++17 builds? @WeiqunZhang

ax3l:
Cool, thanks for fixing those and adding CI :)

.github/workflows/hip.yml
WeiqunZhang: I made the change simply because the name has c++17 in it. `name: HIP ROCm GFortran@9.3 C++17 [tests]`

We don't require C++17.  So we could add another to test C++14.

WeiqunZhang: Or change the name.

WeiqunZhang: Note that we only support recent rocm releases and they all support C++17.

ax3l: Yep, let's just convert the test to C++14 (maybe no need for another for the sake of CI time?) until we officially drop C++14 for HIP? Just requiring C++17 already for HIP (but then we need to doc this) is fine for me, too. We do the same for DPC++ (there we have no choice).

WeiqunZhang: I looked at `hip.yml` again.  We have a gnu make based test there that uses c++14.  So both standards are covered and I think we should leave it as is.

PR #2110, opened 2021-06-14
TinyProfiler: Add spaces to print

WeiqunZhang:
Add more spaces so that the output looks nicer when there are multiple
layers of profilers.

## Checklist

The proposed changes:


kngott (approved):

kngott:
The tab = 2 spaces conversion is a real nice touch. 😄

PR #2016, opened 2021-05-13
Add SENSEI Adaptor for Particles Containers

c-wetterer-nelson:
## Summary
Here, we add two pieces of functionality:
1. The Amr class is currently hard coded to enable SENSEI. This PR adds a new compile flag AMREX_SENSEI_AUTO which allows you to turn off that behavior and then manually instrument an analysis code with SENSEI.

2. The major addition from this PR is a collection of data adaptors and in-situ bridges for adapting Particle Container data via SENSEI. There are two new data adaptors. the `AmrParticleDataAdaptor` and the `AmrParticleMeshDataAdaptor`. The former performs SENSEI data adapting on a ParticleContainer. The latter is a composite adaptor which contains both an AmrParticleDataAdaptor and an AmrDataAdaptor allowing a user to adapt both an Amr object's mesh and the associated Particle Container at the same time. Bridges associated with each of these new data adaptors are also provided.

## Checklist

The proposed changes:
- [ ] fix a bug or incorrect behavior in AMReX
- [X] add new capabilities to AMReX
- [ ] changes answers in the test suite to more than roundoff level
- [ ] are likely to significantly affect the results of downstream AMReX users
- [ ] are described in the proposed changes to the AMReX documentation, if appropriate

PR #2115, opened 2021-06-16
CI: Add job to test SENSEI build

kwryankrattiger:
## Summary

Add CI job for testing AMReX + SENSEI

## Additional background

ref. https://github.com/AMReX-Codes/amrex/pull/2016

## Checklist

The proposed changes:

### DIFF
--- a/.github/workflows/docker/sensei/install_sensei.sh
+++ b/.github/workflows/docker/sensei/install_sensei.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+DOWNLOAD_ROOT=/root/downloads
+EXTRACT_ROOT=/root/downloads/extracted
+mkdir -p ${DOWNLOAD_ROOT}
+mkdir -p ${EXTRACT_ROOT}
+
+# Download and install Sensei (VTK + MPI + Python)
+export CMAKE_GENERATOR=Ninja
+
+cd ${DOWNLOAD_ROOT}
+mkdir sensei
+cd sensei
+git clone https://gitlab.kitware.com/sensei/sensei.git &&
+cd sensei &&
+git checkout ${SENSEI_VERSION} &&
+mkdir build &&
+cd build &&
+cmake .. \
+  -D BUILD_TESTING=OFF \
+  -D ENABLE_PYTHON=ON \
+  -D ENABLE_MANDELBROT=OFF \
+  -D ENABLE_OSCILLATORS=OFF \
+  -D ENABLE_VTK_MPI=ON \
+  -D MPI4PY_INCLUDE_DIR=/usr/lib64/python3.9/site-packages/openmpi/mpi4py/include &&
+cmake --build . &&
+cmake --install . --prefix ${SENSEI_DIR}
+
+rm -rf ${DOWNLOAD_ROOT}
+rm -rf ${EXTRACT_ROOT}
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -87,6 +87,36 @@ jobs:
             -DAMReX_PARTICLES=ON
         make -j 2
 
+  test_sensei:
+    name: SENSEI Adaptor [test]
+    runs-on: ubuntu-18.04
+    env:
+      CXX: clang++
+      CC: clang
+      CXXFLAGS: "-Werror -Wshadow -Woverloaded-virtual -Wunreachable-code -fno-operator-names"
+      CMAKE_GENERATOR: Ninja
+      CMAKE_PREFIX_PATH: /root/install/sensei/develop/lib/cmake
+    container:
+      image: ryankrattiger/sensei:fedora33-vtk-mpi-20210616
+    steps:
+    - uses: actions/checkout@v2
+    - name: Setup
+      run: mkdir build
+    - name: Configure
+      run: |
+        cd build
+        cmake ..                  \
+            -DCMAKE_BUILD_TYPE=Debug \
+            -DAMReX_ENABLE_TESTS=ON  \
+            -DAMReX_FORTRAN=OFF      \
+            -DAMReX_PARTICLES=ON     \
+            -DAMReX_SENSEI=ON        \
+            -DCMAKE_CXX_STANDARD=14
+    - name: Build
+      run: |
+        cd build
+        cmake --build .
+
   # Build libamrex and all tests
   tests_cxx20:
     name: GNU@10.1 C++20 OMP [tests]
--- a/.github/workflows/docker/sensei/Dockerfile
+++ b/.github/workflows/docker/sensei/Dockerfile
@@ -1,0 +1,30 @@
+FROM fedora:33
+
+# Set install root
+ENV PACKAGE_ROOT=/root/install
+
+# Configure MPI environment
+ENV MPI_HOME=/usr/lib64/openmpi/
+
+# Configure Python environment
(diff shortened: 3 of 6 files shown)


kwryankrattiger:
@burlen Here is the SENSEI CI followup for the particle adapters.

atmyers:
Hi @kwryankrattiger, thank you for your PR! It all looks good to me.

As I understand it, this adds a CI test for compiling AMReX with SENSEI support, as well as the necessary configuration files needed to configure a docker container with SENSEI and it's dependencies installed. Is the idea that a correctness test for the SENSEI functionality will be added to #2016?

atmyers (approved):

kwryankrattiger:
> Is the idea that a correctness test for the SENSEI functionality will be added to #2016?

The plan was to add something in the tutorials CI that actually runs code, I believe it will be base off of this: https://github.com/AMReX-Codes/amrex-tutorials/pull/6

As far as correctness tests here, I don't think there is a plan to add more beyond the build.

atmyers:
I see, thank you. I think this can go in as-is then.

burlen:
@atmyers @kwryankrattiger to correct one thing: the plan is to add a correctness test to AMReX library itself NOT the tutorials.

@kwryankrattiger One request for modification here (I don't think it requires a new PR) is to please install the `sensei_config` tool in the VM. This is necessary to test the Makefile build.

.github/workflows/linux.yml
```diff
@@ -87,6 +87,36 @@ jobs:
             -DAMReX_PARTICLES=ON
         make -j 2
 
+  test_sensei:
+    name: SENSEI Adaptor [test]
+    runs-on: ubuntu-18.04
+    env:
+      CXX: clang++
+      CC: clang
+      CXXFLAGS: "-Werror -Wshadow -Woverloaded-virtual -Wunreachable-code -fno-operator-names"
```
burlen, line 96: @kwryankrattiger I don't think -Werror should be on, this would potentially cause the build to fail for reasons unrelated to the SENSEI code.

ax3l: Oh, we add this to avoid that various code sections introduce warnings with new PRs :)
We add this for all tests, so unregulated sections will not fail here w/o also failing a general tests (but in turn, warnings from sections here that are otherwise disabled are truly caught)

.github/workflows/docker/sensei/install_sensei.sh
```diff
@@ -0,0 +1,30 @@
...
+git checkout ${SENSEI_VERSION} &&
+mkdir build &&
+cd build &&
+cmake .. \
+  -D BUILD_TESTING=OFF \
+  -D ENABLE_PYTHON=ON \
+  -D ENABLE_MANDELBROT=OFF \
+  -D ENABLE_OSCILLATORS=OFF \
+  -D ENABLE_VTK_MPI=ON \
+  -D MPI4PY_INCLUDE_DIR=/usr/lib64/python3.9/site-packages/openmpi/mpi4py/include &&
```
ax3l, line 25: Potential generic:
`-D MPI4PY_INCLUDE_DIR="$(which python3) -c 'import mpi4py; print(mpi4py.get_include())'"`

.github/workflows/linux.yml
burlen: Ok, sounds good then!

.github/workflows/docker/sensei/install_sensei.sh
burlen: That is better, however it should not be necessary to specify -DMPI4PY_INCLUDE_DIR manually when configuring SENSEI.  
```
module load mpi/openmpi-x86_64
```
would likely resolve the issue.

kwryankrattiger: This container does not have environment modules support, so it is not possible to module load anything. Since this is a container that we maintain, I had control over where software is so a hard coded path was just easier.

burlen: Sure, but you could also `sudo dnf install environment-modules` I guess this is a fedora'ism with they way they support multiple MPI.

ax3l: Env modules are a bit tricky in non-login shells though. Do-able, but probably not needed for the one hint.

I was just wondering if this line could be more generic :)

burlen: Yeah, I get it, it's a good thought. the reason why cmake (which is intertnally doing what you're suggesting) can't find mpi4py is that it's not in the pythonpath by default on fedora. installing modules and loading the mpi module would solve that.

I've not seen any issues with environment modules in our other fedora ci images ;-)

.github/workflows/linux.yml
```diff
@@ -87,6 +87,36 @@ jobs:
...
+    - name: Configure
+      run: |
+        cd build
+        cmake ..                  \
+            -DCMAKE_BUILD_TYPE=Debug \
+            -DAMReX_ENABLE_TESTS=ON  \
+            -DAMReX_FORTRAN=OFF      \
+            -DAMReX_PARTICLES=ON     \
+            -DAMReX_SENSEI=ON        \
+            -DCMAKE_CXX_STANDARD=14
```
ax3l, line 114: X-ref: https://github.com/ECP-WarpX/WarpX/pull/2210

The current need to pass this flag looks like a bug to me.
We use `target_compile_features(... PUBLIC cxx_std_14)` for the `amrex` target. Maybe a similar requirement is missing of an auxiliary SENSEI target in AMReX.
https://github.com/AMReX-Codes/amrex/search?q=target_compile_features

@c-wetterer-nelson @kwryankrattiger @burlen Can one of you please take this on?

kwryankrattiger: From what I could tell at the time I added this AMReX was using C++11 by default. In CMake, you should be able to just set the option globally `set(CMAKE_CXX_STANDARD 14)` if you want to use C++14 for the project.

I am not 100% on the project needs. Is there a reason you are setting it per target?

ax3l: Thanks @kwryankrattiger - yes, AMReX was C++11 at the time but now transitioned to C++14 as well.
For AMReX, we avoid setting `CMAKE_CXX_STANDARD` because the sets the C++ standard of the whole build - but downstream projects transition faster to newer C++ standards and might prefer C++17/20 - or want to leave the exact choice open to a package manager. Thus, `target_compile_features` is the better choice, which only declares a minimal requirement.

I tried removing `-DCMAKE_CXX_STANDARD=14`, but found it did not compile. Thus, there is somewhere a `target_compile_features` call missing in AMReX when SENSEI is built, I think.

ax3l: I think I found the reason for this: SENSEI overwrites our optimization flags and injects a C++ standard flag in the CXX flags:
```
/root/install/sensei/develop/lib/cmake/SENSEIConfig.cmake:  set(CMAKE_CXX_FLAGS "-fPIC -std=c++11 -Wall -Wextra -O3 -march=native -mtune=native"
```

We need to fix this upstream in SENSEI, we cannot inject these flags unconditionally into downstream builds:
https://github.com/SENSEI-insitu/SENSEI/blob/73873211eadc290991dcd62a5edddc45dc09d2bb/CMake/build.cmake#L18

ax3l: Upstream bug filed in https://github.com/SENSEI-insitu/SENSEI/issues/15

PR #2121, opened 2021-06-23
Parser

WeiqunZhang:
Add amrex::Parser that can be used at runtime to evaluate math expressions
given by strings.  It works on both host and device.  This is moved from
WarpX with additional performance optimization of eliminating recursion.

## Checklist

The proposed changes:


drummerdoc:
Nice!

Src/Base/Parser/README
```diff
@@ -0,0 +1,44 @@
...
+   WarpX, because the files have already been generated.
+
+** wp_parser.tab.c & wp_parser.tab.h
+
+   They are generated by bison.
+
+** wp_parser_y.c & wp_parser_y.h
+
+   These contain C codes that are used to evaluate a mathematical
+   expression given in string format.
```
jmsexton03: WarpX -> AMReX ?

WeiqunZhang: Thanks.  Forgot to update the README.

WeiqunZhang:
Added support for local variables so that one can do things like `r2=x**2+y**2; r=sqrt(r2); r2*log(r)`.

atmyers (approved):

PR #2137, opened 2021-06-30
Make ParticleToMesh and MeshToParticle support iMultifab.

atmyers:
Also adds a few missing methods to iMultiFab, like SumBoundary and the version of Copy that takes an IntVect number of ghost cells.

The proposed changes:


jmsexton03 (approved):

Src/Base/AMReX_iMultiFab.cpp
```diff
@@ -51,6 +51,19 @@ iMultiFab::Copy (iMultiFab&       dst,
...
 
+void
+iMultiFab::Copy (iMultiFab& dst, const iMultiFab& src,
+                 int srccomp, int dstcomp, int numcomp, const IntVect& nghost)
+{
+// don't have to BL_ASSERT(dst.boxArray() == src.boxArray());
+    BL_ASSERT(dst.distributionMap == src.distributionMap);
+    BL_ASSERT(dst.nGrowVect().allGE(nghost));
+
+    BL_PROFILE("MultiFab::Copy()");
```
WeiqunZhang:
```suggestion
    BL_PROFILE("iMultiFab::Copy()");
```

WeiqunZhang:
Maybe we can move SumBoundary into FabArray class.

WeiqunZhang:
Now, `AMReX_FabArray.H` and `AMReX_FabArrayUtility.H` includes each other.  That will cause issues.   Moving `MFInfo` from `FabArray` to `FabArrayUtility` does not solve the issue.  It only avoids issue if the files are included in the following order
```
#include <AMReX_FabArray.H>
#include <AMReX_FabArrayUtility.H>
```
But it will fail if they are included as
```
#include <AMReX_FabArrayUtility.H>
#include <AMReX_FabArray.H>
```
I think the solution is to move `amrex::copy` from `AMReX_FabArrayUtility.H` to `AMReX_FabArray.H` with some forward declarations, or to a new file.

WeiqunZhang (approved):

PR #2139, opened 2021-06-30
add an option for hypre solver for Tests/Linearsolvers/NodalPoisson

b94209002:
## Summary
Add input-rt.hypre for NodalPoisson example
## Checklist

The proposed changes:
- [v] add new capabilities to AMReX


Tests/LinearSolvers/NodalPoisson/MyTest.H
```diff
@@ -34,7 +37,7 @@ private:
     int max_iter = 100;
     int max_fmg_iter = 0;
     amrex::Real reltol = 1.e-11;
-
+    amrex::Real bottom_reltol = 1.e-11;
```
WeiqunZhang: My suggestion is the default here should be negative.  In the MLMG setup, if `bottom_reltol` is negative, do not call `MLMG::setBottomTolerance`.  This way, it will keep its default behavior and allow the user to change it with the ParmParse parameter.  `bottom_reltol = 1.e-11` is probably not a good default also for other reasons.  At the bottom of our multigrid, bicgstab may have difficulty to reduce the residual by 1.e-11.  Another thing is hypre uses 2-norm for convergence test, whereas we use inf-norm.  So it is possible that the hypre solution that reduces the 2-norm by 1.e-11 is not considered a converged solution by MLMG.  Then a second iteration will need to be done and it might be hard for hypre to reduce the residual by another 1e-11.

WeiqunZhang (changes requested):

Tests/LinearSolvers/NodalPoisson/MyTest.H
b94209002: I removed control for bottom_reltol in this pull-request.

WeiqunZhang (approved):

PR #2144, opened 2021-07-02
Fix crashes with Conduit Blueprint Code

### LINKED ISSUE #2133: Crash In Conduit
Hi,

I'm attempting to add Ascent support to https://github.com/Exawind/amr-wind , and I've encountered a crash in the `MultiLevelToBlueprint` Conduit code, and I'm a little confused. 

Looking at `AMReX_Conduit_Blueprint.cpp` with the for loop that starts here: 

https://github.com/AMReX-Codes/amrex/blob/14dde9aaf741f84cf85adb67a7cdc5bef4c0764a/Src/Extern/Conduit/AMReX_Conduit_Blueprint.cpp#L363-L376

We have box_offsets starting out as zero-sized. In the first loop `i` is `0`, so box_offsets gets a `0` pushed back, so its size is 1. 

The next run through the loop hits this line

https://github.com/AMReX-Codes/amrex/blob/14dde9aaf741f84cf85adb67a7cdc5bef4c0764a/Src/Extern/Conduit/AMReX_Conduit_Blueprint.cpp#L374

Which causes the crash as box_offsets is size 1, and `i == 1`, which doesn't exist. As my code passes the asserts at the top of the function, I don't think I'm violating any contract, but it seems to me like this would always crash. 

I was able to fix it in my current patch by adding `box_offsets.resize(n_levels);` before the loop and changing the push back to `box_offsets[i] = 0;`, but I can't be sure I'm reading the code correctly. Any help would be appreciated. 

Thanks,
Nicholas

### LINKED ISSUE #2135: Nestsets Conduit Crash
Hello, 

While using the `MultiLevelToBlueprint` function, I have encountered another issue. The code has a function to build nest sets:

https://github.com/AMReX-Codes/amrex/blob/14dde9aaf741f84cf85adb67a7cdc5bef4c0764a/Src/Extern/Conduit/AMReX_Conduit_Blueprint.cpp#L80-L123

The crash happens at these lines:

https://github.com/AMReX-Codes/amrex/blob/14dde9aaf741f84cf85adb67a7cdc5bef4c0764a/Src/Extern/Conduit/AMReX_Conduit_Blueprint.cpp#L116-L120

Specifically, `level + 1` is out of bounds for the `ref_ratios` list that I am passing in. In this case, I have a `n_levels == 3` and `ref_ratios.size() == 2`. Now, based on the assert here 
https://github.com/AMReX-Codes/amrex/blob/14dde9aaf741f84cf85adb67a7cdc5bef4c0764a/Src/Extern/Conduit/AMReX_Conduit_Blueprint.cpp#L354
this appears to be a legal configuration. However, when we get to lines 116-120 with `level == 1`, this will be out of bounds for my `ref_ratios` list. 

My current fix, which successfully runs and has correct output is to change the if at line 80 from 
https://github.com/AMReX-Codes/amrex/blob/14dde9aaf741f84cf85adb67a7cdc5bef4c0764a/Src/Extern/Conduit/AMReX_Conduit_Blueprint.cpp#L80
to 
`if((level < n_levels - 1) && (level < ref_ratio.size() - 1))`.

But with my limited understanding of the codebase, this is just me grasping in the dark. What do you folks think?

I apologize in advance if I have done something very stupid.

nicholasbl:
## Summary

This PR contributes two fixes to the Conduit Blueprint code.

### Nestsets fix (Issue AMReX-Codes/amrex#2135)

The first fix here is within `Nestsets`, which resolves a one-past the end access on `ref_ratio`.

The problematic lines are 116 to 120, of the form `window["ratio/i"] = ref_ratio[level+1][0];` For some inputs, this is one-past the end of the `ref_ratio` array. It was advised in https://github.com/AMReX-Codes/amrex/issues/2135#issuecomment-872692909 to resolve this by removing the `+1`.

### MultiLevelToBlueprint fix (Issue AMReX-Codes/amrex#2133)

The second fix resolves a bug inside `MultiLevelToBlueprint`, where the `box_offsets` array is not built correctly.

The problematic code is here:

https://github.com/AMReX-Codes/amrex/blob/14dde9aaf741f84cf85adb67a7cdc5bef4c0764a/Src/Extern/Conduit/AMReX_Conduit_Blueprint.cpp#L363-L376

During the construction of the `box_offsets` array the first round of the loop at line 364, with `i == 0` will size the `box_offsets` array to 1. Subsequent rounds through the loop will attempt to write to non-existent slots of the vector here:

https://github.com/AMReX-Codes/amrex/blob/14dde9aaf741f84cf85adb67a7cdc5bef4c0764a/Src/Extern/Conduit/AMReX_Conduit_Blueprint.cpp#L374

This fix sizes `box_offsets` before the loop starts, and changes all accesses to the vector by index, not by appending. This appears to be the right fix as advised here: https://github.com/AMReX-Codes/amrex/issues/2133#issuecomment-871501258

## Additional background

Changes shouldn't impact any other code; the test suite was run and did not fail.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Extern/Conduit/AMReX_Conduit_Blueprint.cpp
+++ b/Src/Extern/Conduit/AMReX_Conduit_Blueprint.cpp
@@ -113,11 +113,11 @@ bool Nestsets(const int level,
         {
             window["dims/k"] = overlap.size()[2];
         }
-        window["ratio/i"] = ref_ratio[level+1][0];
-        window["ratio/j"] = ref_ratio[level+1][1];
+        window["ratio/i"] = ref_ratio[level][0];
+        window["ratio/j"] = ref_ratio[level][1];
         if(dims == 3)
         {
-            window["ratio/k"] = ref_ratio[level+1][2];
+            window["ratio/k"] = ref_ratio[level][2];
         }
       }
     }
@@ -361,13 +361,16 @@ MultiLevelToBlueprint (int n_levels,
 
     Vector<const BoxArray*> box_arrays;
     Vector<int> box_offsets;
+
+    box_offsets.resize(n_levels);
+
     for(int i = 0; i < n_levels; i++)
     {
       const BoxArray &boxs = mfs[i]->boxArray();
       box_arrays.push_back(&boxs);
       if(i == 0)
       {
-        box_offsets.push_back(0);
+        box_offsets[i] = 0;
       }
       else
       {


nicholasbl:
Hmm, I'm not seeing any details on the failed test.

WeiqunZhang:
It was just a glitch in github action.  I just restarted the failed Windows tests.

WeiqunZhang (approved):

PR #2155, opened 2021-07-09
YAFluxRegister with blocking factor of 1

WeiqunZhang:
Fix YAFluxRegister for the case that the blocking factor is 1.  In that
case, a coarse cell might have fine neighbors at both x-lo and x-hi faces.

## Checklist

The proposed changes:


WeiqunZhang:
This addresses the issue in #2140.

atmyers (approved):

PR #2183, opened 2021-07-21
Introduction of the experimental parfor into NodeLaplacian interpolation

PaulMullowney:
All algorithms fully implemented including  interpadd_aa, semi_interpadd_aa, interp_c, interp_rap, and interpadd_ha.

### DIFF
--- a/Src/LinearSolvers/MLMG/AMReX_MLNodeLaplacian.cpp
+++ b/Src/LinearSolvers/MLMG/AMReX_MLNodeLaplacian.cpp
@@ -323,55 +323,133 @@ MLNodeLaplacian::interpolation (int amrlev, int fmglev, MultiFab& fine, const Mu
         AMREX_ALWAYS_ASSERT(regular_coarsening);
     }
 
-#ifdef AMREX_USE_OMP
-#pragma omp parallel if (Gpu::notInLaunchRegion())
-#endif
-    for (MFIter mfi(fine, TilingIfNotGPU()); mfi.isValid(); ++mfi)
-    {
-        Box const& bx = mfi.tilebox();
-        Array4<Real> const& ffab = fine.array(mfi);
-        Array4<Real const> const& cfab = cmf->const_array(mfi);
-        Array4<int const> const& mfab = dmsk.const_array(mfi);
+    auto fine_ma = fine.arrays();
+    auto crse_ma = cmf->const_arrays();
+    auto msk_ma = dmsk.const_arrays();
+
+    if (Gpu::inLaunchRegion()) {
+        IntVect ng( 0);
         if (m_coarsening_strategy == CoarseningStrategy::RAP)
         {
-            Array4<Real const> const& stfab = stencil->const_array(mfi);
-            AMREX_HOST_DEVICE_PARALLEL_FOR_3D(bx, i, j, k,
+            auto sten_ma = stencil->const_arrays();
+            experimental::ParallelFor(fine, ng,[=] AMREX_GPU_DEVICE(int box_no, int i, int j, int k) noexcept
             {
-                mlndlap_interpadd_rap(i,j,k,ffab,cfab,stfab,mfab);
+                mlndlap_interpadd_rap(i, j, k, fine_ma[box_no], crse_ma[box_no], sten_ma[box_no], msk_ma[box_no]);
             });
         }
         else if (sigma[0] == nullptr)
         {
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
@PaulMullowney Could you merge the latest development branch into this?  One hip ci job is failing because of changes in cmake 3.21.

PaulMullowney:
Hip succeeds but now Intel is failing??

Src/LinearSolvers/MLMG/AMReX_MLNodeLaplacian.cpp
```diff
@@ -323,55 +323,133 @@ MLNodeLaplacian::interpolation (int amrlev, int fmglev, MultiFab& fine, const Mu
...
+        IntVect ng( 0);
         if (m_coarsening_strategy == CoarseningStrategy::RAP)
         {
-            Array4<Real const> const& stfab = stencil->const_array(mfi);
-            AMREX_HOST_DEVICE_PARALLEL_FOR_3D(bx, i, j, k,
+            auto sten_ma = stencil->const_arrays();
+            experimental::ParallelFor(fine, ng,[=] AMREX_GPU_DEVICE(int box_no, int i, int j, int k) noexcept
             {
-                mlndlap_interpadd_rap(i,j,k,ffab,cfab,stfab,mfab);
+                mlndlap_interpadd_rap(box_no, i, j, k, fine_ma, crse_ma, sten_ma, msk_ma);
```
WeiqunZhang: Could we make some changes here and other places?
```
mlndlap_interpadd_rap(i, j, k, fine_ma[box_no], crse_ma[box_no], sten_ma[box_no], msk_ma[box_no]);
```

Src/LinearSolvers/MLMG/AMReX_MLNodeLaplacian.cpp
```diff
@@ -323,55 +323,133 @@ MLNodeLaplacian::interpolation (int amrlev, int fmglev, MultiFab& fine, const Mu
...
+#endif
+            for (MFIter mfi(fine, TilingIfNotGPU()); mfi.isValid(); ++mfi)
+            {
+                int box_no = mfi.index();
+                Box const& amrex_i_box = mfi.tilebox();
+                LoopConcurrentOnCpu(amrex_i_box,[=] (int i, int j, int k) noexcept
                 {
-                    mlndlap_interpadd_aa(i,j,k,ffab,cfab,sfab,mfab);
+                    mlndlap_interpadd_rap(box_no, i, j, k, fine_ma, crse_ma, sten_ma, msk_ma);
                 });
```
WeiqunZhang: I suggest we use the traditional approach for cpu code.  There is also a bug here. `mfi.index()` returns the global box index, but we need to use the local box index to access the `Array4` inside `MultiArray4`. It would be simpler to just use the traditional approach.

WeiqunZhang (changes requested):

Src/LinearSolvers/MLMG/AMReX_MLNodeLaplacian.cpp
PaulMullowney: When I use mfi.LocalTileIndex(), the example I'm running segfaults.

WeiqunZhang: Tile index is different.  I believe `LocalIndex` is the one that could be used here.  It can be confusing.  That's why I think the traditional `MultiFab[MFIter]` approach is clearer when we are not using the MF ParallelFor.

Src/LinearSolvers/MLMG/AMReX_MLNodeLaplacian.cpp
WeiqunZhang: If we do it this way, we no longer need to modify `AMReX_MLNodeLap_?D_K.H` add new versions of `mlndlap_interpadd_*`.

PaulMullowney: I see what you want. Hopefully the latest commit addresses your concerns.

Src/LinearSolvers/MLMG/AMReX_MLNodeLaplacian.cpp
PaulMullowney: LocalIndex works. Thanks

WeiqunZhang (approved):

PR #2185, opened 2021-07-21
Make CreateGhostParticles use GPU functions

jmsexton03:
## Checklist

The proposed changes:


Src/Particle/AMReX_ParticleTransformation.H
```diff
@@ -394,12 +394,55 @@ auto filterParticles (DstTile& dst, const SrcTile& src, Pred&& p,
     amrex::ParallelForRNG(np,
     [=] AMREX_GPU_DEVICE (int i, amrex::RandomEngine const& engine) noexcept
     {
-        p_mask[i] = particle_detail::call_f(p, src_data, i, engine);
+        p_mask[i] = particle_detail::call_f(p, src_data, src_start+i, engine);
```
atmyers, line 394: Are you sure this is correct? It seems to me that the next function call, to the version of `filterParticles` that takes a mask, assumes the mask will be filled everywhere. Maybe we can init the mask to `0`, then only call the function over the specified range.

Src/Particle/AMReX_ParticleTransformation.H
```diff
@@ -394,12 +394,55 @@ auto filterParticles (DstTile& dst, const SrcTile& src, Pred&& p,
...
+ * \param dst the destination tile
+ * \param src the source tile
+ * \param mask pointer to the mask - 1 means copy, 0 means don't copy
+ * \param f defines the transformation that will be applied to the particles on copy
+ *
+ */
+template <typename DstTile, typename SrcTile, typename Index, typename F,
+          std::enable_if_t<std::is_integral<Index>::value, int> foo = 0>
+Index filterAndTransformParticles (DstTile& dst, const SrcTile& src, Index* mask, F&& f,
+                                   Index src_start, Index dst_start) noexcept
```
atmyers, line 417: Could the already existing `filterAndTransformParticles` function, that doesn't take `src_start` and `dst_start`, now be implemented in terms of this function (and likewise for the other overload you added below)?

Src/Particle/AMReX_ParticleTransformation.H
```diff
@@ -394,12 +394,55 @@ auto filterParticles (DstTile& dst, const SrcTile& src, Pred&& p,
     amrex::ParallelForRNG(np,
     [=] AMREX_GPU_DEVICE (int i, amrex::RandomEngine const& engine) noexcept
     {
-        p_mask[i] = particle_detail::call_f(p, src_data, i, engine);
+        p_mask[i] = particle_detail::call_f(p, src_data, src_start+i, engine);
     });
-
+    Gpu::streamSynchronize();
```
atmyers: In versions of these that are implemented in terms of another overload, I'm not sure these syncs are needed, because the inner-most function that actually implements the copy will do so.

Src/Particle/AMReX_ParticleContainerI.H
```diff
@@ -796,26 +851,24 @@ ParticleContainer<NStructReal, NStructInt, NArrayReal, NArrayInt, Allocator>
...
-                SuperParticleType p = src.getSuperParticle(pindex);
-                p.id() = GhostParticleID;
-                ghosts.push_back(p);
-            }
-        }
+        auto np = src_tile.numParticles();
+        ghosts.resize(ghost_offset+np);
+        ghost_offset += filterAndTransformParticles(ghosts, src_tile, AssignGridFilter(assign_grid,gid,level,nGrow), TransformerGhost(),0,ghost_offset);
+        ghosts.resize(ghost_offset);
+        Gpu::synchronize();
```
atmyers: `resize` implies a `sync`, so you don't need to do so explicitly here.

Src/Particle/AMReX_ParticleContainerI.H
```diff
@@ -772,6 +772,61 @@ ParticleContainer<NStructReal, NStructInt, NArrayReal, NArrayInt, Allocator>
...
+        dst.m_aos[dst_i] = src.m_aos[src_i];
+        for (int j = 0; j < DstData::NAR; ++j)
+            dst.m_rdata[j][dst_i] = src.m_rdata[j][src_i];
+        for (int j = 0; j < dst.m_num_runtime_real; ++j)
+            dst.m_runtime_rdata[j][dst_i] = src.m_runtime_rdata[j][src_i];
+        for (int j = 0; j < DstData::NAI; ++j)
+            dst.m_idata[j][dst_i] = src.m_idata[j][src_i];
+        for (int j = 0; j < dst.m_num_runtime_int; ++j)
+            dst.m_runtime_idata[j][dst_i] = src.m_runtime_idata[j][src_i];
+        ParticleIDWrapper((dst.m_aos[dst_i]).m_idcpu) = GhostParticleID;
```
atmyers: I think you can do `dst.m_aos[dst_i].id() = GhostParticleID;`

Tests/Particles/GhostsAndVirtuals/main.cpp
```diff
@@ -107,6 +415,188 @@ void test_ghosts_and_virtuals (TestParams& parms)
...
+        const int ngrow = 1;
+        const int src_lev = 0;
+        const int dst_lev = 1;
+        MyParticleContainer ghostPC(geom, dmap, ba, rr);
+        MyParticleContainer::ParticleTileType ghosts;
+        myPC.CreateGhostParticles(src_lev, ngrow, ghosts);
+        ghostPC.AddParticlesAtLevel(ghosts, dst_lev, ngrow);
+        Real sum_test = amrex::ReduceSum(ghostPC, [=] AMREX_GPU_HOST_DEVICE (const PType& p) -> Real { return (amrex::Math::abs(p.rdata(0))+amrex::Math::abs(p.rdata(1))+amrex::Math::abs(p.rdata(2))+amrex::Math::abs(p.rdata(3))); });
+        amrex::ParallelAllReduce::Sum(sum_test,ParallelContext::CommunicatorSub());
+        amrex::Print().SetPrecision(18)<<"Found sum of ghosts: "<<sum_test<<std::endl;
```
atmyers, line 604: Thank you! This test is much better now.

Src/Particle/AMReX_ParticleContainerI.H
jmsexton03: Pushed update commit, using id directly resulted in strange cpu or id values when debugging at one point, but I can't reproduce that behavior anymore

Src/Particle/AMReX_ParticleTransformation.H
jmsexton03: I believe it's correct, the `filterParticles` call loops over `mask[src_start+i]`  for `i=0,..,n-1`.

https://github.com/AMReX-Codes/amrex/pull/2185/files/039eb01c9bdc40bbedc0dfe5b6c4cb7d88f7ec53#diff-8e42446c3fbcc60f09f881dbf68c630fc1019e5481eb1c9e253710b948e8e3a3L319-L345

Src/Particle/AMReX_ParticleTransformation.H
jmsexton03: Pushed update commit that removes synchronizes in outer function

Src/Particle/AMReX_ParticleTransformation.H
jmsexton03: Pushed update commit that uses the overloaded function

Src/Particle/AMReX_ParticleContainerI.H
jmsexton03: Pushed update commit, removed extra resize and extra sync

Src/Particle/AMReX_ParticleTransformation.H
atmyers: This `ParallelForRNG` call goes over the full `np` in the src tile, so if `src_start` is not 0, doesn't that go out of bounds? Likewise, the `exclusive_sum` in the nested called to `filterParticles` goes over the full range of the mask, from 0 to np, so I think those values all need to be filled with 0 or 1.

You could make this so that both `mask` and `offsets` are size `n`, which in principle would be more efficient in memory and in terms of the number of times the predicate is applied, but I don't think you need to do that for this PR.

jmsexton03: I misread it, I thought both loops were over `n`, not `np`. Is there a compelling reason not to treat the `ParallelForRNG(np` as a typo that should actually be `ParallelForRNG(n`? If both `mask` and `offsets` are size `n`, then it's consistent with the comments above the function, and seems more logical to have a function that would have a `src_start` that's non-zero.

atmyers (approved):

PR #2184, opened 2021-07-21
Add ROCMClang as an LLVM compiler to cmake

WeiqunZhang:
Since CMake >= 3.21, `hipcc` is identified as ROCMClang and unwrapped to `clang++`.
AMD's `clang++` is still identified as vanilla `Clang`.

## References

- https://gitlab.kitware.com/cmake/cmake/-/blob/v3.21.0/Modules/CMakeDetermineCompilerId.cmake#L153-159
- https://github.com/AMReX-Codes/amrex/pull/2184

## Checklist

The proposed changes:
- [ ] fix a bug or incorrect behavior in AMReX
- [ ] add new capabilities to AMReX
- [ ] changes answers in the test suite to more than roundoff level
- [ ] are likely to significantly affect the results of downstream AMReX users
- [ ] are described in the proposed changes to the AMReX documentation, if appropriate

PR #2187, opened 2021-07-22
ROCMClang: work-around missing C++ Std Flag

ax3l:
## Summary

Manually set std flag.

I usually try to avoid this fixed setting since the usual CMake logic allows users to explicitly set a specific CXX standard on the command line (e.g. `-DCMAKE_CXX_STANDARD=20`) or raise it in their subprojects further via `target_compile_features(warpx PUBLIC cxx_std_20)`.

This hacks in `-std=c++17` to work-around the `hipcc` bug reported in:
  https://gitlab.kitware.com/cmake/cmake/-/issues/22460

Also treats the CMake 3.21+ ROCMClang as `hipcc` with respect to the Fortran fix that will land in ROCm 4.4

Follow-up to #2184

## Checklist

The proposed changes:

### DIFF
--- a/Tools/CMake/AMReXParallelBackends.cmake
+++ b/Tools/CMake/AMReXParallelBackends.cmake
@@ -178,6 +178,20 @@ if (AMReX_HIP)
 
    set(_valid_hip_compilers clang++ hipcc nvcc CC)
    get_filename_component(_this_comp ${CMAKE_CXX_COMPILER} NAME)
+   
+   # work-arounds for breaking Fortran compiles
+   #   https://github.com/ROCm-Developer-Tools/HIP/pull/2280
+   set(IS_PRE_ROCM4_HIPCC FALSE)
+   # CMake<3.21 and hipcc as CXX
+   if(${_this_comp} STREQUAL hipcc)
+      set(IS_PRE_ROCM4_HIPCC TRUE)
+   endif()
+   # CMake >=3.21 and hipcc as CXX but then set by cmake to clang++
+   if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "ROCMClang")
+      if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS 4.4.0)
+          set(IS_PRE_ROCM4_HIPCC TRUE)
+      endif()
+   endif()
 
    if (NOT (_this_comp IN_LIST _valid_hip_compilers) )
       message(WARNING "\nCMAKE_CXX_COMPILER (${_this_comp}) is likely "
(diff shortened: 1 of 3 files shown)


Tools/CMake/AMReXParallelBackends.cmake
```diff
@@ -229,16 +243,16 @@ if (AMReX_HIP)
    endif()
    # AMD's or mainline clang++ with support for "-x hip"
    # Cray's CC wrapper that points to AMD's clang++ underneath
-   if(NOT ${_this_comp} STREQUAL hipcc)
+   if(NOT IS_PRE_ROCM4_HIPCC)
        target_link_libraries(amrex PUBLIC hip::device)
```
ax3l, line 247: Hm, that unwrapping to `clang++` now adds no `-x hip` flags for CXX targets and this cannot be linked until the Fortran parts are fixed.

ax3l: Maybe we can set the CXX compiler for <4.4 back to `hpcc` if `ROCMClang` is detected...

Tools/CMake/AMReXParallelBackends.cmake
```diff
@@ -178,6 +178,20 @@ if (AMReX_HIP)
...
+   
+   # work-arounds for breaking Fortran compiles
+   #   https://github.com/ROCm-Developer-Tools/HIP/pull/2280
+   set(IS_PRE_ROCM4_HIPCC FALSE)
+   # CMake<3.21 and hipcc as CXX
+   if(${_this_comp} STREQUAL hipcc)
+      set(IS_PRE_ROCM4_HIPCC TRUE)
+   endif()
+   # CMake >=3.21 and hipcc as CXX but then set by cmake to clang++
+   if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "ROCMClang")
```
ax3l, line 190: Becomes vanilla Clang again in 3.21.3+:
https://gitlab.kitware.com/cmake/cmake/-/merge_requests/6533

ax3l:
Problem solved in upstream LLVM/Clang after my reports:
- https://gist.github.com/ax3l/9489132#gistcomment-3758576
- https://reviews.llvm.org/D103221
- https://github.com/llvm/llvm-project/commit/f7e87dd6ff0c1c4a41a6b654d2460557ae97ab8d
- https://github.com/ROCm-Developer-Tools/HIP/issues/2278

Tools/CMake/AMReXParallelBackends.cmake
ax3l: Fixed because this change was reverted after my report.

PR #2189, opened 2021-07-25
MFCellConsLinInterp

WeiqunZhang:
Add MFCellConsLinInterp that is similar to CellConsLinInterp.  The
difference is that MFCellConsLinInterp works on MultiFab, whereas
CellConsLinInterp works on FArrayBox.  The new interpolater launches fewer
kernels, therefore has better performance on GPU when there are many boxes.

## Checklist

The proposed changes:


WeiqunZhang:
This PR has passed all the nightly regression tests at CCSE.

jmsexton03 (approved):

PR #2195, opened 2021-07-27
Make CreateVirtualParticles use GPU functions for aggregation_type=None

jmsexton03:
## Checklist

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_ParticleContainerI.H
+++ b/Src/Particle/AMReX_ParticleContainerI.H
@@ -646,6 +646,37 @@ ParticleContainer<NStructReal, NStructInt, NArrayReal, NArrayInt, Allocator>::Re
   }
 }
 
+struct TransformerVirt
+{
+    int aggregation_type_int;
+    /**
+       \brief This copies the particle but multiplies all the idata by m_factor
+       *
+     **/
+    TransformerVirt ()
+    { }
+
+    template <typename DstData, typename SrcData>
+    AMREX_GPU_HOST_DEVICE
+    void operator() (DstData& dst, const SrcData& src,
+                     int src_i, int dst_i) const noexcept
+    {
+        dst.m_aos[dst_i] = src.m_aos[src_i];
+        for (int j = 0; j < DstData::NAR; ++j)
+            dst.m_rdata[j][dst_i] = src.m_rdata[j][src_i];
+        for (int j = 0; j < dst.m_num_runtime_real; ++j)
+            dst.m_runtime_rdata[j][dst_i] = src.m_runtime_rdata[j][src_i];
+        for (int j = 0; j < DstData::NAI; ++j)
+            dst.m_idata[j][dst_i] = src.m_idata[j][src_i];
+        for (int j = 0; j < dst.m_num_runtime_int; ++j)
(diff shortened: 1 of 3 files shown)


Src/Particle/AMReX_ParticleContainerI.H
```diff
@@ -646,6 +646,36 @@ ParticleContainer<NStructReal, NStructInt, NArrayReal, NArrayInt, Allocator>::Re
   }
 }
 
+struct TransformerVirt
+{
+    /**
+       \brief This copies the particle but replaces the id with VirtualParticleID
+       *
+     **/
+    TransformerVirt ()
```
atmyers: I think you can remove this and let the compiler generate this for you.

Src/Particle/AMReX_ParticleContainerI.H
```diff
@@ -646,6 +646,36 @@ ParticleContainer<NStructReal, NStructInt, NArrayReal, NArrayInt, Allocator>::Re
...
+    {}
+
+    template <typename DstData, typename SrcData>
+    AMREX_GPU_HOST_DEVICE
+    void operator() (DstData& dst, const SrcData& src,
+                     int src_i, int dst_i) const noexcept
+    {
+        dst.m_aos[dst_i] = src.m_aos[src_i];
```
atmyers: I think you can save some code here by using the `copyParticle` function in `AMReX_ParticleTransformation.H`.

Src/Particle/AMReX_ParticleContainerI.H
```diff
@@ -646,6 +646,36 @@ ParticleContainer<NStructReal, NStructInt, NArrayReal, NArrayInt, Allocator>::Re
...
+{
+    /**
+       \brief This copies the particle but replaces the id with VirtualParticleID
+       *
+     **/
+    TransformerVirt ()
+    {}
+
+    template <typename DstData, typename SrcData>
+    AMREX_GPU_HOST_DEVICE
```
atmyers: `AMREX_FORCE_INLINE`?

Src/Particle/AMReX_ParticleContainerI.H
jmsexton03: Pushed a commit which removes this

Src/Particle/AMReX_ParticleContainerI.H
jmsexton03: Pushed a commit which does this

Src/Particle/AMReX_ParticleContainerI.H
jmsexton03: Pushed a commit which adds the inline, and updates other inlines in Src/Particles for consistency

Src/Particle/AMReX_ParticleContainerI.H
atmyers: Nice - not needed for this PR, but you could also update the ghost one to do the same thing.

jmsexton03: I will include that update as well (got confused on doing both removing the constructor and updating the copy)

atmyers (approved):

PR #2213, opened 2021-08-03
Parser: noinline

WeiqunZhang:
Make [i]parser_exe_eval noinline.

For some WarpX kernels, having parser calls inlined makes them much slower even when they
are not used at runtime.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_Extension.H
+++ b/Src/Base/AMReX_Extension.H
@@ -4,6 +4,11 @@
 
 #if !defined(BL_LANG_FORT)
 
+#if defined(__INTEL_COMPILER) && defined(__EDG__)
+// Disable "routine is both inline and noinline" warning from classical Intel compiler
+#pragma warning disable 2196
+#endif
+
 // restrict
 
 #ifdef __cplusplus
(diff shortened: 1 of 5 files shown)


ax3l:
@atmyers do you want to run the KPP with that one again to check if this helps? :)

atmyers:
I already have - this PR, by itself, not does fix the performance regression observed on the KPP. It may still be a good idea to merge it for other reasons, but I think we also need to address this on the WarpX side by having GatherAndPush, etc, be templated on whether we use the parser or not.

WeiqunZhang:
I will test this on summit trying to understand why it does not help v100, though it helps a100.

WeiqunZhang:
I tested this PR on a single GPU of summit.  Contrary to what Andrew observed, the performance was good.  Maybe it's because we are using different inputs? @atmyers Could you give me the exact setup?

atmyers:
It's at `/gpfs/alpine/scratch/atmyers2/aph114/ParserTest` - you could probably make it smaller, but that's the exact test I was using.

WeiqunZhang:
I don't have access to the directory.

WeiqunZhang:
I moved `AMREX_NO_INLINE` to a different function.  For the WarpX test that does not use parser, the performance is not affected by this change.  But for https://github.com/WeiqunZhang/amrex-microbenchmarks/tree/main/parser, the new noinline way is significantly faster than the previous noinline way.

WeiqunZhang:
The issue is addressed by #2229.

PR #2218, opened 2021-08-04
Fixes to ensure covered cells don't get included

drangara:
## Checklist

The proposed changes:


drangara:
Only patches covered cells from getting included and doesn't address the root problem which appears to be that face centroids of neighbors are not updated when covering small cells

PR #2219, opened 2021-08-04
Make CreateVirtualParticles use GPU functions for aggregation_type=Cell

jmsexton03:
* Uses ParticleToMesh with MultiFab, iMultiFab
* Use ExclusiveSum to map between MultiFab and particle tile
* Uses ParticleLocator to check no-aggregation buffer
* Uses DeviceVector type for better GPU memory use
* Uses Gpu::Atomic::Add
* Fixes minor gridding bugs in GhostAndVirtuals test

## Checklist

The proposed changes:


Src/Particle/AMReX_ParticleContainerI.H
```diff
@@ -701,87 +722,142 @@ ParticleContainer<NStructReal, NStructInt, NArrayReal, NArrayInt, Allocator>
...
+        const auto& geom = Geom(level);
+        const auto plo = geom.ProbLoArray();
+        const auto dxi = geom.InvCellSizeArray();
+        const auto domain = geom.Domain();
+
         BoxList bl_buffer;
         bl_buffer.complementIn(Geom(level).Domain(), ParticleBoxArray(level));
         BoxArray buffer(std::move(bl_buffer));
         buffer.grow(aggregation_buffer);
+        Gpu::synchronize();
```
atmyers: Are you certain that you need all these synchronize() calls? I think all three can be removed, but you should test to make sure.

Src/Particle/AMReX_ParticleContainerI.H
```diff
@@ -701,87 +722,142 @@ ParticleContainer<NStructReal, NStructInt, NArrayReal, NArrayInt, Allocator>
...
+
+               for (int i = 0; i < AMREX_SPACEDIM; ++i)
+               {
+                   amrex::Gpu::Atomic::AddNoRet(&partData(iv,i), static_cast<Real>((p.rdata(0)*p.pos(i))));
+               }
+               for (int i = 1; i < NStructReal; ++i)
+               {
+                   amrex::Gpu::Atomic::AddNoRet(&partData(iv,AMREX_SPACEDIM+i), static_cast<Real>((p.rdata(0)*p.rdata(i))));
+               }
+               for (int i = 0; i < 1; ++i)
```
atmyers, line 776: As we discussed at the meeting today, throughout this PR, could you use brackets, even if the for loop only contains 1 line?

Src/Particle/AMReX_ParticleContainerI.H
```diff
@@ -701,87 +722,142 @@ ParticleContainer<NStructReal, NStructInt, NArrayReal, NArrayInt, Allocator>
...
+        {
+
+            auto iv = getParticleCell(p, plo_loc, dxi_loc, domain);
+            if(assign_buffer_grid(iv)==-1)
             {
-                virts.push_back(agg_particle.second);
+            //if this cell has no particle id info, do a straight copy
+            if(partData(iv,0)==0)
+            {
+          //                amrex::Gpu::Atomic::AddNoRet(&partData(iv,0), VirtualParticleID);
```
atmyers: Could you remove the commented-out line?

Src/Particle/AMReX_ParticleContainerI.H
```diff
@@ -701,87 +722,142 @@ ParticleContainer<NStructReal, NStructInt, NArrayReal, NArrayInt, Allocator>
...
+    for (MFIter mfi(mf); mfi.isValid(); ++mfi)
+    {
+        const auto bx = mfi.tilebox();
+        const auto partData = mf.array(mfi);
+        const auto imf_arr = imf.array(mfi);
+
+        Gpu::DeviceVector<int> offsets(bx.numPts());
+        auto offsets_ptr = offsets.dataPtr();
+        int next_offset = Scan::ExclusiveSum((int) bx.numPts(),(imf_arr.ptr(bx.smallEnd(),0)),(offsets.dataPtr()),Scan::retSum);
+        Gpu::synchronize();
```
atmyers: I think this sync can also be removed, as ExclusiveSum should do this already.

Src/Particle/AMReX_ParticleContainerI.H
```diff
@@ -701,87 +722,142 @@ ParticleContainer<NStructReal, NStructInt, NArrayReal, NArrayInt, Allocator>
...
+    }
+
+    // last_offset should equal virts.numParticles()
+    auto virts_offset = last_offset;
+    for(ParConstIterType pti(*this, level); pti.isValid(); ++pti)
+    {
+        const auto& src_tile = ParticlesAt(level, pti);
+
+        auto np = src_tile.numParticles();
+        Gpu::synchronize();
```
atmyers: I think only one of these is needed.

Src/Particle/AMReX_ParticleContainerI.H
```diff
@@ -701,87 +722,142 @@ ParticleContainer<NStructReal, NStructInt, NArrayReal, NArrayInt, Allocator>
             virts_offset += np;
         }
     }
-
     if (aggregation_type == "Cell")
     {
+        //Components would be based on
+        MultiFab mf;
```
atmyers: You could do this all on the same line, instead of default constructing + calling define.

jmsexton03: Removed use of define

Src/Particle/AMReX_ParticleContainerI.H
jmsexton03: These synchronize calls were for sanity, now that I'm using DeviceVector instead of Vector I can't reproduce the failure I thought these might guard against.

Src/Particle/AMReX_ParticleContainerI.H
jmsexton03: Added braces to make if and for loops not 1-line

Src/Particle/AMReX_ParticleContainerI.H
jmsexton03: Removed

Src/Particle/AMReX_ParticleContainerI.H
jmsexton03: Removed

Src/Particle/AMReX_ParticleContainerI.H
jmsexton03: Removed sync preceding resize

Src/Particle/AMReX_ParticleContainerI.H
```diff
@@ -701,87 +723,162 @@ ParticleContainer<NStructReal, NStructInt, NArrayReal, NArrayInt, Allocator>
...
+                 }
+                 //Set rdata(n>0) with the normalized weighted field for NArrayReal
+                 for (int n = 0; n < NArrayReal; ++n)
+                 {
+                     p.rdata(NStructReal+n) = static_cast<ParticleReal>(partData(i,j,k,AMREX_SPACEDIM+NStructReal+n));
+                 }
+                 //Set idata with the "first" particles idata field for NStructInt
+                 for (int n = 0; n < NStructInt; ++n)
+                 {
+                     p.idata(n) = (long unsigned int) imf_arr(i,j,k,1+n);
```
atmyers: Why do you need to do this cast? Is it to avoid a compiler warning when `NStructInt` is `0`?

jmsexton03: I believe this cast was from a compiler complaining about casting from signed to unsigned

atmyers (approved):

PR #2229, opened 2021-08-09
Parser: Noinline transcendental functions

WeiqunZhang:
It's too expensive to have them inlined.

## Checklist

The proposed changes:


WeiqunZhang:
This appears to be a better fix than #2213.  In my testing, it has the same performance as #2213 and the old WarpX Parser for the WarpX KPP test.  But for https://github.com/WeiqunZhang/amrex-microbenchmarks/tree/main/parser, it's only 10% slower than the fully inlined version.

ax3l (approved):
LGTM :rocket: 
Thank you for digging into this! :+1:

atmyers:
This seems to fix the issue on my 320-node runs as well.

atmyers (approved):

PR #2232, opened 2021-08-09
CI: Build Embedded Boundaries

ax3l:
## Summary

- [x] Add coverage for EB in CI
- [x] fix DPC++ single precision build
- [x] ~~fix `LinearSolvers` array bounds w/ GCC 7.5~~ -> false positive

## Checklist

The proposed changes:

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -8,6 +8,7 @@ on: [push, pull_request]
 
 jobs:
   # Build and install libamrex as AMReX CMake project
+  # Note: this is an intentional "minimal" build that does not enable (many) options
   library:
     name: GNU@7.5 C++17 Release [lib]
     runs-on: ubuntu-18.04
@@ -32,6 +33,8 @@ jobs:
 
         export PATH=/tmp/my-amrex/bin:$PATH
         which fcompare
+
+        ctest --output-on-failure
 
   library_clang:
     name: Clang@6.0 C++14 SP NOMPI Debug [lib]
@@ -67,6 +70,8 @@ jobs:
         export PATH=/tmp/my-amrex/bin:$PATH
         which fcompare
 
+        ctest --output-on-failure -E CNS
+
   # Build libamrex and all tests
   tests_build:
     name: GNU@7.5 C++14 Debug Fortran [tests]
@@ -89,6 +94,8 @@ jobs:
             -DAMReX_PARTICLES=ON
         make -j 2
 
+        ctest --output-on-failure -E CNS
+
   test_sensei:
     name: SENSEI Adaptor [test]
     runs-on: ubuntu-18.04
@@ -104,21 +111,19 @@ jobs:
     - uses: actions/checkout@v2
     - name: Setup
       run: mkdir build
-    - name: Configure
+    - name: Configure & Build
       run: |
         cd build
         cmake ..                  \
             -DCMAKE_BUILD_TYPE=Debug \
-            -DAMReX_EB=ON            \
             -DAMReX_ENABLE_TESTS=ON  \
             -DAMReX_FORTRAN=OFF      \
             -DAMReX_PARTICLES=ON     \
             -DAMReX_SENSEI=ON        \
             -DCMAKE_CXX_STANDARD=14
-    - name: Build
-      run: |
-        cd build
-        cmake --build .
+        cmake --build build -j 2
+
+        ctest --output-on-failure
 
   # Build libamrex and all tests
   tests_cxx20:
@@ -146,6 +151,8 @@ jobs:
             -DCMAKE_CXX_COMPILER=$(which g++-10)            \
             -DCMAKE_Fortran_COMPILER=$(which gfortran-10)
         make -j 2
+
+        ctest --output-on-failure -E CNS
 
   tests_clang:
     name: Clang@6.0 C++14 SP Particles DP Mesh Debug [tests]
@@ -175,6 +182,8 @@ jobs:
             -DCMAKE_Fortran_COMPILER=$(which gfortran)
         make -j 2
 
+        ctest --output-on-failure -E CNS
+
   # Build libamrex and all tests w/o MPI
   tests-nonmpi:
     name: GNU@7.5 C++14 NOMPI [tests]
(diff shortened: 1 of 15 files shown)


WeiqunZhang:
Do we want some tests that do not use EB?  It's possible to have errors that appear only when EB is not enabled.

ax3l:
> Do we want some tests that do not use EB? It's possible to have errors that appear only when EB is not enabled.

Good idea - in WarpX I add a single CI test that builds a "minimal" build - disabling as much features as possible.
That way, we can cover that we can truly disable features at compile-time.

ax3l:
@WeiqunZhang can you take a look at the `-Werror=array-bounds` error in `Src/LinearSolvers/MLMG/AMReX_MLEBABecLap_F.cpp`?
```
Src/LinearSolvers/MLMG/AMReX_MLEBABecLap_F.cpp
In file included from /home/runner/work/amrex/amrex/Src/Base/AMReX_ArrayLim.H:9:0,
                 from /home/runner/work/amrex/amrex/Src/Base/AMReX_Box.H:7,
                 from /home/runner/work/amrex/amrex/Src/Base/AMReX_FabFactory.H:6,
                 from /home/runner/work/amrex/amrex/Src/EB/AMReX_EBFabFactory.H:5,
                 from /home/runner/work/amrex/amrex/Src/LinearSolvers/MLMG/AMReX_MLEBABecLap.H:5,
                 from /home/runner/work/amrex/amrex/Src/LinearSolvers/MLMG/AMReX_MLEBABecLap.cpp:2:
/home/runner/work/amrex/amrex/Src/Base/AMReX_Array.H: In member function ‘virtual void amrex::MLEBABecLap::applyBC(int, int, amrex::MultiFab&, amrex::MLLinOp::BCMode, amrex::MLLinOp::StateMode, const amrex::MLMGBndry*, bool) const’:
/home/runner/work/amrex/amrex/Src/Base/AMReX_Array.H:114:23: error: array subscript is above array bounds [-Werror=array-bounds]

**Comment by WeiqunZhang** [comment]
I suspect this is false positive.  GCC's `-Werror=array-bounds` has a lot of false positives.  I could not reproduce it on my desktop with gcc 9.3.0.  May have to use gcc 7.5.

**Comment by ax3l** [comment]
Thanks for the check, see it now as well in the other entries.
I would say for the sake of compiler coverage and since we have also newer GCCs in CI, I would add `-Wno-array-bounds` to the one GCC 7.5 build that has the false positive.

Alternatively, I can also just disable EB in that one build again, to keep the very first build small and minimal. I pushed that for now.

What do you think?

**Comment by WeiqunZhang** [comment]
Sounds good.  When I get time, I will install gcc 7.5 and figure out what it complains about.

**Comment by ax3l** [comment]
Looking where this is coming from:
```
36/38 Test #36: LinearSolvers_NodalPoisson ..........***Failed    0.72 sec
MPI initialized with 1 MPI processes
AMReX (21.08.0) initialized
terminate called after throwing an instance of 'std::bad_cast'
  what():  std::bad_cast
SIGABRT
See Backtrace.0 file for details
Abort(6) on node 0 (rank 0 in comm 0): application called MPI_Abort(MPI_COMM_WORLD, 6) - process 0
```

**Comment by WeiqunZhang** [comment]
Possible reason is that the non-EB test does work when EB is enabled.

**Comment by WeiqunZhang** [comment]
I can push a fix to this PR.

**Comment by ax3l** [comment]
I think the bad cast is here:
https://github.com/AMReX-Codes/amrex/blob/0ca401daa57fd80e7ae7a12eb5fd28b9d10ce11b/Src/EB/AMReX_algoim.cpp#L25

(With the last push it might now be one call earlier.)

**Comment by WeiqunZhang** [comment]
That's right.  I added a test before calling that function and that should also work.  The CI now fails for a different reason.

**Comment by ax3l** [comment]
Is it possible the EB unit tests are pretty large in terms of runtime? I see the CI on ICC crunching on it for 20+ min

**Comment by WeiqunZhang** [comment]
Yes, some EB tests are too large for CI.

**Comment by ax3l** [comment]
Those are the biggest runtime eaters:
```
30/38 Test #30: CNS_Combustor .......................   Passed  145.31 sec
      Start 31: CNS_Pulse
32/38 Test #32: CNS_ShockRef ........................   Passed  4259.57 sec
      Start 33: CNS_Sod
33/38 Test #33: CNS_Sod .............................   Passed  112.03 sec
```

@WeiqunZhang can we just reduce their resolution and steps for the checked in inputs?
If they need to be large they might be better candidates for tutorials than the tests in the AMReX repo?

**Comment by WeiqunZhang** [comment]
Yes, we can try to reduce the size or exclude some inputs files.

I also want to look at the EB coverage.  We already have some CI tests for EB in development.  I think we are converting too many non-EB tests to EB now.

**Comment by ax3l** [comment]
For this PR to finish, shall I exclude all `CNS*` tests for now and we reduce in a follow-up?

**Comment by ax3l** [comment]
Excluded the `CNS_*` test in the last commit from CI.
Let's re-enable them in #2237 once we reduce the inputs.

**Comment by ax3l** [comment]
@WeiqunZhang ready for merge :)

**Comment by WeiqunZhang** [comment]
In AMReX, the EB codes are either put inside `#ifdef AMREX_USE_EB` or excluded in the build so that non-EB builds could work.  The issue in this PR is now the majority of the tests has enabled EB.  So, the CI tests may not be able to catch non-EB build errors.  Currently in the development branch, we are already testing EB builds in 4 GNU make CI tests with a combination of 2d, 3d, mpi, omp, hip, and cuda.  We want to add more coverage especially with cmake, but I think we should still leave most of the CI tests as non-EB builds.

File: .github/workflows/linux.yml [hunk 1]
```diff
@@ -143,6 +153,8 @@ jobs:
             -DCMAKE_Fortran_COMPILER=$(which gfortran-10)
         make -j 2
 
+        ctest --output-on-failure -E Particles_GhostsAndVirtuals
```
ax3l, line 156: Those should become addressed with #2247 (issue: #2243).
We can comment those in again once the linked PR is merged

ax3l:
@WeiqunZhang green now :)

WeiqunZhang (approved):

PR #2234, opened 2021-08-10
Fix derive data in AmrLevel::writePlotFile

WeiqunZhang:
Previously, it only worked with single-component DeriveList.

## Checklist

The proposed changes:


atmyers (approved):

philip-blakely:
Thanks; I've tested the patch in our code and it all works as expected.

PR #2238, opened 2021-08-11
Enable mixed (face) GpuBndryFuncFab

esclapez:
## Summary
Allow PhysBCFunct to fill face-centered MultiFabs ghost cells. A set of boundary conditions for faces is implemented in FilFC.

## Additional background
Currently available fillfc for non user-defined Dirichlet conditions include: `foextrap`, `hoextrap`, `reflect_even` and `reflect_odd`. Note that {`hoextrap`,`reflect_even`,`reflect_odd`} acts differently on `idim`-indexed ghost faces in the `idim` direction and faces indexed in the transverse directions.

## Checklist

The proposed changes:


zingale:
tested this with Castro with the face-centered magnetic field stuff and it works

Src/Base/AMReX_PhysBCFunct.H
```diff
@@ -619,6 +626,295 @@ GpuBndryFuncFab<F>::nddoit (Box const& bx, FArrayBox& dest,
...
+    Array4<Real> const& fab = dest.array();
+    const auto geomdata = geom.data();
+
+#ifdef AMREX_USE_GPU
+    AsyncArray<BCRec> bcr_aa(bcr.data()+bcomp, numcomp);
+    BCRec* bcr_p = bcr_aa.data();
+
+    const auto f_user = m_user_f;
+
+    // filcc on the faces first
```
WeiqunZhang:
```suggestion
    // filfc on the faces first
```

Src/Base/AMReX_PhysBCFunct.H
```diff
@@ -202,7 +209,7 @@ GpuBndryFuncFab<F>::operator() (Box const& bx, FArrayBox& dest,
     } else if (bx.ixType().nodeCentered()) {
         nddoit(bx,dest,dcomp,numcomp,geom,time,bcr,bcomp,orig_comp);
     } else {
-        amrex::Abort("GpuBndryFuncFab: mixed type not supported");
+        fcdoit(bx,dest,dcomp,numcomp,geom,time,bcr,bcomp,orig_comp);
```
WeiqunZhang: I think we can remove a lot of duplicated codes by changing `filcc_cell`'s interface to make it the same as `filfc_face`.  That's we add `IndexType` to the argument of `filcc_cell`.  Then the code here becomes
```
    if (bx.ixType().cellCentered()) {
        ccfcdoit(bx,dest,dcomp,numcomp,geom,time,bcr,bcomp,orig_comp,filcc_cell);
    } else if (bx.ixType().nodeCentered()) {
        nddoit(bx,dest,dcomp,numcomp,geom,time,bcr,bcomp,orig_comp);
    } else {
        ccfcdoit(bx,dest,dcomp,numcomp,geom,time,bcr,bcomp,orig_comp,filfc_face);
    }
```
We can remove `ccdoit `, rename `fcdoit` to `ccfcdoit` and upddate it.
```
template <class F>
template <class FF>
void
GpuBndryFuncFab<F>::ccfcdoit (Box const& bx, FArrayBox& dest,
                            const int dcomp, const int numcomp,
                            Geometry const& geom, const Real time,
                            const Vector<BCRec>& bcr, const int bcomp,
                            const int orig_comp, FF&& ff)
```

Src/Base/AMReX_FilFC_1D_C.H
```diff
@@ -0,0 +1,107 @@
...
+#include <AMReX_FArrayBox.H>
+#include <AMReX_BCRec.H>
+#include <AMReX_Geometry.H>
+
+namespace amrex {
+
+AMREX_GPU_DEVICE
+inline
+void
+filfc_face (const IntVect& iv, const IndexType &idxType,
```
WeiqunZhang: Because `IndexType` is so small, we can pass it by value.

WeiqunZhang (commented):
A question.  Are we supposed to set value on the domain faces?  It looks like the lower domain faces are not set, but the higher domain faces are set.  Note the domain box in Geometry is always cell-centered.  So its big end's index plus one is right on the domain face.

WeiqunZhang:
Maybe we can change the interface of `filcc_cell` and `filfc_cell` to
```
void (const IntVect& iv, Array4<Real> const& q,
         const int dcomp, const int numcomp,
         Box const& domain_box, // domain box with the index type of the data
         const BCRec* bcr, const int bcomp) noexcept
```
I think then we don't even need to test on index type inside `filfc_face`.  We can also remove `time` and `orig_comp` from the parameters because they are not used.

esclapez:
I can address most of those changes but your original question: do we fill the domain faces ? I was working in the context of MAC velocities after the projection so did not intent to change the domain faces (despite the error of not converting the domain_box). But maybe setting the domain face is actually the primary intent ?

WeiqunZhang:
My feeling is for `reflect_odd`, it does seem to make sense to set the value on the domain face to zero. For others we should not modify them.  What do others think?

zingale:
That makes sense to me -- for `reflect_odd`, I think the user would expect that behavior.

Src/Base/AMReX_PhysBCFunct.H
```diff
@@ -228,22 +232,147 @@ GpuBndryFuncFab<F>::ccdoit (Box const& bx, FArrayBox& dest,
...
+                              const int dcomp, const int numcomp,
+                              Geometry const& geom, const Real time,
+                              const Vector<BCRec>& bcr, const int bcomp,
+                              const int orig_comp, FF&& fillfunc)
+{
+    const IntVect& len = bx.length();
+
+    IndexType idxType = bx.ixType();
+    Box const& domain = (idxType.cellCentered()) ? amrex::convert(geom.Domain(),idxType)
+                                                 : geom.Domain();
```
WeiqunZhang:
```suggestion
    Box const& domain = amrex::convert(geom.Domain(), idxType);
```

Src/Base/AMReX_FilFC_1D_C.H
```diff
@@ -0,0 +1,120 @@
...
+    {
+        const int i = iv[0];
+
+        // Domain box is indexed according to the face currently treating
+        const auto& domain_lo = domain_box.loVect();
+        const auto& domain_hi = domain_box.hiVect();
+        const int ilo = domain_lo[0];
+        const int ihi = domain_hi[0];
+        const int is = amrex::max(q.begin.x,ilo);
+        const int ie = amrex::min(q.end.x-1,ihi);
```
WeiqunZhang:
```suggestion
```
They are not used.

WeiqunZhang (approved):

PR #2244, opened 2021-08-13
Bug fix of rarely met issue from Tools/CMake/AMReXUtils.cmake file

### LINKED ISSUE #2240: CMake Error at while installing amr-wind: string sub-command STRIP requires two arguments.
I posted such an issue at the amr-wind project, but then were told that the issue comes from  the original AMReX. That seems to be true.

I was following instruction from here:
https://exawind.github.io/amr-wind/user/build.html

Running command
$ cmake -DAMR_WIND_ENABLE_TESTS:BOOL=ON ../

has led to the following error:
CMake Error at submods/amrex/Tools/CMake/AMReXUtils.cmake:20 (string):
string sub-command STRIP requires two arguments.

Investigation of the mentioned file has shown the difference between strings containing STRIP command:

image

string( STRIP ${_tmp} _tmp ) at line 20
VS
string(STRIP "${_tmp}" _tmp ) at line 32

Solution:
I followed a tip described here:
https://stackoverflow.com/questions/26585513/cmake-error-with-string-sub-command-strip-requires-two-arguments

adding the same quotas to line 20 seemed to solve the problem.

Original amr-wind issue: https://github.com/Exawind/amr-wind/issues/472

maxshatsky:
## Summary

Everything is already described here:
https://github.com/AMReX-Codes/amrex/issues/2240

## Checklist

The proposed changes:

### DIFF
--- a/Tools/CMake/AMReXUtils.cmake
+++ b/Tools/CMake/AMReXUtils.cmake
@@ -17,7 +17,7 @@ function (get_amrex_version)
       execute_process ( COMMAND git describe --abbrev=12 --dirty --always --tags
          WORKING_DIRECTORY ${CMAKE_CURRENT_LIST_DIR}
          OUTPUT_VARIABLE _tmp )
-      string( STRIP ${_tmp} _tmp )
+      string( STRIP "${_tmp}" _tmp )
       # filter invalid descriptions in shallow git clones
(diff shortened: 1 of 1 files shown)


WeiqunZhang (approved):

ax3l:
Thanks a lot, Maxim! :tada:

PR #2245, opened 2021-08-13
Use new ParallelFor in FabArray

WeiqunZhang:
* Add FabArrayBase::isFusingCandidate() that is fine tuned for MI100, V100
  and A100.

* Convert some FabArray functions to using the new ParallelFor

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_FabArrayBase.cpp
+++ b/Src/Base/AMReX_FabArrayBase.cpp
@@ -2484,6 +2484,30 @@ FabArrayBase::is_cell_centered () const noexcept
     return boxArray().ixType().cellCentered();
 }
 
+bool
+FabArrayBase::isFusingCandidate () const noexcept
+{
+#ifdef AMREX_USE_GPU
+    // This is fine tuned on MI100.
+    // For V100 and A100, it is not very sensitive to the choice here.
+    const int n = local_size();
+    if (n <= 1) {
+        return false;
+    } else if (n > 8) {
+        return true;
+    } else {
+        for (int i = 0; i < n; ++i) {
+            if (boxarray[indexArray[i]].numPts() <= Long(65*65*65)) {
+                return true;
+            }
+        }
+        return false;
+    }
+#else
+    return false;
+#endif
+}
+
 #ifdef AMREX_USE_GPU
 
 FabArrayBase::ParForInfo::ParForInfo (const FabArrayBase& fa, const IntVect& nghost, int nthreads)
--- a/Src/Base/AMReX_FabArray.H
+++ b/Src/Base/AMReX_FabArray.H
@@ -229,20 +229,34 @@ template <class FAB,
 void
 Copy (FabArray<FAB>& dst, FabArray<FAB> const& src, int srccomp, int dstcomp, int numcomp, const IntVect& nghost)
 {
+#ifdef AMREX_USE_GPU
+    if (Gpu::inLaunchRegion() && dst.isFusingCandidate()) {
+        auto const& srcarr = src.const_arrays();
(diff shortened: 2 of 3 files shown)


Src/Base/AMReX_FabArrayBase.cpp
```diff
@@ -2484,6 +2484,30 @@ FabArrayBase::is_cell_centered () const noexcept
...
+                return true;
+            }
+        }
+        return false;
+    }
+#else
+    return false;
+#endif
+}
+
```
kngott, line 2510: Would it make sense to right now setup to allow different versions for different build types? i.e:
`#ifdef AMREX_USE_CUDA
...
#elif defined(AMREX_USE_HIP)`
or even do it for individual hardware? I assume MI100 and MI200 might have very different optimizations here (although we might not care about MI100 at all once we get MI200s).

kngott:
Just curious: why these functions?

Also, if we do it this way, we might want to clearly mark somewhere very "official" that macros are simple launchers and *DO NOT* contain fusing or other fancier features. If for some reason we have a request to make that change in a few months/a year, I can see forgetting the way this is implemented and messing this up, accidentally.

Otherwise... extremely cool!

Src/Base/AMReX_FabArrayBase.cpp
WeiqunZhang: Yes, in the future, we might use macros that depend on hardware and software (e.g., rocm versions).  For V100 and A100, the choice here does not appear to matter much.  We could even simply return true for all cases.

WeiqunZhang:
As a first step, I want to replace the use of `*_FUSIBLE` macros.  Then we can remove the old fusing approach entirely.  The new approach is far more superior.

WeiqunZhang:
I am doing this in multiple PRs so that we can review the changes more carefully.  If anything is broken, it would be easier to track it down.

kngott:
Ah, I see. Makes sense.

Src/Base/AMReX_FabArrayBase.cpp
kngott: Really? Even for 2 128^3's? That's impressive. Sounds like something we should get the AMD folks looking at. Is there a "fused" benchmark?

WeiqunZhang: For the test here, https://github.com/WeiqunZhang/amrex-devtests/blob/main/mfparfor_test/main.cpp using https://github.com/WeiqunZhang/amrex/tree/fuse_mfutil, on A100 I get,
```
$ ./main*.ex box_sizes="128" nboxes="2"
...
    Kernel run time is 1.8165400000e-04.
```
If I make the following changes to amrex
```
--- a/Src/Base/AMReX_FabArrayBase.cpp
+++ b/Src/Base/AMReX_FabArrayBase.cpp
@@ -2487,6 +2487,7 @@ FabArrayBase::is_cell_centered () const noexcept
 bool
 FabArrayBase::isFusingCandidate () const noexcept
 {
+    return true;
 #ifdef AMREX_USE_GPU
     // This is fine tuned on MI100.
     // For V100 and A100, it is not very sensitive to the choice here.
```
I get
```
$ ./main*.ex box_sizes="128" nboxes="2"
...
    Kernel run time is 1.7696400000e-04.
```

kngott: Huh. Interesting. Not what I would expect. I wonder if the A100s have enough registers that both of the boxes run at the same time, reducing the additional fusing overheads, maybe?

kngott (approved):

PR #2247, opened 2021-08-14
Make sure all ba are defined in test, and only call Init once per PC

jmsexton03:
## Summary

* Fixes a potential indexing issue in rr
* Makes ba[lev] definition use for loop instead of if to allow for nlevs>2
* Don't call ParticleInit functions on non-empty ParticleContainers

## Additional background
This is meant to fix #2243 
The assertions now depend on PR #2248 

## Checklist

The proposed changes:
- [X] fix a bug or incorrect behavior in AMReX
- [ ] add new capabilities to AMReX
- [ ] changes answers in the test suite to more than roundoff level
- [ ] are likely to significantly affect the results of downstream AMReX users
- [ ] are described in the proposed changes to the AMReX documentation, if appropriate

PR #2252, opened 2021-08-15
Fix Mixed Precision Particle CI Tests

WeiqunZhang:
A number of CI tests had issues with mixing single precision particle data
with double precision amrex::Real.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_ParticleInit.H
+++ b/Src/Particle/AMReX_ParticleInit.H
@@ -1004,7 +1004,7 @@ InitRandom (Long                    icount,
                         r = amrex::Random();
                         x = geom.ProbLo(i) + (r * len[i]);
                     }
-                    while (x < xlo[i] || x > xhi[i]);
+                    while (static_cast<ParticleReal>(x) < static_cast<ParticleReal>(xlo[i]) || static_cast<ParticleReal>(x) >= static_cast<ParticleReal>(xhi[i]));
 
                     pos[j*AMREX_SPACEDIM + i] = static_cast<ParticleReal>(x);
                 }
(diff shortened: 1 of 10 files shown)


jmsexton03 (approved):

ax3l:
@WeiqunZhang Thanks a lot for fixing these! :tada: 

Can you please rebase against `development`? That should fix the `Particles_GhostsAndVirtuals` test that fails, I guess (#2247) :)

Src/Particle/AMReX_ParticleInit.H
```diff
@@ -1004,7 +1004,7 @@ InitRandom (Long                    icount,
                         r = amrex::Random();
                         x = geom.ProbLo(i) + (r * len[i]);
                     }
-                    while (x < xlo[i] || x > xhi[i]);
+                    while (static_cast<ParticleReal>(x) < static_cast<ParticleReal>(xlo[i]) || static_cast<ParticleReal>(x) >= static_cast<ParticleReal>(xhi[i]));
```
ax3l, line 1007: That's confusing, `x`, `xlo` and `xhi` are all of `Real` type for this comparison, no? Why do we need the casts here, too?

Src/Particle/AMReX_ParticleInit.H
```diff
@@ -1140,7 +1140,7 @@ InitRandom (Long                    icount,
                     r = amrex::Random();
                     x = geom.ProbLo(i) + (r * len[i]);
                 }
-                while (x < xlo[i] || x > xhi[i]);
+                while (static_cast<ParticleReal>(x) < static_cast<ParticleReal>(xlo[i]) || static_cast<ParticleReal>(x) >= static_cast<ParticleReal>(xhi[i]));
```
ax3l, line 1143: Just noted as above.

WeiqunZhang:
I already rebased.  The failure of GhostAndVirtural is due to mixing precision.  I just disabled the test in the Clang SP/DP CI.  @jmsexton03 will fix it later.

Src/Particle/AMReX_ParticleInit.H
WeiqunZhang: Because later we assign the double precision x to a single precision variable.

ax3l: I see.
I think since casts are expensive in this inner loop, I would in that case store `x`, `xhi` and `xlo` as `ParticleReal` and cast on assignment.

WeiqunZhang: I am hoping the compiler could figure it out.

ax3l: I fear they usually do not :-S

ax3l:
Thanks again for the fix :tada: :)

PR #2258, opened 2021-08-18
Remove amr adaptor/bridge when AMRLEVEL is off

### LINKED ISSUE #2263: SENSEI: Should not imply AMRLevel?
WarpX builds without the `AMRLEVEL` feature set of AMReX (ref: https://github.com/ECP-WarpX/WarpX/pull/1764).

When we compile with SENSEI, this currently triggers a dependency on AMRLEVEL, which we would like to avoid unless really needed.

Seen in https://github.com/ECP-WarpX/WarpX/pull/2209 (and cross-linked comments/issues).

cc @c-wetterer-nelson @burlen

c-wetterer-nelson:
## Summary
This PR moves the AMR data adaptor and bridge code behind the AMReX_AMRLEVEL compiler option so that it is not compiled if the end application does not need those optional features.

Fixes https://github.com/AMReX-Codes/amrex/issues/2263

## Additional background
This addresses the problem noted in https://github.com/ECP-WarpX/WarpX/pull/2192#issuecomment-901285722. 
This PR ought to render https://github.com/ECP-WarpX/WarpX/pull/2209 unnecessary.
 
## Checklist

The proposed changes:

### DIFF
--- a/Src/Extern/SENSEI/AMReX_AmrParticleDataAdaptor.H
+++ b/Src/Extern/SENSEI/AMReX_AmrParticleDataAdaptor.H
@@ -9,7 +9,6 @@
 
 #include <AMReX_AmrMesh.H>
 #include <AMReX_MultiFab.H>
-#include <AMReX_AmrDataAdaptor.H>
 #include <AMReX_ParticleDataAdaptor.H>
 // sensei includes
 #include "DataAdaptor.h"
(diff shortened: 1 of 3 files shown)


c-wetterer-nelson:
@ax3l This ought to address the dependency issue we ran into. Let me know what you think!

ax3l (approved):
Awesome, thanks a lot! :tada:

PR #2259, opened 2021-08-19
Update contributing md

etpalmer63:
## Summary
Adds style guide and short doxygen example to CONTRIBUTING.md 

## Checklist

The proposed changes:


CONTRIBUTING.md
```diff
@@ -237,6 +237,75 @@ After saving and then exiting the editor, `git log` should show a clean history
...
+
+       for (int n=0; n<10; ++n) Print() << "Not like this.";
+```
+  or
+```cpp
+       for (int n=0; n<10; ++n)
+          Print() << "Not like this.";
+```
+  * Add a space after function definitions and declarations (but
+not when simply calling the function). For example:
```
WeiqunZhang, line 269: This does not seem to be entirely accurate.  The space is in the middle of function definitions and declarations between the function name and `(`.

WeiqunZhang (approved):

PR #2264, opened 2021-08-20
Add new machine to GNUMake based on Make.olcf

jmsexton03:
## Checklist

The proposed changes:

### DIFF
--- a/Tools/GNUMake/Make.machines
+++ b/Tools/GNUMake/Make.machines
@@ -127,3 +127,8 @@ ifneq (,$(filter $(host_name_env),poplar redwood tulip))
   which_computer := $(host_name_env)
 endif
 
+ifeq ($(findstring birch, $(host_name)),birch)
+  which_site := frontier-coe
+  which_computer := birch
+endif
+
--- a/Tools/GNUMake/Make.machines
+++ b/Tools/GNUMake/Make.machines
@@ -127,8 +127,10 @@ ifneq (,$(filter $(host_name_env),poplar redwood tulip))
   which_computer := $(host_name_env)
 endif
 
+ifeq ($(findstring "cray.com", $(host_name)),"cray.com")
 ifeq ($(findstring birch, $(host_name)),birch)
   which_site := frontier-coe
-  which_computer := birch
+  which_computer := $(host_name_env)
+endif
 endif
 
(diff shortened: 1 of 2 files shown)


Tools/GNUMake/Make.machines
```diff
@@ -127,3 +127,8 @@ ifneq (,$(filter $(host_name_env),poplar redwood tulip))
   which_computer := $(host_name_env)
 endif
 
+ifeq ($(findstring birch, $(host_name)),birch)
```
WeiqunZhang, line 131: Could we make the test to be that `host_name` contains both `birch` and `cray.com`?

Tools/GNUMake/Make.machines
```diff
@@ -127,3 +127,8 @@ ifneq (,$(filter $(host_name_env),poplar redwood tulip))
...
+  which_site := frontier-coe
+  which_computer := birch
+endif
+
```
jmsexton03: Should other frontier-coe sites also check for `cray.com`, and / or be combined into one if test?
```suggestion
ifeq ($(findstring "cray.com", $(host_name)),"cray.com")
ifeq ($(findstring birch, $(host_name)),birch)
  which_site := frontier-coe
  which_computer := $(host_name_env)
endif
endif

```

Tools/GNUMake/Make.machines
jmsexton03: Yes, made a separate comment / suggestion in response since I couldn't come up with a readable one-line if statement to check both things.

Tools/GNUMake/Make.machines
WeiqunZhang: I cannot remember how the host name for other coe machines looked like before the recent changes to the systems.

WeiqunZhang: If `hostname -f` all contains `cray.com` now, we can put them together.

jmsexton03: when I check `hostname -f` it looks like that won't work for tulip. Will just adding an extra check for birch be sufficient for now?

WeiqunZhang: Yes.

WeiqunZhang (approved):

PR #2269, opened 2021-08-22
Fine-tune FabArray::setBndry

WeiqunZhang:
Use either ParallelFor(MF) or ParallelFor(tag) depending on box sizes.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_FabArray.H
+++ b/Src/Base/AMReX_FabArray.H
@@ -1875,7 +1875,7 @@ FabArray<FAB>::setBndry (value_type val)
 }
 
 template <class FAB>
-template <class F, typename std::enable_if<IsBaseFab<F>::value,int>::type>
+template <class F, typename std::enable_if<IsBaseFab<F>::value,int>::type Z>
 void
 FabArray<FAB>::setBndry (value_type val,
                          int        strt_comp,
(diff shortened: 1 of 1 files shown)


Src/Base/AMReX_FabArray.H
```diff
@@ -1875,20 +1875,102 @@ FabArray<FAB>::setBndry (value_type val)
...
+                if (boxarray[indexArray[0]].numPts() > Long(65*65*65)) {
+                    use_mfparfor = false;
+                }
+            } else {
+                for (int i = 0; i < nboxes; ++i) {
+                    const Long npts = boxarray[indexArray[i]].numPts();
+                    if (npts >= Long(64*64*64)) {
+                        use_mfparfor = false;
+                        break;
+                    } else if (npts <= Long(17*17*17)) {
```
kngott, line 1900: Not sure about this for mixed-size box arrays. If big is first or small is first, could get two different implementations for essentially the same box array. And 64^3 and 16^3 seem easily mix-able for AMR-built levels.

Instead, search for biggest and smallest box in the array, and base the test on that? Is big more important or small, or is number of boxes more important?

WeiqunZhang: Even if we know the distribution of box sizes, it's not clear which approach should be taken for mixed-size box arrays.  The reason I put two early `break`s there is to minimize the work on CPU.  When there are a lot of boxes, the amount of time spend on iterating all the boxes on CPU can be comparable to the difference in the two approaches.  In that case, even if we could make the right choice, there is no performance gain.

kngott: After plotting the results, I think this strategy makes sense.

Except for the question of mixed box arrays, but we'll deal with that when we get there.

kngott (approved):

PR #2272, opened 2021-08-25
Protect against nullptr when calling PODVector dtor.

atmyers:
This should allow DeviceVector, etc, to be used as global variables provided we swap with an empty vector before `amrex::Finalize()` is called.

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_PODVector.H
+++ b/Src/Base/AMReX_PODVector.H
@@ -212,7 +212,7 @@ namespace amrex
             a_vector.m_capacity = 0;
         }
 
-        ~PODVector () noexcept { deallocate(m_data, capacity()); }
+        ~PODVector () noexcept { if (m_data != nullptr) { deallocate(m_data, capacity()); }}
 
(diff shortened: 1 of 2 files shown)


Src/Base/AMReX_PODVector.H
```diff
@@ -212,7 +212,7 @@ namespace amrex
             a_vector.m_capacity = 0;
         }
 
-        ~PODVector () noexcept { deallocate(m_data, capacity()); }
+        ~PODVector () noexcept { if (m_data != nullptr) { deallocate(m_data, capacity()); }}
```
ax3l, line 215: For our own allocators, we could also perform this check `if(!nullptr)` in `deallocate`.

That way, `deallocate` would match the behavior of `delete` in std C++: https://en.cppreference.com/w/cpp/language/delete

atmyers: good idea, done.

WeiqunZhang (approved):

PR #2268, opened 2021-08-21
Port Tests/Amr/Advection_AmrLevel from Fortran to C++ GPU

wyphan:
## Summary
The Advection_AmrLevel test/tutorial has been fully ported from Fortran to C++ (except for `initdata()` subroutine that stays in Fortran).

## Additional background
A good chunk of the code was copy-pasted and/or adapted from the AmrCore and GPU/CNS test/tutorials.
All four tests {SingleVortex,UniformVelocity} x {2D,3D} have been tested to work on the CPU, NVIDIA GPU on Gigan, and AMD GPU on Spock (`fcompare` shows differences of the order of 1e-14).

The proposed changes:
- [ ] fix a bug or incorrect behavior in AMReX
- [x] add new capabilities to AMReX
- [ ] changes answers in the test suite to more than roundoff level
- [ ] are likely to significantly affect the results of downstream AMReX users
- [x] include documentation in the code and/or rst files, if appropriate

PR #2273, opened 2021-08-25
Add EditorConfig section for rst files

wyphan:
## Summary
Add a section for `*.rst` files for EditorConfig.

## Additional background
When editing the Sphinx `rst` source files for AmrLevel (PR #2268), I found out that the root `.editorconfig` file doesn't have a section for `rst` files. This PR adds that section, roughly based on the guidelines at https://docs.typo3.org/m/typo3/docs-how-to-document/master/en-us/GeneralConventions/CodingGuidelines.html .
Note that I haven't gauged the "disruption" from whitespace changes that might be caused by this PR.

## Checklist

The proposed changes:

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -43,9 +43,6 @@ indent_size = 3
 # Clean up trailing whitespace
 trim_trailing_whitespace = true
 
-# Automatically wrap lines at 80 columns
-max_line_length = 80
-
 [Makefile]
 # TABs are part of its syntax
 indent_style = tab
--- a/.editorconfig
+++ b/.editorconfig
@@ -26,6 +26,26 @@ insert_final_newline = true
 trim_trailing_whitespace = false
 
 
+[*.rst]
+# Enforce UTF-8 encoding
+charset = utf-8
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -26,6 +26,26 @@ insert_final_newline = true
...
+
+# 3 space indentation
+indent_style = space
+indent_size = 3
+
+# Clean up trailing whitespace
+trim_trailing_whitespace = true
+
+# Automatically wrap lines at 80 columns
+max_line_length = 80
```
WeiqunZhang: Do we want to limit the line length to 80?  https://linux.slashdot.org/story/20/05/31/211211/linus-torvalds-argues-against-80-column-line-length-coding-style-as-linux-kernel-deprecates-it

wyphan: What would you prefer it to be? 100? 132?

WeiqunZhang: We do not have to set it.

WeiqunZhang (approved):

PR #2274, opened 2021-08-25
Parser: Hide local symbols

WeiqunZhang:
Previously `Parser::symbols()` returned a list of symbols including local
symbols.  This is problematic because WarpX makes sure that all symbols are
either a registered variable or constant.  But the local variables are
unknown to WarpX.  In this commit, the local symbols are excluded from the
return value of `Parser::symbols()`.

## Checklist

The proposed changes:


ax3l (approved):
:tada: Thanks a lot!

PR #2283, opened 2021-08-26
fix bug in AmrLevel exclusion logic

c-wetterer-nelson:
## Summary
Fixes a bug in the build logic regarding optional inclusion of the AMRLEVEL components, and return a required  dependency to AMReX_AmrParticleDataAdaptor.H

## Additional background
This bug was introduced in https://github.com/AMReX-Codes/amrex/pull/2258 where the SENSEI adaptors and bridges that processed data from the Amr class. A key dependency was removed, in the refactor. This Pr adds that dependency back in, and moves `AMReX_AmrParticleInSituBridge.H` behind the `AMReX_AMRLEVEL` compile option where it belongs.
## Checklist

The proposed changes:

### DIFF
--- a/Src/Extern/SENSEI/AMReX_AmrParticleDataAdaptor.H
+++ b/Src/Extern/SENSEI/AMReX_AmrParticleDataAdaptor.H
@@ -5,6 +5,7 @@
 #include <AMReX_Config.H>
 #ifdef AMREX_PARTICLES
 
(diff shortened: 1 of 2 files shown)


c-wetterer-nelson:
@ax3l whoops! I didn't catch this error. let me know what you think here.

WeiqunZhang (approved):

PR #2285, opened 2021-08-26
Add SENSEI adaptor and bridge for amrMesh+Particles

c-wetterer-nelson:
## Summary
This PR adds a new composite data adaptor which enables processing AmrMesh data alongside ParticleContainer data. 

Currently tested working in WarpX with the [3D Beam-driven electron acceleration](https://warpx.readthedocs.io/en/latest/usage/examples.html) using the WarpX version in https://github.com/ECP-WarpX/WarpX/pull/2243.

This PR also reworks how the ParticleDataAdapter loops over particle data. 

## Checklist

The proposed changes:

### DIFF
--- a/Src/Extern/SENSEI/AMReX_ParticleDataAdaptorI.H
+++ b/Src/Extern/SENSEI/AMReX_ParticleDataAdaptorI.H
@@ -266,7 +266,7 @@ int ParticleDataAdaptor<NStructReal, NStructInt, NArrayReal, NArrayInt>::GetMesh
     SENSEI_ERROR("No mesh named \"" << meshName << "\"")
     return -1;
   }
-  vtkMultiBlockDataSet *mb = vtkMultiBlockDataSet::New();
+  vtkMultiBlockDataSet* mb = vtkMultiBlockDataSet::New();
 
   if (structureOnly)
   {
@@ -656,15 +656,15 @@ vtkPolyData* ParticleDataAdaptor<NStructReal, NStructInt, NArrayReal, NArrayInt>
 
   // allocate vertex storage for particles
 #ifdef AMREX_SINGLE_PRECISION_PARTICLES
-  vtkFloatArray *coords = vtkFloatArray::New();
+  vtkNew<vtkFloatArray> coords;
 #else
-  vtkDoubleArray *coords = vtkDoubleArray::New();
+  vtkNew<vtkDoubleArray> coords;
 #endif
   coords->SetName("coords");
   coords->SetNumberOfComponents(3);
   coords->SetNumberOfTuples(numParticles);
 #ifdef AMREX_SINGLE_PRECISION_PARTICLES
-  float *pCoords = coords->GetPointer(0);;
+  float *pCoords = coords->GetPointer(0);
 #else
   double *pCoords = coords->GetPointer(0);
 #endif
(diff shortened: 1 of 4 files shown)


c-wetterer-nelson:
@ax3l WarpX PR inbound.

c-wetterer-nelson:
@burlen Here's that AmrMesh + particle data adaptor! I've been working to make sure it works in WarpX, but I've been unsuccessful getting it working in our em PIC example you put together in amrex-tutorials. Still struggling with valid particles there.

Src/Extern/SENSEI/AMReX_ParticleDataAdaptorI.H
```diff
@@ -735,44 +706,41 @@ vtkPolyData* ParticleDataAdaptor<NStructReal, NStructInt, NArrayReal, NArrayInt>
...
+      auto& aos = pti.GetArrayOfStructs();
+      auto numReal = pti.numParticles();
       for (long long i = 0; i < numReal; ++i)
       {
-        const auto &part = parts[i];
+        const auto &part = aos[i];
         if (part.id() > 0)
         {
-          partIds[i] = part.id();
+          idArray->SetTuple1(ptId, part.id());
```
burlen: This changes from dereferencing a pointer to calling a virtual method to pass a single value. This adds unnecessary overhead and I had put the effort to refactor the code from the virtual methods to using pointers. Please don't use the virtual API when a pointer can be used instead.

Src/Extern/SENSEI/AMReX_ParticleDataAdaptorI.H
```diff
@@ -787,38 +755,33 @@ int ParticleDataAdaptor<NStructReal, NStructInt, NArrayReal, NArrayInt>::AddPart
...
+      auto& aos = pti.GetArrayOfStructs();
+      auto numReal = pti.numParticles();
       for (long long i = 0; i < numReal; ++i)
       {
-        const auto &part = parts[i];
+        const auto &part = aos[i];
         if (part.id() > 0)
         {
-          partCpu[i] = part.cpu();
+          cpuArray->SetTuple1(ptId, part.cpu());
```
burlen: Same request. Please use pointer to the data rather than virtual method call to set values on the vtk array

Src/Extern/SENSEI/AMReX_ParticleDataAdaptorI.H
```diff
@@ -944,41 +893,35 @@ int ParticleDataAdaptor<NStructReal, NStructInt, NArrayReal, NArrayInt>::AddPart
...
+      const auto &intData = particle_attributes.GetIntData(index);
 
       for (long long i = 0; i < numReal; ++i)
       {
-        const auto &part = parts[i];
+        const auto &part = aos[i];
         if (part.id() > 0)
         {
-          pData[i] = intData[i];
+          data->SetTuple1(ptId, intData[i]);
```
burlen: same request, avoid virtual methods when passing array elements.

Src/Extern/SENSEI/AMReX_ParticleDataAdaptorI.H
```diff
@@ -654,78 +654,49 @@ vtkPolyData* ParticleDataAdaptor<NStructReal, NStructInt, NArrayReal, NArrayInt>
...
           // add a vertex type cell
           vertex->InsertNextCell(1);
           vertex->InsertCellPoint(ptId);
-
           // copy the partilce coordinates
 #if (AMREX_SPACEDIM == 1)
-          pCoords[0] = part.pos(0);
-          pCoords[1] = amrex_particle_real(0);
-          pCoords[2] = amrex_particle_real(0);
+          points->SetPoint(ptId, part.pos(0), 0.0, 0.0);
```
burlen: this changes from dereferencing a pointer to a virtual method call to pass point values. Please don't use virtual methods to set the values of an array.

burlen (changes requested):
WarpX is just one use case of many possible. Had you tested this with the neighbor particle example as well? While it is great that it's working with WarpX I think we need to get it working with the other demos before merging. @c-wetterer-nelson Let's make some time for a hacking session to debug the issues you're experiencing in the other codes.

I also would like us to not use virtual method calls to set the individual values of arrays when pointers (or non-virtual API) could be used.

c-wetterer-nelson:
Hi @burlen, thanks for the review! I appreciate your concern with the virtual function calls. When I was trying to get this thing working in WarpX, I was having all sort of strange crashes and memory issues, so I went through and refactored the whole thing and the result was a working adaptor in warpx! However, that refactor  included the above noted virtual function calls _and_ redoing how we iterate over particles to line up with the AMReX tutorials. I have a hunch we can keep the updated iteration and roll back to your array assignment organization. better yet, maybe we take the opportunity to make this all zero copy?

burlen:
> Hi @burlen, thanks for the review! I appreciate your concern with the virtual function calls. When I was trying to get this thing working in WarpX, I was having all sort of strange crashes and memory issues, so I went through and refactored the whole thing and the result was a working adaptor in warpx! However, that refactor included the above noted virtual function calls _and_ redoing how we iterate over particles to line up with the AMReX tutorials. I have a hunch we can keep the updated iteration and roll back to your array assignment organization. better yet, maybe we take the opportunity to make this all zero copy?

Glad you got it working, SetTuple can be replaced with pointer manipulations(after all it's what goes on inside the SetTuple calls), your other changes are what made the difference.

zero copy won't be possible until we modify VTK to handle AOS case. VTK is missing stride that tells the offset between array elements. VTK's AOS arrays currently assume a stride of NumberOfTuples which works for VTK's representation of  vectors but does not work in general for C structs. This is one of the motivations for forking VTK.

c-wetterer-nelson:
Update to the particle adaptor. This addresses the virtual function call concerns above and keeps the updated iterator layout. Successfully tested in the Neighborlist demo! 

Tests:
- WarpX 3D example problem: Working
- amrex-tutorials/SENSEI/Advection_AmrLevel: Working
- amrex-tutorials/SENSEI/Particles/Neighborlist: Working
- amrex-tutorials/SENSEI/Particles/ElectromagneticPIC: still on the fritz (pretty sure it's not the adaptor's fault)

burlen:
@c-wetterer-nelson this is awersome! thanks for testing with the various tutorials and avoiding VTK data array virtaul API. We have yet to get the EM PIC tutorial working so if you agree we can leave that for future work.

I don't see how to change my review so I'll leave this in a comment.

c-wetterer-nelson:
thanks @burlen! I think you can start a new review and the new review will replace the old one. 

I agree on the EMPIC future work.

burlen (approved):
this looks good to me.

atmyers (approved):

PR #2289, opened 2021-08-27
WIP: Eb slopes grown

asalmgren:
## Summary
Split the AMReX_EB_slopes_K.H file into two files -- one for 2D and one for 3D.   Also add "grown" functionality which will use a larger stencil for the least squares slope calculation.
## Checklist

The proposed changes:


asalmgren:
We will instead move the slopes routines out of amrex itself and into AMReX-Hydro.

PR #2291, opened 2021-08-27
Update inputs files of Advection_AmrLevel

WeiqunZhang:
There are no longer probin files.

## Checklist

The proposed changes:


wyphan (approved):
LGTM. I already deleted the `probin` files anyway after integrating the contents into the `tagging` (UV and SV) and `prob` (UV only) blocks for feeding into `ParmParse` from the main input files.

PR #2296, opened 2021-08-28
Add gradient error tagging in Tests/Advection_AmrLevel

wyphan:
## Summary
Add proper `FillPatch()`ing for gradient checks in `AmrLevelAdv::errorEst()`.

## Additional background
While working on #2268, I "accidentally" discovered that even though gradient error tagging is coded in `Advection_AmrLevel`, this check is never actually performed. This PR tries to fix that by adapting the error tagging code from `Tests/GPU/CNS`.
Tested on local for CPU (error = exactly 0) and on Gigan for NVIDIA GPU (error = 1e-14 range).

## Checklist

The proposed changes:

### DIFF
--- a/Tests/Amr/Advection_AmrLevel/Exec/SingleVortex/inputs
+++ b/Tests/Amr/Advection_AmrLevel/Exec/SingleVortex/inputs
@@ -41,5 +41,7 @@ amr.plot_int          = 100    # number of timesteps between plot files
 adv.do_tracers = 0
 
 # ERROR TAGGING
-tagging.phierr =  1.01  1.1   1.5
-tagging.max_phierr_lev = 10
+tagging.phierr  =  1.01  1.1   1.5
+tagging.phigrad =  1e+20 1e+20 1e+20
+tagging.max_phierr_lev  = 10
+tagging.max_phigrad_lev = 10
--- a/Tests/Amr/Advection_AmrLevel/Exec/SingleVortex/inputs
+++ b/Tests/Amr/Advection_AmrLevel/Exec/SingleVortex/inputs
@@ -41,5 +41,5 @@ amr.plot_int          = 100    # number of timesteps between plot files
 adv.do_tracers = 0
 
 # ERROR TAGGING
-tagging.phierr  =  1.01  1.1   1.5
-tagging.max_phierr_lev  = 10
+tagging.phierr =  1.01  1.1   1.5
+tagging.max_phierr_lev = 10
--- a/Tests/Amr/Advection_AmrLevel/Exec/SingleVortex/inputs
+++ b/Tests/Amr/Advection_AmrLevel/Exec/SingleVortex/inputs
@@ -42,6 +42,4 @@ adv.do_tracers = 0
 
 # ERROR TAGGING
 tagging.phierr  =  1.01  1.1   1.5
-tagging.phigrad =  1e+20 1e+20 1e+20
 tagging.max_phierr_lev  = 10
-tagging.max_phigrad_lev = 10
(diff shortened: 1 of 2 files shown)


Tests/Amr/Advection_AmrLevel/Exec/SingleVortex/inputs
```diff
@@ -41,5 +41,7 @@ amr.plot_int          = 100    # number of timesteps between plot files
 adv.do_tracers = 0
 
 # ERROR TAGGING
-tagging.phierr =  1.01  1.1   1.5
-tagging.max_phierr_lev = 10
+tagging.phierr  =  1.01  1.1   1.5
+tagging.phigrad =  1e+20 1e+20 1e+20
+tagging.max_phierr_lev  = 10
+tagging.max_phigrad_lev = 10
```
WeiqunZhang: Since the default of phigrad are huge numbers, I think we should still use `-1` as default for max_phigrad_lev  so that we do not do unnecessary work.

Tests/Amr/Advection_AmrLevel/Source/AmrLevelAdv.cpp
```diff
@@ -710,6 +710,11 @@ AmrLevelAdv::errorEst (TagBoxArray& tags,
 {
     MultiFab& S_new = get_new_data(Phi_Type);
 
+    // Create temporary multifab with properly filled patches
+    const Real cur_time = state[Phi_Type].curTime();
+    MultiFab phi(S_new.boxArray(), S_new.DistributionMap(), NUM_STATE, 1);
+    FillPatch(*this, phi, phi.nGrow(), cur_time, Phi_Type, 0, NUM_STATE);
```
WeiqunZhang: We do not need to create a new temporary MulitFab here.  It's OK to call FillPatch on S_new.  We also do not need to fill all ghost cells, just one is sufficient for computing grad phi.  Finally, if level >= max_phigrad_lev, we do not need to call FillPatch.

WeiqunZhang (changes requested):

Tests/Amr/Advection_AmrLevel/Source/AmrLevelAdv.cpp
wyphan: I can't seem to find a way around not creating a temporary `MultiFab` for the gradient check (assertion error trying to fill in ghost cells). I think I can only implement the latter two.

WeiqunZhang: OK.  Didn't realize the StateData in this example does not have ghost cells.

Tests/Amr/Advection_AmrLevel/Source/AmrLevelAdv.cpp
```diff
@@ -710,19 +710,26 @@ AmrLevelAdv::errorEst (TagBoxArray& tags,
 {
     MultiFab& S_new = get_new_data(Phi_Type);
 
+    // Properly fill patches and ghost cells for phi gradient check.
+    MultiFab phi(S_new.boxArray(), S_new.DistributionMap(), NUM_STATE, 1);
+    if (level < max_phigrad_lev) {
+        const Real cur_time = state[Phi_Type].curTime();
+        FillPatch(*this, phi, 1, cur_time, Phi_Type, 0, NUM_STATE);
+    }
+
```
WeiqunZhang: I think you should do something like this to avoid allocating a temporary MultiFab when level is not less than max_phigrad_lev. 
```
MultiFab phitmp;
if (level < max_phigrad_lev) {
    phitmp.define(....);
    FillPatch(....)
}

MultiFab const& phi = (level < max_phigrad_lev) ? phitmp : S_new;
```

Then use phi in both tagging functions in the code following this.

wyphan: I like that approach, uses less memory. Lemme commit it real quick.

WeiqunZhang (approved):

PR #2301, opened 2021-09-01
Gpu::Atomic::If

atmyers:
This adds support for *conditionally* doing atomic operations. For example, to atomically subtract one from a value, but only if the result would be non-negative:

```
Gpu::Atomic::If(&data[i], 1.0, amrex::Minus<amrex::Real>(),
                          [=] AMREX_GPU_DEVICE (amrex::Real res) noexcept
                          {
                              return (res >= 0.0);
                          });
```

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_GpuAtomic.H
+++ b/Src/Base/AMReX_GpuAtomic.H
@@ -49,7 +49,7 @@ namespace detail {
 
     template <typename R, typename I, typename Op, typename Cond>
     AMREX_GPU_DEVICE AMREX_FORCE_INLINE
-    bool atomic_op_if (R* const address, R const val, Op const op, Cond const cond) noexcept
+    bool atomic_op_if (R* const address, R const val, Op&& op, Cond&& cond) noexcept
     {
 #if defined(__SYCL_DEVICE_ONLY__)
         constexpr auto mo = sycl::memory_order::relaxed;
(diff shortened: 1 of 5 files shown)


Src/Base/AMReX_GpuAtomic.H
```diff
@@ -145,6 +196,77 @@ namespace detail {
...
+*  but only if "cond" is true.
+*
+* \tparam T the type pointed to by add
+* \tparam Op callable that takes two T argument and combines them
+* \tparam Cond callable that takes a "T" a returns whether to do the update
+*
+* \param add address to atomically update
+* \param value value to combine
+* \param op callable specifying the operation to use to combine *add and value
+* \param cond callable specifying the condition to test on first.
```
WeiqunZhang, line 236: Maybe we can make it clear that the test is performed on the would-be combined value.

Src/Base/AMReX_GpuAtomic.H
```diff
@@ -145,6 +196,77 @@ namespace detail {
...
+                   std::forward<Op>(op), std::forward<Cond>(cond));
+    }
+
+    template <typename T, typename Op, typename Cond>
+    AMREX_GPU_DEVICE AMREX_FORCE_INLINE
+    bool If_device (T* const sum, T const value, Op&& op, Cond&& cond) noexcept
+    {
+        return detail::atomic_op_if<T, T>(sum, value,
+                   std::forward<Op>(op), std::forward<Cond>(cond));
+    }
```
WeiqunZhang: I think we can simplify this to something like
```
    template <typename T, typename Op, typename Cond,
              std::enable_if_t<sizeof(T) == sizeof(unsigned int), int> foo = 0>
    bool If_device (T* const sum, T const value, Op&& op, Cond&& cond) noexcept
    {
        return detail::atomic_op_if<T, unsigned int>(sum, value,
                   std::forward<Op>(op), std::forward<Cond>(cond));
    }

// another one for sizeof(unsigned long long)
```

atmyers: Yes, that does seem to work.

WeiqunZhang (approved):

PR #2303, opened 2021-09-01
Update Profiling API docs.

kngott:
## Summary
Based on previous discussion that TinyProfiler's REGION wasn't well documented, this adds in the discussion about the various methods and macros available for instrumentation.

Please proofread! I _keep_ catching small typos in here!

## Checklist

The proposed changes:

### DIFF
--- a/Docs/sphinx_documentation/source/AMReX_Profiling_Tools.rst
+++ b/Docs/sphinx_documentation/source/AMReX_Profiling_Tools.rst
@@ -7,12 +7,12 @@
 Types of Profiling
 ==================
 
-AMReX's build-in profiling works through objects that start and stop timers
-based on user-placed macros or in the objects' constructor and destructor.
-The results of these timers are stored in a global list that is consolidated
-and printed during finalization, or at a user-defined flush.
+AMReX's built-in profiling works through objects that start and stop timers
+based on user-placed macros or an object's constructor and destructor.
+The results from these timers are stored in a global list that is consolidated
+and printed during finalization, or at a user-defined flush point.
 
-Currently, AMReX has two options for build-in profiling:
+Currently, AMReX has two options for built-in profiling:
 :ref:`sec:tiny:profiling` and :ref:`sec:full:profiling`.
 
 .. _sec:tiny:profiling:
@@ -20,34 +20,38 @@ Currently, AMReX has two options for build-in profiling:
 Tiny Profiling
 ----------------------
 
-To enable "Tiny Profiling", if using GNU Make then set
+To enable "Tiny Profiling" with GNU Make, in `GNUMakefile` set,
 
 ::
 
   TINY_PROFILE = TRUE
   PROFILE      = FALSE
 
-in your GNUMakefile.   If using cmake then set the following cmake flags
+If building with CMake, set the following cmake flags,
 
 ::
 
   AMReX_TINY_PROFILE = ON
   AMReX_BASE_PROFILE = OFF
 
-Note that if you set ``PROFILE = TRUE``  (or ``AMReX_BASE_PROFILE =
-ON``) then this will override the ``TINY_PROFILE`` flag and tiny profiling will
-be disabled.
+.. note::
+   If you set ``PROFILE = TRUE`` (or ``AMReX_BASE_PROFILE =
+   ON``) to enable full profiling then this will override the ``TINY_PROFILE`` flag
+   and tiny profiling will be disabled.
 
-At the end of the run, a summary of exclusive and inclusive function times will
-be written to stdout.  This output includes the minimum and maximum (over
+Output
+~~~~~~
+
+At the end of a run, a summary of exclusive and inclusive function times will
+be written to ``stdout``. This output includes the minimum and maximum (over
 processes) time spent in each routine as well as the average and the maximum
-percentage of total run time.   See :ref:`sec:sample:tiny` for sample output.
+percentage of total run time. See :ref:`sec:sample:tiny` for sample output.
 
-The tiny profiler automatically writes the results to stdout at the end of your
+The tiny profiler automatically writes the results to ``stdout`` at the end of your
 code, when ``amrex::Finalize();`` is reached. However, you may want to write
 partial profiling results to ensure your information is saved when you may fail
 to converge or if you expect to run out of allocated time. Partial results can
-be written at user-defined times by inserting the line:
+be written at user-defined points in the code by inserting the line:
 
 ::
 
@@ -73,23 +77,23 @@ Full Profiling
 If you set ``PROFILE = TRUE`` then a ``bl_prof`` directory will be written that
 contains detailed per-task timings for each processor.  This will be written in
 ``nfiles`` files (where ``nfiles`` is specified by the user).  In addition, an
-exclusive-only set of function timings will be written to stdout.
+exclusive-only set of function timings will be written to ``stdout``.
 
 Trace Profiling
 ~~~~~~~~~~~~~~~
 
-   If, in addition to ``PROFILE = TRUE``, you set ``TRACE_PROFILE = TRUE``,
+   If you set ``TRACE_PROFILE = TRUE`` in addition to ``PROFILE = TRUE``,
    then the profiler keeps track of when each profiled function is called and
    the ``bl_prof`` directory will include the function call stack. This is
    especially useful when core functions, such as :cpp:`FillBoundary` can be
-   called from many different regions of the code. Part of the trace profiling
-   is the ability to set regions in the code which can be analyzed for
+   called from many different regions of the code. Using trace profiling
+   allows one to specify regions in the code that can be analyzed for
    profiling information independently from other regions.
 
 Communication Profiling
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-  If, in addition to ``PROFILE = TRUE``, you set ``COMM_PROFILE = TRUE``, then
+  If you set ``COMM_PROFILE = TRUE`` in addition to ``PROFILE = TRUE``, then
   the ``bl_prof`` directory will contain additional information about MPI
   communication (point-to-point timings, data volume, barrier/reduction times,
   etc.). ``TRACE_PROFILE = TRUE`` and ``COMM_PROFILE = TRUE`` can be set
@@ -105,7 +109,7 @@ AMReX profiler objects are created and managed through :cpp:`BL_PROF` macros.
 
 .. highlight:: c++
 
-To start with, you must at least instrument main(), i.e.:
+To start, you must at least instrument main(), i.e.:
 
 ::
 
@@ -120,7 +124,9 @@ To start with, you must at least instrument main(), i.e.:
       amrex::Finalize();
     }
 
-    // OR
+or
+
+::
 
     void main_main()
     {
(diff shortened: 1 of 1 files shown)


Docs/sphinx_documentation/source/AMReX_Profiling_Tools.rst
```diff
@@ -157,20 +164,24 @@ time an entire function. For example:
     }    // <------ Timer goes out of scope here, calling stop and returning the function time.
 
 Note that all AMReX timers are scoped and will call "stop" when the corresponding object is destroyed.
-This macro is unique because it can _only_ stop when it goes out of scope.
+This macro is unique because it can *only* stop when it goes out of scope.
+
+.. HACK check meaning of _only_ here.
```
etpalmer63: Were the underscores on "only" for emphasis?

kngott: Yup. So, I agree with that change.

Docs/sphinx_documentation/source/AMReX_Profiling_Tools.rst
```diff
@@ -299,11 +320,18 @@ inside :cpp:`MyFunc0` and :cpp:`MyFunc1`:
...
+
 
+.. admonition:: A Combined Tiny-Full Region Profiling Strategy
+
+  We recommend using scoped regions in a few select places that will be timed when either the
+  Tiny or Full Profiler is activated. Because named regions will be isolated by the Full Profiler
+  and ignored by the Tiny Profiler, they can be used for regions that you wish to instrument only
+  when the Full Profiler is active.
+
+.. HACK verify intended meaning.
```
etpalmer63: I want to verify I understood this correctly.

Docs/sphinx_documentation/source/AMReX_Profiling_Tools.rst
```diff
@@ -321,10 +349,12 @@ with the following calls:
 
 Note that the start and stop calls must be matched and the profiling output
 will warn of any :fortran:`bl_proffortfuncstart` calls that were not stopped
-with :fortran:`bl_proffortfuncstop` calls (in debug mode only). You will need
-to add :fortran:`bl_proffortfuncstop` before any returns and at the end of the
+with :fortran:`bl_proffortfuncstop` calls (when in debug mode only). You will need
+to add :fortran:`bl_proffortfuncstop` before any returns, at the end of the
 function or at the point in the function you want to stop profiling.
 
+.. HACK check meaning change "and" to ","
```
etpalmer63: Based off the context, I assumed it only needed a single 'bl_proffortfuncstop' in one of the three listed places. So I changed the "and" to a ",". Please correct me if I misunderstood.

Docs/sphinx_documentation/source/AMReX_Profiling_Tools.rst
kngott: Not... quite. Regions don't add timers, just isolate sets of timers into a separate analysis. So, they do something like: "in this code block, these timers were found and took this long".

So, I would change line 329 to: "We recommend using scoped regions in a few select places that will be implemented when either the" (or something else that removed "timed", as regions don't add a timer.)

Docs/sphinx_documentation/source/AMReX_Profiling_Tools.rst
kngott: Again, not quite. This is just a poorly written sentence. :)

There are two requirements: There must be a stop, and it must be reached by all possible code paths (before any returns). So, perhaps that's what we say instead?

"need to add ... to ensure there is always a matching stop before leaving the scope of the corresponding start, taking into account all possible code paths."

Docs/sphinx_documentation/source/AMReX_Profiling_Tools.rst
WeiqunZhang: Unfortunately, tiny profiler does add a timer to a region.

kngott: Ah, yes. To get the "total time" of the region. There's isn't the extra REGION+TIMER variation, like in the Full Profiler.

I'd still lean away from using the word "timer", as we don't want people putting in REGIONS to act like "timers". Unless we want a full section discussing the output in substantial detail, this might be the way to go.

Docs/sphinx_documentation/source/AMReX_Profiling_Tools.rst
etpalmer63: Thanks for taking the time to explain. What do you think about it like this?

Note that the start and stop calls must be matched before leaving the scope of the corresponding start. Moreover, it is necessary to take into account all possible code paths. Therefore, you may need to add :f:`bl_stop` in multiple locations, such as before any returns, at the end of the function and at the point in the function where you want to stop profiling. The profiling output will only warn of any :f:`bl_start` calls that were not stopped with :f:`bl_stop` calls when in debug mode.

kngott: Sounds good!

Docs/sphinx_documentation/source/AMReX_Profiling_Tools.rst
WeiqunZhang: Then I think we should delete this line.

Docs/sphinx_documentation/source/AMReX_Profiling_Tools.rst
```diff
@@ -107,59 +118,220 @@ You must at least instrument main(), i.e
...
+
+      < Your Function Code Block>
+
+    }    // <------ Timer goes out of scope here, calling stop and returning the function time.
+
+Note that all AMReX timers are scoped and will call "stop" when the corresponding object is destroyed.
+This macro is unique because it can *only* stop when it goes out of scope.
 
-For other timers within an already instrumented function, add:
+.. HACK check meaning of _only_ here.
```
WeiqunZhang:
```suggestion
```

Docs/sphinx_documentation/source/AMReX_Profiling_Tools.rst
```diff
@@ -107,59 +118,220 @@ You must at least instrument main(), i.e
...
+
+
+.. admonition:: A Combined Tiny-Full Region Profiling Strategy
+
+  We recommend using scoped regions in a few select places that will be timed when either the
+  Tiny or Full Profiler is activated. Because named regions will be isolated by the Full Profiler
+  and ignored by the Tiny Profiler, they can be used for regions that you wish to instrument only
+  when the Full Profiler is active.
+
+.. HACK verify intended meaning.
```
WeiqunZhang:
```suggestion
```
Don't forget to remove this line before merging.

Docs/sphinx_documentation/source/AMReX_Profiling_Tools.rst
```diff
@@ -177,10 +349,12 @@ with the following calls:
 
 Note that the start and stop calls must be matched and the profiling output
 will warn of any :fortran:`bl_proffortfuncstart` calls that were not stopped
-with :fortran:`bl_proffortfuncstop` calls (in debug mode only). You will need
-to add :fortran:`bl_proffortfuncstop` before any returns and at the end of the
+with :fortran:`bl_proffortfuncstop` calls (when in debug mode only). You will need
+to add :fortran:`bl_proffortfuncstop` before any returns, at the end of the
 function or at the point in the function you want to stop profiling.
 
+.. HACK check meaning change "and" to ","
```
WeiqunZhang:
```suggestion
```
Don't forget to remove this line.

Docs/sphinx_documentation/source/AMReX_Profiling_Tools.rst
etpalmer63: Sorry, that my fault. I will clean it up.

Docs/sphinx_documentation/source/AMReX_Profiling_Tools.rst
```diff
@@ -287,10 +287,39 @@ macros in the following way:
...
+    -------------------------------------------------------------
+    Name          NCalls  Incl. Min  Incl. Avg  Incl. Max   Max %
+    -------------------------------------------------------------
+    REG::MyFuncs    1000      8.809      8.809      8.809  28.39%
+    MyFunc0         1000      4.402      4.402      4.402  14.19%
+    MyFunc1         1000       4.39       4.39       4.39  14.15%
+    -------------------------------------------------------------
+
+    END REGION MyFuncs
+
```
kngott, line 341: Should this go down under the "Sample Output" section? Seems out of place here, given that section down below.

etpalmer63: Ah, I've been so focused on the named regions I missed that. 

Actually, how about we combine the Sample Output Section with the Tiny profile section?  Then the additional region sample would make more sense.

Docs/sphinx_documentation/source/AMReX_Profiling_Tools.rst
```diff
@@ -324,8 +353,9 @@ the Tiny Profiler output and to allow instrumentation strategies without needing
 
 .. admonition:: A Combined Tiny-Full Region Profiling Strategy
 
-  We recommend using scoped regions in a few select places that will be timed when either the
-  Tiny or Full Profiler is activated. Because named regions will be isolated by the Full Profiler
+  We recommend using scoped regions in a few select places that will be isolated for analysis
+  when either the Tiny or Full Profiler is activated.
+  Because only named regions will be isolated by the Full Profiler
```
etpalmer63: I wasn't able to get this to work for myself, so I want to verify that "only named regions will be isolated by the Full Profiler... " is correct?

kngott:
These look reasonable. Are we still going to re-write the REGIONS part into Tiny and Full profiling sections?

etpalmer63:
At this point I'm content with leaving it as it is because it seems to me that this gives an accurate introduction to most the current functionality that is being used. My reasoning is that: I think its ok not to explicitly describe some features if there isn't a lot of use (or we don't expect a lot of use). If later we discover users want to use more of the features I think we can add it then.  However, I'm very concerned about making inaccurate statements or misjudging which features users want to/are using. So please let me know your opinion.  

"When using the Tiny Profiler, the only available region macro is the scoped macro."   -- This one is still correct right? 

If it is, and you're ok with it. I think these changes look ready.

kngott:
I believe things are accurate as written here. So, sure3: we can go ahead and approve this.

Docs/sphinx_documentation/source/AMReX_Profiling_Tools.rst
etpalmer63: Removed this in a later version.

Docs/sphinx_documentation/source/AMReX_Profiling_Tools.rst
etpalmer63: Removed this in a later version.

Docs/sphinx_documentation/source/AMReX_Profiling_Tools.rst
etpalmer63: Got it.

Docs/sphinx_documentation/source/AMReX_Profiling_Tools.rst
etpalmer63: got it.

etpalmer63 (approved):

PR #2307, opened 2021-09-02
Update DiffMultiFab tool

WeiqunZhang:
Print the location of min and max diff.  Make it work with MPI.  An example
of output is below,

```
diffmultifab3d.gnu.TEST.ex infile1=mf1  infile2=mf2 ngrow=0

Reading mf1
Reading mf2
Component 0
    Min and max of the diff are -0.5738299806 and 0.3011247047
    Min Index: (63,127,127)
    Max Index: (75,127,127)
    Min and max of 1st mf are 0.1249966244 and 1.000010767
    Min and max of 2nd mf are 0.125 and 1
Component 1
    Min and max of the diff are -1.273924814e-05 and 0.3954904262
    Min Index: (31,127,127)
    Max Index: (64,127,127)
    Min and max of 1st mf are -1.273924814e-05 and 0.3954904262
    Min and max of 2nd mf are 0 and 0
Component 2
    Min and max of the diff are 0 and 0
    Min and max of 1st mf are 0 and 0
    Min and max of 2nd mf are 0 and 0
Writing mfdiff
```

## Checklist

The proposed changes:

### DIFF
--- a/Tools/C_util/DiffMultiFab/diffmultifab.cpp
+++ b/Tools/C_util/DiffMultiFab/diffmultifab.cpp
@@ -7,17 +7,10 @@
 #include <AMReX_VisMF.H>
 #include <AMReX_ParmParse.H>
 
-#ifdef AMREX_USE_MPI
-#define MY_USE_MPI 1
-#else
-#define MY_USE_MPI 0
-#endif
-
 using namespace amrex;
 
 void
-print_usage (int,
-             char* argv[])
+print_usage (int, char* argv[])
 {
     std::cerr << "usage:\n";
     std::cerr << argv[0] << " infile1=f1 infile2=f2, ngrow=ngrow" << std::endl;
@@ -26,11 +19,11 @@ print_usage (int,
 
 int main(int argc, char* argv[])
 {
-    static_assert(MY_USE_MPI!=1, "cannot work with MPI" );
     amrex::Initialize(argc,argv);
     {
-        if (argc < 3)
+        if (argc < 3) {
             print_usage(argc,argv);
+        }
 
         std::string name1, name2;
         int ngrow = -1;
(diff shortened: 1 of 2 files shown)


wyphan:
While you're at it, do you mind converting the three arguments to positional arguments?
It kinda sucks not being able to use tab completion if there is the `infile1=` part.

jmsexton03:
> While you're at it, do you mind converting the three arguments to positional arguments?
> It kinda sucks not being able to use tab completion if there is the `infile1=` part.

That sounds like it could be a breaking change for existing workflows. A workaround for tab complete I've used before is to add an extra space between the equals, like: `infile= plt0` . The extra space can be deleted later if required. You can also change how your particular shell takes into account context with auto-complete.

drummerdoc:
I use bash and the tab-auto-complete seems to work fine for that the ``infile=XXX`` arguments, AND the space after the equal sign and before the name is removed in the ParmParse processing, Jean's workaround shouldn't be an issue either.  Maybe there's a configuration setting you can add to your shell?

Tools/C_util/DiffMultiFab/diffmultifab.cpp
```diff
@@ -72,31 +74,40 @@ int main(int argc, char* argv[])
             mf2_min[icomp] = mf2.min(icomp,ngrow);
             mf2_max[icomp] = mf2.max(icomp,ngrow);
         }
-*/
-
-        MultiFab mfdiff(mf1.boxArray(), mf1.DistributionMap(), ncomp, ngrow);
 
+        MultiFab mfdiff(mf1.boxArray(), mf2.DistributionMap(), ncomp, ngrow);
```
wyphan, line 72: Is this a typo? `mfdiff` uses the `BoxArray` for `mf1` but the `DistributionMap` for `mf2`.

WeiqunZhang:
> While you're at it, do you mind converting the three arguments to positional arguments?
> It kinda sucks not being able to use tab completion if there is the `infile1=` part.

Only if existing users approve the change.

https://lkml.org/lkml/2012/12/23/75

WeiqunZhang:
As @drummerdoc has mentioned, ParmParse is OK with `a = b` or `a= b` or `a =b`.

wyphan:
@drummerdoc I can confirm that `ParmParse` automatically removes the space after `=`. Thanks for the tip!
My `.bashrc` and `.profile` files can be viewed [here](https://gist.github.com/wyphan/7dfc254f48f6089dc8b4f494fbc62fb4), if you wanna help me debug the tab completion when I don't add the space after `=`.

WeiqunZhang:
`shopt -u progcomp`

wyphan:
> https://lkml.org/lkml/2012/12/23/75

That's a pretty strong reaction from Linus lol

wyphan:
> `shopt -u progcomp`

I wish this is a permanent work around. It disables programmable completion, which is useful for `make` to list all valid targets, for instance.
Maybe somebody can write a bash completion script instead that we can `source`?

WeiqunZhang:
I reverted @wyphan's changes to `print_usage` because

1. It's inconsistent with what the code actually does.
2. If we change the interface, we should ask other people's opinion, because it's a breaking change and all other tools in `Tools/C_Util/` have similar `ParmParse` based interface.
3. Even if we do change the interface, there is still `f1 f2 ngrow=1` vs. `[-g 1] f1 f2`.  The later is more consistent with `fcompare`'s interface.

wyphan (approved):

Tools/C_util/DiffMultiFab/diffmultifab.cpp
WeiqunZhang: No, it's not.  In both MPI and non-MPI case, we want to use mf2's DistributionMapping.

wyphan:
@WeiqunZhang Ok, I was just trying to make the help text more descriptive.

Tools/C_util/DiffMultiFab/diffmultifab.cpp
WeiqunZhang: Otherwise, `mfdiff.minus(mf2, 0, ncomp, ngrow);` might fail because it's not a parallel function.

PR #2316, opened 2021-09-06
Add const to Array's sum and product functions

WeiqunZhang:
Since C++14, `constexpr` does not imply `const`.  This is a follow-up on #2217.

## Checklist

The proposed changes:


wyphan:
That's really weird, how `constexpr` does not imply `const` automatically. Anyway, LGTM.

wyphan (approved):
That's really weird, how `constexpr` does not imply `const` automatically. Anyway, LGTM.

atmyers (approved):

PR #2320, opened 2021-09-08
Add comment on "is_periodic" to doxygen

etpalmer63:
## Summary

Adds a short comment to Doxygen docs about `is_periodic`  based off the answer to a user's question. 

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_Geometry.H
+++ b/Src/Base/AMReX_Geometry.H
@@ -137,7 +137,7 @@ public:
      * domains. Useful for when the Geometry object has been constructed with the default constructor.
      *
      * \param dom cell-centered `Box` specifying the index space bounds of the domain.
-     * \param rb A `RealBox` specifying the real space bounds of the domain. Optional. If not specified,
+     * \param rb A pointer to a `RealBox` specifying the real space bounds of the domain. Optional. If not specified,
      *        will be ParmParsed from "geometry.prob_lo" and "geometry.prob_hi".
      * \param coord Specifies the coordinate system type. Optional. Can be one of:
      *        CoordSys::cartesian
--- a/Src/Base/AMReX_Geometry.H
+++ b/Src/Base/AMReX_Geometry.H
@@ -154,7 +154,7 @@ public:
      * domains. Useful for when the Geometry object has been constructed with the default constructor.
      *
      * \param dom cell-centered `Box` specifying the index space bounds of the domain.
-     * \param rb A `RealBox` specifying the real space bounds of the domain. Optional. If not specified,
+     * \param rb A pointer to a `RealBox` specifying the real space bounds of the domain. Optional. If not specified,
      *        will be ParmParsed from "geometry.prob_lo" and "geometry.prob_hi".
      * \param coord Specifies the coordinate system type. Optional. Can be one of:
      *        CoordSys::cartesian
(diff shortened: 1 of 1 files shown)


Src/Base/AMReX_Geometry.H
```diff
@@ -62,7 +63,7 @@ public:
     RealBox prob_domain;
     Box domain;
     Real dx[AMREX_SPACEDIM];
-    int is_periodic[AMREX_SPACEDIM];
+    int is_periodic[AMREX_SPACEDIM]; /**< For each dimension, 0 if the domain is non-periodic and 1 if it is. */
```
atmyers, line 66: Does it make sense to add doxygen comments to the constructor and `define()` methods explaining this, instead of / in addition to commenting the member?

etpalmer63: Yes. I don't see why it couldn't be added to both. 

A positive of documenting the member this way is that you can search `is_periodic` and it will take you to the information. 

I added to the doxy comments on `define()` because it seemed natural to explain each variable there. What information would you want to put with the constructor? or should we just add to the class description?

atmyers: See my push for what I had in mind. I think docstrings for the class itself should contain what the class does, while the constructor docstrings should explain all the arguments, like with other methods.

Src/Base/AMReX_Geometry.H
```diff
@@ -100,9 +132,40 @@ public:
     static void ResetDefaultPeriodicity (const Array<int,AMREX_SPACEDIM>& is_per) noexcept;
     static void ResetDefaultCoord (int coord) noexcept;
 
-    //! Set the rectangular domain after using default constructor.
+    /**
+     * Defines a geometry object that describes the problem domain and coordinate system for rectangular problem
+     * domains. Useful for when the Geometry object has been constructed with the default constructor.
+     *
+     * \param dom cell-centered `Box` specifying the index space bounds of the domain.
+     * \param rb A `RealBox` specifying the real space bounds of the domain. Optional. If not specified,
```
atmyers:
```suggestion
     * \param rb A pointer to a `RealBox` specifying the real space bounds of the domain. Optional. If not specified,
```

Src/Base/AMReX_Geometry.H
```diff
@@ -100,9 +132,40 @@ public:
...
+     *        Optional. Default is non-periodic.
+     */
     void define (const Box& dom, const RealBox* rb = 0, int coord = -1, int const* is_per = nullptr) noexcept;
+
+    /**
+     * Defines a geometry object that describes the problem domain and coordinate system for rectangular problem
+     * domains. Useful for when the Geometry object has been constructed with the default constructor.
+     *
+     * \param dom cell-centered `Box` specifying the index space bounds of the domain.
+     * \param rb A `RealBox` specifying the real space bounds of the domain. Optional. If not specified,
```
atmyers:
```suggestion
     * \param rb A pointer to a `RealBox` specifying the real space bounds of the domain. Optional. If not specified,
```

Src/Base/AMReX_Geometry.H
etpalmer63: Looks great! I also think this is a good general guideline. I'll add it to the Doxygen best practices guide.

atmyers (approved):

PR #2321, opened 2021-09-08
CMake: Always CUDA Perf-Neutral Debug

ax3l:
## Summary

Always default-ON for performance-neutral debug info to CUDA builds with CMake, even in `Release` mode.

## Additional background

We used `RelWithDebInfo` before but switched downstream to `Release` by default as well to avoid `-O2`/`-O3` complications in flags.

## Checklist

The proposed changes:

### DIFF
--- a/Tools/CMake/AMReXCUDAOptions.cmake
+++ b/Tools/CMake/AMReXCUDAOptions.cmake
@@ -78,16 +78,9 @@ cuda_print_option(AMReX_CUDA_COMPILATION_TIMER)
 option(AMReX_CUDA_DEBUG "Generate debug information for device code (optimizations: off)" OFF)
 cuda_print_option(AMReX_CUDA_DEBUG)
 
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
@maxpkatz  What does `--source-in-ptx` do?  The nvcc documentation on this is simply
```
Interleave source in PTX.
May only be used in conjunction with --device-debug or --generate-line-info.
```

maximumcats:
> What does --source-in-ptx do?

The PTX will be annotated with the lines of source it corresponds to. It uses the same underlying functionality as what `-lineinfo` does, so it doesn't add anything new, but it does allow you to directly inspect the generated PTX (if you use `-keep`) and see the annotations. There isn't really any reason for an end user to use it that I can think of, since I assume end users don't normally inspect PTX directly but rather would use Nsight Compute for understanding kernels.

WeiqunZhang (approved):

PR #2327, opened 2021-09-12
ParReduce

WeiqunZhang:
* Add ParReduce functions for MultiFab/FabArray and 1D iteration space.

* The implementation of ParReduce(MF) uses new ReduceOps::eval functions that launch one
  GPU kernel for the whole MultiFab/MultiFab.

* Use the new ParReduce functions in MultiFab/FabArray reduction functions.

## Additional background

The name `ParallelReduce` has been used for the namespace containing MPI reduce wrappers.  So we have to use a different name.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_MultiFab.cpp
+++ b/Src/Base/AMReX_MultiFab.cpp
@@ -43,24 +43,27 @@ MultiFab::Dot (const MultiFab& x, int xcomp,
 
     BL_PROFILE("MultiFab::Dot()");
 
-    Real sm = 0.0;
+    Real sm = Real(0.0);
 #ifdef AMREX_USE_GPU
     if (Gpu::inLaunchRegion()) {
-        sm = amrex::ReduceSum(x, y, nghost,
-        [=] AMREX_GPU_HOST_DEVICE (Box const& bx, Array4<Real const> const& xfab, Array4<Real const> const& yfab) -> Real
+        auto const& xma = x.const_arrays();
+        auto const& yma = y.const_arrays();
+        sm = ParReduce(TypeList<ReduceOpSum>{}, TypeList<Real>{}, x, IntVect(nghost),
+        [=] AMREX_GPU_DEVICE (int box_no, int i, int j, int k) noexcept -> GpuTuple<Real>
         {
-            Real t = 0.0;
-            AMREX_LOOP_4D(bx, numcomp, i, j, k, n,
-            {
+            Real t = Real(0.0);
+            auto const& xfab = xma[box_no];
+            auto const& yfab = yma[box_no];
+            for (int n = 0; n < numcomp; ++n) {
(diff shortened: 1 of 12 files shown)


Src/Base/AMReX_MultiFab.cpp
```diff
@@ -115,22 +143,49 @@ MultiFab::Dot (const iMultiFab& mask,
...
-            t += xfab(i,j,k,xcomp+n) * yfab(i,j,k,ycomp+n) * mi;
+            Real t = Real(0.0);
+            if (mma[box_no](i,j,k)) {
+                auto const& xfab = xma[box_no];
+                auto const& yfab = yma[box_no];
+                for (int n = 0; n < numcomp; ++n) {
+                    t += xfab(i,j,k,xcomp+n) * yfab(i,j,k,ycomp+n);
+                }
+            }
+            return t;
```
jmsexton03, line 163: This changes behavior if mskfab has integers other than 0 and 1.

Src/Base/AMReX_MultiFab.cpp
```diff
@@ -115,22 +143,49 @@ MultiFab::Dot (const iMultiFab& mask,
...
+        for (MFIter mfi(x,true); mfi.isValid(); ++mfi)
+        {
+            Box const& bx = mfi.growntilebox(nghost);
+            Array4<Real const> const& xfab = x.const_array(mfi);
+            Array4<Real const> const& yfab = y.const_array(mfi);
+            Array4<int const> const& mfab = mask.const_array(mfi);
+            AMREX_LOOP_4D(bx, numcomp, i, j, k, n,
+            {
+                if (mfab(i,j,k)) {
+                    sm += xfab(i,j,k,xcomp+n) * yfab(i,j,k,ycomp+n);
```
jmsexton03, line 180: This changes behavior if mskfab has integers other than 0 and 1.

jmsexton03 (approved):
Some suggestions (possibly for a future PR):

- Possibly add a sentence or two on ParReduce to documentation as well as doxygen if ParReduce is meant to be user-facing
- Update or Remove MultiFab::Dot example from documentation
- Possibly add Doxygen or documentation to clarify that masks will be treated as bools, not ints (or multipy the mask)

https://amrex-codes.github.io/amrex/docs_html/GPU.html#multifab-reductions

Src/Base/AMReX_MultiFab.cpp
WeiqunZhang: I don't think the behavior has changed.  In the old code, the int was converted to bool first and then to int.

jmsexton03: Thanks for clarifying, I see the `static_cast` nesting now

WeiqunZhang:
Yes, we definitely need to update the reduction documentation.

PR #2332, opened 2021-09-16
NodeLaplacian Updates

PaulMullowney:
In this branch, I make the number of smoothing steps a user defined parameter. Previously it was fixed at 4 for GPUs and 2 for CPUs. I'm not sure why GPU and CPU had different settings. I set the default to 2 for both.

I also modified the NodalPoisson test to 
1) use this new parameter
2) include additional parameters that control plot file dumping and the number of trials of the test to be run.

## Checklist

The proposed changes:

### DIFF
--- a/Src/LinearSolvers/MLMG/AMReX_MLNodeLaplacian.H
+++ b/Src/LinearSolvers/MLMG/AMReX_MLNodeLaplacian.H
@@ -82,6 +82,10 @@ public :
         if (m_const_sigma == Real(0.0)) m_coarsening_strategy = cs;
     }
 
+    void setSmoothNumSweeps (int nsweeps) noexcept {
+        m_smooth_num_sweeps = nsweeps;
+    }
+
     virtual BottomSolver getDefaultBottomSolver () const final override {
         return (m_coarsening_strategy == CoarseningStrategy::RAP) ?
             BottomSolver::bicgcg : BottomSolver::bicgstab;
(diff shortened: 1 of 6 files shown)


WeiqunZhang:
Can we change the default to 4 for gpu and 2 for cpu?  Because of race conditions in Gauss-Seidel smoothing, we found we had to do more sweeps for the solver to converge nicely.

PaulMullowney:
> Can we change the default to 4 for gpu and 2 for cpu? Because of race conditions in Gauss-Seidel smoothing, we found we had to do more sweeps for the solver to converge nicely.

So, I have seen results which show that while it takes more MLMG items to converge (with less smoothing), the overall run time is quite a bit faster. That's one of the reasons why I wanted to make this parameter tunable. I'm trying to do a more thorough sweep to see if this result holds up.

WeiqunZhang:
I cannot remember the detail.  But I believe we used 4 as default for GPU builds because some EB tests could not converge with just 2 sweeps.

PaulMullowney:
In AMReX_MLNodeLinOp.H, I initialize the variable to 4 (GPU) or 2 (CPU). I don't understand why the Intel build failed. This appears unrelated to my commit.

WeiqunZhang:
Please merge development into your branch.  Intel just release a new version of oneAPI that has deprecated some functions and we have Werror in the compiler flag.

WeiqunZhang (approved):

PR #2346, opened 2021-09-21
OverlapSync nowait and finish.

kngott:
## Summary

Adds OverlapSync_nowait and OverlapSync_finish functionality.

## Additional background

Because this is not a FabArray member function, it requires returning the temporary and passing it to `_finish`, along with the target FabArray. If the function was in FabArray.H, it could have a traditional blank `_finish` function signature.  Any opinions?

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_FabArrayUtility.H
+++ b/Src/Base/AMReX_FabArrayUtility.H
@@ -1377,7 +1377,7 @@ void
 OverrideSync_nowait (FabArray<FAB> & fa, FabArray<IFAB> const& msk, const Periodicity& period)
 {
     BL_PROFILE("OverrideSync_nowait()");
-    AMREX_ASSERT_WITH_MESSAGE(fa.os_temp, "OverrideSync_nowait() called when already in progress.");
+    AMREX_ASSERT_WITH_MESSAGE(!fa.os_temp, "OverrideSync_nowait() called when already in progress.");
 
     if (fa.ixType().cellCentered()) return;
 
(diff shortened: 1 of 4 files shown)


Src/Base/AMReX_FabArrayUtility.H
```diff
@@ -1397,12 +1413,31 @@ OverrideSync (FabArray<FAB> & fa, FabArray<IFAB> const& msk, const Periodicity&
...
+
+template <class FAB, class bar = std::enable_if_t<IsBaseFab<FAB>::value> >
+void
+OverrideSync_finish (FabArray<FAB> & fa, std::unique_ptr< FabArray<FAB> >& tmpmf)
+{
+    BL_PROFILE("OverrideSync_finish()");
+
+    // If !pcd, PC is done (all local), so skip finish.
+    if ( (fa.ixType().cellCentered()) && (tmpmf->pcd) )
+    {
```
WeiqunZhang:
```suggestion
    if (tmpmf)
    {
```

kngott:
Okay, new version:

To not have a handler, there's a new pointer in FabArray that stores a pointer to the temporary FabArray. Then, everything goes through the Utility implementation, either `fa.OverrideSync_finish()` or `amrex::OverrideSync_finish(fa)`. How's that?

I was thinking about re-using the PCData space to avoid adding more to FabArray, but that seems like a bad option, given PC & OSync is another potential overlap, the real MF is only a source here (not a dest, where PCData is stored now), and the checks for consistent nowait/finish use would have to become rather complex.

WeiqunZhang (approved):

PR #2347, opened 2021-09-21
Port CellConservativeProtected::protect() to C++ with GPU support

wyphan:
## Summary
CellConservativeProtected interpolater has been ported to C++ with GPU support.
For the CPU-only version on local, the errors are exactly 0.
For the GPU-enabled version on Gigan, the errors are in ~1e-11 range for IAMR 2D bubble test case, and ~1e-6 range for IAMR 3D Poiseuille test case.

## Checklist

The proposed changes:


WeiqunZhang:
There are `#ifndef BL_NO_FORT`s around `CellConservtiveProtected` class and functions that need to be removed.

wyphan:
@WeiqunZhang Yes, I changed that on purpose, since it has been ported to C++, and the original Fortran subroutine has been removed.

WeiqunZhang:
There are lots of warnings like below
```
../../../../Src/AmrCore/AMReX_Interpolater.cpp:730:2: warning: embedding a directive within macro arguments has undefined behavior [-Wembedded-directive]
#if (AMREX_SPACEDIM == 2)
 ^
```

This is because `#if (AMREX_SPACEDIM == 2)` is used inside `AMREX_HOST_DEVICE_PARALLEL_FOR_4D_FLAG`.  You can either split them into two calls (i.e., move `#if (AMREX_SPACEDIM == 2)` outside the `AMREX_HOST_DEVICE_PARALLEL_FOR_4D_FLAG` macro, or use the `AMREX_2D_ONLY` macro instead of `#if (AMREX_SPACE_DIM == 2)`.  I like the first approach better and I would put code block passed to `AMREX_HOST_DEVICE_PARALLEL_FOR_4D_FLAG` in separate functions.

WeiqunZhang:
> @WeiqunZhang Yes, I changed that on purpose, since it has been ported to C++, and the original Fortran subroutine has been removed.

I know you changed in one place, but not in other places.  That's one of the reasons why the CI tests are failing.

Src/AmrCore/AMReX_Interp_3D_C.H
```diff
@@ -381,6 +381,187 @@ face_linear_interp_z (int i, int j, int k, int n, Array4<T> const& fine,
     }
 }
 
+template <typename T>
+AMREX_GPU_DEVICE AMREX_FORCE_INLINE
```
WeiqunZhang: I think the reason the CIs fail is that these functions need to be `AMREX_GPU_HOST_DEVICE`.

wyphan: Ah, right. Thanks for catching this!

Src/AmrCore/AMReX_Interpolater.cpp
```diff
@@ -663,68 +665,85 @@ CellConservativeProtected::protect (const FArrayBox& crse,
...
-        fine_geom.GetEdgeVolCoord(fvc[dir],target_fine_region,dir);
-        crse_geom.GetEdgeVolCoord(cvc[dir],crse_bx,dir);
+    // Check coarse-to-fine ratios against rMAX
+#if (AMREX_SPACEDIM == 2)
+    const int rMAX = 32;
+#else
+    const int rMAX = 16;
+#endif
+    if ( amrex::max(AMREX_D_DECL(ratio[0], ratio[1], ratio[2])) > rMAX ) {
+        amrex::Abort("rMAX in CellConservativeProtected::protect");
```
WeiqunZhang: We can remove this check.  It was there to make sure the fixed size array in the Fortran routine were big enough.  But the Fortran routines have been removed now.

WeiqunZhang (commented):
Three kernels are launched here.  But I think they can be merged into one and there is no need to use a temporary `BaseFab<char>` to hold the `redo_me` flag.  In the original Fortran routine, everything is inside a single nested loops construct like below, and the iterations are independent of each other.
```
      do kc = cblo(3), cbhi(3)
      do jc = cblo(2), cbhi(2)
      do ic = cblo(1), cbhi(1)
          ! ...
      end do
      end do
      end do
```
This can be done with
```
   AMREX_HOST_DEVICE_PARALLEL_FOR_3D_FLAG(runon, crse_box, ic, jc, kc,
   {
   });
```

WeiqunZhang:
There are still two kernel launches.  I don't think there is any reason that they cannot be merged like this
```
AMREX_HOST_DEVICE_PARALLEL_FOR_3D_FLAG(runon, cs_bx, ic, jc, kc,
{
   // kernel 1
   for (int n ....) {
   }

   // kernel 2
});
```
Alternatively, we can turn the second kernel launch into
```
AMREX_HOST_DEVICE_PARALLEL_FOR_3D_FLAG(runon, target_fine_region, i, j, k,
{
});
```
And then there is no need for the `for (int k = klo; k <= khi; ++k)` loop etc. in `ccprotect_set_sync`.

I don't know which one is better, but both are better than the current approach.  The first one launches only one kernel.  The second kernel in the second approach has coalesced memory access.

Also, I noticed that the kernels are split into many small functions.  I would argue this is a bad style because the following code is repeated 3 times.
```
    // Calculate bounds for interpolation                                                                                                                                                                  
    Dim3 fnbxlo = lbound(fine_bx);
    Dim3 fnbxhi = ubound(fine_bx);
    int ilo = amrex::max(ratio[0]*ic,              fnbxlo.x);
    int ihi = amrex::min(ratio[0]*ic+(ratio[0]-1), fnbxhi.x);
    int jlo = amrex::max(ratio[1]*jc,              fnbxlo.y);
    int jhi = amrex::min(ratio[1]*jc+(ratio[1]-1), fnbxhi.y);
    int klo = amrex::max(ratio[2]*kc,              fnbxlo.z);
    int khi = amrex::min(ratio[2]*kc+(ratio[2]-1), fnbxhi.z);
```
I would merge them into one function and simply call them like this
```
AMREX_HOST_DEVICE_PARALLEL_FOR_3D_FLAG(runon, cs_bx, ic, jc, kc,
{
    the_new_merged_function(...);
});
```

Src/AmrCore/AMReX_Interp_2D_C.H
```diff
@@ -197,6 +198,258 @@ face_linear_interp_y (int i, int j, int /*k*/, int n, Array4<T> const& fine,
...
+                    } else {
+                        // RZ
+                        const Real* fn_problo = fn_geomdata.ProbLo();
+                        Real rp = fn_problo[0] + (i+0.5)*fn_dx[0];
+                        Real rm = fn_problo[0] + (i-0.5)*fn_dx[0];
+                        fvol = (rp*rp - rm*rm) * fn_dx[1];
+                    }
+
+                    // Calculate sums
+                    crseTot += fvol * fine(i,j,n);
```
WeiqunZhang:
```suggestion
                    crseTot += fvol * fine(i,j,0,n);
```

WeiqunZhang (approved):

PR #2354, opened 2021-09-23
Gpu Comments to Doxygen Docs

etpalmer63:
## Summary
Added some comments about `Gpu::setLaunchRegion`, `Gpu::synchronize()` and `Gpu::streamSynchronize()` from the new and previous Gpu Debugging write-up. Converted some comments to Doxygen comments. 

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_GpuControl.H
+++ b/Src/Base/AMReX_GpuControl.H
@@ -87,8 +87,8 @@ namespace Gpu {
     /**
      * Enable/disable GPU kernel launches.
      *
-     * \note This will only switch from GPU to CPU for kernel launch
-     * macros. Functions like `amrex::ParallelFor` will be unaffected.
+     * \note This will only switch from GPU to CPU for kernels launched
+     * with macros. Functions like `amrex::ParallelFor` will be unaffected.
      * Therefore it should not be used for comparing GPU to non-GPU performance
      * or behavior.
      *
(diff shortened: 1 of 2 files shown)


etpalmer63:
I added the comment about the `GPU::setLaunchRegion` because we decided to remove the information from the AMReX user's guide and I didn't want it to be lost. If we don't want to put it at this place in the code another option would be to add a designated page to Doxygen documentation to keep developer information.

Src/Base/AMReX_GpuControl.H
```diff
@@ -84,6 +84,24 @@ namespace Gpu {
...
+     *
+     * \code
+     *  Gpu::setLaunchRegion(0);
+     *
+     *  //...
+     *
+     *  Gpu::setLaunchRegion(1);
+     * \endcode
+     *
+     * Will disable the launching of GPU kernels between the calls.
```
WeiqunZhang, line 103: We want to be clear that it only switches to CPU for kernels launch with macros, but not functions like `amrex::ParallelFor`.

etpalmer63: The new description reads: 

This will only switch from GPU to CPU for kernels launched with macros. Functions like `amrex::ParallelFor` will be unaffected. Therefore it should not be used for comparing GPU to non-GPU performance or behavior.

WeiqunZhang (approved):

PR #2355, opened 2021-09-24
CMake CUDA Debug: -G by Default

ax3l:
## Summary

In the past, this often did not compile at all, was very sensitive to further set options, or compiled super slowly; in some cases, such as recursive function usage, apps need to increase `cudaLimitStackSize` in order to not stack overflow with device debug symbols (this costs some extra DRAM).

Nonetheless, for CUDA approx. 11.0+, we see the opposite: we have very slow Debug builds with CUDA if we do not activate -G for some `LinearSolvers/MLMG` objects: `AMReX_MLNodeLaplacian.cpp` & `AMReX_MLNodeLaplacian_sten.cpp`. Thus, we default-on now to -G in Debug builds.

- [x] testing: now some other TUs compile a bit longer (1-2min), but that's better than the other two 5min TUs above without `-G` in Debug

## Additional background

Thanks to @WeiqunZhang and @maxpkatz to looking into this with me.

**Note:** This means the new CMake `Debug` mode will disable GPU optimizations!
That's already the default in GNUmake, so we sync up.

## Checklist

The proposed changes:

### DIFF
--- a/Docs/sphinx_documentation/source/GPU.rst
+++ b/Docs/sphinx_documentation/source/GPU.rst
@@ -215,7 +215,7 @@ check the :ref:`table <tab:cmakecudavar>` below.
    +------------------------------+-------------------------------------------------+-------------+-----------------+
    | AMReX_CUDA_COMPILATION_TIMER |  CSV table with time for each compilation phase | NO          | YES, NO         |
    +------------------------------+-------------------------------------------------+-------------+-----------------+
-   | AMReX_CUDA_DEBUG             |  Device debug information (optimizations: off)  | NO          | YES, NO         |
+   | AMReX_CUDA_DEBUG             |  Device debug information (optimizations: off)  | YES: Debug  | YES, NO         |
    +------------------------------+-------------------------------------------------+-------------+-----------------+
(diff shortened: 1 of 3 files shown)


WeiqunZhang (approved):

ax3l:
```bash
make LOG_BUILD_TIME=TRUE USE_GPU=TRUE DEBUG=TRUE
```
Without `-G`:
```
('AMReX_MLNodeLaplacian_sten.cpp', 308.0)
('AMReX_MLNodeLaplacian.cpp', 304.0)
...
('AMReX_MLTensorOp.cpp', 184.0)
('PhysicalParticleContainer.cpp', 103.0)
('AMReX_MLNodeLaplacian_sync.cpp', 46.0)
...
('MultiParticleContainer.cpp', 126.0)
```
With `-G`:
```
('AMReX_MLNodeLaplacian.cpp', 59.0)
('AMReX_MLNodeLaplacian_sten.cpp', 41.0)
...
('AMReX_MLTensorOp.cpp', 129.0)
('PhysicalParticleContainer.cpp', 124.0)
('AMReX_MLNodeLaplacian_sync.cpp', 112.0)
```

PR #2369, opened 2021-09-30
Port CellQuadratic to C++ with GPU support

wyphan:
## Summary
The CellQuadratic interpolater has been ported from Fortran to C++ with GPU support.
The errors for the CPU-only version on local are in the ~1e-15 range for AmrLevel SingleVortex 2D, ~1e-15 for IAMR 2D bubble Cartesian, and ~1e-12 for IAMR 2D bubble RZ.
The errors for the GPU-enabled version on Gigan are in the ~1e-14 range for AmrLevel SingleVortex2D, ~1e-10 for IAMR 2D bubble Cartesian, and ~1e-9 for IAMR 2D bubble RZ.

## Additional background
This port is heavily influenced by the CellConservativeLinear interpolater. It reuses two of the same slope kernels.
With this PR, _all_ of the interpolaters now work on the GPU!
Also, thanks to @cgilet for helping with debugging.

## Checklist

The proposed changes:


Src/AmrCore/AMReX_Interpolater.cpp
```diff
@@ -481,68 +473,58 @@ CellQuadratic::interp (const FArrayBox& crse,
...
     //
     Box target_fine_region = fine_region & fine.box();
 
+    // Make Boxes for slopes.
     Box crse_bx(amrex::coarsen(target_fine_region,ratio));
-    Box fslope_bx(amrex::refine(crse_bx,ratio));
+    // Box fslope_bx(amrex::refine(crse_bx,ratio));
     Box cslope_bx(crse_bx);
     cslope_bx.grow(1);
     BL_ASSERT(crse.box().contains(cslope_bx));
```
WeiqunZhang: I saw your latest commit to avoid the out-of-bound error.  I would suggest you revert that commit and instead change the code here to be simply
```
Box cslope_bx = amrex::coarsen(target_fine_region,ratio);
```
without growing the Box.

wyphan: Will do!

WeiqunZhang (approved):

PR #2376, opened 2021-10-01
Fix bug in the particle communication routines when the number of chars in the message is > MAXINT.

atmyers:
I believe this bug was introduced in PR#1569.

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_ParticleCommunication.H
+++ b/Src/Particle/AMReX_ParticleCommunication.H
@@ -485,15 +485,14 @@ void communicateParticlesStart (const PC& pc, ParticleCopyPlan& plan, const Buff
     {
         if (i == MyProc) continue;
         const auto Who  = i;
-        Long nbytes     = plan.m_snd_num_particles[i]*psize;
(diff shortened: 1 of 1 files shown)


Src/Particle/AMReX_ParticleCommunication.H
```diff
@@ -467,7 +467,7 @@ void communicateParticlesStart (const PC& pc, ParticleCopyPlan& plan, const Buff
         const auto offset = rOffset[i];
         Long nbytes       = plan.m_rcv_num_particles[Who]*psize;
         std::size_t acd   = ParallelDescriptor::alignof_comm_data(nbytes);
```
WeiqunZhang, line 469: This may trigger an unused variable warning because later `acd` is only used in `AMREX_ASSERT`.

atmyers: I think it's okay because it's still used in the line below.

WeiqunZhang: Oh, yes.  How did I miss it?

WeiqunZhang (approved):

PR #2380, opened 2021-10-06
Add floor and ceil to parser

NeilZaim:
## Summary

I wanted to use the `floor` function in the parser for a WarpX automated test that I'm developing, so I've added `floor` and `ceil` in the parser. Hopefully I've done it right (I've tested compiling and using with WarpX on my computer and it looks like both functions are working well).

## Checklist

The proposed changes:


WeiqunZhang:
Could you update https://github.com/AMReX-Codes/amrex/blob/f0a8fc785ff798cac6bdfe4913ad0fa5db988dfe/Docs/sphinx_documentation/source/Basics.rst?

NeilZaim:
Thanks, I've just done that (also added that there is support for Bessel functions of the first kind).

WeiqunZhang (approved):

PR #2386, opened 2021-10-06
MLEBNodeFDLaplacian Coarsening

WeiqunZhang:
Limit the maximum coarsening level so that EB objects do not disappear due
to coarsening.  The solver would fail to converge if the coarsening causes
EB objects to disappear entirely.

## Checklist

The proposed changes:

### DIFF
--- a/Src/LinearSolvers/MLMG/AMReX_MLEBNodeFDLap_2D_K.H
+++ b/Src/LinearSolvers/MLMG/AMReX_MLEBNodeFDLap_2D_K.H
@@ -177,6 +177,20 @@ void mlebndfdlap_gsrb(int i, int j, int k, Array4<Real> const& x,
     }
 }
 
+AMREX_GPU_DEVICE AMREX_FORCE_INLINE
+int mlebndfdlap_check_coarsening (int i, int j, int, Array4<int const> const& mask) noexcept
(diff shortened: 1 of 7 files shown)


WeiqunZhang:
This solves the WarpX issue.

Src/LinearSolvers/MLMG/AMReX_MLEBNodeFDLap_2D_K.H
```diff
@@ -177,6 +177,20 @@ void mlebndfdlap_gsrb(int i, int j, int k, Array4<Real> const& x,
     }
 }
 
+AMREX_GPU_DEVICE AMREX_FORCE_INLINE
+int mlebndfdlap_check_coarsening (int i, int j, int, Array4<int const> const& mask) noexcept
+{
+    if ((i%2 == 0) || (j%2 == 0)) {
```
atmyers, line 183: Does this only work for ref_ratio = 2?

WeiqunZhang: That's multigrid coarsening ratio.  We always use 2.

atmyers (approved):

PR #2388, opened 2021-10-07
Make fcompare print usage and exit if only one plotfile is provided.

cgilet:
## Checklist

The proposed changes:

### DIFF
--- a/Tools/Plotfile/fcompare.cpp
+++ b/Tools/Plotfile/fcompare.cpp
@@ -75,7 +75,7 @@ int main_main()
         plotfile_b = amrex::get_command_argument(farg++);
     }
 
-    if (plotfile_a.empty() && plotfile_b.empty()) {
+    if (plotfile_a.empty() || plotfile_b.empty()) {
         amrex::Print()
             << "\n"
             << " Compare two plotfiles, zone by zone, to machine precision\n"
--- a/Tools/Plotfile/fcompare.cpp
+++ b/Tools/Plotfile/fcompare.cpp
@@ -15,6 +15,30 @@ struct ErrZone {
     IntVect cell;
 };
 
+void PrintUsage()
+{
+    amrex::Print()
+        << "\n"
+        << " Compare two plotfiles, zone by zone, to machine precision\n"
+        << " and report the maximum absolute and relative errors for each\n"
+        << " variable.\n"
+        << "\n"
+        << " usage:\n"
+        << "    fcompare [-n|--norm num] [-d|--diffvar var] [-z|--zone_info var] [-a|--allow_diff_grids] [-r|rel_tol] [--abs_tol] file1 file2\n"
+        << "\n"
+        << " optional arguments:\n"
+        << "    -n|--norm num         : what norm to use (default is 0 for inf norm)\n"
+        << "    -d|--diffvar var      : output a plotfile showing the differences for\n"
+        << "                            variable var\n"
+        << "    -z|--zone_info var    : output the information for a zone corresponding\n"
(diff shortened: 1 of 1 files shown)


wyphan:
Would you mind adding a check like this too, while you're at it?
```
if (plotfile_a == "-h" || plotfile_a == "--help")
```
Edit: #2387 partially enables this functionality. Nice!

cgilet:
@wyphan fcompare now parses "-h" and "--help"

Tools/Plotfile/fcompare.cpp
```diff
@@ -15,6 +15,30 @@ struct ErrZone {
...
+        << "    -n|--norm num         : what norm to use (default is 0 for inf norm)\n"
+        << "    -d|--diffvar var      : output a plotfile showing the differences for\n"
+        << "                            variable var\n"
+        << "    -z|--zone_info var    : output the information for a zone corresponding\n"
+        << "                            to the maximum error for the given variable\n"
+        << "    -a|--allow_diff_grids : allow different BoxArrays covering the same domain\n"
+        << "    -r|--rel_tol rtol     : relative tolerance (default is 0)\n"
+        << "    --abs_tol atol        : absolute tolerance (default is 0)\n"
+        << std::endl;
+    exit(EXIT_FAILURE);
```
WeiqunZhang: I don't think `PrintUsage` should call `exit` or return an error.

cgilet: Hmm, that's what all the C_util PrintUsage functions do. I just kinda assumed that was the norm. ... GNU standards say program should exit successfully 
https://www.gnu.org/prep/standards/html_node/_002d_002dhelp.html

cgilet: So maybe just change EXIT_FAILURE -> EXIT_SUCCESS ?

WeiqunZhang: I think `PrintUsage` should do what the function name suggests.  That is to print usage.  Otherwise it should be called `PrintUsageAndExit`.  As for the return code, I think `-h` and no plotfiles at all should return success, otherwise failure.

Tools/Plotfile/fcompare.cpp
```diff
@@ -42,7 +65,10 @@ int main_main()
     int farg = 1;
     while (farg <= narg) {
         const std::string fname = amrex::get_command_argument(farg);
-        if (fname == "--infile1") {
+        if (fname == "-h" || fname == "--help"){
+            PrintUsage();
+            exit(EXIT_SUCCESS);
```
WeiqunZhang:
```suggestion
            return EXIT_SUCCESS;
```

Tools/Plotfile/fcompare.cpp
```diff
@@ -75,27 +101,9 @@ int main_main()
...
-            << "    -z|--zone_info var    : output the information for a zone corresponding\n"
-            << "                            to the maximum error for the given variable\n"
-            << "    -a|--allow_diff_grids : allow different BoxArrays covering the same domain\n"
-            << "    -r|--rel_tol rtol     : relative tolerance (default is 0)\n"
-            << "    --abs_tol atol        : absolute tolerance (default is 0)\n"
-            << std::endl;
-        return 0;
+    if (plotfile_a.empty() || plotfile_b.empty()) {
+        PrintUsage();
+        exit(EXIT_FAILURE);
```
WeiqunZhang:
```suggestion
        return EXIT_FAILURE;
```

WeiqunZhang (approved):

PR #2391, opened 2021-10-08
Move info on sharing the command line into the ParmParse section.

cgilet:
## Summary
It had been residing in the initialization section.

## Checklist

The proposed changes:


etpalmer63 (commented):
Totally agree with this change! Thanks for catching it and doing the PR. While we're looking at it, I would like to make some edits to the wording. I am not sure the best way to go about. Maybe we can talk about it Monday morning. Thanks!

etpalmer63 (approved):
Looks great! Thanks for your efforts!

PR #2393, opened 2021-10-08
Default of the Arena's Release Threshold

WeiqunZhang:
Change the default of the Arena's release threshold from max (i.e., no
release) to the maximum of the initial allocation size and 3/4 of the GPU
memory.  This helps the memory oversubscription issue observed in some WarpX
runs.  Once we have to allocate additional device memory because the initial
allocation, which is 3/4 of the total global memory, has been used up, the
new additions to the Arena are not big enough for efficient reuse, resulting
in fragmentation in the Arena.  Because at that point we are very close to
use up all device memory, it's much more important to make sure we do not
oversubscribe due to memory fragmentation by releasing memory back to the
system than the performance gain we might have in memory allocation beyond
the initialize size.

### DIFF
--- a/Docs/sphinx_documentation/source/GPU.rst
+++ b/Docs/sphinx_documentation/source/GPU.rst
@@ -461,15 +461,17 @@ specific type of GPU memory:
 
 .. table:: Memory Arenas
 
-    +---------------------+----------------------------+
-    | Arena               |        Memory Type         |
-    +=====================+============================+
-    | The_Arena()         |  managed or device memory  |
-    +---------------------+----------------------------+
-    | The_Managed_Arena() |  managed memory            |
-    +---------------------+----------------------------+
-    | The_Pinned_Arena()  |  pinned memory             |
-    +---------------------+----------------------------+
+    +---------------------+---------------------------------------------------+
+    | Arena               |        Memory Type                                |
+    +=====================+===================================================+
+    | The_Arena()         |  managed or device memory                         |
+    +---------------------+---------------------------------------------------+
+    | The_Device_Arena()  |  device memory, could be an alias to The_Arena()  |
+    +---------------------+---------------------------------------------------+
+    | The_Managed_Arena() |  managed memory, could be an alias to The_Arena() |
+    +---------------------+---------------------------------------------------+
+    | The_Pinned_Arena()  |  pinned memory                                    |
+    +---------------------+---------------------------------------------------+
 
 .. raw:: latex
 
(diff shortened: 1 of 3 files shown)


jmsexton03:
Can this also be in the documentation/doxygen? I've been testing out setting a specific release threshold in Nyx for the_arena and the_pinned_arena, which has helped with oversubscription in Nyx. If the default is changed I'll be more careful about if I'm specifically setting this parameter.
Also, it might help if there was some sort of default release threshold for pinned as well (if that's in scope)

WeiqunZhang:
I will update the document and add @etpalmer63 as a reviewer.

WeiqunZhang:
> Also, it might help if there was some sort of default release threshold for pinned as well (if that's in scope)

The current default release threshold for the pinned arena is the size of the total global memory.

atmyers (approved):

Docs/sphinx_documentation/source/GPU.rst
```diff
@@ -497,6 +499,30 @@ gradually. The behavior of :cpp:`The_Managed_Arena()` likewise depends on the
...
+being used.  By default, the release threshold for :cpp:`The_Arena()` is set
+to be the maximum of initial allocation size and 3/4 of the total device
+memory, and it can be changed with a parameter,
+``amrex.the_arena_release_threshold``.  For :cpp:`The_Pinned_Arena()`, the
+default release threshold is the size of the total device memory, and the
+runtime parameter is ``amrex.the_pinned_arena_release_threshold``.  If it is
+a separate arena, :cpp:`The_Device_Area()` or :cpp:`The_Managed_Arena()`
+does not release memory by default, and this can be changed with
+``amrex.the_device_arena_release_threshold`` or
+``arrex.the_managed_arena_release_threashold``.
```
jmsexton03: arrex => amrex

Docs/sphinx_documentation/source/GPU.rst
```diff
@@ -497,6 +499,30 @@ gradually. The behavior of :cpp:`The_Managed_Arena()` likewise depends on the
 ``amrex.the_arena_is_managed=1``, :cpp:`The_Managed_Arena()` is simply aliased
 to :cpp:`The_Arena()` to reduce memory fragmentation.
 
+In :cpp:`amrex::Initialize`, a large amount of GPU device memory is
+allocated and is kept in :cpp:`The_Arena()`.  The default is 3/4 of the
+total device memory, and it can be changed with a :cpp:`ParmParse`
+parameter, ``amrex.the_arena_init_size``, in the unit of bytes.  The default
+initial size for other arenas are 8 MB.  For :cpp:`The_Managed_Arena()` and
```
jmsexton03: It might be helpful to give an example for what number for `amrex.the_pinned_arena_init_size=######`  would produce the same 8MB initial size, and potentially state the units for all the other flags listed here as well. Is there any easy way to put in the size requested in MB like the Arena usage prints?

drummerdoc: @jmsexton03 Or rename the variable to ``amex.the_pinned_arena_init_size_in_MB``

WeiqunZhang: I will update it to make it clear that the units are all in bytes.  I guess we could allow `*size = 1024 GB`, `MB` etc. and use bytes by default when there are no units to maintain backward compatibility.  Maybe will do it next time.

kngott: Do we have a ParmParse templated on size_t?

We could make that template read as a int + optional char, or parse a  string, look for `K`, `M`, `G`, `B`, etc. at the end and convert the number to bytes as part of the reading.

etpalmer63 (approved):
Looks good! I believe the information in the paragraph (502-526) might work as a table also. I'll give it a try on Monday, but no need to wait for me.

etpalmer63:
I have the tables on a local branch. I think they look good but I'm not sure of the best way to include them. When its convenient for you, can let me know the best way to put them up? Thanks.

WeiqunZhang:
I might cancel this PR, because testing shows it might be better to leave the threshold at maximum and strategically insert calls to `The_Arena()->freeUnused()` around regions with big memory allocations.  If I do, I will resubmit a PR for the documentation part of the change.

PR #2395, opened 2021-10-09
Support fmod in Parser

WeiqunZhang:
## Checklist

The proposed changes:


Docs/sphinx_documentation/source/Basics.rst
```diff
@@ -382,7 +382,7 @@ AMReX provides a parser in ``AMReX_Parser.H`` that can be used at runtime to eva
 expressions given in the form of string.  It supports ``+``, ``-``, ``*``,
 ``/``, ``**`` (power), ``^`` (power), ``sqrt``, ``exp``, ``log``, ``log10``,
 ``sin``, ``cos``, ``tan``, ``asin``, ``acos``, ``atan``, ``sinh``, ``cosh``,
-``tanh``, ``abs``, ``floor`` and ``ceil``.  The minimum and maximum of two
+``tanh``, ``abs``, ``floor``, ``ceil`` and ``fmod``.  The minimum and maximum of two
```
ax3l, line 385: Random thought: although we need to call `std::fmod` underneath, could we implement in the parser modulo also as `%` for our reals? Or is that at bad/confusing idea?

WeiqunZhang: Since python supports `%` for reals, we could consider using `%`.  However, python's `%` is different from `std::fmod` in the following way.
```
std::fmod(-5.1, 3.) => -2.1
(-5.1) % 3 => 0.9  in python

std::fmod(5.1, -3.) => 2.1
5. % (-3.) => -0.9 in python 
```
So it might be very confusing.

ax3l: ah, the conventions for negative modulo are always tricky.
We could invert the sign though in the parser to match :)

WeiqunZhang: If we search for `C++ fmod` on the web, the top results have very high quality.  If we search for `python %`, `python modulo` or `python remainder`, it's mostly noise.  (I guess python is too popular.)  So I think using `fmod` is better.

ax3l: Discussed on Zoom today: `fmod` is simpler to google and remember than the non-uniform definitions, even if we document our definition of `%`. Thus, we only go with `fmod` for now.

ax3l (approved):

PR #2397, opened 2021-10-12
Move Tools/C_util/Convergence/WritePlotFile to Src/Extern/amrdata.

cgilet:
This contains WritePlotFile functions that take an AmrData object
as a parameter, and thus Src/Extern/amrdata is the most logical
place to put it when codes in Tools/C_util/Convergence and
Tools/Postprocessing/C_src use it.

## Checklist

The proposed changes:


WeiqunZhang:
We need to put WritePlotFile functions into `namespace amrex`.

WeiqunZhang:
I don't think it compiles, because the function definitions are not in namespace amrex.  Could you also make `thePlotFileType()` in AMReX_WritePlorFile.cpp a static function.

WeiqunZhang (approved):

PR #2404, opened 2021-10-13
Fix dot product mask in EB nodal solver

WeiqunZhang:
We should set the dot product mask on all Dirichlet nodes to zero so that
they do not contribute to the sum used in fixing singular systems.

## Checklist

The proposed changes:


asalmgren (approved):

esclapez:
Sorry, I didn't see that earlier ! I'll test the fix, thanks !

WeiqunZhang:
Note that there is a separate issue of potentially coarsening too much.  So you may have to limit the number of coarsening if the domain is fully enclosed inside boundary before it's fixed.

esclapez:
Definitely ! But I tried a before/after this PR and it marginally helps MLMG when I'm not sending the problem to Hypre. Fully enclosed EB are still pretty tricky.

PR #2405, opened 2021-10-14
Add Filesystem::DirectorySeparator

ax3l:
## Summary

Add a public function to that returns the platform specific directory separator. We currently need this in WarpX and defining this in a central place makes sense for re-use.

It looks like plotfile writing works on Windows as of now through std filesystem operations, so no further changes are needed at this point unless we observe otherwise down the road.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_FileSystem.H
+++ b/Src/Base/AMReX_FileSystem.H
@@ -13,6 +13,20 @@ typedef unsigned short mode_t;
 namespace amrex {
 namespace FileSystem {
 
+/** The Platform-Specific Directory Separator
+ *
+ * Return a backslash on Windows and a slash on all other platforms.
+ */
+static constexpr char
+DirectorySeparator ()
+{
+#ifdef _WIN32
+    return '\\';
+#else
+    return '/';
+#endif
+}
+
 bool
 CreateDirectories (std::string const& filename, mode_t mode, bool verbose = false);
 
--- a/Src/Base/AMReX_FileSystem.cpp
+++ b/Src/Base/AMReX_FileSystem.cpp
@@ -82,7 +82,7 @@ CreateDirectories (std::string const& path, mode_t mode, bool verbose)
     bool retVal(false);
     Vector<std::pair<std::string, int> > pathError;
 
-    const char* path_sep_str = "/";
+    const char* path_sep_str = DirectorySeparator();
 
     if (path.length() == 0 || path == path_sep_str) {
         return true;
--- a/Src/Base/AMReX_FileSystem.cpp
+++ b/Src/Base/AMReX_FileSystem.cpp
@@ -82,7 +82,7 @@ CreateDirectories (std::string const& path, mode_t mode, bool verbose)
     bool retVal(false);
     Vector<std::pair<std::string, int> > pathError;
 
-    const char* path_sep_str = DirectorySeparator();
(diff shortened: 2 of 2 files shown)


Src/Base/AMReX_FileSystem.cpp
```diff
@@ -82,7 +82,7 @@ CreateDirectories (std::string const& path, mode_t mode, bool verbose)
     bool retVal(false);
     Vector<std::pair<std::string, int> > pathError;
 
-    const char* path_sep_str = "/";
+    const char* path_sep_str = DirectorySeparator();
 
     if (path.length() == 0 || path == path_sep_str) {
```
ax3l: I hope I don't break calls into `CreateDirectories` if I use the updated `path_sep_str` here?
Otherwise we can also leave this place as-is for now.

Src/Base/AMReX_FileSystem.H
```diff
@@ -13,6 +13,20 @@ typedef unsigned short mode_t;
 namespace amrex {
 namespace FileSystem {
 
+/** The Platform-Specific Directory Separator
+ *
+ * Return a backslash on Windows and a slash on all other platforms.
+ */
+static constexpr char
```
ax3l, line 20: `std::string` can only be `constexpr` in C++20, thus I just return a `char` for now.
https://stackoverflow.com/a/43335753/2719194

WeiqunZhang:
Do you really need it?  `/` is a portable directory separator even on Windows, if one uses `filesystem`.  https://en.cppreference.com/w/cpp/filesystem/path/path  You can also convert it with https://en.cppreference.com/w/cpp/filesystem/path/make_preferred.

ax3l:
Yeah, that's the reason it works so far, I guess.

But `std::filesystem` is C++17 and needs often extra linking. I need this in https://github.com/ECP-WarpX/WarpX/pull/2424 to silence a warning.

Src/Base/AMReX_FileSystem.cpp
```diff
@@ -82,7 +82,7 @@ CreateDirectories (std::string const& path, mode_t mode, bool verbose)
     bool retVal(false);
     Vector<std::pair<std::string, int> > pathError;
 
-    const char* path_sep_str = "/";
+    const char* path_sep_str = DirectorySeparator();
```
ax3l: Not needed and I just want the new universal function for general use.
```suggestion
    const char* path_sep_str = "/";
```

WeiqunZhang:
But we have this in `AMReX_Config.cmake` to force the std to be 17.
```
   # minimum: C++14 on Linux, C++17 on Windows, C++17 for dpc++ and hip                                                                                                                                    
   if (AMReX_DPCPP OR AMReX_HIP)
      target_compile_features(amrex PUBLIC cxx_std_17)
   else ()
      target_compile_features(amrex PUBLIC $<IF:$<STREQUAL:$<PLATFORM_ID>,Windows>,cxx_std_17,cxx_std_14>)
   endif ()
```

And in `AMReX_FileSystem.cpp`
```
#if defined(_WIN32) // || __cplusplus >= 201703L                                                                                                                                                           
#include <filesystem>
```
So it won't build without C++17.

ax3l:
Oh, I forgot that we already require C++17 on Windows, true.

PR #2416, opened 2021-10-19
Update MLNodeTensorLap for WarpX

WeiqunZhang:
Change the stencil of the linear operator so that L=D*G.  This could make
the solution satisfy div E = rho up to solver tolerance.  Previously, the
stencil was similar to that of the nodal solver for approximate projection

## Checklist

The proposed changes:


WeiqunZhang:
Ready for review.  But don't merge until we verify with WarpX.

WeiqunZhang:
@RemiLehe

WeiqunZhang:
It works as expected for WarpX.

atmyers:
This replaces the old stencil, right? Is it not worth keeping that one around?

WeiqunZhang:
I don't think we need to keep the old one.  The operator is very WarpX specific.

atmyers (approved):

PR #2417, opened 2021-10-19
Fix two IO overflow bugs uncovered by NyX.

atmyers:
This fixes two IO bugs that have to do with integer overflow. 

1. In AMReX_WriteBinaryParticleData.H, the number of reals numbers written to disk can overflow an `int` even if the number of particles on a box doesn't. This fixes this behavior. 

2. The AMReX_FabConv.cpp, the conversion functions take a `Long` specifying the number of objects to convert. However, this number is inappropriately cast to int before checking its size. This PR fixes this as well. 

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_FabConv.cpp
+++ b/Src/Base/AMReX_FabConv.cpp
@@ -992,7 +992,7 @@ RealDescriptor::convertToNativeFormat (Real*                 out,
 
     while (nitems > 0)
     {
-        Long get = nitems > Long(readBufferSize) ? Long(readBufferSize) : nitems;
+        int get = nitems > readBufferSize ? readBufferSize : int(nitems);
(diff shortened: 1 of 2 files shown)


Src/Base/AMReX_FabConv.cpp
```diff
@@ -992,7 +992,7 @@ RealDescriptor::convertToNativeFormat (Real*                 out,
 
     while (nitems > 0)
     {
-        int get = int(nitems) > readBufferSize ? readBufferSize : int(nitems);
+        int get = Long(nitems) > readBufferSize ? readBufferSize : int(nitems);
```
atmyers: Should `get` be a `Long` as well?

jmsexton03 (approved):

PR #2430, opened 2021-10-22
Cleanup of amrdata

WeiqunZhang:
* Avoid using std and amrex in amrdata headers.

* Fix some warnings.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Extern/amrdata/AMReX_XYPlotDataList.H
+++ b/Src/Extern/amrdata/AMReX_XYPlotDataList.H
@@ -14,7 +14,6 @@
 #include <list>
 #include <string>
 
-//using namespace amrex;
 
 namespace amrex {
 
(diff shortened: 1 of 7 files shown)


Src/Extern/amrdata/AMReX_XYPlotDataList.H
```diff
@@ -13,10 +13,8 @@
 
 #include <list>
 #include <string>
-using std::string;
-using std::list;
 
-using namespace amrex;
+//using namespace amrex;
```
atmyers: Remove rather than comment out?

atmyers (approved):

Src/Extern/amrdata/AMReX_XYPlotDataList.H
WeiqunZhang:
```suggestion
```

PR #2432, opened 2021-10-22
New HDF5 schema with separate datasets for different variables

houjun:
## Summary

- Add a new function WriteMultiLevelPlotfileHDF52 to write plot files with separate HDF5 datasets for different variables. The new format cannot be read by Visit directly like the previous format. The function also allows passing compression parameters and apply them on the datasets. See Tests/HDF5Benchmark for the example usage.
- Relocate HDF5 related files to Src/Extern/HDF5

## Checklist

The proposed changes:


Src/Extern/HDF5/AMReX_PlotFileUtilHDF5.H
```diff
@@ -0,0 +1,83 @@
...
+                                      const Vector<int> &level_steps,
+                                      const Vector<IntVect> &ref_ratio,
+                                      const std::string &compression = "None#0",
+                                      const std::string &versionName = "HyperCLaw-V1.1",
+                                      const std::string &levelPrefix = "Level_",
+                                      const std::string &mfPrefix = "Cell",
+                                      const Vector<std::string>& extra_dirs = Vector<std::string>());
+
+
+    void WriteSingleLevelPlotfileHDF52 (const std::string &plotfilename,
```
atmyers: Could you give these version `2` functions more descriptive names?

houjun: How about WriteSingleLevelPlotfileHDF5SingleDset and WriteSingleLevelPlotfileHDF5MultiDset?

atmyers: Sounds good to me.

atmyers (approved):

atmyers:
Thanks for the new features and the re-organization!

Src/Extern/HDF5/AMReX_WriteBinaryParticleDataHDF5.H
```diff
@@ -0,0 +1,459 @@
...
+    ret  = H5Aread(attr, dtype, data);
+    if (ret < 0) {
+        amrex::Print() << " Error with H5Aread [" << name << "]\n";
+        return -1;
+    }
+    H5Aclose(attr);
+    return 1;
+}
+
+#ifdef BL_USE_MPI
```
WeiqunZhang, line 79: We should add `#include <AMReX_Config.H>` at the beginning of this header, so that `BL_USE_MPI` is always defined.

houjun: This file is only included in Src/Particle/AMReX_WriteBinaryParticleData.H which has #include <AMReX_Config.H>

PR #2433, opened 2021-10-25
Advection_AmrLevel Test: Convert Prob.f90 to Prob.cpp

etpalmer63:
## Summary
Converts the only Fortran file in the AmrLevel test to cpp. 

## Checklist

The proposed changes:


Tests/Amr/Advection_AmrLevel/CMakeLists.txt
```diff
@@ -29,7 +29,8 @@ list(TRANSFORM _sources PREPEND Source/)
 ###############################################################################
 set(_sv_exe_dir Exec/SingleVortex/)
 
-set(_sv_sources face_velocity_${AMReX_SPACEDIM}d_K.H Prob_Parm.H Adv_prob.cpp Prob.f90)
+#set(_sv_sources face_velocity_${AMReX_SPACEDIM}d_K.H Prob_Parm.H Adv_prob.cpp Prob.f90)
```
WeiqunZhang: Can we remove the commented line here and the one below for UniformVelocity?

Tests/Amr/Advection_AmrLevel/Exec/SingleVortex/Prob.H
```diff
@@ -0,0 +1,8 @@
+#ifndef _Prob_H_
+#define _Prob_H_
```
WeiqunZhang: Can we use `PROB_H_` instead of `_Prob_H_`?  Symbols start with `_` followed by a capital letter is reserved by the compiler.  Do that to UniformVortex too.

Tests/Amr/Advection_AmrLevel/Exec/SingleVortex/Prob.H
```diff
@@ -0,0 +1,8 @@
...
+
+using namespace amrex;
+
+void initdata(MultiFab& S_tmp, const Geometry& geom);
```
WeiqunZhang: Can we remove `using namespace amrex;` in headers? UnifromVortex too.

Tests/Amr/Advection_AmrLevel/Exec/SingleVortex/Prob.cpp
```diff
@@ -0,0 +1,47 @@
...
+    for (MFIter mfi(S_tmp); mfi.isValid(); ++mfi)
+    {
+        const Box& box = mfi.validbox();
+        const Array4<Real>& phi = S_tmp.array(mfi);
+
+        // Populate MultiFab data
+        ParallelFor(box, [=] AMREX_GPU_DEVICE ( int i, int j, int k) noexcept
+        {
+
+            Real x = prob_lo[0] + (i + 0.5) * dx[0];
```
WeiqunZhang: Can we use `0.5_rt` or `Real(0.5) instead of `0.5` so that they become single precision when `Real` is single precision?

WeiqunZhang (commented):
Could we remove `Adv_F.H`?

Could we remove `OR NOT CMAKE_Fortran_COMPILER_LOADED` from `CMakeLists.txt`?

Coule we add `BL_NO_FORT = TRUE` to `amrex/Tests/Amr/Advection_AmrLevel/Exec/Make.Adv`, just like `amrex/Tests/Amr/Advection_AmrCore/Exec/Make.Adv`?

WeiqunZhang:
AmrLevel is not supported on Windows.  So we should skip the test for Windows builds.

Tests/Amr/Advection_AmrLevel/CMakeLists.txt
```diff
@@ -1,8 +1,15 @@
-if ( (AMReX_SPACEDIM EQUAL 1) OR NOT CMAKE_Fortran_COMPILER_LOADED )
-   return()
+if (AMReX_SPACEDIM EQUAL 1)
+        message([WARNING] "Advection_AmrLevel Tests require AMReX_SPACEDIM = 2 or 3." )
```
WeiqunZhang: I don't think we should print a warning message here.  Users will see the warning if they do `-DAMReX_ENABLE_TESTS:BOOL=ON -DAMReX_SPACEDIM=1` when building amrex.  The message seems to imply there is something wrong, but that's not the case.

etpalmer63: I see your point here. Maybe we can change it to `[NOTICE]`, change the message to something else, or both? I was actually thinking the opposite way: What if the user wants to run the AmrLevel test with the same `-DAMReX_ENABLE_TESTS:BOOL=ON -DAMReX_SPACEDIM=1` ? Then the executable just doesn't show up and there is no explanation. My concern was about addressing this situation.

WeiqunZhang: If you are worried about that, it can be explained in the README file in that test.

etpalmer63: Got it. I removed the warning and put a short explanation in the README file.

WeiqunZhang:
LGTM

etpalmer63:
Sorry forgot to click the "Ready" button. @WeiqunZhang, thanks for all your help!

WeiqunZhang (approved):

WeiqunZhang:
Realized there is an issue after merging.

PR #2435, opened 2021-10-26
Update MPI config checking to include Intel MPI wrappers for DPCPP

jmsexton03:
## Summary

For Make.unknown, if USE_DPCC=TRUE and mpiicpc is found with icpc as the compiler, use Intel MPI wrappers. Otherwise use mpicxx, etc

## Checklist

The proposed changes:

### DIFF
--- a/Tools/GNUMake/Make.defs
+++ b/Tools/GNUMake/Make.defs
@@ -1016,10 +1016,6 @@ FINAL_LIBS = $(libraries)
 
 ifeq ($(USE_DPCPP),TRUE)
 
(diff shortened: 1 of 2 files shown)


WeiqunZhang:
Can I push some changes to your branch?

jmsexton03:
> Can I push some changes to your branch?

Sure

WeiqunZhang:
I made some changes so that we do not see error message of `mpiicpc -show` when it does not exist.

WeiqunZhang (approved):

PR #2442, opened 2021-10-27
Mark NeighborParticles as a 'selfTest'

atmyers:
The proposed changes:

### DIFF
--- a/Tools/RegressionTesting/AMReX-cuda-tests.ini
+++ b/Tools/RegressionTesting/AMReX-cuda-tests.ini
@@ -361,7 +361,6 @@ useOMP = 0
 compileTest = 0
 selfTest = 1
 doComparison = 0
-diffDir = neighbor_particles.0
 doVis = 0
 testSrcTree = C_Src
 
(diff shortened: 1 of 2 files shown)


jmsexton03 (approved):

Tools/RegressionTesting/AMReX-cuda-tests.ini
```diff
@@ -359,6 +359,7 @@ useMPI = 1
 numprocs = 1
 useOMP = 0
 compileTest = 0
+selfTest = 1
 doComparison = 0
 diffDir = neighbor_particles.0
```
WeiqunZhang: Do we still need `diffDir`?

atmyers:
```suggestion
```

Tools/RegressionTesting/AMReX-hip-tests.ini
```diff
@@ -357,6 +357,7 @@ useMPI = 0
 numprocs = 1
 useOMP = 0
 compileTest = 0
+selfTest = 1
 doComparison = 0
 diffDir = neighbor_particles.0
```
atmyers:
```suggestion
```

Tools/RegressionTesting/AMReX-cuda-tests.ini
atmyers: I think it will be ignored, but no we don't need it any more.

atmyers: I think it will be ignored, but no we don't need it any more.

PR #2443, opened 2021-10-28
Fix use-after-free

WeiqunZhang:
The parallel copy meta-data cache is store in std::multimap<BDKey,CPC*>.  A
CPC* might be stored multiple times in the multimap, and we need to delete
the pointers before clearing the multimap.  The previous way of avoiding
double free was incorrect, because it had a use-after-free.  In this commit,
we fix it by pushing a unique copy of the pointers to a vector first while
iterating the multimap.

This bug only manifests if a user's code has a memory leak of
MultiFab/FabArray.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_FabArrayBase.cpp
+++ b/Src/Base/AMReX_FabArrayBase.cpp
@@ -555,12 +555,16 @@ FabArrayBase::flushCPC (bool no_assertion) const
 void
 FabArrayBase::flushCPCache ()
 {
+    std::vector<CPC*> cpcs;
     for (CPCacheIter it = m_TheCPCache.begin(); it != m_TheCPCache.end(); ++it)
     {
         if (it->first == it->second->m_srcbdk) {
             m_CPC_stats.recordErase(it->second->m_nuse);
-            delete it->second;
+            cpcs.push_back(it->second);
         }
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
ROCm 4.5 was just released.  The CI tests are failing now.
```
2021-10-28T07:04:22.2999712Z CMake Error at Tools/CMake/AMReXParallelBackends.cmake:251 (find_package):
2021-10-28T07:04:22.3001192Z   Could not find a package configuration file provided by "rocrand" with any
2021-10-28T07:04:22.3002047Z   of the following names:
2021-10-28T07:04:22.3002502Z 
2021-10-28T07:04:22.3003083Z     rocrandConfig.cmake
2021-10-28T07:04:22.3004435Z     rocrand-config.cmake
```
This might be temporary because it looks like some rocm packages installed in CI are still in 4.4.  I am going to make the HIP tests non-required for now.

@ax3l

atmyers (approved):

ax3l:
I just saw this in WarpX as well. Looks like they either did an inconsistent rollout or changed the name of the rocrand package?

ax3l:
Fix in #2445

PR #2444, opened 2021-10-28
WIP : Implementation of MLPoisson::Fsmooth using ParallelFor(MultiFab)

PaulMullowney:
I had to write wrapper implementations that execute the loops over the box, when compiling for CPU. Also, the compactify method does not compile for GPU. So, the 3D impl with hasHiddenDimension is executed in the old manner. Lastly, the failing HIP tests seem unrelated to this PR.

## Checklist

The proposed changes:


WeiqunZhang:
This is done in #3071 now.

PR #2448, opened 2021-10-28
AMREX_RELEASE_NUMBER

WeiqunZhang:
Add a new macro `AMREX_RELEASE_NUMBER`.  It is a six digits number.  The
first two are for the year, and the next two are for the month.  The last
two are usually 00, unless a patch other than the monthly release is
released.  If the AMReX source is a git version that's not the release
version, the last release number will be used.

For git version obtained with git --describe, we define something like

    #define AMREX_GIT_VERSION "21.10-65-gfbb2edc95551-dirty"
    #define AMREX_RELEASE_NUMBER 211000

For git version obtained from the CHANGES file, we define something like

    #define AMREX_GIT_VERSION "21.10"
    #define AMREX_RELEASE_NUMBER 211000

or

    #define AMREX_GIT_VERSION "21.10.1"
    #define AMREX_RELEASE_NUMBER 2110001

or

    #define AMREX_GIT_VERSION "21.10.11"
    #define AMREX_RELEASE_NUMBER 211011

## Checklist

The proposed changes:


ax3l:
Thx!

Please don't forget to add testing for sources that users obtained via:
- tarballs or
- shallow git clones

those usually give unexpected `git` results that can quickly break the logic for such macros.

WeiqunZhang:
Thanks for the comment.  I think tarball works.  The shallow clone probably does not.

WeiqunZhang:
The CMake build does work with shallow clone.  But the gnu make does not.  Need to fix it.

Tools/CMake/AMReXUtils.cmake
```diff
@@ -48,6 +52,27 @@ function (get_amrex_version)
    set( AMREX_PKG_VERSION "${_pkg_version}" CACHE INTERNAL "" )
    unset(_pkg_version)
 
+   if (AMREX_GIT_VERSION)
```
ax3l, line 55: For this path, it might be good to check that this does not error out if the `AMREX_GIT_VERSION` returns `YY.MM` without any `-`.

Escaping looks good to me, but maybe some functions need `if` guards.
`string(FIND` will probably return `-1` in that case (https://cmake.org/cmake/help/latest/command/string.html#find), but `string(SUBSTRING` might count from the back (https://cmake.org/cmake/help/latest/command/string.html#substring), etc.

ax3l: On the other hand, I see that we use the same logic already above? ^^

ax3l (commented):
The CMake part looks good to me besides one more corner-case that we might want to check in both GNUmake and CMake (parsing a newly released AMReX `git` version without any prefixes)?

ax3l (approved):

WeiqunZhang:
Do you mean something like this
```
$ git describe --abbrev=12 --dirty --always --tags
21.12
```
That works.

PR #2450, opened 2021-10-29
Set to never write fill values to the dataset for HDF5 output

brtnfld:
## Summary

When creating datasets and groups with a single rank, it is good practice to not write fill values.

On Summit, with 2688 ranks for:

ncells=2688 max_grid_size=8 ncomp=6 nlevs=1 nppc=1 nplotfile=5 nparticlefile=2

The total program exec time went from 5m3 to 3m56.

The proposed changes:


atmyers:
Thank you for this PR! This change sounds good to me - @houjun, would you mind having a look and letting me know what you think?

houjun:
Yes, this is good.

atmyers (approved):

PR #2421, opened 2021-10-21
Update WarpX EB solver

WeiqunZhang:
Change the coarsening strategy for much better performance.

## Checklist

The proposed changes:
- [x] fix a bug or incorrect behavior in AMReX
- [ ] add new capabilities to AMReX
- [ ] changes answers in the test suite to more than roundoff level
- [ ] are likely to significantly affect the results of downstream AMReX users
- [ ] include documentation in the code and/or rst files, if appropriate

PR #2453, opened 2021-11-01
Fix WarpX EB solver's grad phi computation

WeiqunZhang:
## Summary

Must set the Dirichlet mask on covered nodes to a negative value, because
the function that computes the gradient relies on the information to detect
EB.  This is a bug introduced recently in #2421.

## Additional background

https://github.com/ECP-WarpX/WarpX/issues/2512

## Checklist

The proposed changes:


ax3l (approved):
Thank you! :sparkles:

PR #2459, opened 2021-11-03
DPCPP: AMREX_INTEL_ARCH

WeiqunZhang:
When AOT is on, set the target to AMREX_INTEL_ARCH, which is required if the
SYCL backend is used and AOT is on.

## Checklist

The proposed changes:

### DIFF
--- a/Docs/sphinx_documentation/source/GPU.rst
+++ b/Docs/sphinx_documentation/source/GPU.rst
@@ -319,7 +319,7 @@ we provide the helper function ``setup_target_for_cuda_compilation()``:
 
 
 
-Enabling HIP support
+Enabling HIP Support
 ^^^^^^^^^^^^^^^^^^^^
 
 To build AMReX with HIP support in CMake, add
@@ -350,7 +350,7 @@ Below is an example configuration for HIP on Tulip:
    cmake --build build -j 6
 
 
-Enabling SYCL support
+Enabling SYCL Support
 ^^^^^^^^^^^^^^^^^^^^^
 
 To build AMReX with SYCL support in CMake, add
--- a/Docs/sphinx_documentation/source/GPU.rst
+++ b/Docs/sphinx_documentation/source/GPU.rst
@@ -319,8 +319,8 @@ we provide the helper function ``setup_target_for_cuda_compilation()``:
 
 
 
-Enabling HIP support (experimental)
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Enabling HIP support
+^^^^^^^^^^^^^^^^^^^^
 
 To build AMReX with HIP support in CMake, add
 ``-DAMReX_GPU_BACKEND=HIP -DAMReX_AMD_ARCH=<target-arch> -DCMAKE_CXX_COMPILER=<your-hip-compiler>``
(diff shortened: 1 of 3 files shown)


Tools/CMake/AMReXOptions.cmake
```diff
@@ -157,6 +157,20 @@ cmake_dependent_option( AMReX_DPCPP_ONEDPL "Enable DPCPP's oneDPL algorithms"  O
    "AMReX_GPU_BACKEND STREQUAL SYCL" OFF)
 print_option(  AMReX_DPCPP_ONEDPL )
 
+if (AMReX_DPCPP)
+   set(AMReX_INTEL_ARCH_DEFAULT "IGNORE")
```
ax3l, line 161: Do you think we could use a default here?

Tools/CMake/AMReXOptions.cmake
```diff
@@ -157,6 +157,20 @@ cmake_dependent_option( AMReX_DPCPP_ONEDPL "Enable DPCPP's oneDPL algorithms"  O
...
+   if (DEFINED ENV{AMREX_INTEL_ARCH})
+      set(AMReX_INTEL_ARCH_DEFAULT "$ENV{AMREX_INTEL_ARCH}")
+   endif()
+
+   set(AMReX_INTEL_ARCH ${AMReX_INTEL_ARCH_DEFAULT} CACHE STRING
```
ax3l, line 166: Let's document `AMReX_INTEL_ARCH` in `Docs/sphinx_documentation/source/GPU.rst`, so users find it :)

ax3l (commented):
Implementation looks good, just the manual docs are missing.

We could as a follow-up also implement a logic to try to find the local architecture as a heuristic, if that is potentially of any help?

Tools/CMake/AMReXOptions.cmake
WeiqunZhang: We have something in GNU Make that uses `cat /sys/devices/cpu/caps/pmu_name` to determine the intel device type.  But that is not a reliable way for the new hardware.  There is also concern about having the new hardware name in the public repo.  We can revisit this later.  Let's leave this as is for now.

Tools/CMake/AMReXOptions.cmake
WeiqunZhang: I will do it.

Tools/CMake/AMReXOptions.cmake
ax3l: Agreed.
When we approach this later on, we could try to keep the heuristic in a similar order as for CUDA.

Docs/sphinx_documentation/source/GPU.rst
```diff
@@ -350,8 +350,8 @@ Below is an example configuration for HIP on Tulip:
    cmake --build build -j 6
 
 
-Enabling SYCL support (experimental)
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Enabling SYCL support
```
etpalmer63: I think "support" should be capitalized as "Support". This way the titles remain consistent.

Docs/sphinx_documentation/source/GPU.rst
```diff
@@ -319,8 +319,8 @@ we provide the helper function ``setup_target_for_cuda_compilation()``:
 
 
 
-Enabling HIP support (experimental)
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Enabling HIP support
```
etpalmer63: I think "support" should be capitalized as "Support". This way the titles remain consistent.

etpalmer63:
Sorry for the delay on looking this over. It got buried in my inbox by GitHub spam.

Docs/sphinx_documentation/source/GPU.rst
WeiqunZhang:
```suggestion
Enabling HIP Support
```

Docs/sphinx_documentation/source/GPU.rst
WeiqunZhang:
```suggestion
Enabling SYCL Support
```

ax3l (approved):

PR #2473, opened 2021-11-11
Fix multi-level EB nodal solver solvability fix

WeiqunZhang:
For the multi-level EB nodal solver with a singular matrix, we restrict the
fine level RHS to the coarse level, and fix the coarse level's solvability
by making the total sum of RHS is zero.  The restriction is done with the
stencil instead of simply geometric restriction because we need to take the
EB into account.  However, the previous way is incorrect if the sigma
coefficient is not a constant, because the stencil includes sigma.  But the
restriction should be based on EB information only.  In this commit, we fix
this by constructing a special constant sigma stencil.  We use this new
stencil for the restriction of RHS, and for consistence the residual as
well.  Because we used to use the original stencil for the restriction of
residual, this commit will change the answer at the tolerance level for
multi-level problems.

## Checklist

The proposed changes:


asalmgren (approved):

esclapez (approved):
Solved my unsolvable problems. Thanks Weiqun !

PR #2474, opened 2021-11-12
Fix Advection_AmrLevel test

WeiqunZhang:
There is no need to use pinned memory in initData anymore after #2433.

## Checklist

The proposed changes:


etpalmer63 (approved):
:+1:

PR #2489, opened 2021-11-19
Fix a corner case in fixing small cells.

WeiqunZhang:
During the iteration of fixing small cells, the edge types and level sets
could become inconsistent if the fixed small cells are too close to periodic
boundaries.  The function that updates the edge types did not take that into
account.  This commit fixes that.

## Checklist

The proposed changes:


etpalmer63 (approved):
Looks great! Thank you for showing me. I had never heard of a quiet nan before.

PR #2495, opened 2021-11-24
iMultiFab I/O

WeiqunZhang:
Add functions for writing and reading iMultiFab.

## Checklist

The proposed changes:


Src/Base/AMReX_VisMF.H
```diff
@@ -380,6 +365,380 @@ std::ostream& operator<< (std::ostream& os, const VisMF::Header& hd);
...
+                }
+            }
+        }
+        ofs << fod;
+    }
+}
+
+namespace detail {
+template <typename FAB>
+void read_fab(FAB& fab, VisMF::FabOnDisk const& fod, std::string const& name)
```
atmyers:
```suggestion
void read_fab (FAB& fab, VisMF::FabOnDisk const& fod, std::string const& name)
```

Src/Base/AMReX_VisMF.H
```diff
@@ -380,6 +365,380 @@ std::ostream& operator<< (std::ostream& os, const VisMF::Header& hd);
...
+    int ncomp;
+    IntVect ngrow;
+    Vector<VisMF::FabOnDisk> fod;
+    {
+        std::string header_file_name = name + "_H";
+        Vector<char> header_file_chars;
+        ParallelDescriptor::ReadAndBcastFile(header_file_name, header_file_chars);
+        std::string header_file_string(header_file_chars.data());
+        std::stringstream ifs(header_file_string, std::istringstream::in);
+
```
atmyers, line 572: Should we check the status of `ifs` here (and elsewhere where we read / write from streams in this PR?

WeiqunZhang: It seems that we don't usually check the state of stringstream.  I guess it was assumed that stringstream is unlikely to fail unlike fstream.  Yes, it does not hurt to check.

WeiqunZhang: In this case, there is an assertion later on the version string read from the stream.  So I guess it would fail the assertion if the stringstream is not good.

atmyers (approved):

PR #2502, opened 2021-11-29
Fix HIP CI

atmyers:
It's not correctly being passed the "-std=c++17" flag.

The proposed changes:


WeiqunZhang:
I think we should add `-std=c++17` to `CXXFLAGS`.
```
env: {CXXFLAGS: "-std=c++17 ...
```

atmyers:
I think I prefer this fix as its consistent with the other HIP test. It looks like it works now, with the `-std=c++17` being passed correctly.

WeiqunZhang (approved):

PR #2505, opened 2021-11-30
Update NVector_MultiFab functionality from ATPESC-codes and add to Src/Extern/SUNDIALS

jmsexton03:
## Summary

Adds NVector_MultiFab from https://github.com/AMReX-Codes/ATPESC-codes
Adds a check for `SUNDIALS_FOUND` in the CMake setup of SUNDIALS as a ThirdPartyLibrary

## Additional background

Listed authors from the original code are:

  Authors (alphabetical):
    David Gardner (gardner48@llnl.gov)
    John Loffeld (loffeld1@llnl.gov)
    Daniel Reynolds (reynolds@smu.edu)
    Donald Willcox (dewillcox@lbl.gov)

## Checklist

The proposed changes:


Src/Extern/SUNDIALS/AMReX_NVector_MultiFab.H
```diff
@@ -0,0 +1,107 @@
...
+    John Loffeld (loffeld1@llnl.gov)
+    Daniel Reynolds (reynolds@smu.edu)
+    Donald Willcox (dewillcox@lbl.gov)
+
+  --------------------------------------------------------------------
+  Header file for N_Vector wrap of AMReX 'MultiFab' structure.
+  --------------------------------------------------------------------*/
+
+#ifndef NVECTOR_MULTIFAB_H
+#define NVECTOR_MULTIFAB_H
```
WeiqunZhang:
```suggestion
#ifndef AMREX_NVECTOR_MULTIFAB_H
#define AMREX_NVECTOR_MULTIFAB_H
```

Src/Extern/SUNDIALS/AMReX_NVector_MultiFab.H
```diff
@@ -0,0 +1,107 @@
...
+
+#include <stdio.h>
+#include <sundials/sundials_nvector.h>
```
WeiqunZhang: Could we move these after the amrex headers?  Could we include `cstdio` instead and move it after the sundials header?

Src/Extern/SUNDIALS/AMReX_NVector_MultiFab.H
```diff
@@ -0,0 +1,107 @@
...
+  amrex::MultiFab *mf;   /* wrapped MultiFab        */
+};
+
+typedef struct _N_VectorContent_MultiFab *N_VectorContent_MultiFab;
+
+/* -----------------------------------------------------------------
+ * Accessor Macros
+ * -----------------------------------------------------------------*/
+
+#define NV_CONTENT_M(v) ( (N_VectorContent_MultiFab)(v->content) )
```
WeiqunZhang: We need to add `AMREX_` prefix to these macros.

Src/Extern/SUNDIALS/AMReX_NVector_MultiFab.H
```diff
@@ -0,0 +1,107 @@
...
+#endif
+
+namespace amrex {
+namespace sundials {
+
+/* -----------------------------------------------------------------
+ * Vector content structure
+ * -----------------------------------------------------------------*/
+
+struct _N_VectorContent_MultiFab {
```
WeiqunZhang: We should in principle avoid names starting with `_` followed by a capital letter.  It's reserved by the C++ standard.

Src/Extern/SUNDIALS/AMReX_NVector_MultiFab.cpp
```diff
@@ -0,0 +1,810 @@
...
+    John Loffeld (loffeld1@llnl.gov)
+    Daniel Reynolds (reynolds@smu.edu)
+    Donald Willcox (dewillcox@lbl.gov)
+
+  --------------------------------------------------------------------
+  Implementation file for N_Vector wrap of AMReX 'MultiFab' structure.
+  --------------------------------------------------------------------*/
+
+#include <stdio.h>
+#include <stdlib.h>
```
WeiqunZhang:
```suggestion
#include <cstdio>
#include <cstdlib>
```

Src/Extern/SUNDIALS/AMReX_NVector_MultiFab.cpp
```diff
@@ -0,0 +1,810 @@
...
+      Array4<Real> const& x_fab = mf_x->array(mfi);
+      Array4<Real> const& z_fab = mf_z->array(mfi);
+      const auto lo = lbound(bx);
+      const auto hi = ubound(bx);
+
+      for (int c = 0; c < ncomp; ++c) {
+         for (int k = lo.z; k <= hi.z; ++k) {
+            for (int j = lo.y; j <= hi.y; ++j) {
+               for (int i = lo.x; i <= hi.x; ++i) {
+                  z_fab(i,j,k,c) = SUNRabs(x_fab(i,j,k,c));
```
WeiqunZhang: Why do we explicitly compute this on the CPU?  It seems that we should do it on GPU?  We should also update other functions for GPU.

jmsexton03: It looks like there's a FabArray function that should be used instead for the Abs routine:
https://github.com/AMReX-Codes/amrex/blob/development/Src/Base/AMReX_FabArrayUtility.H#L1288-L1294

For things like the weighted root mean square norm, I was looking at copying the GPU strategy used in `norminfmask` by changing the norm to multiply the weight instead of doing the max, and then having the squares and sqrt and dividing at the end by number of elements.
https://github.com/AMReX-Codes/amrex/blob/development/Src/Base/AMReX_BaseFab.H#L2271-L2310

For the more specialized routines, does it make a substantial difference if I use ParallelFor vs converting the entire MFIter loop to use ParallelForMF (given these should be large box sizes)?

Src/Extern/SUNDIALS/AMReX_NVector_MultiFab.H
jmsexton03: Added AMREX_ prefix, and then removed macros

Src/Extern/SUNDIALS/AMReX_NVector_MultiFab.H
jmsexton03: Renamed struct to avoid `_` followed by a capital letter.

Src/Extern/SUNDIALS/AMReX_NVector_MultiFab.cpp
jmsexton03: Changed cpu-only routines to use `ParallelFor`, `amrex::ReduceMin`, `amrex::ParReduce` etc.

Src/Base/AMReX_MultiFabUtil.H
```diff
@@ -329,6 +329,137 @@ void average_down_nodal (const FabArray<FAB>& fine, FabArray<FAB>& crse,
     }
 }
 
+   /**
+    * \brief Returns part of a norm based on two MultiFabs
+    * The MultiFabs MUST have the same underlying BoxArray.
+    * The Predicate pf is used to test the mask
```
WeiqunZhang: This one has no predicate pf.  But we can probably remove this version because we should be able to call the other one with a pf that is known to the compiler to return true.

WeiqunZhang: Never mind my comment.  The predicate requires a mask multifab.  But the comment still needs to be fixed.

WeiqunZhang (approved):

PR #2506, opened 2021-11-30
Extend EB signed distance reach

esclapez:
Add an optional argument to the FillSignedDistance functions to control
how far away from the EB the distance is built. Default value is 1.0 and higher values incur increased computations.

## Checklist

The proposed changes:


WeiqunZhang:
The issue is the EB information data need to have enough ghost cells, otherwise the distance to EB could be wrong because it won't see anything that is not in the valid+ghost region.

esclapez:
@WeiqunZhang I see what you mean. I've found a workaround by propagating the LS front manually afterward, so I'll close this PR.

PR #2509, opened 2021-12-03
ANSI Escape Codes

WeiqunZhang:
Add commonly used ANSI escape codes to AMReX_ANSIEscCode.H.  They can be
used to print colored output.  For example,

    amrex::Print() << Font::Bold << FGColor::Red << "Hello world from AMReX\n"
                   << ResetDisplay << Font::Faint << FGColor::BrightCyan
                   << Font::Underline << "Hello world again from AMReX "
                   << ResetDisplay << BGColor::White << FGColor::Black << amrex::Version()
                   << ResetDisplay << std::endl;

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_ANSIEscCode.H
+++ b/Src/Base/AMReX_ANSIEscCode.H
@@ -1,0 +1,63 @@
+#ifndef AMREX_ANSI_ESC_CODE_H_
+#define AMREX_ANSI_ESC_CODE_H_
+
+// https://en.wikipedia.org/wiki/ANSI_escape_code
+
+namespace amrex {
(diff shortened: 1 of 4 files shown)


zingale (approved):

cgilet:
Question @WeiqunZhang : what happens if I don't do the intermediate `ResetDisplay` and, say, just try to change the font color with `FGColor::BrightCyan`?

zingale:
it will persist with that color setting until you change it next.  So if you want to do a block of output in a color, you can change it once, do all of the `std::cout`, and then reset.

PR #2515, opened 2021-12-05
Fix missing namespace

jmsexton03:
## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_TimeIntegrator.H
+++ b/Src/Base/AMReX_TimeIntegrator.H
@@ -78,10 +78,10 @@ public:
         switch (integrator_type)
         {
             case IntegratorTypes::ForwardEuler:
-                integrator_ptr = std::make_unique<FEIntegrator<T> >(S_data);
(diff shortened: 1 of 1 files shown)


Src/Base/AMReX_TimeIntegrator.H
```diff
@@ -78,10 +78,10 @@ public:
         switch (integrator_type)
         {
             case IntegratorTypes::ForwardEuler:
-                integrator_ptr = std::make_unique<FEIntegrator<T> >(S_data);
+                integrator_ptr = std::make_unique<amrex::FEIntegrator<T> >(S_data);
```
WeiqunZhang, line 81: Is this really needed?  This function is already inside `namespace amrex {}`.

jmsexton03: You're right, I wasn't using `make realclean` so I was getting the wrong / old header file. Closing PR

PR #2523, opened 2021-12-09
Add ParallelCopyToGhost and fix a bug in FillPatch

WeiqunZhang:
* Add ParallelCopyToGhost function that copies data to ghost cells only.

* There was a bug in the Array version of FillPatch for Face data.  The
  issue is, if the destination is the same FabArray as one of the source
  FabArrays on the fine level, some valid data on valid box faces are
  overwritten with the interpolated values.  This is now fixed with the use
  of ParallelCopyToGhost in that case.

## Checklist

The proposed changes:

### DIFF
--- a/Src/AmrCore/AMReX_FillPatchUtil_I.H
+++ b/Src/AmrCore/AMReX_FillPatchUtil_I.H
@@ -573,7 +573,17 @@ namespace {
 
                 for (int d=0; d<AMREX_SPACEDIM; ++d)
                 {
-                    mf[d]->ParallelCopy(mf_refined_patch[d], 0, dcomp, ncomp, IntVect{0}, nghost);
+                    bool aliasing = false;
+                    for (auto const& fmf_a : fmf) {
+                        aliasing = aliasing || (mf[d] == fmf_a[d]);
+                    }
+                    if (aliasing) {
+                        mf[d]->ParallelCopyToGhost(mf_refined_patch[d], 0, dcomp, ncomp,
+                                                   IntVect{0}, nghost);
+                    } else {
+                        mf[d]->ParallelCopy(mf_refined_patch[d], 0, dcomp, ncomp,
+                                            IntVect{0}, nghost);
+                    }
                 }
             }
         }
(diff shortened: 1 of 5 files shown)


WeiqunZhang:
@cgilet Could you try this to see if it fixes the issue you have?

cgilet:
@weiqun This fixes the issue for me. Thanks

cgilet (approved):

Src/AmrCore/AMReX_FillPatchUtil_I.H
```diff
@@ -573,15 +573,25 @@ namespace {
 
                 for (int d=0; d<AMREX_SPACEDIM; ++d)
                 {
-                    mf[d]->ParallelCopy(mf_refined_patch[d], 0, dcomp, ncomp, IntVect{0}, nghost);
+                    bool aliasing = false;
+                    for (auto const& fmf_a : fmf) {
+                        aliasing = aliasing || (mf[d] == fmf_a[d]);
+                    }
```
kngott, line 579: This is directly comparing the pointers, correct? Is that sufficient? That this will likely only trigger when the function is called with explicitly identical pointers? Any other cases we want to catch?

Also, do we expect cases where `mf[0] != fmf[0] && mf[0] == fmf[1]`? The nested AMREX_SPACEDIM loops feel off and like they might confuse things if we take a look later. (This beckons back to our discussion about what `ASSERT`s are appropriate for this function.)

Src/Base/AMReX_FabArrayCommI.H
```diff
@@ -260,6 +260,23 @@ FabArray<FAB>::ParallelCopy (const FabArray<FAB>& src,
     ParallelCopy_finish();
 }
 
+template <class FAB>
+void
+FabArray<FAB>::ParallelCopyToGhost (const FabArray<FAB>& src,
```
kngott, line 265: _nowait and _finish versions?

Src/AmrCore/AMReX_FillPatchUtil_I.H
WeiqunZhang: This is sufficient unless people use alias MultiFabs in `mf` or `fmf`.  In that case, we will have to compare the data pointers to the data.  We could do that, but we could argue that's very unlikely to happen.

I think it's possible `mf[0] != fmf[0] && mf[0] == fmf[1]`.

Src/Base/AMReX_FabArrayCommI.H
WeiqunZhang: Sure.  We could.

Src/AmrCore/AMReX_FillPatchUtil_I.H
kngott: I think this is ready to go if we've decided that we aren't concerned with aliases at this point. I'll approve and you can decide. 😃

kngott (approved):

PR #2525, opened 2021-12-09
CMake 3.22+: Policy CMP0127

ax3l:
## Summary

Fix a warning with CMake 3.22+.

~~We use use sometimes ;-connected AND syntax in `cmake_dependent_option`, so we need to update some of those options once we require CMake 3.22+ only.~~
We use simple syntax in `cmake_dependent_option`, so we are compatible with the extended syntax in CMake 3.22+
  https://cmake.org/cmake/help/v3.22/policy/CMP0127.html

## Additional background

First seen on Perlmutter.

## Checklist

The proposed changes:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,13 +8,12 @@ if(POLICY CMP0091)
 endif()
 
 #
-# We use use sometimes ;-connected AND syntax in cmake_dependent_option,
-# so we need to update some of those options once we require CMake 3.22+
-# only
+# We use simple syntax in cmake_dependent_option, so we are compatible with the
+# extended syntax in CMake 3.22+
 # https://cmake.org/cmake/help/v3.22/policy/CMP0127.html
 #
 if(POLICY CMP0127)
-    cmake_policy(SET CMP0127 OLD)
+    cmake_policy(SET CMP0127 NEW)
 endif()
 
 #
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,6 +8,16 @@ if(POLICY CMP0091)
 endif()
 
 #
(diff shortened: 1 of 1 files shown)


ax3l:
@etpalmer63 we can probably also set this to `NEW` and update some of the `cmake_dependent_option` calls for CMake <3.22 already.

Since actually, CMake 3.22+ also still supports the `;` syntax:
https://cmake.org/cmake/help/v3.22/module/CMakeDependentOption.html#command:cmake_dependent_option

So we could check if everything still works with CMake 3.22 if we set this to `NEW`... :)
- Done :)

CMakeLists.txt
```diff
@@ -7,6 +7,16 @@ if(POLICY CMP0091)
   cmake_policy(SET CMP0091 NEW)
 endif()
 
+#
+# We use use sometimes ;-connected AND syntax in cmake_dependent_option,
```
ax3l: Actually, the `;` syntax is still supported in 3.22+ as well:
https://cmake.org/cmake/help/v3.22/module/CMakeDependentOption.html#command:cmake_dependent_option

etpalmer63:
I see the ; lines as you said.  

For my own understanding, is it that the conflict comes from lines like `Tools/CMake/AMReXOptions.cmake` line 65, which has:
```
cmake_dependent_option(CUDA_FORWARD_UNKNOWN_FLAGS_HOST
    "Forward unknown NVCC flags to the host compiler" ON
    "CMAKE_CUDA_COMPILER_VERSION VERSION_GREATER_EQUAL 10.2.89;CMAKE_VERSION VERSION_LESS 3.17" OFF)

```
 Otherwise, it looks like a good fix. Thanks for tagging me on it! :smile:

WeiqunZhang (approved):

PR #2528, opened 2021-12-10
Add another overload for MakeSimilarDM.

atmyers:
This one works with an input `ba` and `dm` instead of a `MultiFab.`

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_DistributionMapping.cpp
+++ b/Src/Base/AMReX_DistributionMapping.cpp
@@ -1888,15 +1888,20 @@ DistributionMapping::writeOn (std::ostream& os) const
 
 DistributionMapping MakeSimilarDM (const BoxArray& ba, const MultiFab& mf, const IntVect& ng)
 {
-    Vector<int> pmap(ba.size());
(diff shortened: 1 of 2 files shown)


Src/Base/AMReX_DistributionMapping.cpp
```diff
@@ -1888,15 +1888,20 @@ DistributionMapping::writeOn (std::ostream& os) const
...
 {
-    Vector<int> pmap(ba.size());
     const DistributionMapping& mf_dm = mf.DistributionMap();
     const BoxArray& mf_ba = convert(mf.boxArray(),ba.ixType());
+    return MakeSimilarDM(ba, mf_ba, mf_dm, ng);
+}
 
+DistributionMapping MakeSimilarDM (const BoxArray& ba, const BoxArray& src_ba,
+                                   const DistributionMapping& src_dm, const IntVect& ng)
+{
```
WeiqunZhang, line 1898: Maybe we can an assertion here.  If the two boxarrays have different types, the intersections will fail at assertion.  But it's might be clear to catch it here.

WeiqunZhang (approved):

PR #2535, opened 2021-12-13
Fix order of processing of `add_par` function and parsing inputs file.

etpalmer63:
## Summary
Fix order of processing `add_par` function and inputs file, so that the
`add_par` function is evaluated first. This allows the variables, 
if specified in the inputs file, to be overwriten. 

## Additional background
The behavior of the code now matches the description in the docs.

## Checklist

The proposed changes:


etpalmer63:
Rather than change the code, I will change the docs to properly describe the current behavior.

PR #2537, opened 2021-12-14
fix WritePlotfileToASCII compilation

ajnonaka:
## Summary
fix an issue with compilation of WritePlotfileToASCII on certain systems

## Checklist

The proposed changes:
- [ x ] fix a bug or incorrect behavior in AMReX

### DIFF
--- a/Tools/Postprocessing/C_Src/Make.package
+++ b/Tools/Postprocessing/C_Src/Make.package
@@ -1,6 +1,7 @@
 CEXE_sources += ${EBASE}.cpp
 
 ifneq ($(EBASE), particle_compare)
+ifneq ($(EBASE), WritePlotfileToASCII)
 
   INCLUDE_LOCATIONS += $(AMREX_HOME)/Src/Base
   include $(AMREX_HOME)/Src/Base/Make.package
@@ -13,4 +14,6 @@ ifneq ($(EBASE), particle_compare)
   ifeq ($(NEEDS_f90_SRC),TRUE)
     f90EXE_sources += ${EBASE}_nd.f90
   endif
+
+endif
 endif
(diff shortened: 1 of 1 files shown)


Tools/Postprocessing/C_Src/Make.package
```diff
@@ -1,6 +1,7 @@
 CEXE_sources += ${EBASE}.cpp
 
 ifneq ($(EBASE), particle_compare)
+ifneq ($(EBASE), WritePlotfileToASCII)
```
jmsexton03, line 5: What about just selecting the ones with `_nd.f90` files currently in development, instead of excluding executables?
```suggestion
ifeq ($(EBASE), PtwisePltTransform)
ifeq ($(EBASE), PlotfileToTurb)
```

Tools/Postprocessing/C_Src/Make.package
```diff
@@ -1,6 +1,7 @@
 CEXE_sources += ${EBASE}.cpp
 
-ifneq ($(EBASE), particle_compare)
+ifeq ($(EBASE), PtwisePltTransform)
+ifeq ($(EBASE), PlotfileToTurb)
```
WeiqunZhang: No EBASE will pass both tests.  So it's always false.

ajnonaka: Doh!  Guess we need the previous logic then of nested ifneq's?

jmsexton03: yeah, it looks like some but not all of the other EBASE options do need amrdata, which is the main contribution of that part of the loop (like `PlotfileToMatLab`). My original suggestion was wrong for that reason too.

ajnonaka: What's pushed at the moment works with all EBASE options.

WeiqunZhang (approved):

PR #2539, opened 2021-12-18
Update FillPatchTwoLevels for single component face-centered data

cgilet:
## Summary

Allow the non-array version of FillPatchTwoLevels to operate on face-centered data and use the valid fine values on the coarse-fine interface during the interpolation to ghost cells. 

This also changes the FaceLinear Interpolater to use the valid fine data on the coarse-fine interface in the interpolation to ghost cells. 

## Checklist

The proposed changes:


Src/AmrCore/AMReX_Interpolater.cpp
```diff
@@ -190,29 +250,76 @@ void FaceLinear::interp_arr (Array<FArrayBox*, AMREX_SPACEDIM> const& crse,
...
+                  AMREX_LOOP_3D(bx2, i, j, k,
+                  {
+                      for (int n=0; n<ncomp; ++n)
+                      {
+                          face_linear_face_interp(i,j,k,n,2,fine_arr[2],crse_arr[2],mask_arr[2],ratio);
+                      }
+                  });
+              });
+
+    // FIXME? We need data from the previous lambda in this one. Do we need a sync here? They're on different boxes...
```
WeiqunZhang: No, we do not need a sync here, because they are on the same stream.  Box sizes are irrelevant here. 
```suggestion
```

Src/AmrCore/AMReX_Interpolater.H
```diff
@@ -57,6 +57,39 @@ public:
...
+    * \param runon
+    */
+    virtual void interp_face (const FArrayBox& /*crse*/,
+                              const int        /*crse_comp*/,
+                              FArrayBox&       /*fine*/,
+                              const int        /*fine_comp*/,
+                              const int        /*ncomp*/,
+                              const Box&       /*fine_region*/,
+                              const IntVect&   /*ratio*/,
+                              IArrayBox* const& /*solve_mask*/,
```
WeiqunZhang: Let's change this (and overrides) from `IArrayBox* const&` to `IArratBox const*`.   The problem of `IArrayBox* const& solve_mask` is it actually allows for `solve_mask->setVal(0)`.  So it's the wrong kind of `const.

Src/AmrCore/AMReX_Interp_C.H
```diff
@@ -10,4 +10,122 @@
...
+namespace amrex {
+
+//
+// Fill fine values with piecewise-constant interpolation of coarse data.
+// Operate only on faces that overlap--ie, only fill the fine faces that make up each
+// coarse face, leave the in-between faces alone.
+//
+template<typename T>
+AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE void
+face_linear_face_interp (int ci, int cj, int ck, int n, int idir, Array4<T> const& fine,
```
WeiqunZhang: We could split this into 3 functions for performance.

Src/AmrCore/AMReX_Interpolater.H
cgilet: Maybe a better option would be `const IArrayBox&` ? For this single component case, `solve_mask` doesn't need to be a pointer at all.

WeiqunZhang: Yes, agreed.

WeiqunZhang (approved):

PR #2541, opened 2021-12-20
add a tool to compute the volume integral of a plotfile field

zingale:
## Checklist

The proposed changes:

### DIFF
--- a/Tools/Plotfile/fvolumesum.cpp
+++ b/Tools/Plotfile/fvolumesum.cpp
@@ -57,6 +57,8 @@ void main_main()
 
     if (varname_arg.empty()) {
         var_name = "density";
+    } else {
+        var_name = varname_arg;
     }
 
     // make sure that variable name is valid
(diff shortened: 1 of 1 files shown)


Tools/Plotfile/fvolumesum.cpp
```diff
@@ -0,0 +1,200 @@
...
+                                        // Cartesian
+                                        for (int idim = 0; idim < AMREX_SPACEDIM; ++idim) {
+                                            vol *= dx[idim];
+                                        }
+                                    } else if (coord == 1) {
+                                        // axisymmetric V = pi (r_r**2 - r_l**2) * dz
+                                        //                = pi dr * dz * (r_r + r_l)
+                                        //                = 2 pi r dr dz
+                                        vol = 2 * M_PI * p[0] * dx[0] * dx[1];
+                                    }
```
WeiqunZhang, line 179: There is no `coord == 2` here.  But it is supported for `lev < fine_level`.

zingale: good catch.  Fixed.

WeiqunZhang (approved):

philip-blakely:
Hello; I've just tried using the fvolumesum tool for checking conservation in our AMReX-based code.

Very useful, but I've found a small error: you need to initialise Real vol=1; (in two places).
Otherwise the final integral is undefined (often zero) in the Cartesian case.
Also, fvolumesum ideally needs to be added to the GNUMakefile.

zingale:
you are right -- I had implemented it and tested it for axisymmetric, but the vol initialization for Cartesian was wrong.  That is fixed in #2596

PR #2548, opened 2021-12-23
Resize vector inside GatherLayoutDataToVector

kngott:
## Summary

The output vector of GatherLayoutDataToVector doesn't resize the resulting vector; the user has to pass a minimally `nrank` sized vector themselves. This PR resizes the result vector inside the function.

The only potential issue is if someone has been passing in something larger than `nranks`, attempting to use it for some type of concatenation with whatever is hanging over the end of the vector. But, I doubt we want that to be an actual use case for this function. If I'm wrong, we can make the resize conditional.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_ParallelDescriptor.H
+++ b/Src/Base/AMReX_ParallelDescriptor.H
@@ -971,6 +971,7 @@ ParallelDescriptor::GatherLayoutDataToVector (const LayoutData<T>& sendbuf,
 
     int nprocs = ParallelContext::NProcsSub();
     Vector<int> recvcount(nprocs, 0);
+    recvbuf.resize(nprocs);
     const Vector<int>& old_pmap = sendbuf.DistributionMap().ProcessorMap();
     for (int i=0; i<old_pmap.size(); ++i)
     {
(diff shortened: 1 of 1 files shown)


Src/Base/AMReX_ParallelDescriptor.H
```diff
@@ -971,6 +971,7 @@ ParallelDescriptor::GatherLayoutDataToVector (const LayoutData<T>& sendbuf,
 
     int nprocs = ParallelContext::NProcsSub();
     Vector<int> recvcount(nprocs, 0);
+    recvbuf.resize(nprocs);
```
WeiqunZhang: The size is wrong.  I think it should be
```suggestion
    recvbuf.resize(sendbuf.size());
```

WeiqunZhang (changes requested):
For consistence, we should change the non-MPI version too.

Src/Base/AMReX_ParallelDescriptor.H
kngott: Ah yes, number of boxes, not number of ranks. I've been mixing that up a lot lately with this graph stuff.

Fixed.

WeiqunZhang:
> For consistence, we should change the non-MPI version too.

What about this comment?

kngott:
Huh.. that comment was tucked away non-obviously in the initial email. I didn't know that's how that worked/looks. Good to know. :)

WeiqunZhang (approved):

PR #2549, opened 2021-12-23
Add Gatherv for USE_MPI=FALSE.

kngott:
## Summary

This adds a Gatherv implementation for when USE_MPI=FALSE.

## Additional background

This is causing a runtime error in the GitHub checks for my graph object _only_ in the simple processing tools, like fcompare. Not sure why isn't not throwing an error in other USE_MPI=FALSE tests, and I'm really curious about it, but adding in the USE_MPI=FALSE implementation for completion can't hurt.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_ParallelDescriptor.H
+++ b/Src/Base/AMReX_ParallelDescriptor.H
@@ -1169,6 +1169,17 @@ Gather(const T& t, int /*root*/)
 
 template <class T>
 void
+Gatherv (const T* send, int sc,
+         T* recv, const std::vector<int>& /*rc*/,
+         const std::vector<int>& /*disp*/, int /*root*/)
+{
+    std::vector<T> resl(sc);
+    for (int j=0; j<sc; ++j) { recv[j] = send[j]; }
+    return resl;
+}
+
+template <class T>
+void
 GatherLayoutDataToVector (const LayoutData<T>& sendbuf,
                           Vector<T>& recvbuf, int /*root*/)
 {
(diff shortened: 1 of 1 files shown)


Src/Base/AMReX_ParallelDescriptor.H
```diff
@@ -1167,6 +1167,17 @@ Gather(const T& t, int /*root*/)
...
 
+template <class T>
+void
+Gatherv (const T* send, int sc,
+         T* recv, const std::vector<int>& /*rc*/,
+         const std::vector<int>& /*disp*/, int /*root*/)
+{
+    std::vector<T> resl(sc);
+    for (int j=0; j<sc; ++j) { recv[j] = send[j]; }
+    return resl;
```
WeiqunZhang: This function's return type is `void`, not `std::vector`.

WeiqunZhang (changes requested):

Src/Base/AMReX_ParallelDescriptor.H
kngott: Oops. Too many small PRs. Went too fast. :)
Fixed.

Src/Base/AMReX_ParallelDescriptor.H
```diff
@@ -1167,6 +1167,16 @@ Gather(const T& t, int /*root*/)
     return resl;
 }
 
+template <class T>
+void
+Gatherv (const T* send, int sc,
+         T* recv, const std::vector<int>& /*rc*/,
+         const std::vector<int>& /*disp*/, int /*root*/)
+{
+    std::vector<T> resl(sc);
```
WeiqunZhang:
```suggestion
```

WeiqunZhang (approved):

PR #2550, opened 2021-12-23
Embed more EB data in EBFArrayBox

WeiqunZhang:
Before this, we only had access to EBCellFlag given an EBFArrayBox.  This
adds access to more EB data via EBFArrayBox.  This could be useful for
functions with Fab parameters (e.g., AmrLevel's derive functions).

## Checklist

The proposed changes:

### DIFF
--- a/Src/EB/AMReX_EBFArrayBox.H
+++ b/Src/EB/AMReX_EBFArrayBox.H
@@ -14,7 +14,7 @@ class EBFArrayBox
 {
 public:
     EBFArrayBox ();
-    EBFArrayBox (Arena* ar);
+    explicit EBFArrayBox (Arena* ar);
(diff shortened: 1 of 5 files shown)


Src/EB/AMReX_EBFArrayBox.H
```diff
@@ -7,14 +7,16 @@
...
 
 class EBCellFlagFab;
+class EBFArrayBoxFactory;
 
 class EBFArrayBox
     : public FArrayBox
 {
 public:
     EBFArrayBox ();
     EBFArrayBox (Arena* ar);
```
atmyers, line 16: Maybe this should be marked `explicit`?

atmyers (approved):

PR #2551, opened 2021-12-28
Update to SUNDIALS 6

balos1:
## Summary

Update AMReX to use SUNDIALS 6. We could make this backwards compatible with SUNDIALS 5.7.0/5.8.0 without too much effort (I think),  if that is desired.

## Additional background

SUNDIALS 6 introduces a `SUNContext` class (needed by all/most other SUNDIALS classes) that should be associated with only one SUNDIALS simulation at a time. Thus, when using multiple threads, multiple `SUNContext` objects are needed. This is why we create one (and one `SUNMemoryHelper`) for each OpenMP thread.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX.cpp
+++ b/Src/Base/AMReX.cpp
@@ -543,7 +543,11 @@ amrex::Initialize (int& argc, char**& argv, bool build_parm_parse,
 #endif
 
 #ifdef AMREX_USE_SUNDIALS
-    sundials::Initialize();
+#ifdef AMREX_USE_OMP
+    sundials::Initialize(omp_get_max_threads());
+#else
+    sundials::Initialize(1);
+#endif
 #endif
 
     if (system::verbose > 0)
--- a/Src/Base/AMReX.cpp
+++ b/Src/Base/AMReX.cpp
@@ -543,11 +543,7 @@ amrex::Initialize (int& argc, char**& argv, bool build_parm_parse,
 #endif
 
 #ifdef AMREX_USE_SUNDIALS
-#ifdef AMREX_USE_OMP
-    sundials::Initialize(omp_get_max_threads());
-#else
-    sundials::Initialize(1);
-#endif
+    sundials::Initialize(amrex::OpenMP::get_max_threads());
 #endif
 
     if (system::verbose > 0)
--- a/Src/Extern/SUNDIALS/AMReX_Sundials.H
+++ b/Src/Extern/SUNDIALS/AMReX_Sundials.H
@@ -4,5 +4,16 @@
 #include <AMReX_SUNMemory.H>
 #include <AMReX_NVector_MultiFab.H>
 static_assert(std::is_same<amrex::Real,realtype>::value, "amrex::Real must be the same as SUNDIALS realtype");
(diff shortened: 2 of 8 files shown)


balos1:
Suggesting we get reviews/feedback from @jmsexton03 and @esclapez and @gardner48 (they are all well aware of this change). CC:  @jrood-nrel.

Src/Base/AMReX.cpp
```diff
@@ -543,7 +543,11 @@ amrex::Initialize (int& argc, char**& argv, bool build_parm_parse,
 #endif
 
 #ifdef AMREX_USE_SUNDIALS
-    sundials::Initialize();
+#ifdef AMREX_USE_OMP
+    sundials::Initialize(omp_get_max_threads());
+#else
+    sundials::Initialize(1);
+#endif
```
jmsexton03: Using a wrapper may be more concise:
https://github.com/AMReX-Codes/amrex/blob/development/Src/Base/AMReX_OpenMP.H
```suggestion
    sundials::Initialize(amrex::OpenMP::get_max_threads());
```

Src/Extern/SUNDIALS/AMReX_Sundials.H
```diff
@@ -1,8 +1,24 @@
...
+namespace amrex {
+namespace sundials {
+
+void Initialize(int nthreads);
+void Finalize();
+#ifdef AMREX_USE_OMP
+::sundials::Context* The_Sundials_Context(int i = omp_get_thread_num());
+#else
+::sundials::Context* The_Sundials_Context(int i = 0);
+#endif
```
jmsexton03: Using a wrapper may be more concise:
https://github.com/AMReX-Codes/amrex/blob/development/Src/Base/AMReX_OpenMP.H
```suggestion
::sundials::Context* The_Sundials_Context(int i = amrex::OpenMP::get_thread_num());
```

Src/Extern/SUNDIALS/AMReX_Sundials.cpp
```diff
@@ -0,0 +1,56 @@
+#include <AMReX_Sundials.H>
+
+namespace amrex {
+namespace sundials {
+
+namespace {
+  std::vector<bool> initialized;
+  std::vector<::sundials::Context*> the_sundials_context;
```
jmsexton03: Is there a strong reason not to use `amrex::Vector` here? 
https://github.com/AMReX-Codes/amrex/blob/135010c8baab688b9a190f008e5098a50a91edb2/Src/Base/AMReX_Vector.H#L18-L22

WeiqunZhang: Note that `Vector<bool>` does not work.  If you choose to use `Vector`, you can use `Vector<int>` instead.

balos1: I switched this over to `Vector<int>`

esclapez (approved):

gardner48 (approved):

WeiqunZhang (approved):

PR #2559, opened 2022-01-05
Kernel fusing in FabArray Comm

WeiqunZhang:
Note that amrex::Add has been moved from AMReX_FabArrayUtility.H to
AMReX_FabArray.H so that it can be used by AMReX_FabArrayCommI.H.

## Checklist

The proposed changes:


kngott:
Do you believe fusing (like this) would be possible/useful in the non-simple, "short-circuit" cases? If so, I can add it to the MPI list.

kngott (approved):

PR #2560, opened 2022-01-05
Return a zero-size vector in ParallelDescriptor::Gather.

kngott:
## Summary

The Gather implementation that returns a vector currently returns an un-initialized, 1-element vector to the non-root ranks. This seems unintuitive as a user and requires very careful consideration of the returned vector on non-root ranks.

This PR changes the non-root return to a zero-size vector instead, which seems safer and more intuitive than a one-element vector with un-initialized data that, for example, may be accessed in a range-loop over the vector. (Who would do such a thing? ... cough, cough... 😃). 

## Additional background

I just chose this style for simplicity and readability. Another option is set `std::vector<T> resl(0)` initially, and use `.front()` in the Gather call, but `.front()` is undefined if `size() == 0`, which seems less clear and potentially corner-case bug-inducing, so I opted for this.

The best implementation would probably be to return something with bounds-checking as well, such as an `amrex::Vector`. I assume `std::vector<T> a = amrex::Vector<T> b` should work everywhere and so such a change wouldn't break anything. Any thoughts?

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_ParallelDescriptor.H
+++ b/Src/Base/AMReX_ParallelDescriptor.H
@@ -933,7 +933,12 @@ ParallelDescriptor::Gather (const T& t, int root)
                                root,
                                Communicator()) );
     BL_COMM_PROFILE(BLProfiler::GatherTi, sizeof(T), root, BLProfiler::NoTag());
-    return resl;
+
+    if ( root == MyProc() ) {
+        return resl;
+    } else {
+        return std::vector<T>;
+    }
 }
 
 template <class T>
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
May I suggest
```
std::vector<T> resl;
if (root == MyProc()) { resl.resize(NProcs()); }
....
    resl.data(),  // instead of &resl[0]
....
return resl;
```
It's OK to call `data()` on an empty vector, and its value is ignored by MPI_Gather on non-root processes.

kngott:
Yeah, that looks good to me!

WeiqunZhang (approved):

PR #2563, opened 2022-01-06
Support implicit function class defined on host when using device memory explicitly. 

hengjiew:
## Summary 
AMReX needs to use managed memory if the user-defined implicit function (IF) class's member functions are only callable by the host. This PR allows users to use IF classes defined on the host with device memory.

## Checklist

The proposed changes:

### DIFF
--- a/Src/EB/AMReX_EB2_GeometryShop.H
+++ b/Src/EB/AMReX_EB2_GeometryShop.H
@@ -331,12 +331,32 @@ public:
     void fillFab (BaseFab<Real>& levelset, const Geometry& geom, RunOn,
                   Box const& bounding_box) const noexcept
     {
+#ifdef AMREX_USE_GPU
+        if (!levelset.arena()->isHostAccessible()) {
+            const Box& bx = levelset.box();
+            BaseFab<Real> h_levelset(bx, levelset.nComp(), The_Pinned_Arena());
+            fillFab_Cpu(h_levelset, geom, bounding_box);
+            Gpu::htod_memcpy_async(levelset.dataPtr(), h_levelset.dataPtr(),
+                                   levelset.nBytes(bx, levelset.nComp()));
+            Gpu::streamSynchronize();
+        }
+        else
+#endif
+        {
+            fillFab_Cpu(levelset, geom, bounding_box);
+        }
+    }
+
+    void fillFab_Cpu(BaseFab<Real>& levelset, const Geometry& geom,
+                     Box const& bounding_box) const noexcept
+    {
         const auto problo = geom.ProbLoArray();
         const auto dx = geom.CellSizeArray();
         const Box& bx = levelset.box();
-        const auto& a = levelset.array();
         const auto blo = amrex::lbound(bounding_box);
         const auto bhi = amrex::ubound(bounding_box);
+
+        const auto& a = levelset.array();
         amrex::LoopOnCpu(bx, [&] (int i, int j, int k) noexcept
         {
             a(i,j,k) = m_f(RealArray{AMREX_D_DECL(problo[0]+amrex::Clamp(i,blo.x,bhi.x)*dx[0],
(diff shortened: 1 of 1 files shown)


Src/EB/AMReX_EB2_GeometryShop.H
```diff
@@ -331,12 +331,33 @@ public:
     void fillFab (BaseFab<Real>& levelset, const Geometry& geom, RunOn,
                   Box const& bounding_box) const noexcept
     {
+#ifdef AMREX_USE_GPU
+        const Box& bx = levelset.box();
```
WeiqunZhang: Minor comment.  We can move the line above inside the if block.

Src/EB/AMReX_EB2_GeometryShop.H
```diff
@@ -387,35 +409,64 @@ public:
...
-        const auto bhi = amrex::ubound(bounding_box);
         for (int idim = 0; idim < AMREX_SPACEDIM; ++idim) {
             Array4<Real> const& inter = inter_arr[idim];
             Array4<Type_t const> const& type = type_arr[idim];
+            // When GShop is not GPUable, the intercept is either in managed memory or
+            // in device memory and needs to be copied to host and setup there.
+#ifdef AMREX_USE_GPU
             const Box bx{inter};
-            amrex::LoopOnCpu(bx, [&] (int i, int j, int k) noexcept
+            if (!The_Arena()->isManaged()) {
```
WeiqunZhang:
```suggestion
            if (!The_Arena()->isHostAccessible()) {
```

Src/EB/AMReX_EB2_GeometryShop.H
```diff
@@ -387,35 +409,64 @@ public:
...
-        auto const& problo = geom.ProbLoArray();
-        const auto blo = amrex::lbound(bounding_box);
-        const auto bhi = amrex::ubound(bounding_box);
         for (int idim = 0; idim < AMREX_SPACEDIM; ++idim) {
             Array4<Real> const& inter = inter_arr[idim];
             Array4<Type_t const> const& type = type_arr[idim];
+            // When GShop is not GPUable, the intercept is either in managed memory or
+            // in device memory and needs to be copied to host and setup there.
+#ifdef AMREX_USE_GPU
             const Box bx{inter};
```
WeiqunZhang: We can move this Box inside the if block.

Src/EB/AMReX_EB2_GeometryShop.H
```diff
@@ -387,35 +409,64 @@ public:
...
-            amrex::LoopOnCpu(bx, [&] (int i, int j, int k) noexcept
+            if (!The_Arena()->isManaged()) {
+                BaseFab<Real> h_inter_fab(bx, 1, The_Pinned_Arena());
+                BaseFab<Type_t> h_type_fab(bx, 1, The_Pinned_Arena());
+                Gpu::dtoh_memcpy(h_type_fab.dataPtr(), type.dataPtr(), h_type_fab.nBytes(bx, 1));
+                const auto& h_inter = h_inter_fab.array();
+                const auto& h_type  = h_type_fab.array();
```
WeiqunZhang:
```suggestion
                const auto& h_type  = h_type_fab.const_array();
```

WeiqunZhang (changes requested):
It might be better to use the async version of the dtoh and htod memcpy and add an explicit stream sync at the end of the scope for the temporary fabs in the pinned memory.

hengjiew:
> It might be better to use the async version of the dtoh and htod memcpy and add an explicit stream sync at the end of the scope for the temporary fabs in the pinned memory.

Hello @WeiqunZhang, do you mean sth like this?
```
dtoh_memcpy_async(h_levelset.dataPtr(), levelset.dataPtr(), ...);
fillFab_cpu(h_levelset, ...);
htod_memcpy_async(d_levelset.dataPtr(), h_levelset.dataPtr(), ...);
streamSynchronize();
```
I am not sure if this will work because the host function `fillFab_cpu` and the copy to pinned memory have data dependence (h_levelset) and need to be executed in serial.

WeiqunZhang:
Yes, you are right that we need an extra `streamSynchronize` in `getIntercept` if we use the async version of memcpy.  However, for the code shown above, I don't think we need to call `dtoh` memcpy at all in `fillFab` because we don't need to initialize the `levelset` fab before calling `fillFab_Cpu`.

WeiqunZhang (approved):

PR #2575, opened 2022-01-12
`add_par` precendence correction in Docs.

etpalmer63:
## Summary
Correct description of precendence for processing inputs with `add_par` type functions. Reword to
emaphsis recommended usage.
## Checklist

The proposed changes:

### DIFF
--- a/Docs/sphinx_documentation/source/Basics.rst
+++ b/Docs/sphinx_documentation/source/Basics.rst
@@ -295,35 +295,48 @@ run with:
 Setting Parameter Values Inside Functions
 -----------------------------------------
 
-Sometimes an application code may want to set a default that differs from the
-default in AMReX.  In this case, it is often convenient to define a function that
-sets the variable(s), and pass the name of that function to :cpp:`amrex::Initialize`.
-As an example, we may define :cpp:`add_par` to set :cpp:`extend_domain_face`
-to false if it hasn't already been set in the inputs file.
+An application code may want to set values or defaults that differ from the
+those in AMReX in a function. This is accomplished in two steps:
 
-.. highlight:: c++
+- First, define a function that sets the variable(s).
 
-::
+- Second, pass the name of that function to :cpp:`amrex::Initialize`.
 
-    void add_par () {
-       ParmParse pp("eb2");
-       if(not pp.contains("extend_domain_face")) {
-          pp.add("extend_domain_face",false);
-       }
-    };
+The example function below sets variable values using two different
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
Although one could argue it's not very clear, the current documentation is NOT wrong.  The example uses `if (not pp.contains(".."))` before calling `add`.  If the value is already set in an inputs file or command line, the test will be false.  Maybe we can add another case to the example to show that we can also ignore values set in the inputs files and command line.  We may also delete the note and add comments instead to the code.

etpalmer63:
I agree the docs as currently written are not wrong; I should have titled the PR better. The two example idea is also a good solution for emphasizing the effect of different usages. Shall I use the method you mentioned before for the other example?

Adding,

```
{
    ParmParse pp;
    pp.add(...);
}
```
before `amrex::Initialize`.

WeiqunZhang:
No, we don't need to show the ParmParse before Initialize way.

etpalmer63:
The commit history is no good on this one. I will fix it.

etpalmer63:
Ready for review.

WeiqunZhang:
I am confused.  A warning is added,
```
Setting values inside a function in this way will overide (sic) AMReX defaults and values set in the inputs file.
```
What exactly is *this* way?  There is only one example in this section.  I assume that is the *this*.  But that example does NOT override the values set in the inputs file.

WeiqunZhang:
My suggestion is to remove the note/warning and have them as comments in the example.
```
// comments
if (! pp.contains("...")) {
    pp.add(...);
}

// comments
pp.add(...)
```

WeiqunZhang:
The initial confusion a while ago came from the users only read the note but did not follow the example. Make the note a comment right above the code should help.

etpalmer63:
Sorry, it looks like I did the rebase wrong and lost all my changes. I have a backup, I will try to fix it.

etpalmer63:
I couldn't come up with a good short comment to add to the second code. I would say `Overwrite` but if the variable isn't already set, it will just make one. And I can't say `write` because it will also overwrite. So I figured I would let the code speak for itself and instead leave it explained in the docs below.

WeiqunZhang:
I think it's clearer if we put both cases in one `add_par` function.

```
// `abc` set in inputs file or command line is ignored.
if (! pp.contains("abc")) {
    pp.add("abc", 1);
}

// `xyz` can be overidden by inputs file and command line argument.
pp.add("xyz", 1)
```

WeiqunZhang (approved):

PR #2581, opened 2022-01-13
Enable CUDA in CMakeList.txt for Spack smoke test.

etpalmer63:
## Summary
Adds the necessary lines to CMakeLists.txt to enable CMake to compile the Spack smoke test with CUDA. 

## Additional background
Tested on spack/develop. Passed, but requires an extra step to make sure the CUDA compilers match. Next step is to adjust the smoke test script for AMReX in Spack. 

## Checklist

The proposed changes:

### DIFF
--- a/Tests/SpackSmokeTest/CMakeLists.txt
+++ b/Tests/SpackSmokeTest/CMakeLists.txt
@@ -17,43 +17,61 @@ set(_src_dir   ${_base_dir}/Source)
 set(_input_dir ${_base_dir}/Exec)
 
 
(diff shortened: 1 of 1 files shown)


sayerhs (approved):
The code changes look fine. Would it be possible to link the spack script to this PR for future reference? In particular, it would be nice to see what arguments are passed to CMake via spack.

WeiqunZhang:
https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/amrex/package.py

PR #2584, opened 2022-01-13
[WIP] HIP RDC: Add --hip-link

ax3l:
## Summary

As of ROCm 4.5 on Crusher, we saw that `--hip-link` is not always coming in (via hip::device) to link flags on Cray compilers, but this flag is required if `-fgpu-rdc` is used.

## Additional background

https://github.com/ROCm-Developer-Tools/HIP/issues/2203#issuecomment-740163037

Trying to avoid
```
ld.lld: error: undefined symbol: __hip_fatbin
```
errors with the Cray prg env wen using RDC: https://github.com/ECP-WarpX/WarpX/pull/2741

## Checklist

The proposed changes:

### DIFF
--- a/Tools/CMake/AMReXParallelBackends.cmake
+++ b/Tools/CMake/AMReXParallelBackends.cmake
@@ -289,15 +289,19 @@ if (AMReX_HIP)
    # device variable support (for codes that use global variables on device)
    # as well as our kernel fusion in AMReX, e.g. happening likely in amr regrid
    # As of ROCm 4.1, we cannot enable this with hipcc, as it looks...
+   # As of ROCm 4.5 on Crusher, we saw that `--hip-link` is not always coming
+   # in (via hip::device) to link flags on Cray compilers, but this flag is
+   # required if `-fgpu-rdc` is used.
+   #   https://github.com/ROCm-Developer-Tools/HIP/issues/2203#issuecomment-740163037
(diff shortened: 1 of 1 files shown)


ax3l:
Tried this and lands on the link lines, but does not solve the problem in `PrgEnv-cray` when compiling with RDC.

ax3l:
Solved this by using the Cray compiler wrappers: https://warpx.readthedocs.io/en/latest/install/hpc/crusher.html

Wondering if we should still pro-actively add the flag. We tried again today in MFIX to use this as a work-around on their Crusher build, but the problem persists as well. For them, the current work-around is to use the legacy `hipcc` compiler on Crusher as `CXX` (and [hint MPICH as in the OLCF docs](https://docs.olcf.ornl.gov/systems/crusher_quick_start_guide.html#compiling-with-hipcc) via `CXXFLAGS` & `LDFLAGS`).

HPE's `CC` has a `--hip-link` flag, but it does not really seem to resolve the symbol issue...

ax3l:
Might be an HPE/Cray compiler wrapper bug. I opened an OLCF Ticket: OLCFHELP-5465

ax3l:
Oh, one has to load the `cce` module _after_ loading the `rocm` module for this to work on Crusher (`PrgEnv-cray`, default):
https://github.com/ECP-WarpX/WarpX/pull/2770

PR #2586, opened 2022-01-13
CMake: remove nonexistent header file, update minimum sundials version

jmsexton03:
## Summary
This makes CMake more consistent with expecting Sundials 6.x

## Checklist

The proposed changes:


Tools/CMake/AMReXThirdPartyLibraries.cmake
```diff
@@ -81,7 +81,8 @@ if (AMReX_SUNDIALS)
     if (SUNDIALS_FOUND)
         message(STATUS "SUNDIALS_FOUND is true, assuming nvecserial or gpu-specific vector found for version 5.7.0")
```
WeiqunZhang: Does this message need to be updated?

jmsexton03: yes, pushed a fix

jmsexton03:
I may also need to remove sundials_context.h from here:
https://github.com/jmsexton03/amrex/blob/bb17c90ad82090b63df6df31a677b265d4932a9a/Src/Extern/SUNDIALS/AMReX_SUNMemory.H#L4

Tools/CMake/AMReXThirdPartyLibraries.cmake
```diff
@@ -79,9 +79,10 @@ endif ()
 #
 if (AMReX_SUNDIALS)
     if (SUNDIALS_FOUND)
-        message(STATUS "SUNDIALS_FOUND is true, assuming nvecserial or gpu-specific vector found for version 5.7.0")
+        message(STATUS "SUNDIALS_FOUND is true, assuming nvecserial or gpu-specific vector found for version 6.0.0 or higher")
```
sayerhs, line 82: Does sundials provide a variable/components that could be checked here? We could also throw an error by checking `SUNDIALS_VERSION` (or equivalent) here to enforce `6.0.0` or higher?

Tools/CMake/AMReXThirdPartyLibraries.cmake
```diff
@@ -79,9 +79,10 @@ endif ()
...
     else ()
-        find_package(SUNDIALS 5.7.0 REQUIRED)
+       set(SUNDIALS_MINIMUM_VERSION 6.0.0 CACHE INTERNAL "Minimum required SUNDIALS version")
```
sayerhs, line 84: This is more of a curiosity. What is the reason to use a variable here?

Tools/CMake/AMReXThirdPartyLibraries.cmake
```diff
@@ -79,9 +79,10 @@ endif ()
...
+       find_package(SUNDIALS ${SUNDIALS_MINIMUM_VERSION} CONFIG QUIET )
```
sayerhs, line 85: Same question as above, should we pass components to `find_package` so that `nveccuda` target exists when it is a CUDA build? It is better to raise error here than encounter an error in line 88?

Tools/CMake/AMReXThirdPartyLibraries.cmake
jmsexton03: In theory, when sundials is installed as a library for the versions of interest a SUNDIALSConfig.cmake file is available. Part of the explicit check on SUNDIALS_FOUND was intended as a workaround for super builds containing both sundials and amrex, where sundials may not be installed to the filesystem.
https://sundials.readthedocs.io/en/latest/Install_link.html?highlight=sundialsconfig#using-sundials-as-a-third-party-library-in-other-cmake-projects

I would prefer to defer using the config file until the next minor Sundials release, as the current config file may depend non-trivially on the new sundials profiling capability. @gardner48 @balos1 thoughts?

Tools/CMake/AMReXThirdPartyLibraries.cmake
jmsexton03: Previous code being reused? No strong reason

Tools/CMake/AMReXThirdPartyLibraries.cmake
jmsexton03: It's likely worth checking for components earlier if possible in order to error cleanly. For the case where SUNDIALS might be built with more functionality than we're compiling the amrex application with, we wouldn't want to link anything extra.

sayerhs (approved):

PR #2588, opened 2022-01-18
Add headings and steps to plot vector field in ParaView doc section

etpalmer63:
## Summary
Adds directions for ploting a vector field from 3 separate scalar values using the Calculator filter.
Also split up the ParaView section by adding some headings.

## Checklist

The proposed changes:


WeiqunZhang:
Could we make the png file smaller?  When we make a compressed tar ball of amrex, about 70% of the size is used by png files in Docs/.

etpalmer63:
It should be down to about 136K now.

etpalmer63:
I think I could also similarly reduce some of the other images without a lot of degredation using indexed colors. Let me know if you want me to give it a try.

WeiqunZhang:
The problem of that is although it reduces the tar ball size, it will increase the git repo size because both versions will exist in git.

WeiqunZhang (approved):

PR #2594, opened 2022-01-20
Overloading reducesum with 4 multifabs

ashesh2512:
## Summary
Overloading reduce sum to support 4 multifabs

## Additional background
With recent developments in amr-wind, we have noticed a need for using reducesum with 4 multifabs. I have coded this based on the other reducesum overloading calls. Please let me know if there would be a cleaner way for me to implement this. 

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_FabArrayUtility.H
+++ b/Src/Base/AMReX_FabArrayUtility.H
@@ -327,6 +327,91 @@ ReduceSum (FabArray<FAB1> const& fa1, FabArray<FAB2> const& fa2,
 }
 #endif
 
+template <class FAB1, class FAB2, class FAB3, class FAB4, class F,
+          class bar = std::enable_if_t<IsBaseFab<FAB1>::value> >
+typename FAB1::value_type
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
We have introduced a new function `ParReduce`.  If it's not for backward compatibility, we might even remove the various ReduceSum functions.  Here is how you do this in the new way to compute the sum of the product of four MultiFabs.
```
auto const& ma1 = mf1.const_arrays();
auto const& ma2 = mf2.const_arrays();
auto const& ma3 = mf3.const_arrays();
auto const& ma4 = mf4.const_arrays();
Real result = ParReduce(TypeList<ReduceOpSum>{}, TypeList<Real>{}, mf1, nghost, // nghost is IntVect
                        [=] AMREX_GPU_DEVICE (int box_no, int i, int j, int k) -> GpuTuple<T>
                        {
                            return { ma1[box_no](i,j,k) * ma2[..](...) * ma3.... * ma4...  };
                        });
```

ashesh2512:
@WeiqunZhang Thank you for the update. I was able to use the new approach for our purposes. Closing this PR.

PR #2598, opened 2022-01-21
Disable EB solver's phi-on-centroid for hip for now

WeiqunZhang:
We need this to work around the out-of-register issue with current
compilers.

## Checklist

The proposed changes:

### DIFF
--- a/Src/LinearSolvers/MLMG/AMReX_MLEBABecLap_F.cpp
+++ b/Src/LinearSolvers/MLMG/AMReX_MLEBABecLap_F.cpp
@@ -111,6 +111,10 @@ MLEBABecLap::Fapply (int amrlev, int mglev, MultiFab& out, const MultiFab& in) c
 
             if (treat_phi_as_on_centroid) {
 #ifdef AMREX_USE_HIP
+                // This causes an abort in HIP 4.5 but works in earlier versions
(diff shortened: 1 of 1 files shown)


Src/LinearSolvers/MLMG/AMReX_MLEBABecLap_F.cpp
```diff
@@ -110,6 +110,13 @@ MLEBABecLap::Fapply (int amrlev, int mglev, MultiFab& out, const MultiFab& in) c
             bool treat_phi_as_on_centroid = ( phi_on_centroid && (mglev == 0) );
 
             if (treat_phi_as_on_centroid) {
+#ifdef AMREX_USE_HIP
+                amrex::Abort("MLEBABecLap::Fapply: phi on centroid not supported for HIP");
```
ax3l, line 118: Let's add a comment, so we remember when to try and put it in again? :)
```suggestion
                // This causes an abort in HIP 4.5 but works in earlier versions
                // A ROCm follow-up release will fix this.
                // Error message for ROCm 4.5:
                //   lld: error: ran out of registers during register allocation
                amrex::Abort("MLEBABecLap::Fapply: phi on centroid not supported for HIP");
```

jmsexton03 (approved):

PR #2602, opened 2022-01-27
Spack install instructions for Building section.

etpalmer63:
## Summary
Adds instructions for how to install AMReX using Spack to the Building with AMReX section. 

## Additional background 
If we decide to include something like this, this draft is what I was thinking. 

## Checklist

The proposed changes:

### DIFF
--- a/Docs/sphinx_documentation/source/BuildingAMReX.rst
+++ b/Docs/sphinx_documentation/source/BuildingAMReX.rst
@@ -743,3 +743,38 @@ This capability is not supported on Windows.
 (3) Memory profiling is an optional feature in AMReX that is not enabled by default.  It reads
 memory system information from the OS to give us a summary of our memory usage.  This is not
 supported on Windows.
+
+
+.. _sec:build:spack:
+
+Spack
+=====
+
+AMReX can be installed using the scientific software package manager Spack. Spack
+supports multiple versions and configurations of AMReX across a wide variety of platforms
+and environments. To learn more about Spack visit http://www.spack.io. For system requirements and
+installation instructions please see https://spack.readthedocs.io/.
+
+Once Spack has been downloaded and the Spack environment enabled, AMReX can be
+installed with the command,
+
+.. code-block:: bash
+
+   spack install amrex
+
+This will install the latest release of AMReX and required dependencies if needed.
+
+AMReX can be built in several combinations of versions and configurations. Available options can
+be viewed by typing,
+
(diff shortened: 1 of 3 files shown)


Docs/sphinx_documentation/source/BuildingAMReX.rst
```diff
@@ -743,3 +743,38 @@ This capability is not supported on Windows.
...
+
+   spack info amrex
+
+For example, suppose we want to install the development version of AMReX for a two dimensional
+simulation with Cuda support for Cuda Architecture ``sm_60``. Then we would
+use the install commands,
+
+.. code-block:: bash
+
+   spack install amrex@develop cuda cuda_arch=60
```
WeiqunZhang: By default, the dimension is 3.  To build for 2d (as mentioned earlier), one must explicit specify `dimensions=2`.  Also I don't think the command above works, because it should be `+cuda` instead of `cuda`.

Also, there are two places mentioning spack already.  https://amrex-codes.github.io/amrex/docs_html/search.html?q=spack&check_keywords=yes&area=default  Maybe in those two places we could refer the readers to this for more details.

WeiqunZhang (changes requested):

Docs/sphinx_documentation/source/BuildingAMReX.rst
```diff
@@ -743,3 +743,38 @@ This capability is not supported on Windows.
...
 memory system information from the OS to give us a summary of our memory usage.  This is not
 supported on Windows.
+
+
+.. _sec:build:spack:
+
+Spack
+=====
+
+AMReX can be installed using the scientific software package manager Spack. Spack is designed
```
WeiqunZhang: Maybe it's more appropriate to say `Spack supports ...` instead.

Docs/sphinx_documentation/source/BuildingAMReX.rst
etpalmer63: Yes, this one is no good. Thank you for catching the error.

In the latest push, I added links and wording to refer readers from those two places; `GettingStarted.rst` and `BuildingAMReX_Chapter.rst`.

Docs/sphinx_documentation/source/BuildingAMReX.rst
etpalmer63: Yes, that is better.

WeiqunZhang (approved):

PR #2603, opened 2022-01-27
Update FabArray::copyTo

WeiqunZhang:
Remove the two versions taking a sub box.  I don't think those versions are
used by any codes.  This allows for simplification of the implementation.
Reimplement copyTo to reduce the number of bcast to just one.  Also the new
versions work without managed memory.

## Checklist

The proposed changes:


WeiqunZhang:
@michaeljbrazell @rporcu

atmyers (approved):

rporcu:
thank you!

PR #2624, opened 2022-02-09
Update checkpoint format to account for expanded particle ids.

atmyers:
See discussion here: https://github.com/ECP-WarpX/WarpX/issues/2785

Note that this does not alter plotfiles yet, only checkpoint / restart.

The proposed changes:

### DIFF
--- a/Src/Extern/HDF5/AMReX_ParticleHDF5.H
+++ b/Src/Extern/HDF5/AMReX_ParticleHDF5.H
@@ -60,7 +60,7 @@ ParticleContainer<NStructReal, NStructInt, NArrayReal, NArrayInt, Allocator>
                           [=] AMREX_GPU_HOST_DEVICE (const SuperParticleType& p) -> int
                           {
                               return p.id() > 0;
-                          });
+                          }, true);
 }
 
 template <int NStructReal, int NStructInt, int NArrayReal, int NArrayInt,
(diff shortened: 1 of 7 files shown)


Src/Extern/HDF5/AMReX_ParticleHDF5.H
```diff
@@ -1299,9 +1307,19 @@ ParticleContainer<NStructReal, NStructInt, NArrayReal, NArrayInt, Allocator>
...
-
+        if (convert_ids) {
+            union punner {
+                std::int32_t i;
+                std::uint32_t u;
+            };
+            punner x, y;
+            x.i = iptr[0];
+            y.i = iptr[1];
+            p.m_idcpu = ((std::uint64_t)x.u) << 32 | y.u;
```
WeiqunZhang: Strictly speaking, this is undefined behavior. https://en.cppreference.com/w/cpp/language/union
> It's undefined behavior to read from the member of the union that wasn't most recently written. Many compilers implement, as a non-standard language extension, the ability to read inactive members of a union.

atmyers: I think this is one of those things that this in principal UB but that in practice compilers basically have to make work or a ton of real-world code would break. I agree it's better to avoid if possible - can you suggest a way of implementing this that avoids UB?

WeiqunZhang:
Maybe using `memcpy`?

atmyers:
> Maybe using `memcpy`?

See update.

atmyers:
@houjun I've tried to update the HDF5 format as well - would you mind having a look and seeing if everything looks okay with that?

houjun:
Looks good.
Also, I'm working on adding SZ compression to the HDF5 output, I will submit a pull request after this gets merged.

peterscherpelz:
@atmyers I've now tested this on my code and verified that it fixes the restart issue, see https://github.com/ECP-WarpX/WarpX/issues/2785#issuecomment-1036519233 comment. Thanks!!!

WeiqunZhang (approved):

PR #2629, opened 2022-02-15
Update SpackSmokeTest to run with Fortran and/or HIP.

etpalmer63:
## Summary
Updates the CMake file for the Spack smoke test so that it will run when either fortran and/or the gpu backend is hip. 

## Additional background
Notice the CMake version has been raised to 3.22.

## Checklist

The proposed changes:

### DIFF
--- a/Tests/SpackSmokeTest/CMakeLists.txt
+++ b/Tests/SpackSmokeTest/CMakeLists.txt
@@ -6,7 +6,7 @@
 # against a currently installed version of AMReX. The resulting
 # executable can then be ran to test functionality.
 
-cmake_minimum_required(VERSION 3.14)
+cmake_minimum_required(VERSION 3.22)
 
 project(amrex-test-install)
 
@@ -17,9 +17,20 @@ set(_src_dir   ${_base_dir}/Source)
 set(_input_dir ${_base_dir}/Exec)
 
 
+if (AMReX_FORTRAN_INTERFACES)
+    enable_language(Fortran)
+endif()
+
 if (AMReX_GPU_BACKEND STREQUAL "CUDA")
     enable_language(CUDA)
     find_package(AMReX REQUIRED CUDA)
+
+elseif(AMReX_GPU_BACKEND STREQUAL "HIP")
+    find_package(AMReX REQUIRED HIP)
+    find_package(rocrand REQUIRED CONFIG)
+    find_package(rocprim REQUIRED CONFIG)
+    find_package(hiprand REQUIRED CONFIG)
+
 else()
    # Use installed version of AMReX
    find_package(AMReX REQUIRED)
@@ -62,7 +73,6 @@ target_sources(install_test
     )
 
 target_link_libraries(install_test PUBLIC AMReX::amrex)
-
 
 # Additional CUDA configuration commands
 if (AMReX_GPU_BACKEND STREQUAL "CUDA")
--- a/Tests/SpackSmokeTest/CMakeLists.txt
+++ b/Tests/SpackSmokeTest/CMakeLists.txt
@@ -6,7 +6,7 @@
 # against a currently installed version of AMReX. The resulting
 # executable can then be ran to test functionality.
 
-cmake_minimum_required(VERSION 3.22)
+cmake_minimum_required(VERSION 3.14)
 
 project(amrex-test-install)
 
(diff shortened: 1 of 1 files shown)


Tests/SpackSmokeTest/CMakeLists.txt
```diff
@@ -6,7 +6,7 @@
 # against a currently installed version of AMReX. The resulting
 # executable can then be ran to test functionality.
 
-cmake_minimum_required(VERSION 3.14)
+cmake_minimum_required(VERSION 3.22)
```
ax3l: Can you summarize again why we do this, just for documentation? I cannot spot the reason from the `diff` alone right now.

Should we generally raise the CMake dependency at some point? I wonder how confusing it is to have multiple components use different constrains (but would be ok if we leave the smoke test a bit higher in constrains for now).

Tests/SpackSmokeTest/CMakeLists.txt
```diff
@@ -17,9 +17,20 @@ set(_src_dir   ${_base_dir}/Source)
 set(_input_dir ${_base_dir}/Exec)
 
 
+if (AMReX_FORTRAN_INTERFACES)
```
ax3l, line 20: Where is this variable set again for the smoke test?

Tests/SpackSmokeTest/CMakeLists.txt
etpalmer63: Bumping it up made the HIP stuff work for me on Spock. I recall searching and seeing something about HIP support improvements in 3.21, but let me double check and I will add it here. 

I wasn't very concerned about moving the version up here for exactly the reason you mentioned (smoke test). On top of that, Spack usually has easy access to recent versions of CMake. But I can/will test some lower versions to see when it breaks.

Tests/SpackSmokeTest/CMakeLists.txt
etpalmer63: This variable isn't set in the smoke test. It was set with the original AMReX install. 

I put it there because without it, the smoke test fails when you install with `spack install amrex+fortran`.

Tests/SpackSmokeTest/CMakeLists.txt
ax3l: Ok, in that case I would probably recommend to at least bump the version constrain for CMake in Spack itself during https://github.com/spack/spack/pull/28967 :)

We could also add `message(FATAL_ERROR ...)` here on `CMAKE_VERSION` and keep 3.14 for CPU builds.
According to Spack, we also need CMake 3.17+ for CUDA already.

Tests/SpackSmokeTest/CMakeLists.txt
ax3l: I see, I was just wondering where it gets injected because this looks like a standalone `CMakeLists.txt`. But I think we call it from somewhere with arguments... Can you point me to where we do this?

etpalmer63: In the Spack amrex package (amrex/package.py), in the self test section around line 294. The command `args.extend(self.cmake_args())` contains the  FORTRAN_INTERFACES flag and several others.

Tests/SpackSmokeTest/CMakeLists.txt
etpalmer63: I was able to test it with CMake 3.20 and it worked when I turned off MPI. It turned out going from 3.21 to 3.22 fixed something about finding MPI compilers, but I suspect that is probably an issue specific to the system I'm testing on. What do you think about going to 3.20?

ax3l: Where did you run, on Crusher / a HPE/Cray?

I merged this to CMake 3.22, thus the Cray FindMPI nice-ness:
https://gitlab.kitware.com/cmake/cmake/-/merge_requests/6264

ax3l: You probably need 3.20 due to the other bugs we fixed that required `SHELL:` support with HIP.

So I would say: leave 3.14 for now on top. Use `CMAKE_VERSION` to
- check for CMake 3.20+ if either HIP or HIP+Fortran are requested.
- check for CMake 3.17+ if CUDA is requested

Tests/SpackSmokeTest/CMakeLists.txt
ax3l: Thanks!! :)

Tests/SpackSmokeTest/CMakeLists.txt
etpalmer63: Yes, I was running on Spock. So that fix was your doing :grin: , thanks for that! 

I added the tests and fatal_error messages as you suggested.

ax3l (approved):
Looking good, thanks a lot! :)

PR #2630, opened 2022-02-16
ParticleContainer::make_alike

ax3l:
## Summary

This creates a new AMReX particle container type with same compile-time and run-time attributes. But, it can change its allocator. This is helpful when creating temporary particle buffers for filter operations and device-to-host copies.

## Additional background

https://github.com/ECP-WarpX/WarpX/pull/2860

## Checklist

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_Particles.H
+++ b/Src/Particle/AMReX_Particles.H
@@ -1220,6 +1220,35 @@ public:
     int NumRealComps () const { return NArrayReal + NumRuntimeRealComps(); }
     int NumIntComps  () const { return NArrayInt  + NumRuntimeIntComps() ; }
 
+    /** type trait to translate one particle container to another, with changed allocator */
+    template <template<class> class NewAllocator=amrex::DefaultAllocator>
+    using ContainerLike = amrex::ParticleContainer<NStructReal, NStructInt, NArrayReal, NArrayInt, NewAllocator>;
+
+    /** Create an empty particle container
+     *
+     * This creates a new AMReX particle container type with same compile-time
+     * and run-time attributes. But, it can change its allocator. This is
+     * helpful when creating temporary particle buffers for filter operations
+     * and device-to-host copies.
+     *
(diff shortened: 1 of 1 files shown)


atmyers:
What do you think about making this a static member function, so it would be used like

```
auto new_pc = ParticleContainer::make_alike<amrex::PinnedArenaAllocator>(pc);
```

i.e. more similar to, `std::make_unique`, etc...

ax3l:
Yes, I thought about this as well but that implementation needed a `const_cast` if `pc` is passed as a `const` argument. That's indeed how I first added it to WarpX:
https://github.com/ECP-WarpX/WarpX/pull/2860/commits/71d33f41f37f01f07735c75b7084dd081c952695

```C++
    static ParticleContainer
    make_alike (ParticleContainer const * pc)
    {
        auto gdb = const_cast<amrex::ParGDBBase*>(pc->GetParGDB());
        ParticleContainer tmp(gdb);

        // add runtime real comps to tmp
        for (int ic = 0; ic < pc->NumRuntimeRealComps(); ++ic) { tmp.AddRealComp(false); }

        // add runtime int comps to tmp
        for (int ic = 0; ic < pc->NumRuntimeIntComps(); ++ic) { tmp.AddIntComp(false); }

        return tmp;
    }
```
but let me try again here, the original code was more templated.

I think the main issue resides that we would like to pass `pc` as `ParticleContainer const *` not `ParticleContainer *`.

Another nitch: since functions do not support partial specialization, this needs to read somthing like:
```C++
auto tmp = ParticleContainer<..., amrex::PinnedArenaAllocator>::make_alike(pc);
```

Alternatively, we could maybe do an interface of the form:
```C++
auto tmp = ParticleContainer::make_alike(pc, amrex::PinnedArenaAllocator{});
```
...but the `const_cast` need would remain for the `GetParGDB()`.

atmyers:
As discussed offline, let's just leave this as a non-static member function, since it seems not that straightforward the other way.

atmyers (approved):

PR #2635, opened 2022-02-25
Doc + CMake: AMREX_EXPORT_DYNAMIC

ax3l:
## Summary

- [x] Add the CMake control for `AMREX_EXPORT_DYNAMIC`
- [ ] Add documentation (help wanted)
- [ ] Is this still needed on modern macOS or can we assume that we can roll with `dlopen` and these imports?

## Additional background

CMake builds currently generate empty backtraces on macOS/Darwin.
cc @RTSandberg

## Checklist

The proposed changes:

### DIFF
--- a/Docs/sphinx_documentation/source/BuildingAMReX.rst
+++ b/Docs/sphinx_documentation/source/BuildingAMReX.rst
@@ -495,6 +495,8 @@ The list of available options is reported in the :ref:`table <tab:cmakevar>` bel
    +------------------------------+-------------------------------------------------+-------------------------+-----------------------+
    | AMReX_BOUND_CHECK            |  Enable bound checking in Array4 class          | NO                      | YES, NO               |
    +------------------------------+-------------------------------------------------+-------------------------+-----------------------+
+   | AMReX_EXPORT_DYNAMIC         |  Enable backtrace on macOS                      | NO (unless Darwin)      | YES, NO               |
+   +------------------------------+-------------------------------------------------+-------------------------+-----------------------+
    | AMReX_SENSEI                 |  Enable the SENSEI in situ infrastucture        | NO                      | YES, NO               |
    +------------------------------+-------------------------------------------------+-------------------------+-----------------------+
    | AMReX_NO_SENSEI_AMR_INST     |  Disables the instrumentation in amrex::Amr     | NO                      | YES, NO               |
--- a/Tools/CMake/AMReXOptions.cmake
+++ b/Tools/CMake/AMReXOptions.cmake
@@ -331,6 +331,14 @@ print_option( AMReX_ASSERTIONS )
 option(AMReX_BOUND_CHECK  "Enable bound checking in Array4 class" OFF)
 print_option( AMReX_BOUND_CHECK )
 
+if("${CMAKE_SYSTEM_NAME}" MATCHES "Darwin")
+    set(AMReX_EXPORT_DYNAMIC_DEFAULT ON)
+else()
+    set(AMReX_EXPORT_DYNAMIC_DEFAULT OFF)
+endif()
+option( AMReX_EXPORT_DYNAMIC "Enable Backtrace for macOS/Darwin" ${AMReX_EXPORT_DYNAMIC_DEFAULT})
+print_option( AMReX_EXPORT_DYNAMIC )
+
 #
 # Profiling options  =========================================================
 #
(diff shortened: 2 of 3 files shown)


Docs/sphinx_documentation/source/BuildingAMReX.rst
```diff
@@ -495,6 +495,8 @@ The list of available options is reported in the :ref:`table <tab:cmakevar>` bel
    +------------------------------+-------------------------------------------------+-------------------------+-----------------------+
    | AMReX_BOUND_CHECK            |  Enable bound checking in Array4 class          | NO                      | YES, NO               |
    +------------------------------+-------------------------------------------------+-------------------------+-----------------------+
+   | AMREX_EXPORT_DYNAMIC         |  Something backtrace on macOS                   | NO (unless Darwin)      | YES, NO               |
```
ax3l: Fixme

Tools/CMake/AMReXOptions.cmake
```diff
@@ -331,6 +331,14 @@ print_option( AMReX_ASSERTIONS )
 option(AMReX_BOUND_CHECK  "Enable bound checking in Array4 class" OFF)
 print_option( AMReX_BOUND_CHECK )
 
+if("${CMAKE_SYSTEM_NAME}" MATCHES "Darwin")
+    set(AMREX_EXPORT_DYNAMIC_DEFAULT ON)
+else()
+    set(AMREX_EXPORT_DYNAMIC_DEFAULT OFF)
+endif()
+option( AMREX_EXPORT_DYNAMIC "Something something for macOS/Darwin" ${AMREX_EXPORT_DYNAMIC_DEFAULT})
```
ax3l: Fixme

Tools/CMake/AMReXSetDefines.cmake
```diff
@@ -76,6 +76,9 @@ add_amrex_define( AMREX_USE_ASSERTION NO_LEGACY IF AMReX_ASSERTIONS )
 # Bound checking
 add_amrex_define( AMREX_BOUND_CHECK NO_LEGACY IF AMReX_BOUND_CHECK )
 
+# Backtraces on macOS
```
ax3l, line 79: Fixme

Tools/CMake/AMReXOptions.cmake
etpalmer63: How about this for line 339:
```
option( AMReX_EXPORT_DYNAMIC "Enable backtrace for macOS/Darwin" ${AMREX_EXPORT_DYNAMIC_DEFAULT})
```

Docs/sphinx_documentation/source/BuildingAMReX.rst
etpalmer63:
```
   | AMReX_EXPORT_DYNAMIC         |  Enable backtrace on macOS                   |  NO (YES, if Darwin OS)     | YES, NO               | 
```
It's hard to succinctly describe that variable default option so I thought I would at least put up another version to think about.

Tools/CMake/AMReXSetDefines.cmake
etpalmer63: Do we want a lower case 'e' on the last one for line 80, i.e. `AMReX_EXPORT_DYNAMIC`?

Docs/sphinx_documentation/source/BuildingAMReX.rst
ax3l: Yes, I was wondering if there is more to it, thus I kept a placeholder that would raise suspicion for PR review :)

Tools/CMake/AMReXSetDefines.cmake
ax3l: yes, good catch! fixed :)

ax3l:
Discussed on Slack with @WeiqunZhang :
> Yes, it's related to dynamic loading of shared library.  Not sure if it's still needed.  Don't have a mac for testing.  I remember I used @jlvay's mac for the initial development of backtracing on mac.

So we could revisit if we hard-code the `dlopen` requirement at some point, maybe that was for older macs or it's still up-to-date and depends on the toolchain.

Anyhow, good enough for us to merge for now :) My mac is at work and I have my remote VNC not yet set up.

etpalmer63 (approved):
Looks good to me

ax3l:
Dang, I forgot to add this to `Tools/CMake/AMReX_Config.H.in`: #2650

PR #2649, opened 2022-03-08
Add int overflow assert check before Bcast wrapper

jmsexton03:
## Summary
* Adds AMREX_ASSERT in the functions which call ParallelDescriptior::Bcast
* Adds amrex::Abort() for InitRandom when serialize=true would cause too many particles to be sent from IOProc

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_FabArrayCommI.H
+++ b/Src/Base/AMReX_FabArrayCommI.H
@@ -585,7 +585,7 @@ FabArray<FAB>::copyTo (FAB& dest, int scomp, int dcomp, int ncomp, int nghost) c
     using T = typename FAB::value_type;
     if (ParallelContext::NProcsSub() > 1) {
         Long count = dest.numPts()*ncomp;
-        AMREX_ASSERT(count < std::numeric_limits<int>::max())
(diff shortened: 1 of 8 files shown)


Src/Base/AMReX_FabArrayCommI.H
```diff
@@ -585,6 +585,7 @@ FabArray<FAB>::copyTo (FAB& dest, int scomp, int dcomp, int ncomp, int nghost) c
     using T = typename FAB::value_type;
     if (ParallelContext::NProcsSub() > 1) {
         Long count = dest.numPts()*ncomp;
+        AMREX_ASSERT(count < std::numeric_limits<int>::max());
```
atmyers: I'm not sure that these asserts are needed, since the one in Bcast here will do the same thing: https://github.com/AMReX-Codes/amrex/blob/development/Src/Base/AMReX_ParallelDescriptor.H#L843-L864

and the message isn't really any more descriptive. However, the one in InitRandom does add useful information when this goes wrong.

atmyers (approved):

PR #2653, opened 2022-03-11
Introduce `AMReX_Version.H`

ax3l:
## Summary

The defined `AMREX_GIT_VERSION` and `AMREX_RELEASE_NUMBER` change in every commit. This modifies the `AMReX_Config.H` file as well for every commit, which is included in every header file.

Using always the latest commit counteracts `ccache` since the hash of a file included in every compilation call changes every commit.

Move this define and `AMREX_RELEASE_NUMBER` to a separate version file, which is included only where needed (currently in a single `.cpp` file).

- [x] CMake
- [x] GNUmake

## Additional background

https://github.com/ECP-WarpX/WarpX/pull/2959

## Checklist

The proposed changes:


jmsexton03:
Is there a compelling reason not to include this in `AMReX_buildInfo.cpp.in` as well, or did I just miss the connection?

ax3l:
You mean instead of `AMReX_Version.cpp`?
I took that route since we I did the same in WarpX, where `::Version` is a class member function.

But I see here it's a standalone function in the `amrex::` namespace, so we could indeed merge `AMReX_Version.cpp` into `AMReX_buildInfo.cpp.in`, I think...

Honestly, I am not sure. buildInfo is a bit involved for me.
I would let Weiqun comment on your idea tomorrow when he updates the GNUmake part.

etpalmer63 (approved):
I am in over my head on this one, but I didn't see anything that jumped out as odd. -- Thanks for including me!

sayerhs:
@jmsexton03 I am not sure buildinfo is a good idea as it is something application codes generate and include as part of their executables. The way @ax3l has setup we will always have `amrex::Version()` even when AMReX is built as an external library. This also maintains backward compatibility with the current `amrex` library for codes that don't necessarily use/generate BuildInfo as part of their executables.

jmsexton03:
@sayerhs @ax3l The current code looks good to me. My original curiosity was about whether something similar should be done for BuildInfo to better leverage `ccache`, but after further consideration the application codes which generate that probably wouldn't want those functions to be cached.

ax3l:
Ah, I understand. BuildInfo should only change in a generated `.cpp` file, thus that should be `ccache` friendly.

Src/Base/AMReX_Version.cpp
```diff
@@ -0,0 +1,14 @@
+#include <AMReX.H>
```
WeiqunZhang: We don't need to include AMReX.H here.

ax3l: I would, because it declares `amrex::Version ();`, no?
https://github.com/AMReX-Codes/amrex/blob/c5dce693676eb7ac505b64eb1f88fa9a8873b67c/Src/Base/AMReX.H#L63

ax3l: We could move that one to `AMReX_Version.H` though :)

sayerhs: You can define functions without declaring them? Was that @WeiqunZhang 's comment?

WeiqunZhang: If we move that to AMReX_Version.H, the users would have to include a new header to be able to call amrex::Version().

ax3l: Oh right, then let me remove my last commit again :D

WeiqunZhang: I would agree that we should include `AMReX_Version.H` so that the function declaration is included, if it were C.  But because of function overloading in C++, it's not really useful to have the declaration in AMReX_Version.cpp.

WeiqunZhang: I think we should keep the declaration of `amrex::Version()` in `AMReX.H` for backward compatibility.

ax3l: > You can define functions without declaring them? Was that @WeiqunZhang 's comment?

Probably. I always include the declaration when I define things, but we can remove including `AMReX.H` if you think it's unnecessary here.

ax3l: Updated. I will stop push-ing to the branch now, so you can work on GNUmake :)

ax3l: Ah dang. In that case let me rewrite the `.cpp` file once more:
```
../Src/Base/AMReX_Version.cpp:7:13: error: use of undeclared identifier 'amrex'
```

ax3l: @WeiqunZhang fixed. now for real

WeiqunZhang (approved):

ax3l:
Thank you @WeiqunZhang :blush:

PR #2656, opened 2022-03-15
Removed the use of H5Aget_storage_size

brtnfld:
## Summary
The HDF5 DAOS VOL does not support H5Aget_storage_size, so it was removed in favor of H5Tget_size instead.
With this change, the HDF5Benchmark test passes with the HDF5 DAOS VOL at ANL.

## Checklist

The proposed changes:


atmyers:
Does this look okay to you, @houjun?

houjun:
Looks good

atmyers (approved):

PR #2659, opened 2022-03-16
Doc: AMReX_TP_PROFILE

ax3l:
## Summary

Document third-party profiling options.

Doc-string taken from CMake `option()`.

## Additional background

This option confused one of our developers because `TP` also sounds like `TinyProfiler`, which is unrelated.
We should consider renaming this in a follow-up to `AMReX_THIRDPARTY_PROFILE` (both CMake & GNUmake).

## Checklist

The proposed changes:


kngott:
I'd argue for just changing it now, or removing it. It hasn't been maintained or used in a while, and I wouldn't want a bunch of users suddenly demanding that it do special things with it.

Maybe leave it but state it's for user customization?

@WeiqunZhang: What do you think?

WeiqunZhang:
I am OK with removing it.

WeiqunZhang (approved):

PR #2661, opened 2022-03-16
EB Flow

drangara:
## Summary
There are 2 sets of changes in this PR to support flow from/through embedded boundary faces

1) support for setting EB Dirichlet value in the nodal solver - this includes computation of surface integrals and adding the EB flow contribution to the divergence.
2) adding a new call for the method `EB_computeDivergence` which takes in the velocity at EB faces and adds the EB flow contribution to the divergence.

## Checklist

The proposed changes:


Src/EB/AMReX_EBMultiFabUtil.cpp
```diff
@@ -758,6 +758,50 @@ void EB_computeDivergence (MultiFab& divu, const Array<MultiFab const*,AMREX_SPA
...
+#endif
+    for (MFIter mfi(divu,info); mfi.isValid(); ++mfi)
+    {
+        const Box& bx = mfi.tilebox();
+        const auto& flagfab = flags[mfi];
+
+        if (flagfab.getType(bx) == FabType::singlevalued) {
+            const auto& vfrac = factory.getVolFrac();
+            const auto& bnorm = factory.getBndryNormal();
+            const auto& barea = factory.getBndryArea();
```
WeiqunZhang: We could move this outside MFIter.

Src/LinearSolvers/MLMG/AMReX_MLNodeLaplacian.cpp
```diff
@@ -82,21 +82,33 @@ MLNodeLaplacian::define (const Vector<Geometry>& a_geom,
...
+    m_eb_phi_dot_n.resize(m_num_amr_levels);
     for (int amrlev = 0; amrlev < m_num_amr_levels; ++amrlev)
     {
 #ifdef AMREX_USE_EB
         m_integral[amrlev] = std::make_unique<MultiFab>(m_grids[amrlev][0],
                                                         m_dmap[amrlev][0],
                                                         ncomp_i, 1, MFInfo(),
                                                         *m_factory[amrlev][0]);
+
+        m_surface_integral[amrlev] = std::make_unique<MultiFab>(m_grids[amrlev][0],
```
WeiqunZhang: We don't need to build this if `m_build_surface_integral` is false.  Right?

WeiqunZhang (changes requested):
Now I see why the surface integral multifab is always built.  However, instead of modifying `mlndlap_divu_eb`, we could add a new function (that is called in a new kernel) to add the extra contribution from surface intergral if there are any.

Could you use `Real(n.n)` or `n.n_rt` instead of `n.n`?  This is to avoid compiler warnings when we build in single precision.

WeiqunZhang (approved):

WeiqunZhang:
There are 2 sets of changes in this PR to support flow from/through embedded boundary faces

   *  support for setting EB Dirichlet value for velocity in the nodal solver - this includes computation of surface integrals and adding the EB flow contribution to the divergence.
   * adding a new call for the method EB_computeDivergence which takes in the velocity at EB faces and adds the EB flow contribution to the divergence.

PR #2667, opened 2022-03-21
BackTrace workarounds

jmsexton03:
## Summary
This fixes two different addr2line issues seen at OLCF using amrex

## Additional background
Uses `-gdwarf-4` instead of `-g` for gcc 11 to fix `/usr/bin/addr2line: Dwarf Error: Can't find .debug_ranges section`. This affects systems where binutils is too old. https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100725
Uses addr2line as a fallback if eu-addr2line gives results such as `??:0`

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_BLBackTrace.cpp
+++ b/Src/Base/AMReX_BLBackTrace.cpp
@@ -180,7 +180,6 @@ BLBackTrace::print_backtrace_info (FILE* f)
     if (strings != NULL) {
         int have_eu_addr2line = 0;
         int have_addr2line = 0;
-        int use_eu_addr2line = 0;
         std::string cmd;
         {
             have_eu_addr2line = file_exists("/usr/bin/eu-addr2line");
@@ -189,29 +188,9 @@ BLBackTrace::print_backtrace_info (FILE* f)
                 // cmd = "/usr/bin/eu-addr2line -C -f -i --pretty-print -p "
                 cmd = "/usr/bin/eu-addr2line -C -f -i -p "
                     + std::to_string(pid);
-
-                for (int i = 0; i < nentries; ++i)
-                {
-
-#if !defined(AMREX_USE_OMP) || !defined(__INTEL_COMPILER)
-                  std::string addr2line_result;
-                  if (amrex::system::call_addr2line && have_eu_addr2line) {
-                    if (bt_buffer[i] != nullptr) {
-                      char print_buff[32];
-                      std::snprintf(print_buff,sizeof(print_buff),"%p",bt_buffer[i]);
-                      const std::string full_cmd = cmd + " " + print_buff;
-                      addr2line_result = run_command(full_cmd);
-                    }
-                  }
-                  if (addr2line_result.find("AMReX_BLBackTrace.cpp:") != std::string::npos) {
-                    use_eu_addr2line = 1;
-                    break;
-                  }
-                }
-#endif
             }
         }
-        if (!have_eu_addr2line || !use_eu_addr2line) {
+        if (!have_eu_addr2line) {
             have_addr2line = file_exists("/usr/bin/addr2line");
             if (have_addr2line) {
                 cmd = "/usr/bin/addr2line -Cpfie " + amrex::system::exename;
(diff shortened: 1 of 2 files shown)


BenWibking:
I just ran into this problem with my code a few days ago. However, it exclusively uses CMake for builds. Any chance this PR could also change the default flags for CMake builds?

jmsexton03:
@BenWibking In principle changing the overall default is straightforward, however as far as I can tell most of the compiler version specific warning flags and so forth are only supported in GNUmake. This would make testing these defaults based on gcc version non-trivial.

I believe this is the relevant part of the [CMake flag defaults in AMReX](https://github.com/AMReX-Codes/amrex/blob/development/Tools/CMake/AMReXUtils.cmake#L121-L147). You could try setting `CMAKE_CXX_FLAGS_DEBUG="-gdwarf-4"` to verify whether this solution would work for your code.

Src/Base/AMReX_BLBackTrace.cpp
```diff
@@ -220,8 +220,14 @@ BLBackTrace::print_backtrace_info (FILE* f)
...
-            } else if (amrex::system::call_addr2line && have_addr2line &&
-                       !amrex::system::exename.empty())
+
+                if (addr2line_result.find('?') != std::string::npos) {
+                    addr2line_result.clear();
+                }
+            }
+
+            if ((!(amrex::system::call_addr2line && have_eu_addr2line) || addr2line_result.empty()) &&
+                amrex::system::call_addr2line && have_addr2line && !amrex::system::exename.empty())
```
WeiqunZhang:
```suggestion
            if (addr2line_result.empty() &&
                amrex::system::call_addr2line && have_addr2line && !amrex::system::exename.empty())
```
I think it can be simplified.

Src/Base/AMReX_BLBackTrace.cpp
```diff
@@ -188,9 +189,29 @@ BLBackTrace::print_backtrace_info (FILE* f)
...
+                    if (bt_buffer[i] != nullptr) {
+                      char print_buff[32];
+                      std::snprintf(print_buff,sizeof(print_buff),"%p",bt_buffer[i]);
+                      const std::string full_cmd = cmd + " " + print_buff;
+                      addr2line_result = run_command(full_cmd);
+                    }
+                  }
+                  if (addr2line_result.find("AMReX_BLBackTrace.cpp:") != std::string::npos) {
+                    use_eu_addr2line = 1;
+                    break;
```
WeiqunZhang: We should not break.

Src/Base/AMReX_BLBackTrace.cpp
```diff
@@ -246,7 +261,7 @@ BLBackTrace::print_backtrace_info (FILE* f)
                     if (!addr.empty()) {
                         const std::string full_cmd = cmd + " " + addr;
                         addr2line_result = run_command(full_cmd);
-                        if (addr2line_result.find('?') != std::string::npos) {
+                        if (addr2line_result.find("AMReX_BLBackTrace.cpp:") != std::string::npos) {
```
WeiqunZhang:
```suggestion
                        if (addr2line_result.find("AMReX_BLBackTrace.cpp:") == std::string::npos) {
```

Src/Base/AMReX_BLBackTrace.cpp
jmsexton03: The break is to avoid checking all the entries; should this check be moved into the for loop below where the print statements are instead of where `have_eu_addr2line` is tested?

jmsexton03: This change would remove the break, and the not equal should stay:
```suggestion
                  if (addr2line_result.find("AMReX_BLBackTrace.cpp:") != std::string::npos) {
                    use_eu_addr2line = 1;
```

WeiqunZhang: If the test is passed, the entry is good.  So we want to keep printing more, not break.

WeiqunZhang: I think it should be something like this
```
if (use_eu_addr2line == 0) { // still the initial value
    if (addr2line_result.find("AMReX_BLBackTrace.cpp:") != std::string::npos)) { // found AMReX_BLBackTrace.cpp:
        use_eu_addr2line = 1;
    }
}
```
Once `use_eu_addr2line` is 1, we don't have to do the test anymore.

Src/Base/AMReX_BLBackTrace.cpp
WeiqunZhang: Also if eu-addr2line is OK. nothing is actually printed.  Not what we want.

WeiqunZhang: Never mind my last comment.

Src/Base/AMReX_BLBackTrace.cpp
WeiqunZhang: Let me checkout the code.  Reading diff is hard.

jmsexton03: It sounds like it would be better to store both the command strings and make the test for BLBacktrace for eu-addr2line be inside the for loop which is printing, rather than trying to check it ahead of printing in another for loop. Copying the code like I did here might be less readable/maintainable

WeiqunZhang:
@jmsexton03 has tested it on Perlmutter and Crusher.  It seems to work well.

WeiqunZhang (approved):

PR #2251, opened 2021-08-15
Shorten CI tests and Enable CNS in CI

WeiqunZhang:
* Split macos ci tests into two checks
* Split hip ci tests
* Change some build types from Debug to RelWithDebInfo
* Enable Assertions and FPE in some tests
* Add some inputs for ci
* Enable CNS in CI

## Checklist

The proposed changes:
- [ ] fix a bug or incorrect behavior in AMReX
- [ ] add new capabilities to AMReX
- [ ] changes answers in the test suite to more than roundoff level
- [ ] are likely to significantly affect the results of downstream AMReX users
- [ ] are described in the proposed changes to the AMReX documentation, if appropriate

PR #1969, opened 2021-04-25
Workaround compile error on Apple M1 (ARM64)

sayerhs:
Fixes #1968

PR #2672, opened 2022-03-22
CI: Apple Silicon

ax3l:
## Summary

Make sure we can cross-compile AMReX from Intel CPUs (x86) to "Apple Silicon"/M1 (arm64). Using fancy universal binaries here (aka fat binaries) that even support both architectures at the same time.

## Additional background

Refs.:
- https://developer.apple.com/documentation/apple-silicon/building-a-universal-macos-binary
- https://cmake.org/cmake/help/latest/envvar/CMAKE_OSX_ARCHITECTURES.html

## Checklist

The proposed changes:

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -9,9 +9,12 @@ concurrency:
 jobs:
   # Build libamrex and all tests
   tests-macos-eb:
-    name: AppleClang@11.0 GFortran@9.3 [tests-eb]
+    name: AppleClang@11.0 GFortran@9.3 Universal [tests-eb]
     runs-on: macos-latest
-    env: {CXXFLAGS: "-fno-operator-names -Werror -Wall -Wextra -Wpedantic -Wnull-dereference -Wfloat-conversion -Wshadow -Woverloaded-virtual -Wextra-semi -Wunreachable-code -Wno-c++17-extensions -Wno-range-loop-analysis -Wno-pass-failed"}
+    env:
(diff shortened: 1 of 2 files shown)


.github/workflows/macos.yml
```diff
@@ -29,9 +32,7 @@ jobs:
             -DCMAKE_CXX_COMPILER_LAUNCHER=$(which ccache)
         cmake --build build --parallel 2
 
-        cd build
-        ctest --output-on-failure
-        cd ..
+        ctest --test-dir build --output-on-failure
 
   # Build libamrex and all tests
```
ax3l, line 37: @WeiqunZhang do we need that 2nd test?
I am ok to only build one macOS build with EB on and cover non-EB code paths on other platforms, but I might have missed the reason to split in #2251.

ax3l: Ah, universal builds need universal dependencies (like MPI) as well. In that case I keep two runs here and make one no-MPI.

ax3l:
@WeiqunZhang ready for merge :)

WeiqunZhang:
Oh, I was waiting on this https://github.com/AMReX-Codes/amrex/pull/1969

ax3l:
ah thanks, didn't see the ping :) Updated accordingly.

PhilMiller:
The amended architecture check macros look to be correct, and I checked that they're comprehensive of everywhere in AMReX (or WarpX) that CPU ISA is tested.

WeiqunZhang (approved):

PR #2673, opened 2022-03-22
Redefine ParticleContainter multifab after setting particle boxarray and dmap

RevathiJambunathan:
## Summary
This PR calls the function to redefine the multifab in particle container base when it is set to have a new particle box array and distribution map.

## Additional background
In WarpX, we have diagnostics that need a different particle box array and distribution map compared to that used for the simulation. The current development branch was leading to a segfault when writing particles because it found that all the particles were out of range (when they were within the range for the diagnostics). This is because the ParIter was iterating over the boxes from the simulation and not from the newly set particle box array. This PR fixes this incorrect behavior.

Thank you @atmyers for your help!

## Checklist

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_ParticleContainerBase.cpp
+++ b/Src/Particle/AMReX_ParticleContainerBase.cpp
@@ -75,6 +75,7 @@ void ParticleContainerBase::SetParticleBoxArray (int lev, const BoxArray& new_ba
                           m_gdb->ParticleBoxArray(), m_gdb->refRatio());
     m_gdb = &m_gdb_object;
     m_gdb->SetParticleBoxArray(lev, new_ba);
+    RedefineDummyMF(lev);
 }
 
 void ParticleContainerBase::SetParticleDistributionMap (int lev, const DistributionMapping& new_dmap)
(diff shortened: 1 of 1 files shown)


Src/Particle/AMReX_ParticleContainerBase.cpp
```diff
@@ -83,6 +84,7 @@ void ParticleContainerBase::SetParticleDistributionMap (int lev, const Distribut
                           m_gdb->ParticleBoxArray(), m_gdb->refRatio());
     m_gdb = &m_gdb_object;
     m_gdb->SetParticleDistributionMap(lev, new_dmap);
+    RedefineDummyMF(lev);
 }
 
 void ParticleContainerBase::SetParticleGeometry (int lev, const Geometry& new_geom)
```
ax3l, line 91: Here as well? Looks so similar... (new `gdb` object assigned)

RevathiJambunathan: Yeah. I had the same question and learned that it is not needed here because the geometry does not affect the iterator

ax3l: Cool, thanks for clarifying! :)

ax3l:
Oh, looks like the `Particles_GhostsAndVirtuals` test fails now.

atmyers (approved):

PR #2677, opened 2022-03-24
Catch CUDA 11.6 bug at compile time

### LINKED ISSUE #2607: AMReX CUDA issues with CUDA 11.6
If you build `amrex-tutorials/ExampleCodes/Amr/Advection_AmrCore` with CUDA 11.6 and run as:

```
./main2d.gnu.MPI.CUDA.ex inputs geometry.is_periodic=1 1
```

we get:

```
amrex::Abort::0::GPU last error detected in file /home/zingale/development/amrex//Src/Base/AMReX_GpuLaunchFunctsG.H line 834: invalid device function !!!
SIGABRT
See Backtrace.0 file for details
```

a similar issue is happening with Castro with the latest CUDA.  This does not appear to be an issue with CUDA 11.4

WeiqunZhang:
Make it a compile time error if CUDA 11.6 is used to instantiate
GpuBndryFuncFab.  This used to be a mysterious runtime error (#2607).  Note
that if GpuBndryFuncFab is not used in your code, you can still use CUDA
11.6.  However, the bug might still be triggered at runtime in other places.

Closes #2607.

### DIFF
--- a/Src/Base/AMReX_BCUtil.cpp
+++ b/Src/Base/AMReX_BCUtil.cpp
@@ -1,6 +1,9 @@
 #include <AMReX_BCUtil.H>
 #include <AMReX_PhysBCFunct.H>
 
+// CUDA 11.6 bug: https://github.com/AMReX-Codes/amrex/issues/2607
+#if !defined(__CUDACC__) || (__CUDACC_VER_MAJOR__ != 11) || (__CUDACC_VER_MINOR__ != 6)
+
 namespace amrex
 {
 
@@ -53,3 +56,5 @@ void FillDomainBoundary (MultiFab& phi, const Geometry& geom, const Vector<BCRec
 }
 
 }
+
+#endif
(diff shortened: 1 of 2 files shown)


jmsexton03 (approved):

Src/Base/AMReX_BCUtil.cpp
```diff
@@ -1,6 +1,9 @@
 #include <AMReX_BCUtil.H>
 #include <AMReX_PhysBCFunct.H>
 
+// CUDA 11.6 bug: https://github.com/AMReX-Codes/amrex/issues/2607
+#if !defined(__CUDACC__) || (__CUDACC_VER_MAJOR__ != 11) || (__CUDACC_VER_MINOR__ != 6)
```
ax3l, line 5: Do we want to also check the patch version, since @maxpkatz said it will be fixed in the next minor release?

https://github.com/AMReX-Codes/amrex/issues/2607#issuecomment-1076540867

As far as I know, "Update" and minor releases in CTK these days are some patch level change.

WeiqunZhang: Definitely, if we know the patch number.

ax3l: @maxpkatz does CUDA 11.6.2 (or even 11.6.1?) fix this/#2607?
https://github.com/spack/spack/pull/29756

WeiqunZhang: Does not seem to be the case.  I was using 11.6.1 for this PR.  I also just tested 11.6.2.  Still saw the same "invalid device function" error.

PR #2679, opened 2022-03-25
Use the roundoff domain in enforcePeriodic.

atmyers:
This fixes some roundoff issues seen with single precision runs on Fugaku.

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_Geometry.H
+++ b/Src/Base/AMReX_Geometry.H
@@ -193,6 +193,14 @@ public:
         return {{AMREX_D_DECL(prob_domain.hi(0),prob_domain.hi(1),prob_domain.hi(2))}};
     }
 
+    GpuArray<Real,AMREX_SPACEDIM> RoundoffLoArray () const noexcept {
+        return {{AMREX_D_DECL(roundoff_domain.lo(0),roundoff_domain.lo(1),roundoff_domain.lo(2))}};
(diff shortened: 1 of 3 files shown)


jmsexton03 (approved):

atmyers:
From @WeiqunZhang in slack:

```
I am still worried about it when Real and ParticelReal are different.
What I am worried about are two things.  (1) The bisection is doing (Real - Real)*Real, whereas the particle codes do (ParticleReal-Real)*Real.  (2) Roundoff domain the hi is stored in Real.  It's cast ParticleReal when we use it in enforcePeriodic.  It's not clear to me that's 100% robust.
So I think instead of RealBox roundoff_domain, we should have GpuArray<double,AMREX_SPACEDIM> roundoff_hi_d and GpuArray<float,AMREX_SPACEDIM> roundoff_hi_f.
The bisection should mimic exactly (double_or_float - Real) * Real.
In particle codes, we can have a function that always return the roundoff hi in ParticleReal.   That function can contains ifdef to decide which one to use, roundoff_hi_f or roundoff_hi_d.
Anymore I think the PR is good enough for WarpX right now.  So feel free to merge it as is.
But in long term, it might bite us again.
```

Since this is time-critical for @lucafedeli88 I will merge as is, but I will address the mixed precision issue soon.

PR #2683, opened 2022-03-30
FillBoundaryAndSync

WeiqunZhang:
Add a new function to FabArray that fills ghost cells and synchronize nodal
data. Ghost regions are filled with data from the intersecting valid
regions. The synchronization will override valid regions by the intersecting
valid regions with a higher precedence.  The smaller the global box index
is, the higher precedence the box has.  With periodic boundaries, for cells
in the same box, those near the lower corner have higher precedence than
those near the upper corner.  This function produces the same result as
OverrideSync followed by FillBoundary in a single pass.

Also modify the implementation of OverrideSync to use this new function,
which is more efficient.

The nodal linear solvers use OverrideSync to sync up the nodal data.  They
have been updated to use the new OverrideSync function that does not use
an owner mask.  So we no longer need to build owner masks except for the
top and bottom multigrid levels that need to use the mask in the computation
of norms to avoid overcounting.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_FabArrayBase.cpp
+++ b/Src/Base/AMReX_FabArrayBase.cpp
@@ -23,6 +23,7 @@
 #endif
 
 #include <algorithm>
+#include <utility>
 
 namespace amrex {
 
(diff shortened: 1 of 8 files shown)


WeiqunZhang:
It has passed regression tests of amrex and incflo on GPU and CPU.

Src/Base/AMReX_FabArrayBase.cpp
```diff
@@ -1030,6 +1034,133 @@ FabArrayBase::FB::define_epo (const FabArrayBase& fa)
...
+        // build tags as the receiver.
+
+        for (auto const& krcv : my_receiver) {
+            tag_one_box(krcv, ba, dm, false);
+        }
+    }
+#endif
+
+    // No need to sort send and recv tags because they are already sorted
+    // dut to the way they are built.
```
atmyers:
```suggestion
    // due to the way they are built.
```

Src/LinearSolvers/MLMG/AMReX_MLNodeLinOp.H
```diff
@@ -127,7 +127,8 @@ protected:
 
     virtual void resizeMultiGrid (int new_size) override;
 
-    Vector<Vector<std::unique_ptr<iMultiFab> > > m_owner_mask;      // ownership of nodes
+    std::unique_ptr<iMultiFab> m_owner_mask_top;  // ownership of nodes
```
atmyers, line 130: Could you add something to the PR description about the changes to MLNodeLinOp?

atmyers (approved):

PR #2684, opened 2022-03-31
Minor update to Docs on Linear solver curvilinear coordinate

cgilet:
support.

## Checklist

The proposed changes:

### DIFF
--- a/Docs/sphinx_documentation/source/LinearSolvers.rst
+++ b/Docs/sphinx_documentation/source/LinearSolvers.rst
@@ -388,13 +388,21 @@ to the ghost cell center; :cpp:`maxorder = 3` uses the boundary value and the fi
 Curvilinear Coordinates
 =======================
 
-The linear solvers support curvilinear coordinates including 1D
+The cell-centered linear solvers `MLABecLaplacian` and `MLPoisson`
+support curvilinear coordinates including 1D
 spherical and 2d cylindrical :math:`(r,z)`.  In those cases, the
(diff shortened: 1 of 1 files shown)


Docs/sphinx_documentation/source/LinearSolvers.rst
```diff
@@ -388,14 +388,22 @@ to the ghost cell center; :cpp:`maxorder = 3` uses the boundary value and the fi
 Curvilinear Coordinates
 =======================
 
-The linear solvers support curvilinear coordinates including 1D
```
WeiqunZhang, line 391: The nodal solver supports RZ too, at least partially.  It does require the user to rescale the problem.  But the solver itself also has to do some additional correction. (`git grep is_rz Src/LinearSolvers/MLMG`)

cgilet: This is what I was trying to say in the next paragraph, but obviously, I was not clear. I'll try again...

cgilet: @WeiqunZhang Can you clarify if `setMetricTerm(false)` actually turns off the correction terms for the nodal solver or not. As I read the code, it appears to me as though that doesn't affect the nodal solver, and instead `setRZCorrection(false)` must be used.

WeiqunZhang: That's correct.

WeiqunZhang (approved):

PR #2690, opened 2022-04-01
Semicoarsening in WarpX Linear Solvers

WeiqunZhang:
Add support for semicoarsening in two linear solvers used by WarpX,
MLNodeTensorLap and EBNodeFDLaplacian.  The semicoarsening approach here is
different from the existing approach for the nodal projection solver.  There
the grids are coarsened regularly as much as possible and then we switch to
coarsening in directions that are still coarsenable, whereas here we do
semicoarsening first in specified direction and then switch to regular
coarsening if the specified maximum semicoarsening level has not been
reached.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_BoxArray.H
+++ b/Src/Base/AMReX_BoxArray.H
@@ -643,9 +643,9 @@ public:
     BoxArray& coarsen (int refinement_ratio);
 
     //! Coarsen each Box in the BoxArray to the specified ratio.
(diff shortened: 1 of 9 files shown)


WeiqunZhang:
It has passed regression tests of AMReX, IAMR, incflo, and MFiX-Exa.

atmyers (approved):

PR #2694, opened 2022-04-05
Use pc.WritePlotFile instead of pc.Checkpoint in Tests.

atmyers:
This allows the particles in these plotfiles to be read by e.g. Paraview.

I've also fixed a bug in the `enable_if` in one of the WritePlotFile overloads that was needed for this to work.

I'll fix the ones in amrex-tutorials in a later PR.

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_ParticleIO.H
+++ b/Src/Particle/AMReX_ParticleIO.H
@@ -234,7 +234,7 @@ WritePlotFile (const std::string& dir, const std::string& name,
 
 template <int NStructReal, int NStructInt, int NArrayReal, int NArrayInt,
           template<class> class Allocator>
-template <class F, typename std::enable_if<!std::is_same<F, Vector<std::string>>::value>::type*>
+template <class F, typename std::enable_if<!std::is_same<F, Vector<std::string>&>::value>::type*>
 void
 ParticleContainer<NStructReal, NStructInt, NArrayReal, NArrayInt, Allocator>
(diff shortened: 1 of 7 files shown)


etpalmer63 (approved):

etpalmer63:
Just a note for Paraview particle compatibility: **By default**, the directory inside the plotfile has to be named "particles" for Paraview to process it correctly. This can be done by renaming after the simulation runs, or by modifying the source code (i.e. `TracerPC->WritePlotFile(dir, "Tracers");` :arrow_right: `TracerPC->WritePlotFile(dir, "particles");`) . **On the other hand**, if your particle directory is called "Tracers", another workaround would be to the ignore the error message when importing the particle data, and change the "Particle Type" field to "Tracers".  When you click apply, Paraview will read the particle data from the correct directory [see [AMReX User's Guide - Paraview](https://amrex-codes.github.io/amrex/docs_html/Visualization.html#paraview)]. 

Second comment: Should we also include the change in the HDF5 Benchmark example? 
```
#ifdef AMREX_USE_HDF5
            myPC.CheckpointHDF5(fname, "particle0", false, particle_realnames, particle_intnames, compression);                     
#else
            myPC.Checkpoint(fname, "particle0", false, particle_realnames, particle_intnames);
            /* myPC.WriteAsciiFile("particle0_ascii"); */
#endif
```

atmyers:
I believe the HDF5 test checks actual restarts, so we should leave it as Checkpoint there.

PR #2695, opened 2022-04-05
Make the ParticleReduce functions not rely on ParIter.

atmyers:
A potentially useful pattern that came up in ImpactX is to add particles to a container on grid 0, compute the extrema of the particle positions, set the box size, then call Redistribute(). However, this currently doesn't work because the ParticleReduce functions rely on ParIter, which require the Redistribute method to have already been called after adding the particles on grid 0. This changes the ParticleReduce functions to not rely on ParIter so the above order of operations will work.

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_ParticleReduce.H
+++ b/Src/Particle/AMReX_ParticleReduce.H
@@ -121,12 +121,19 @@ ReduceSum (PC const& pc, int lev_min, int lev_max, F&& f) -> decltype(f(typename
         for (int lev = lev_min; lev <= lev_max; ++lev)
         {
             const auto& plev = pc.GetParticles(lev);
-#ifdef AMREX_USE_OMP
-#pragma omp parallel if (!system::regtest_reduction) reduction(+:sm)
-#endif
-            for (const auto& kv : plev)
+            Vector<std::pair<int, int> > grid_tile_ids;
+            Vector<const typename PC::ParticleTileType*> ptile_ptrs;
+            for (auto& kv : plev)
             {
-                const auto& tile = plev.at(std::make_pair(kv.first.first, kv.first.second));
+                grid_tile_ids.push_back(kv.first);
+                ptile_ptrs.push_back(&(kv.second));
+            }
+#ifdef AMREX_USE_OMP
+#pragma omp parallel for if (!system::regtest_reduction) reduction(+:sm)
+#endif
+            for (int pmap_it = 0; pmap_it < static_cast<int>(ptile_ptrs.size()); ++pmap_it)
+            {
+                const auto& tile = plev.at(grid_tile_ids[pmap_it]);
                 const auto np = tile.numParticles();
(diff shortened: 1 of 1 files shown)


jmsexton03 (approved):

Src/Particle/AMReX_ParticleReduce.H
```diff
@@ -102,9 +101,10 @@ ReduceSum (PC const& pc, int lev_min, int lev_max, F&& f) -> decltype(f(typename
 
         for (int lev = lev_min; lev <= lev_max; ++lev)
         {
-            for(ParIter pti(pc, lev); pti.isValid(); ++pti)
+            const auto& plev = pc.GetParticles(lev);
+            for (const auto& kv : plev)
             {
-                const auto& tile = pti.GetParticleTile();
+                const auto& tile = plev.at(std::make_pair(kv.first.first, kv.first.second));
```
ax3l, line 109: Possibly add the `<utility>` include for this in the file :)

Src/Particle/AMReX_ParticleReduce.H
```diff
@@ -120,16 +120,25 @@ ReduceSum (PC const& pc, int lev_min, int lev_max, F&& f) -> decltype(f(typename
     {
         for (int lev = lev_min; lev <= lev_max; ++lev)
         {
+            const auto& plev = pc.GetParticles(lev);
+            Vector<std::pair<int, int> > grid_tile_ids;
```
ax3l, line 126: Include we could add: `<AMReX_Vector.H>`

ax3l:
Thanks a lot, awesome :+1: 

> ParticleReduce functions rely on ParIter, which require the Redistribute method to have already been called after adding the particles on grid 0

Just curious, what requirement does the `ParIter` have that `Redistribute` adds?

atmyers:
The issue is that the particles have been added to grid 0, tile 0, which is not a valid on all processes. So `ParIter` can't be used until after those have been migrated to the right place in the Redistribute call.

ax3l:
Ah, that makes sense. Thanks for the pointer & clarifying! :)

Src/Particle/AMReX_ParticleReduce.H
```diff
@@ -102,9 +103,10 @@ ReduceSum (PC const& pc, int lev_min, int lev_max, F&& f) -> decltype(f(typename
 
         for (int lev = lev_min; lev <= lev_max; ++lev)
         {
-            for(ParIter pti(pc, lev); pti.isValid(); ++pti)
+            const auto& plev = pc.GetParticles(lev);
+            for (const auto& kv : plev)
             {
-                const auto& tile = pti.GetParticleTile();
+                const auto& tile = plev.at(std::make_pair(kv.first.first, kv.first.second));
```
WeiqunZhang, line 109: Is `std::make_pair` needed? Is it equivalent to `plev.at(kv.first)`?

PR #2698, opened 2022-04-06
add Vector initialization to RealVect

RTSandberg:
## Summary
add a Vector-based initialization of amrex::RealVect

## Additional background
pyAMReX benefits from a list initialization for the wrapper of RealVect. This is facilitated by an amrex::Vector-based initialization of amrex::RealVect

## Checklist

The proposed changes:
- [X ] add new capabilities to AMReX


WeiqunZhang:
`amrex::Vector` is derived from `std::vector`.  So there is no strong reason to change the constructor of `RealVect` to take `const amrex::Vector<amrex::Real>&` instead of `const std::vector<amrex::Real>&` because the current version in the development branch works for both `amrex::Vector` and `std::vector`, whereas this PR will only work for `amrex::Vector`.  This might break people's codes.

WeiqunZhang:
There are two changes in this PR.  To be clearer, adding `explicit RealVect (const Array<Real,AMREX_SPACEDIM>& ar) noexcept` is fine, but I don't think we should modify the std::vector version.  So I am requesting that the change to the vector version be reverted, unless there is a good reason for it.

RTSandberg:
Thanks for your feedback Weiqun.  I understand the issues with the vector version and have no compelling reason for making a change.  The array version may be useful, but I don't need it and it would be cleaner to do through a new PR.  I'll close this PR

PR #2705, opened 2022-04-09
Handle the case where we don't have enough device memory for the snd_buffer

atmyers:
## Checklist

The proposed changes:


Src/Particle/AMReX_ParticleCommunication.H
```diff
@@ -309,6 +309,10 @@ void packBuffer (const PC& pc, const ParticleCopyOp& op, const ParticleCopyPlan&
     {
         total_buffer_size = plan.m_snd_offsets.back();
     }
+
+    if (total_buffer_size > Gpu::Device::freeMemAvailable()) {
```
WeiqunZhang: (1) Maybe the arena has enough memory even though `Gpu::Device::freeMemAvailable` does not.  (2) The arena rounds up the allocation size.  So the test is not always safe.  (3) The arena might attempt to release memory first.  So maybe we could add a new function to the arena to return if it is able to allocate this much memory.

WeiqunZhang (approved):

PR #2708, opened 2022-04-11
Allow `FillBoundary` to cast messages from double to single precision.

hengjiew:
## Summary
For the GPU implementation, this PR allows `FillBoundary` to pack double-precision data into a single-precision buffer and unpack single-precision messages to double-precision FabArray, e.x. `mf.template FillBoundary<float>(...)`. 

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_FBI.H
+++ b/Src/Base/AMReX_FBI.H
@@ -17,41 +17,40 @@ namespace detail {
 
 #ifdef AMREX_USE_GPU
 
-template <class T>
+template <class T0, class T1>
 struct CellStore
 {
     AMREX_GPU_DEVICE AMREX_FORCE_INLINE void
-    operator() (T* d, T s) const noexcept
+    operator() (T0* d, T1 s) const noexcept
     {
-        *d = s;
+      *d = static_cast<T0>(s);
     }
 };
 
-template <class T>
+template <class T0, class T1>
 struct CellAdd
 {
     AMREX_GPU_DEVICE AMREX_FORCE_INLINE void
-    operator() (T* d, T s) const noexcept
+    operator() (T0* d, T1 s) const noexcept
     {
-        *d += s;
+        *d += static_cast<T0>(s);
     }
 };
 
-template <class T>
+template <class T0, class T1>
 struct CellAtomicAdd
 {
-    template<class U=T,
-             std::enable_if_t<amrex::HasAtomicAdd<U>::value,int> = 0>
+    template<class U0=T0, std::enable_if_t<amrex::HasAtomicAdd<U0>::value,int> = 0>
     AMREX_GPU_DEVICE AMREX_FORCE_INLINE void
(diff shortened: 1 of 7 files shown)


Src/Base/AMReX_FBI.H
```diff
@@ -17,49 +17,49 @@ namespace detail {
...
 };
 
-template <class T>
+template <class T0, class T1>
 struct CellAtomicAdd
 {
-    template<class U=T,
-             std::enable_if_t<amrex::HasAtomicAdd<U>::value,int> = 0>
+    template<class U0=T0, std::enable_if_t<amrex::HasAtomicAdd<U0>::value,int> = 0,
+             class U1=T1, std::enable_if_t<amrex::HasAtomicAdd<U1>::value,int> = 0>
```
WeiqunZhang: We don't need to test if `T1` has atomic add because the value of `T1` will be cast to `T0`.

WeiqunZhang (changes requested):
LGTM.  I suggest we make the second typename required in `CellStore`, `CellAdd`, `CellAtomicAdd`, `fab_to_fab`, fab_to_fab_atomic_cpy`, and `fab_to_fab_atomic_add`.  These are not public, so we don't have to worry about API changes.  Also could we do this for CPU too?  We may need to modify functions like `BaseFab::copyFromMem` by adding an extra optional template parameter.

Src/Base/AMReX_FBI.H
hengjiew: > LGTM. I suggest we make the second typename required in `CellStore`, `CellAdd`, `CellAtomicAdd`, `fab_to_fab`, fab_to_fab_atomic_cpy`, and `fab_to_fab_atomic_add`. These are not public, so we don't have to worry about API changes. Also could we do this for CPU too? We may need to modify functions like `BaseFab::copyFromMem` by adding an extra optional template parameter.

`BaseFab::copyFromMem` returns the size of the copied data. For casting and copying from a float buffer to a double fab, which return value makes more sense, sizeof(float)*N or sizeof(double)*N?

hengjiew:
> LGTM. I suggest we make the second typename required in `CellStore`, `CellAdd`, `CellAtomicAdd`, `fab_to_fab`, fab_to_fab_atomic_cpy`, and `fab_to_fab_atomic_add`. These are not public, so we don't have to worry about API changes. Also could we do this for CPU too? We may need to modify functions like `BaseFab::copyFromMem` by adding an extra optional template parameter.

`BaseFab::copyFromMem` returns the size of the copied data. For casting and copying from a float buffer to a double fab, which return value makes more sense, sizeof(float)*N or sizeof(double)*N?

WeiqunZhang:
The size in the 1d buffer, because the reason for returning the size is for the caller to increment the pointer to the buffer.

WeiqunZhang (approved):

PR #2522, opened 2021-12-08
[WIP] add a python script to remove the finest level from plotfile

zingale:
## Checklist

The proposed changes:
- [ ] fix a bug or incorrect behavior in AMReX
- [x] add new capabilities to AMReX
- [ ] changes answers in the test suite to more than roundoff level
- [ ] are likely to significantly affect the results of downstream AMReX users
- [ ] include documentation in the code and/or rst files, if appropriate

PR #2730, opened 2022-04-26
[WIP] add fscale plotfile tool

BenWibking:
## Summary
This adds `fscale` to Tools/Plotfile. This tool divides the RealBox coordinates (i.e. probLo, probHi) by the input length scale _length_unit_ and writes out a new plotfile with the rescaled geometry.

## Additional background
This is necessary to work around bugs in Paraview for large coordinate values. This happens, e.g., when the geometry is a parsec-size box with coordinates in cgs units. See https://github.com/AMReX-Codes/amrex/discussions/2637

## Checklist

The proposed changes:

### DIFF
--- a/Tools/Plotfile/CMakeLists.txt
+++ b/Tools/Plotfile/CMakeLists.txt
@@ -15,6 +15,7 @@ set(_exe_names
    fsnapshot
    ftime
    fvarnames
+   fscale
    )
 
 # Build targets one by one
--- a/Tools/Plotfile/GNUmakefile
+++ b/Tools/Plotfile/GNUmakefile
@@ -21,6 +21,7 @@ ifeq ($(strip $(programs)),)
   programs += fsnapshot
   programs += ftime
   programs += fvarnames
+  programs += fscale
 endif
 
 include $(AMREX_HOME)/Tools/GNUMake/Make.defs
--- a/Tools/Plotfile/fscale.cpp
+++ b/Tools/Plotfile/fscale.cpp
@@ -89,10 +89,16 @@ void main_main() {
   }
 
   int m_spacedim;
-  int m_finest_level, m_nlevels;
+  is >> m_spacedim;
+  os << m_spacedim << '\n';
+
   Real m_time;
-  is >> m_spacedim >> m_time >> m_finest_level;
-  os << m_spacedim << '\n' << m_time << '\n' << m_finest_level << '\n';
+  is >> m_time;
+  os << std::setprecision(30) << m_time << '\n';
+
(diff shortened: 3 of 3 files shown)


WeiqunZhang:
Since only a few lines of `Header` needs to be modified, maybe we should not read and write the binary data.  Also, the header file may contain more than what's in a generic header written by `WriteMultiLevelPlotfile`.  So I suggest we rename the header file first, read line by line and write to a new header.  For lines related to the physical size, we modify and write to the new header file.

jmsexton03:
This scales only the problem domain, not fields which also might relate to the length scale such as velocity, right?

BenWibking:
> Since only a few lines of `Header` needs to be modified, maybe we should not read and write the binary data. Also, the header file may contain more than what's in a generic header written by `WriteMultiLevelPlotfile`. So I suggest we rename the header file first, read line by line and write to a new header. For lines related to the physical size, we modify and write to the new header file.

That makes sense. I am not familiar with the header file format. Can you write a function for manipulating the header? Or is there an example of this somewhere else?

BenWibking:
> This scales only the problem domain, not fields which also might relate to the length scale such as velocity, right?

Yes, that's correct. This would also mean the user would need to manually rescale the output of fvolumesum, etc. If there's a way to store the scaling factor in the Header file, that could be a partial solution.

WeiqunZhang:
You can use `PlotFileDataImpl::PlotFileDataImpl` as an example of how to read Geometry information.  For most of the lines, you can use `std::getline` and write out the string.  Once the Geometry information has been written out, you can then do `while(std::getline(..))` to read and write the rest of the Header.

zingale:
not sure if it helps, but I started a python header-mangler here:

https://github.com/AMReX-Codes/amrex/pull/2522

BenWibking:
> You can use `PlotFileDataImpl::PlotFileDataImpl` as an example of how to read Geometry information. For most of the lines, you can use `std::getline` and write out the string. Once the Geometry information has been written out, you can then do `while(std::getline(..))` to read and write the rest of the Header.

Ok, I've attempted to do that in the new version. If there's a cleaner way to do it, let me know. Right now, it does not yet change the Box data in the header, only the prob_lo, prob_hi and dx values.

I'm not sure if it should overwrite the existing plotfile or copy the whole directory and then overwrite the Header in the new plotfile. For now, it overwrites or creates the header in the output plotfile directory, assuming it already exists.

BenWibking:
We ended up working around this by avoiding Paraview, and I never got around to cleaning this up. I will close for now, can be re-opened if anyone is interested in the future.

PR #2732, opened 2022-04-26
Add methods for resetting the ParGDB of a ParticleContainer

atmyers:
This is useful for when you want to add or remove levels of refinement, but you have *not* used the `AmrCore` version of the constructor.

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_ParticleContainerBase.H
+++ b/Src/Particle/AMReX_ParticleContainerBase.H
@@ -97,8 +97,8 @@ public:
 
     bool isDefined () const { return m_gdb != nullptr; }
 
-    void reserveData ();
-    void resizeData ();
+    virtual void reserveData ();
+    virtual void resizeData ();
     void RedefineDummyMF (int lev);
 
     MFIter MakeMFIter (int lev, const MFItInfo& info) const {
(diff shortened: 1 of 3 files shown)


WeiqunZhang:
Why do `ParticleContainerBase::resizeData` and `reserveData` need to be virtual now?  The reason I ask is that it seems to hint that one is using a pointer to ParticleContainerBase to call `resizeData`.  In that case, the dtor of ParticleContainerBase might have to be virtual too.

atmyers:
Calling `resizeData` using a base class pointer is not the intended use. The issue is that `resizeData` in the base class cannot resize the `m_particles` member, because that needs to know about the template parameters. However, this does need to be done after changing the number of levels in the container using the new methods. So I want the base class `SetParGDB` methods to call the derived class version of `resizeData`.

WeiqunZhang:
Technically calling `resizeData` in `SetParGDB` is still calling `resizeData` using a base class pointer, because it's `this->resizeData()`.  Anyway, it won't hurt to add `virtual ~ParticleContainerBas()`, since the class already has virtual function table.

RevathiJambunathan (approved):
Thank you for this PR, @atmyers 
The crash I was seeing before with BTD is not seen anymore when I used this branch.
I added the changes in this PR to WarpX https://github.com/ECP-WarpX/WarpX/pull/3056 
Thanks a lot for your help!

WeiqunZhang (approved):

PR #2734, opened 2022-04-27
Turn on managed memory by default in The_Arena for HIP

WeiqunZhang:
This makes it have the same behavior as CUDA and DPC++ builds.

## Checklist

The proposed changes:

### DIFF
--- a/Docs/sphinx_documentation/source/GPU.rst
+++ b/Docs/sphinx_documentation/source/GPU.rst
@@ -451,8 +451,8 @@ Memory Allocation
 
 To provide portability and improve memory allocation performance,
 AMReX provides a number of memory pools.  When compiled without
-CUDA, all :cpp:`Arena`\ s use standard :cpp:`new` and :cpp:`delete`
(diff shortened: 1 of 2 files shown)


jmsexton03:
Should this parameter also be included in the table here to reinforce that the default behavior would be changing?
[Input Parameters section
](https://amrex-codes.github.io/amrex/docs_html/GPU.html?highlight=is_managed#inputs-parameters)
Alternatively, can we add a comment in the code in the section it's already mentioned in, to indicate what the defaults are?[Memory Allocation
](https://amrex-codes.github.io/amrex/docs_html/GPU.html?highlight=is_managed#memory-allocation)

Do we expect managed support to be sufficiently robust that this won't affect existing test setups?

WeiqunZhang:
This is currently in the documentation.
> By default, it allocates managed memory. This can be changed with a boolean runtime parameter amrex.the_arena_is_managed. 

I guess that was written before we had the HIP backend and it was never updated.

As for stability, we will report if we notice any issues.

jmsexton03 (approved):

PR #2736, opened 2022-04-27
Drop CUDA 9 support

WeiqunZhang:
* Modify CI to test CUDA 10.2 instead of CUDA 9.2.

* Remove workarounds for CUDA 9.2

* No need to test for CUDA >= 10.

* Update Documentation.

* Update GNU Make and CMake.

## Checklist

The proposed changes:

### DIFF
--- a/Docs/sphinx_documentation/source/GPU_Chapter.rst
+++ b/Docs/sphinx_documentation/source/GPU_Chapter.rst
@@ -8,8 +8,8 @@ NVIDIA, AMD and Intel GPUs using their native vendor language and therefore
 requires CUDA, HIP/ROCm and DPC++/SYCL, for NVIDIA, AMD and Intel GPUs, respectively.
 Users can also use OpenMP and/or OpenACC in their applications.
 
-AMReX supports NVIDIA GPUs with compute capability >= 6 and CUDA >= 10
-as well as CUDA 9.1.  While HIP and DPC++ compilers are in development in
+AMReX supports NVIDIA GPUs with compute capability >= 6 and CUDA >= 10.
+While HIP and DPC++ compilers are in development in
 preparation for Frontier and Aurora, AMReX only supports the latest
 publicly released versions of those compilers on the Iris and Tulip testbeds.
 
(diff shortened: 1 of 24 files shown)


Docs/sphinx_documentation/source/GPU_Chapter.rst
```diff
@@ -8,8 +8,8 @@ NVIDIA, AMD and Intel GPUs using their native vendor language and therefore
 requires CUDA, HIP/ROCm and DPC++/SYCL, for NVIDIA, AMD and Intel GPUs, respectively.
 Users can also use OpenMP and/or OpenACC in their applications.
 
-AMReX supports NVIDIA GPUs with compute capability >= 6 and CUDA >= 10
-as well as CUDA 9.1.  While HIP and DPC++ compilers are in development in
+AMReX supports NVIDIA GPUs with compute capability >= 6 and CUDA >= 10.
```
jmsexton03, line 11: Is the statement on compute capability here consistent with the update to the CI?

Src/Base/AMReX_GpuAsyncArray.H
```diff
@@ -83,12 +81,9 @@ public:
 #if defined(AMREX_USE_HIP)
                 AMREX_HIP_SAFE_CALL ( hipStreamAddCallback(Gpu::gpuStream(),
                                                            amrex_asyncarray_delete, p, 0));
-#elif defined(__CUDACC__) && (__CUDACC_VER_MAJOR__ >= 10)
+#elif defined(AMREX_USE_CUDA)
```
jmsexton03, line 84: For clarity, we expect `__CUDACC__` and `AMREX_USE_CUDA` to be equivalent? The former is provided by the compiler, and we set the latter when we build, correct?

jmsexton03 (approved):

Docs/sphinx_documentation/source/GPU_Chapter.rst
WeiqunZhang: I changed the compute capability in the CI not because 6 didn't not work.  It was because we were not testing 7.

Src/Base/AMReX_GpuAsyncArray.H
WeiqunZhang: Yes.  I used `AMREX_USE_CUDA` for consistence because the hip is tested with `AMREX_USE_HIP`.

PR #2740, opened 2022-04-28
Use BL_PROFILE rather than explicit NVTX region in Device::Initialize

maximumcats:
## Summary

BL_PROFILE injects NVTX regions automatically when TINY_PROFILE=TRUE and USE_CUDA=TRUE, so there's no longer need to explicitly add one.

The TinyProfiler initialization is also moved up to be at the same time as the full profiler initialization so that this is actually captured.

## Additional background

I think we didn't yet have NVTX support in BL_PROFILE when this code was first added a few years ago.

The fact that TinyProfiler initialization was done at the end of AMReX initialization was added in commit f02acdce177f3c9fa03b510ffefa7dbe7538e945, and the comment in the commit was that this was to avoid costly profiling of the initialization activities, but actually starting the TinyProfiler later is orthogonal to whether the profiler (Nsight Systems) records all of the initialization activities; if that needs to be controlled, it can be done by using the option `nsys profile -c cudaProfilerApi`; the `Device::profilerStart()` call at the end of Device::Initialize() will then take care of starting CUDA profiling.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX.cpp
+++ b/Src/Base/AMReX.cpp
@@ -366,6 +366,7 @@ amrex::Initialize (int& argc, char**& argv, bool build_parm_parse,
     }
 
     BL_PROFILE_INITIALIZE();
+    BL_TINY_PROFILE_INITIALIZE();
 
 #ifndef BL_AMRPROF
     if (build_parm_parse)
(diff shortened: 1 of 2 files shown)


WeiqunZhang:
I am not sure if initializing TinyProfiler before most of the other initialization is a good idea.  It would make the output of TinyProfiler include a lot of one time cost such as random number initialization.  Another thing is TinyProfiler::Initialization uses ParmParse before ParmParse is initialized.

maximumcats:
OK, that's fine. I will resubmit with a narrower change.

PR #2753, opened 2022-05-05
Add HDF5 H5Z-ZFP support in CMake

### LINKED ISSUE #2696: HDF5 ZFP support in CMake
It looks like ZFP support is only currently in GNUmake. Would it be possible to add it to CMake as well?

jrood-nrel:
## Summary

This adds HDF5 ZFP lossy compression support to plot files within CMake.

## Additional background

I've tested this as much as I could on my own. I'm finding that the variable in CMake `HDF5_IS_PARALLEL` may not get populated correctly (at least with CMake 3.23.1 on MacOS), which appears to be a bug in CMake. I think this also depends on whether HDF5 was installed using Autotools or CMake as the installation is totally different between both. This is unfortunate because we want the logic for matching HDF5 with or without MPI to AMReX with or without MPI, but CMake can fail in this logic because it reports parallel HDF5 wrong.

Closes https://github.com/AMReX-Codes/amrex/issues/2696

## Checklist

The proposed changes:


ax3l:
> I've tested this as much as I could on my own. I'm finding that the variable in CMake HDF5_IS_PARALLEL may not get populated correctly (at least with CMake 3.23.1 on MacOS), which appears to be a bug in CMake. I think this also depends on whether HDF5 was installed using Autotools or CMake as the installation is totally different between both

Uff, yes HDF5 is just about to fully modernize their CMake infrastructure and fix all bug reports: https://github.com/HDFGroup/hdf5/issues?q=is%3Aissue+is%3Aopen+cmake

On Unix systems, use Autotools still until this process is complete.

ax3l (commented):
That looks great, thank you!

Please add the new option in the user-facing docs in `Docs/sphinx_documentation/source/BuildingAMReX.rst` as well :)

etpalmer63 (approved):
LGTM. Thanks for the contribution!

PR #2754, opened 2022-05-06
Avoid the use of null stream

WeiqunZhang:
The default stream (e.g., stream used outside MFIter) used to be the null
stream for CUDA and HIP.  By default, there is implicit synchronization
between the null stream and other streams.  To avoid that, the default
stream in AMReX is now no longer the null stream.

The behavior of Gpu::synchronize being device wide synchronization has not
changed.  However, for most of its use cases, it can be replaced by a new
function Gpu::streamSynchronizeAll that will synchronize the activities on
all AMReX streams without performing a device wide synchronization that
could potentially interfere with other libraries (e.g., MPI).

The behavior of [dtod|dtoh|htod]_memcpy has changed.  For CUDA and HIP,
these functions used to call the synchronous version of the memcpy.
However, the exact synchronization behavior depends on the memory types.
For SYCL/DPC++, there is no equivalent form because a queue (i.e., stream)
must be specified.  Furthermore, there is no guarantee of consistence across
different vendor platforms.  This has now changed to calling the
asynchronous form using the current stream followed by a stream
synchronization.

## Checklist

The proposed changes:

### DIFF
--- a/Docs/sphinx_documentation/source/GPU.rst
+++ b/Docs/sphinx_documentation/source/GPU.rst
@@ -18,7 +18,7 @@ for AMD and DPC++ for Intel. This will be designated with ``CUDA/HIP/DPC++``
 throughout the documentation.  However, application teams can also use
 OpenACC or OpenMP in their individual codes.
 
-At this time, AMReX does not support cross-native language compliation
+At this time, AMReX does not support cross-native language compilation
 (HIP for non-AMD systems and DPC++ for non Intel systems).  It may work with
 a given version, but AMReX does not track or guarantee such functionality.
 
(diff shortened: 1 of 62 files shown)


Docs/sphinx_documentation/source/GPU.rst
```diff
@@ -1355,6 +1355,48 @@ will show little improvement or even perform worse. So, this conditional stateme
...
+   Gpu::streamSynchronize();
+
+   // recommended
+   Gpu::copyAsync(Gpu::deviceToHost, ....);
+   Gpu::copyAsync(Gpu::deviceToHost, ....);
+   Gpu::copyAsync(Gpu::deviceToHost, ....);
+   Gpu::streamSynchronize();
+
+In addition to stream synchronization, there is also
+`:cpp:Gpu::synchronize()` that will perform a device wide synchronization.
```
```suggestion
:cpp:`Gpu::synchronize()` that will perform a device wide synchronization.
```

Docs/sphinx_documentation/source/GPU.rst
```diff
@@ -1355,6 +1355,48 @@ will show little improvement or even perform worse. So, this conditional stateme
...
+However, a device wide synchronization is usually an overkill and it might
```
```suggestion
However, a device wide synchronization is usually too excessive and it might
```

atmyers (approved):

PR #2762, opened 2022-05-11
BL/TinyProfiler SyncBeforeComms

kngott:
## Summary
Adds macros and sync locations before FillBoundary and ParallelCopy. This allows codes to get an accurate picture of the time spent in their communications, removing any load imbalance that can be captured in these types of calls. Only works when running TinyProfiler or BLProfiler, and the time the MPI_Barrier took is reported as `SyncBeforeComms`.

Turned on with `amrex.use_profiler_syncs = 1`.

## Additional background
Tested on Perlmutter with TinyProfiler. BLProfiler testing is underway.

```
FabArray::FillBoundary()                                    14262      79.74      82.08      84.49  10.94%
SyncBeforeComms                                             23262       16.8         37      77.84  10.08%
FabArray::ParallelCopy()                                     3000      58.95      68.46      73.11   9.47%
```

Also tweaked the position of the FillBoundary_nowait timer for consistency.
Where else should the sync be placed? (Particle Redistribute and FillPatch?)

## Checklist

The proposed changes:


WeiqunZhang:
Do we really need both `BLProfileSync` and `TinyProfileSync`?

Src/Base/AMReX_FabArrayCommI.H
```diff
@@ -240,6 +243,8 @@ FabArray<FAB>::FillBoundary_finish ()
     fbd.reset();
 
 #endif
+
+    BL_PROFILE_SYNC_STOP();
```
WeiqunZhang, line 247: It might too late because the function might return early.

kngott: Should be fixed.

kngott:
> Do we really need both `BLProfileSync` and `TinyProfileSync`?

I suppose not, if we use BL_PROFILE for the timers and wrap it in a TINY_PROFILE or BASE_PROFILE ifdef. I can't imagine a substantial difference between the two.

Where do you see it living? In TinyProfiler? (because BLProfile is completely (ifdef-else if-else)'d over.)

WeiqunZhang:
I suppose it needs to be in BLProfile and visible to either profiler because AMReX_TinyProfiler.H is not included if `AMREX_TINY_PROFILING` is not included.

WeiqunZhang:
We also need to update the cmake logic.  See #2774.

kngott:
Any thoughts on where the runtime static variables would go?

BLProfiler doesn't exist at all if TinyProfiler is on, but the TinyProfiler.H file isn't loaded when BLProfiler is on. So, there currently isn't a single consistent namespace or class to hold the static values. There also isn't an consistent `Initialize()` with a `ParmParse` to check for the runtime flag in either. Is there a higher-level place that makes sense to keep it? i.e. `system` in `AMReX.H`?

WeiqunZhang:
```
$ git grep "BL_PROFILE_INITIALIZE"
Src/Base/AMReX.cpp:    BL_PROFILE_INITIALIZE();
Src/Base/AMReX_BLProfiler.H:#define BL_PROFILE_INITIALIZE()  amrex::BLProfiler::Initialize();
Src/Base/AMReX_BLProfiler.H:#define BL_PROFILE_INITIALIZE()
Src/Base/AMReX_BLProfiler.H:#define BL_PROFILE_INITIALIZE()
```

```
$ git grep "BL_TINY_PROFILE_INITIALIZE"
Src/Base/AMReX.cpp:    BL_TINY_PROFILE_INITIALIZE();
Src/Base/AMReX_BLProfiler.H:#define BL_TINY_PROFILE_INITIALIZE()
Src/Base/AMReX_BLProfiler.H:#define BL_TINY_PROFILE_INITIALIZE()   amrex::TinyProfiler::Initialize()
Src/Base/AMReX_BLProfiler.H:#define BL_TINY_PROFILE_INITIALIZE()
```

Change to something like `amrex::BLProfiler::Initialize(); amerx::ProfileSync::Initialize();`?

kngott:
Accidentally rebased in here, so will reopen in a new PR that doesn't have a broken history.

PR #2771, opened 2022-05-13
CI--HIP: wget gpg key from https instead of http

etpalmer63:
## Summary
Fixes an issue with HIP CI, now downloads the gpg.key for ROCm sources from https. 

## Checklist

The proposed changes:

### DIFF
--- a/.github/workflows/dependencies/dependencies_hip.sh
+++ b/.github/workflows/dependencies/dependencies_hip.sh
@@ -12,14 +12,15 @@
 
 set -eu -o pipefail
 
+
 # Ref.: https://rocmdocs.amd.com/en/latest/Installation_Guide/Installation-Guide.html#ubuntu
-wget -q -O - http://repo.radeon.com/rocm/rocm.gpg.key \
+wget -q -O - https://repo.radeon.com/rocm/rocm.gpg.key \
   | sudo apt-key add -
 echo 'deb [arch=amd64] http://repo.radeon.com/rocm/apt/debian/ ubuntu main' \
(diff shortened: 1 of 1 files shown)


.github/workflows/dependencies/dependencies_hip.sh
```diff
@@ -12,14 +12,15 @@
 
 set -eu -o pipefail
 
+
 # Ref.: https://rocmdocs.amd.com/en/latest/Installation_Guide/Installation-Guide.html#ubuntu
-wget -q -O - http://repo.radeon.com/rocm/rocm.gpg.key \
+wget -q -O - https://repo.radeon.com/rocm/rocm.gpg.key \
   | sudo apt-key add -
 echo 'deb [arch=amd64] http://repo.radeon.com/rocm/apt/debian/ ubuntu main' \
```
WeiqunZhang: We probably should use https here too?

etpalmer63: Agreed.

ax3l (approved):
I don't see issues, but they updated their docs to now use HTTPS, so that's a good idea.
https://rocmdocs.amd.com/en/latest/Installation_Guide/Installation-Guide.html#ubuntu

ax3l:
triggering CI

etpalmer63:
I put it in because of the error message on PR: https://github.com/AMReX-Codes/amrex/pull/2770 But I think its strange its not showing up for others.

WeiqunZhang (approved):

PR #2779, opened 2022-05-17
Add doc page with input parameters and descriptions

etpalmer63:
## Summary
List of input parameters that relate to plotfile output.

## Checklist

The proposed changes:


etpalmer63:
A couple input parameter descriptions are incomplete and some may been inaccurate. I am happy to have help from others for the description entries.

etpalmer63:
I see this PR as having two issues after reviving it:

1.  It's not part of the table of contents and I don't believe any page in the document currently link to it. So it won't be easy for users to find. 
2.  The information may be out of date and should be re-verified.

Thanks! 

Erik

WeiqunZhang:
I think we should put all parameters into `amrex.` prefix space. Once that's done, I will redo this.

Issue #2709, opened 2022-04-13
numLocalTilesAtLevel broken?


ax3l:
We have a use case from @roelof-groenewald in WarpX where
```C++
int i = 0;
for (WarpXParIter pti(myspc, lev); pti.isValid(); ++pti, ++i) {}
```
works to count particles but
```C++
int it = myspc.numLocalTilesAtLevel(lev);
```
gives wrong results / segfaults the code later on:
https://github.com/ECP-WarpX/WarpX/pull/2531

Is there a rep-condition to `numLocalTilesAtLevel` that one could check if it might be violated?

ax3l:
fixed via #2782 :tada:

PR #2782, opened 2022-05-18
make sure m_particles is sized properly when numLocalTilesAtLevel is called

atmyers:
Addresses Issue #2709.

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_Particles.H
+++ b/Src/Particle/AMReX_Particles.H
@@ -373,9 +373,7 @@ public:
 
     //! \brief The total number of tiles on this rank on this level
     int numLocalTilesAtLevel (int lev) const {
-        reserveData();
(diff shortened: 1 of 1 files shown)


Src/Particle/AMReX_Particles.H
```diff
@@ -372,7 +372,11 @@ public:
     }
 
     //! \brief The total number of tiles on this rank on this level
-    int numLocalTilesAtLevel (int lev) const { return m_particles[lev].size(); }
+    int numLocalTilesAtLevel (int lev) const {
+        reserveData();
+        resizeData();
```
WeiqunZhang: I don't think a `const` function can call non-`const` functions.

WeiqunZhang (approved):

PR #2783, opened 2022-05-18
Fix solvability issue in the nodal solver RAP approach

WeiqunZhang:
In the RAP approach of the nodal solver, the RHS at Neumann boundaries only
includes the integral inside the domain.  That is it's only half of what its
"physical" value is.  The usual way of subtracting a constant from the RHS
does not work for RAP.  We should only subtract half of the constant offset
at Neumann boundaries.  The computation of the constant offset needed for
the solvability fix is also affected by the way how RHS is computed at
Neumann boundaries.  For EB, the computation of the constant offset is an
integral of the RHS multiplied by the volume fraction, and the subtraction
also needs to be weighted by the volume fraction.

## Checklist

The proposed changes:


WeiqunZhang:
This fixes the solvability issues reported by @esclapez.

atmyers (approved):

esclapez (approved):

PR #2784, opened 2022-05-19
Profiler Sync Timers

kngott:
## Summary

Attempt 2. Tested on dogora in all configurations (no profiler, BL & Tiny).

Not the best methodology (adds a BL || Tiny macro test), but seems to do the job.

Will test on Perlmutter as well.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_BLProfiler.H
+++ b/Src/Base/AMReX_BLProfiler.H
@@ -314,7 +314,6 @@ private:
     std::string regname;
 };
 
-
 namespace BLProfilerUtils {
   void WriteHeader(std::ostream &os, const int colWidth,
                    const Real maxlen, const bool bwriteavg);
@@ -388,7 +387,7 @@ inline std::string BLProfiler::CommStats::CFTToString(CommFuncType cft) {
 }
 
 #define BL_PROFILE_INITIALIZE()  amrex::BLProfiler::Initialize();
-#define BL_PROFILE_INITPARAMS()  amrex::BLProfiler::InitParams();
+#define BL_PROFILE_INITPARAMS()  amrex::BLProfiler::InitParams(); amrex::BLProfileSync::InitParams()
 #define BL_PROFILE_FINALIZE()    amrex::BLProfiler::Finalize();
 
 #define BL_TINY_PROFILE_INITIALIZE()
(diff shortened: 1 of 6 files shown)


Src/Base/AMReX_BLProfiler.H
```diff
@@ -475,7 +475,7 @@ inline std::string BLProfiler::CommStats::CFTToString(CommFuncType cft) {
 #define BL_PROFILE_INITPARAMS()
 #define BL_PROFILE_FINALIZE()
 
-#define BL_TINY_PROFILE_INITIALIZE()   amrex::TinyProfiler::Initialize()
+#define BL_TINY_PROFILE_INITIALIZE()   amrex::TinyProfiler::Initialize(); BLProfileSync::InitParams()
```
jmsexton03:
```suggestion
#define BL_TINY_PROFILE_INITIALIZE()   amrex::TinyProfiler::Initialize(); amrex::BLProfileSync::InitParams()
```

Src/Base/AMReX_FabArrayCommI.H
```diff
@@ -168,7 +171,7 @@ FabArray<FAB>::FillBoundary_finish ()
 
     BL_PROFILE("FillBoundary_finish()");
 
-    if (!fbd) { n_filled = IntVect::TheZeroVector(); return; }
+    if (!fbd) { n_filled = IntVect::TheZeroVector(); BL_PROFILE_SYNC_STOP(); return; }
```
WeiqunZhang: The only thing `BL_PROFILE_SYNC_STOP()` does is to decrement the counter.  So I think we could simply do it before this line and remove the one at the end of this function.

WeiqunZhang (changes requested):
Don't you need to modify CMakeLists.txt?
```
diff --git a/Src/Base/CMakeLists.txt b/Src/Base/CMakeLists.txt
index ab86c6dfcc..935be24c8e 100644
--- a/Src/Base/CMakeLists.txt
+++ b/Src/Base/CMakeLists.txt
@@ -260,7 +260,7 @@ target_sources( amrex
 # Profiling
 #   this source file has zero symbols in default conditions, which creates
 #   ranlib warnings, e.g., on macOS
-if(AMReX_BASE_PROFILE OR AMReX_FORTRAN)
+if(AMReX_BASE_PROFILE OR OR AMReX_TINY_PROFILE OR AMReX_FORTRAN)
    target_sources( amrex
      PRIVATE
        AMReX_BLProfiler.cpp
```

kngott:
Yup. Sure do! Was going to test it with WarpX on Perlmutter when I some cycles today, to track down and test the needed changes and get some fun numbers out at the same time. (PM was down last night). Looks like you may have found it, though. 😄

Src/Base/AMReX_FabArrayCommI.H
kngott: Ah, good point! I'll make that tweak.

kngott:
Tested all combinations on Perlmutter. They all work.

WeiqunZhang (approved):

PR #2794, opened 2022-05-26
Geometry: M_PI (`<cmath>`)

ax3l:
## Summary

Missing include leads to compile error on Windows:
```
build\temp.win-amd64-cpython-37\Release\1\_deps\fetchedamrex-src\Src\Base\AMReX_Geometry.H(255,39):
error C2065: 'M_PI': undeclared identifier
[C:\Users\VssAdministrator\AppData\Local\Temp\pip-req-build-9ov6s5p1\build\temp.win-amd64-cpython-37\Release\1\_deps\fetchedamrex-build\Src\amrex.vcxproj]
```

Technically, we also need to define `_USE_MATH_DEFINES` (we do that in WarpX). Added now.

## Additional background

Seen on Nightly installation tests for Windows "pip"-from-source test:
```
2022-05-25T05:37:01.5391654Z   -- Building for: Visual Studio 17 2022
2022-05-25T05:37:14.6537255Z   -- The C compiler identification is MSVC 19.32.31328.0
2022-05-25T05:37:16.7716754Z   -- The CXX compiler identification is MSVC 19.32.31328.0
2022-05-25T05:37:16.8589563Z   -- Detecting C compiler ABI info
2022-05-25T05:37:18.0896710Z   -- Detecting C compiler ABI info - done
2022-05-25T05:37:18.0928251Z   -- Check for working C compiler: C:/Program Files/Microsoft Visual Studio/2022/Enterprise/VC/Tools/MSVC/14.32.31326/bin/Hostx64/x64/cl.exe - skipped
...
2022-05-25T05:37:19.1780344Z   -- Check for working CXX compiler: C:/Program Files/Microsoft Visual Studio/2022/Enterprise/VC/Tools/MSVC/14.32.31326/bin/Hostx64/x64/cl.exe - skipped
...
```

- https://docs.microsoft.com/en-us/cpp/c-runtime-library/math-constants?view=msvc-170
- http://www.cplusplus.com/forum/beginner/105176/

## Checklist

The proposed changes:

### DIFF
--- a/Tools/CMake/AMReXSetDefines.cmake
+++ b/Tools/CMake/AMReXSetDefines.cmake
@@ -186,3 +186,11 @@ if(WIN32 AND BUILD_SHARED_LIBS)
   add_amrex_define(AMREX_IS_DLL NO_LEGACY)
   target_compile_definitions( amrex PRIVATE AMREX_IS_DLL_BUILDING)
 endif()
+
+#
+# Windows math symbols <cmath>: M_PI
+# https://docs.microsoft.com/en-us/cpp/c-runtime-library/math-constants?view=msvc-170
+#
+if(WIN32)
+  target_compile_definitions(amrex PUBLIC _USE_MATH_DEFINES)
+endif()
--- a/Src/Base/AMReX_Geometry.H
+++ b/Src/Base/AMReX_Geometry.H
@@ -12,6 +12,7 @@
 #include <omp.h>
 #endif
 
+#include <cmath>
 #include <iosfwd>
 #include <map>
 


Tools/CMake/AMReXSetDefines.cmake
```diff
@@ -186,3 +186,11 @@ if(WIN32 AND BUILD_SHARED_LIBS)
   add_amrex_define(AMREX_IS_DLL NO_LEGACY)
   target_compile_definitions( amrex PRIVATE AMREX_IS_DLL_BUILDING)
 endif()
+
+#
+# Windows math symbols <cmath>: M_PI
+# https://docs.microsoft.com/en-us/cpp/c-runtime-library/math-constants?view=msvc-170
+#
+if(WIN32)
+  target_compile_definitions(amrex PUBLIC _USE_MATH_DEFINES)
```
WeiqunZhang, line 195: Would it be better to do this in `AMReX_Geometry.H`?

ax3l: I think not, I want to avoid that some TUs are translated with the flag and some without it. That usually means trouble, since I cannot guarantee that they do not do other things in that header that would violate the ODR:
- https://en.cppreference.com/w/cpp/language/definition

Also, setting this generally reduced a lot of Windows-specific surprises and feels more like Linux in general (similar to what we set for the extra C++ keywords `and`, `or`, etc.).

WeiqunZhang: I don't see why this has anything to do with ODR.  I think you misunderstood me.  I am suggesting add
```
#ifdef _WIN32
#define _USE_MATH_DEFINES
#endif
```
to `AMReX_Geometry.H` before `#include <cmath>` instead of trusting cmake.

WeiqunZhang: Maybe we should avoid all the trouble with `M_PI` and define our own `constexpr Real pi = 3.141592653589793238462643383279502884197` in that one function.

WeiqunZhang: I guess my suggestion has a flaw.
```
#include "foo.H"  // which include cmath
#define _USE_MATH_DEFINES  /* won't work because of include guard in cmath */
#include <cmath>
```

I think we should define our own pi in that function.

WeiqunZhang:
After discussing with @ax3l, we are going to fix this in another way.

PR #2801, opened 2022-05-31
Solvability of bottom solver: Follow-up on #2783

WeiqunZhang:
In #2783, the solvability fix at the bottom level of EB nodal projection
solver was disabled because we don't have sufficient geometry information.
This broke a number of tests.  It appears that we should still do the best
we can (i.e., use the old approach at the bottom level).

## Checklist

The proposed changes:


WeiqunZhang:
The regression tests of incflo and IAMR are all good now.

atmyers (approved):

Issue #2750, opened 2022-05-03
AMReX_FabConv.cpp warning writing 1 byte into a region of size 0 [-Wstringop-overflow=]


tanweihou:
Dear AMReX developers,
cc: @ax3l 
I am a `WarpX` user in NERSC.
I observed the following warning message when I was building `WarpX` in Perlmutter.

```
In function 'void amrex::_pd_btrvout(char*, amrex::Long, amrex::Long)',
    inlined from 'void amrex::_pd_insert_field(amrex::Long, int, char*, int, int, int)' at /build2204-hdf5/_deps/fetchedamrex-src/Src/Base/AMReX_FabConv.cpp:462:12,
    inlined from 'void amrex::PD_fconvert(void*, const void*, amrex::Long, int, const Long*, const int*, const Long*, const int*, int, int, int)' at /build2204-hdf5/_deps/fetchedamrex-src/Src/Base/AMReX_FabConv.cpp:743:17:
/build2204-hdf5/_deps/fetchedamrex-src/Src/Base/AMReX_FabConv.cpp:393:7: warning: writing 1 byte into a region of size 0 [-Wstringop-overflow=]
  393 |             *p2 = tmp;
      |       ^
/build2204-hdf5/_deps/fetchedamrex-src/Src/Base/AMReX_FabConv.cpp: In function 'void amrex::PD_fconvert(void*, const void*, amrex::Long, int, const Long*, const int*, const Long*, const int*, int, int, int)':
/build2204-hdf5/_deps/fetchedamrex-src/Src/Base/AMReX_FabConv.cpp:612:6: note: at offset [9, 1073741822] into destination object 'in_long' of size 8
  612 | PD_fconvert (void*       out,
```

Here's the list of modules that I use in Perlmutter.
```
Currently Loaded Modules:
  1) craype-x86-milan                      10) craype/2.7.15
  2) libfabric/1.11.0.4.114                11) cray-dsmml/0.2.2
  3) craype-network-ofi                    12) cray-mpich/8.1.15
  4) perftools-base/22.04.0                13) cray-libsci/21.08.1.2
  5) xpmem/2.3.2-2.2_7.5__g93dd7ee.shasta  14) PrgEnv-gnu/8.3.3
  6) xalt/2.10.2                           15) Nsight-Compute/2022.1.1
  7) darshan/3.3.1                         16) Nsight-Systems/2022.2.1
  8) cmake/3.22.0                          17) cudatoolkit/11.5
[...]

PR #2803, opened 2022-05-31
Add AVX2 instructions flag.

kngott:
## Summary
Currently, Perlmutter builds don't have AVX2 instructions turned on. This is due to it being hidden in the wrappers in a way that isn't easily accessible through the craype flags.

Can see the arch flag by:
`CC -craype-verbose`

but, will throw errors (this is an invocation command) that are a pain to parse in the make files.
(Possible solution if we want automation: build a default empty main() and parse the verbose output of that instead, similar to CMake?)

## Additional background

Checked AVX2 instructions are present (or not present) with:
```
objdump -d binary_name_here > binary.asm
grep ymm binary.asm
```
Note: ymm = AVX2 instructions.

With Axel's help, looks like CMake invocations that set the NVCC compiler to CC work as expected. Should we edit the GNUMake build to behave similarly?

## Checklist

The proposed changes:


WeiqunZhang:
#2750

Tools/GNUMake/sites/Make.nersc
```diff
@@ -24,6 +24,12 @@ ifeq ($(which_computer),$(filter $(which_computer),perlmutter))
   endif
   endif
 
+  ifeq (($USE_CUDA),TRUE)
+    includes += -Xcompiler="-march=znver3"
```
WeiqunZhang: This will work.  But `includes` are for `-Ixxx`.

Tools/GNUMake/sites/Make.nersc
```diff
@@ -24,6 +24,12 @@ ifeq ($(which_computer),$(filter $(which_computer),perlmutter))
   endif
   endif
 
+  ifeq (($USE_CUDA),TRUE)
+    includes += -Xcompiler="-march=znver3" 
+  else
+    includes += -march=znver3
```
WeiqunZhang: This might be gnu only.

Tools/GNUMake/sites/Make.nersc
kngott: Ah true. I'll move this to CXXFLAGS and CFLAGS when I get a chance.

Tools/GNUMake/sites/Make.nersc
kngott: Very true. I can add in that ifeq.

Not sure what to do about Nvidia or the upcoming LLVM, though, other than a nested if statement.

kngott:
For the WarpX PR, that's correct and expected. They already see the arch flag because WarpX uses the CC wrapper as the underlying compiler for NVCC (confirmed by Axel last week). So, the wrapper's flag is captured that way. We could change to do that in GNUMake as well, but I think it would take some heavy editing of nvcc.mak and our current strategy (and I'm that might break other system).

Also, it looks like we made a successful workaround, in that PR (?), so this shouldn't have that negative effect, at least. 😄

WeiqunZhang:
We can filter out the error, `CC -craype-verbose 2> dev/null`.  And we don't need it when `USE_MPI=TRUE USE_CUDA=FALSE`.  But it probably won't hurt.

kngott:
Good catch with the CC when `USE_CUDA=FALSE`. Cleaned it up.

However, the dev/null isn't working for me (both in and out of GNUmake):
```
$ CC -craype-verbose
g++ -march=znver3 
g++: fatal error: no input files
compilation terminated.
```
```
$CC -craype-verbose 2> dev/null
-bash: dev/null: No such file or directory
```

I can get it to capture the value, but it still dumps errors. i.e:

`TEST_THING := $(shell CC -craype-verbose "")`
```
$ make print-TEST_THING
Loading ../../amrex/Tools/GNUMake/comps/gnu.mak...
Loading ../../amrex/Tools/GNUMake/sites/Make.nersc...
/usr/bin/ld: cannot find : No such file or directory
collect2: error: ld returned 1 exit status
TEST_THING is g++ -march=znver3 =  
    origin = file
     value = g++ -march=znver3 =  
```

WeiqunZhang:
Sorry there was a typo in my message.  It should read `/dev/null`, not `dev/null`.

kngott:
I see. That makes much more sense.

The version of that I can get to behave outputs the error message to a file named `,` (a comma), which I assume we don't want. 😃 I'm continuing to tweak to prevent dumping to a file.

kngott:
There we go. Not sure quite what the problem was, but this seems to work.

Tools/GNUMake/sites/Make.nersc
```diff
@@ -24,6 +24,22 @@ ifeq ($(which_computer),$(filter $(which_computer),perlmutter))
...
+  ifeq ($(USE_CUDA),TRUE)
+    ifeq ($(AMREX_CCOMP),gnu)
+      WRAPPER_FLAGS := $(filter-out g++ nvcc, $(shell CC -craype-verbose 2> /dev/null))
+
+      CFLAGS += -Xcompiler='$(WRAPPER_FLAGS)'
+      CXXFLAGS += -Xcompiler='$(WRAPPER_FLAGS)'
+    endif
+  else ifeq ($(USE_MPI),FALSE)
+    ifeq ($(AMREX_CCOMP),gnu)
+      WRAPPER_FLAGS := $(filter-out g++ nvcc, $(shell CC -craype-verbose 2> /dev/null))
```
WeiqunZhang: Maybe we should remove the first word instead of explicitly filtering out `g++` and `nvcc`.  something like `$(wordlist 2,1024,$(shell ...))`.  Then we probably don't need to limit this to `gnu`.

kngott: While I think that's a reasonable adjustment, I don't know if we want to explicitly remove gnu. That might automate this a little too much, i.e. assuming that the flag will always been seen under -craype-verbose is likely not accurate, since the LLVM-based environment isn't a Cray maintained environment.

Until all the PrgEnv's and their wrappers are well defined, I like keeping the gnu check there as a reminder that this was explicitly tailored for that environment for now.

Does that make sense? In either case, which filtering method do you prefer. I can go with either one.

WeiqunZhang: We should definitely go with `wordlist` instead of `filter-out g++ nvcc`.  The latter is too hacky.  In fact I am not sure why `nvcc` is there.  For both PrgEnv-nvidia and PrgEnv-nvhpc (at least the default version), CC invokes nvc++, not nvcc.

As for the gnu check, there are currently five programming environments on perlmutter, aocc, cray, gnu, nvhpc and nvidia.  `CC -craype-verbose` works for all.  I don't think we should be concerned about the upcoming LLVM-based environment, whatever that is.  If it's not Cray maintained environment, `CC`, which we explicitly use in other places, is unlikely to work.  So the makefile would need to be fixed anyway whether there is the gnu check or not.

Also, we probably should use `cc` instead of `CC` for `CFLAGS`.  And, in that case, we probably don't need to define `WRAPPER_FLAGS`.

kngott: Sounds good. How's that? (Tested on Perlmutter GNU in all configs.)

WeiqunZhang: LGTM

WeiqunZhang (approved):

PR #2808, opened 2022-06-02
include rocsparse library when compiling with Hypre compiled with hip

etpalmer63:
## Summary
Includes the rocsparse library for HIP enabled builds with HYPRE 

## Checklist

The proposed changes:

### DIFF
--- a/Tools/CMake/AMReXParallelBackends.cmake
+++ b/Tools/CMake/AMReXParallelBackends.cmake
@@ -246,6 +246,7 @@ if (AMReX_HIP)
    # Link to hiprand -- must include rocrand too
    find_package(rocrand REQUIRED CONFIG)
    find_package(rocprim REQUIRED CONFIG)
+   find_package(rocsparse REQUIRED CONFIG)
    find_package(hiprand REQUIRED CONFIG)
    if(AMReX_ROCTX)
        # To be modernized in the future, please see:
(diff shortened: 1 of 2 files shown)


WeiqunZhang:
We might want to move this into `Make.hypre` just like what we do to CUDA.

etpalmer63:
:+1:

Tools/CMake/AMReXParallelBackends.cmake
```diff
@@ -272,7 +272,7 @@ if (AMReX_HIP)
    target_compile_options(amrex PUBLIC $<$<COMPILE_LANGUAGE:CXX>:-m64>)
 
    # ROCm 4.5: use unsafe floating point atomics, otherwise atomicAdd is much slower
-   # 
+   #
```
ax3l, line 275: See above: `rocrand` et al. or `roctx` as examples.
Guard with `if (AMReX_HYPRE)`.

rocSPARSE refs.:
- https://github.com/ROCmSoftwarePlatform/rocSPARSE
- https://rocsparse.readthedocs.io

ax3l:
General question: why is that needed? Does Hypre expose rocSPARSE in public headers or forgets to add transitive link targets?

etpalmer63:
The need for this may have been an error on my part. I need to do some more testing to be sure though. -- Thanks for your patience :pray:

etpalmer63:
I got AMReX with Hypre and MPI to run on crusher without any of these changes, so its clear they are not needed. Closing this PR.

PR #2813, opened 2022-06-07
CMake: 3.17+

ax3l:
## Summary

We are using `find_package(CUDAToolkit ...)` now, which is only supported in CMake 3.17+.
Follow-up to #2757

## Additional background

Technically, we want to go to 3.20+ soon as well, so that we can drop the old CUDA code paths that we used to forward device flags and CUDA architectures.

## Checklist

The proposed changes:

### DIFF
--- a/Tools/CMake/AMReX_Config.cmake
+++ b/Tools/CMake/AMReX_Config.cmake
@@ -49,7 +49,7 @@ function (configure_amrex)
       target_compile_features(amrex PUBLIC $<IF:$<STREQUAL:$<PLATFORM_ID>,Windows>,cxx_std_17,cxx_std_14>)
    endif ()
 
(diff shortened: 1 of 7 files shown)


atmyers (approved):

etpalmer63 (approved):
LGTM

Tools/CMake/AMReX_Config.cmake
```diff
@@ -132,8 +132,9 @@ function (configure_amrex)
       endif ()
 
       #
-      # Add manually nvToolsExt if tiny profiler or base profiler are on.n
+      # Add manually nvToolsExt if tiny profiler or base profiler are on.
       # CMake >= 3.17 provides the module FindCUDAToolkit to do this natively.
+      # TODO: Modernize me, now that we require CMake 3.17+.
```
ax3l, line 137: Upsi, forgot to add this patch in the last commit.
Follow-up in  #2817

PR #2814, opened 2022-06-07
Clamp particles shifted from plo boundary against rhi, rather than back to plo

PhilMiller:
## Summary

https://github.com/ECP-WarpX/WarpX/issues/3155
Particles found outside the low end of the periodic boundary were shifted up by the domain size to be close to the high end boundary. If they were outside the rounding domain, and would thus trip up integer index calculations, they were being clamped back to the **low** end of the domain, rather than the high end. Looking again, I think this was actually just erroneous.

## Additional background

See https://github.com/AMReX-Codes/amrex/pull/2679#issuecomment-1079283896

## Checklist

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_ParticleUtil.H
+++ b/Src/Particle/AMReX_ParticleUtil.H
@@ -538,7 +538,7 @@ bool enforcePeriodic (P& p,
             }
             // clamp to avoid precision issues;
             if (p.pos(idim) >= rhi[idim]) {
-                p.pos(idim) = static_cast<ParticleReal>(plo[idim]);
+                p.pos(idim) = static_cast<ParticleReal>(rhi[idim]);
             }
             shifted = true;
         }


PhilMiller:
@atmyers @WeiqunZhang @ax3l @KZhu-ME

Src/Particle/AMReX_ParticleUtil.H
```diff
@@ -538,7 +538,7 @@ bool enforcePeriodic (P& p,
             }
             // clamp to avoid precision issues;
             if (p.pos(idim) >= rhi[idim]) {
-                p.pos(idim) = static_cast<ParticleReal>(plo[idim]);
```
WeiqunZhang, line 541: @atmyers I am guessing the reason for the old approach is `rhi` might be >= `phi`, which would fail the assertion later in this function.   Maybe we should change function computeRoundoffDomain to make sure rhi < phi.  We could test if `phi - tolerance` passes the test.  If so, we don't need to the bisect.  If not, we use that as the upper bound for bisect.

atmyers (approved):

Src/Particle/AMReX_ParticleUtil.H
atmyers: I will address this in a follow-on PR, where I'll also handle having both a `roundoff_hi_d` and `roundoff_hi_f`

PR #2819, opened 2022-06-08
Allow StateDataPhysBCFunct to operate on face-centered data

cgilet:
## Summary
This allows the velocity fill function (to fill external Dirichlet BCs) stored in the StateDescriptor to be used to fill the face-centered Umac. 

## Checklist

The proposed changes:

### DIFF
--- a/Src/Amr/AMReX_StateData.cpp
+++ b/Src/Amr/AMReX_StateData.cpp
@@ -528,7 +528,7 @@ StateData::FillBoundary (Box const&      bx,
 {
     BL_PROFILE("StateData::FillBoundary(geom)");
 
-    if (domain.contains(enclosedCells(bx))) return;
+    if (domain.contains(convert(bx,domain.ixType()))) return;
 
     Vector<BCRec> bcr(num_comp);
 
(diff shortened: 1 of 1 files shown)


Src/Amr/AMReX_StateData.cpp
```diff
@@ -527,9 +527,8 @@ StateData::FillBoundary (Box const&      bx,
                          int             num_comp)
 {
     BL_PROFILE("StateData::FillBoundary(geom)");
-    BL_ASSERT(bx.ixType() == desc->getType());
 
-    if (domain.contains(bx)) return;
+    if (domain.contains(enclosedCells(bx))) return;
```
WeiqunZhang: What if the domain is nodal type like Press_Type in IAMR?

cgilet: Good point. Although the function is passed a Geometry, and Geometry's domain is assumed to always be cell-centered (per comments in AMReX_Geometry.H, although this is never enforced or checked within Geometry), here, this is StateData's member variable domain, which can have any index type. ... In any case, I think my change should make this correct for any index type.

WeiqunZhang (approved):

PR #2822, opened 2022-06-09
move ParticleContainer into it's own header file

atmyers:
The proposed changes:


WeiqunZhang (approved):

ax3l (approved):
Great idea.
Probably needs updates in user code like WarpX/Hi pace/ImpactX though.

You could add an include in Particles.H temporarily to avoid immediate breakage and remove that later on.

atmyers:
The new `AMReX_Particles.H` includes `AMReX_ParticleContainer.H`, so as long as apps are including that I think they will be good: https://github.com/AMReX-Codes/amrex/blob/e8951cf2a4620a1199041afa1e80a1a45dcabc9a/Src/Particle/AMReX_Particles.H

ax3l:
Ah, perfect! :)

PR #2825, opened 2022-06-10
Allow plotting only a subset of Amr levels

maximumcats:
## Summary

If the user would like to limit plotfile size on disk, they can plot only a subset of the levels -- this should still be an accurate (though lower resolution) representation of the fluid state since the state is averaged down at the end of the timestep.

fcompare is also updated to allow comparing between plotfiles with different numbers of levels.

## Checklist

The proposed changes:


Tools/Plotfile/fcompare.cpp
```diff
@@ -113,10 +117,15 @@ int main_main()
     AMREX_ALWAYS_ASSERT_WITH_MESSAGE(pf_a.spaceDim() == pf_b.spaceDim(),
                                      "ERROR: plotfiles have different numbers of spatial dimensions");
 
-    const int finest_level = pf_a.finestLevel();
+    const int finest_level = std::min(pf_a.finestLevel(), pf_b.finestLevel());
     const int nlevels = finest_level+1;
-    AMREX_ALWAYS_ASSERT_WITH_MESSAGE(pf_a.finestLevel() == pf_b.finestLevel(),
-                                     "ERROR: number of levels do not match");
+    if (allow_diff_num_levels) {
+        amrex::Print() << "\n WARNING: number of levels do not match\n";
```
WeiqunZhang, line 124: Should there be a test comparing the number of levels before printing the message?

WeiqunZhang:
I think we need to update this line too.
https://github.com/AMReX-Codes/amrex/blob/47347f785f5c26f1f9e65bef8e10f2d383406ef7/Src/Amr/AMReX_AmrLevel.cpp#L221

WeiqunZhang:
My suggestion is to revert the function interface change of `Amr::writePlotFileDoit`, because it's hard for AmrLevel to get the `max_level_to_plot` argument.

maximumcats:
> I think we need to update this line too.
> 
> https://github.com/AMReX-Codes/amrex/blob/47347f785f5c26f1f9e65bef8e10f2d383406ef7/Src/Amr/AMReX_AmrLevel.cpp#L221

Updated

> My suggestion is to revert the function interface change of `Amr::writePlotFileDoit`, because it's hard for AmrLevel to get the `max_level_to_plot` argument.

I think the reason I did it this way is that writePlotFileDoIt doesn't know whether it's a small plot file or a full plot file. But I solved this problem for now by getting rid of small_plot_max_level, it seems unnecessary to have separate controls for both kinds of plotfiles for this particular feature.

WeiqunZhang (approved):

PR #2828, opened 2022-06-11
chore: Set permissions for GitHub actions

neilnaveen:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: neilnaveen <42328488+neilnaveen@users.noreply.github.com>

### DIFF
--- a/.github/workflows/clang.yml
+++ b/.github/workflows/clang.yml
@@ -8,6 +8,9 @@ concurrency:
   group: ${{ github.ref }}-${{ github.head_ref }}-linux-clang
   cancel-in-progress: true
 
+permissions:
+  contents: read
+
 jobs:
   # Build and install libamrex as AMReX CMake project
   # Note: this is an intentional "minimal" build that does not enable (many) options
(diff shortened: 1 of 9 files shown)


atmyers (approved):

WeiqunZhang:
Is it really necessary?  The default is `read` for PR from external forks.  https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/

ax3l:
If someone can write to mainline, couldn't they just revert these lines as well? Or do we speak about someone compromising one of the external workflows we use in our workflows?

What's the exact attack vector / risk scenario that we try to mitigate here?

PR #2829, opened 2022-06-11
make PODVector work with PolymorphicArenaAllocator

atmyers:
This is a follow-on to PR #2705.

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_GpuAllocators.H
+++ b/Src/Base/AMReX_GpuAllocators.H
@@ -23,6 +23,9 @@ namespace amrex {
     template <typename T>
     struct RunOnGpu : std::false_type {};
 
+    template <typename T>
+    struct IsPolymorphic : std::false_type {};
+
     struct ArenaAllocatorTraits {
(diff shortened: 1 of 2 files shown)


atmyers:
@jmusser304 reported on Slack that this fixed the MFIX regression tests (where we first learned about this issue).

Src/Base/AMReX_GpuAllocators.H
```diff
@@ -23,6 +23,9 @@ namespace amrex {
     template <typename T>
     struct RunOnGpu : std::false_type {};
 
+    template <typename T>
+    struct IsPolymorphic : std::false_type {};
```
WeiqunZhang: Could we call it something more explicit like `IsPolymorphicArenaAllocator`?  The name `amrex::IsPolymorphic` might be misleading.

WeiqunZhang (changes requested):

WeiqunZhang (approved):

PR #2831, opened 2022-06-12
Re-implement FaceLinear::interp() for InterpFromCoarseLevel

### LINKED ISSUE #2824: InterpFromCoarseLevel no longer works for face data
[InterpFromCoarseLevel](https://github.com/AMReX-Codes/amrex/blob/22.06/Src/AmrCore/AMReX_FillPatchUtil_I.H#L935-L946) calls into `FillPatchInterp`, which in turn calls the interpolater's `interp` method, but #2539 renamed the method for face data to be `interp_face` rather than `interp`; the `interp` method now errors out with the `FaceLinear` interpolater. So to restore this functionality the corresponding call in `InterpFromCoarseLevel` needs to be made conditionally correct for that interpolater (or we could have FaceLinear's `interp` method be functional again).

cgilet:
## Checklist

The proposed changes:

### DIFF
--- a/Src/AmrCore/AMReX_Interp_C.H
+++ b/Src/AmrCore/AMReX_Interp_C.H
@@ -37,13 +37,10 @@ face_linear_face_interp_x (int fi, int fj, int fk, int n, Array4<T> const& fine,
         int ck = 0;
 #endif
 
-        if (mask) {
-            // Check solve mask to ensure we don't overwrite valid fine data.
-            if (!mask(ci, cj, ck, n))
-            { return; }
+        // Check solve mask to ensure we don't overwrite valid fine data.
+        if (!mask || mask(ci, cj, ck, n)) {
+            fine(fi, fj, fk, n) = crse(ci, cj, ck, n);
         }
-
-        fine(fi, fj, fk, n) = crse(ci, cj, ck, n);
     }
 }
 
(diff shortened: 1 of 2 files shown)


cgilet:
The Castro test shows zero diffs with this.

Src/AmrCore/AMReX_Interp_C.H
```diff
@@ -57,10 +61,14 @@ face_linear_face_interp_y (int fi, int fj, int fk, int n, Array4<T> const& fine,
...
         int ck = 0;
 #endif
-        // Check mask to ensure we don't overwrite valid fine data.
-        if (mask(ci, cj, ck, n)) {
-            fine(fi, fj, fk, n) = crse(ci, cj, ck, n);
+
+        if (mask) {
+            // Check solve mask to ensure we don't overwrite valid fine data.
+            if (!mask(ci, cj, ck, n))
+            { return; }
```
WeiqunZhang: I was first confused when reading this.  It might be more readable like this because it's shorter.
```
if (!mask || mask(ci,cj,ck,n)) {
    fine(...) = crse(...);
}
```

WeiqunZhang (changes requested):

Src/AmrCore/AMReX_Interp_C.H
cgilet: I thought you once told me that you can't do that because some compilers will evaluate `!mask` and `mask(i,j,k,n)` at the same time and then have problems if mask is not allocated.

WeiqunZhang: No, the compilers are not allowed to do that.

WeiqunZhang: > I thought you once told me that you can't do that because some compilers will evaluate !mask and mask(i,j,k,n) at the same time and then have problems if mask is not allocated.

Maybe we were talking about Fortran, which gives compilers the freedom to do whatever they choose.

cgilet: thanks for clarifying

WeiqunZhang (approved):

maximumcats:
Thank you, @cgilet!

PR #2834, opened 2022-06-13
Update sensei CI container for sensei v4.0 integration

kwryankrattiger:
## Summary
SENSEI CI container recipe updates. The CI update will come separately with the rest of the SENSEI updates.

@burlen 

## Additional background
SENSEI v4.0.0 was a major update to SENSEI API.

## Checklist

The proposed changes:

### DIFF
--- a/.github/workflows/docker/sensei/install_sensei.sh
+++ b/.github/workflows/docker/sensei/install_sensei.sh
@@ -1,7 +1,10 @@
 #!/bin/bash
 
-DOWNLOAD_ROOT=/root/downloads
-EXTRACT_ROOT=/root/downloads/extracted
+set -x
+set -e
+
+DOWNLOAD_ROOT=/tmp/downloads
+EXTRACT_ROOT=/tmp/downloads/extracted
 mkdir -p ${DOWNLOAD_ROOT}
 mkdir -p ${EXTRACT_ROOT}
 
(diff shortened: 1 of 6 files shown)


kwryankrattiger:
Had to add a patch for VTK 9.1.0 because this is not a very common build configuration, but the patch is going into upstream so it should be okay to remove in the future, but it is here as a record that it was used for this configuration.

https://gitlab.kitware.com/vtk/vtk/-/merge_requests/9271

WeiqunZhang (approved):

.github/workflows/docker/sensei/install_sensei.sh
```diff
@@ -11,19 +14,21 @@ export CMAKE_GENERATOR=Ninja
...
 mkdir sensei
 cd sensei
-git clone https://gitlab.kitware.com/sensei/sensei.git &&
-cd sensei &&
-git checkout ${SENSEI_VERSION} &&
-mkdir build &&
-cd build &&
+git clone https://github.com/SENSEI-insitu/sensei.git
+cd sensei
+git checkout develop
```
burlen, line 19: @kwryankrattiger We want this to be pinned to an official release so that the usual churn that occurs during sensei development doesn't break amrex builds.

kwryankrattiger: That wasn't possible at the time because the tagged release had a couple bugs in it that had to be fixed for this container to be built and work correctly in CI. The container that is used in CI is static, ie. this is not run each time CI is run, only when the container is rebuilt. The develop that was used to generate the container for amrex CI is the develop branch on 2022-06-13 which is denoted by the container name I pushed and referenced in the other PR for SENSEI updates. I agree it would be more ideal to use a tagged version, if you wanted to tag a new patch version of sensei, v4.0.1, I would be happy to update this and regenerate an image using the tagged version, I think most of the build/install issues have been fleshed out on develop now so that may be worthwhile anyway.

burlen: OK. the container is static so this is not going to run against future changes. Sorry, I misunderstood the situation.

PR #2836, opened 2022-06-14
add fvolumesum to GNUmakefile

BenWibking:
## Summary
`fvolumesum` was missing from the GNUmakefile and CMakeLists.txt inside Tools/Plotfile. It is now built alongside the rest of the plotfile tools.

## Checklist

The proposed changes:

### DIFF
--- a/Tools/Plotfile/fvolumesum.cpp
+++ b/Tools/Plotfile/fvolumesum.cpp
@@ -120,7 +120,7 @@ void main_main()
                                                         problo[2]+static_cast<Real>(k+0.5)*dx[2])};
 
                                     // compute the volume
-                                    Real vol;
+                                    Real vol = NAN;
                                     if (coord == 0) {
                                         // Cartesian
                                         vol = 1.0_rt;
--- a/Tools/Plotfile/fvolumesum.cpp
+++ b/Tools/Plotfile/fvolumesum.cpp
@@ -120,7 +120,7 @@ void main_main()
                                                         problo[2]+static_cast<Real>(k+0.5)*dx[2])};
 
                                     // compute the volume
-                                    Real vol = NAN;
+                                    Real vol = std::numeric_limits<Real>::quiet_NaN();
                                     if (coord == 0) {
                                         // Cartesian
                                         vol = 1.0_rt;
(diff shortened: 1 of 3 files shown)


BenWibking:
CI failed. It looks like this exposed a possible bug:
```
fvolumesum.cpp: In function ‘void main_main()’:
fvolumesum.cpp:185:56: error: ‘vol’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
                                     lsum += fab(i,j,k) * vol;
                                             ~~~~~~~~~~~^~~~~
fvolumesum.cpp:142:56: error: ‘vol’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
                                     lsum += fab(i,j,k) * vol;
                                             ~~~~~~~~~~~^~~~~
```

BenWibking:
This warning looks spurious. It's failing because `vol` is not initialized when declared, and the if statements that test for the coordinate system don't include a catch-all case: https://github.com/AMReX-Codes/amrex/blob/1a2fc3dbee1ce0c168cdb66710458757e92a0605/Tools/Plotfile/fvolumesum.cpp#L123

Maybe this could be fixed by initializing `vol` to NAN.

Tools/Plotfile/fvolumesum.cpp
```diff
@@ -120,7 +120,7 @@ void main_main()
                                                         problo[2]+static_cast<Real>(k+0.5)*dx[2])};
 
                                     // compute the volume
-                                    Real vol;
+                                    Real vol = NAN;
```
WeiqunZhang: How about using `std::numeric_limits<Real>::quiet_NaN()`?

Tools/Plotfile/fvolumesum.cpp
```diff
@@ -163,7 +163,7 @@ void main_main()
                                                     problo[2]+static_cast<Real>(k+0.5)*dx[2])};
 
                                     // compute the volume
-                                    Real vol;
+                                    Real vol = NAN;
```
WeiqunZhang: `std::numeric_limits<Real>::quiet_NaN()`

WeiqunZhang (changes requested):

Tools/Plotfile/fvolumesum.cpp
BenWibking: What is the advantage of doing so? I thought it will only cause a floating-point exception if it's used in an arithmetic operation, which it should never be here.

BenWibking: In any case, I've changed it to quiet_NaN().

WeiqunZhang: `quiet_NaN()` is more explicit than `NAN` that it's a quiet NaN.  Also `NAN` is a float.  The compiler should be able to convert it to double at compile time, but it's probably not guaranteed.

WeiqunZhang (approved):

PR #2839, opened 2022-06-14
Maintain the high end of the 'roundoff domain' in both float and double precision

atmyers:
This way we set the proper value for the high end when mixed precision is used.

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_Geometry.H
+++ b/Src/Base/AMReX_Geometry.H
@@ -71,12 +71,12 @@ public:
         template <typename T>
         T bisect_prob_hi (amrex::Real plo, amrex::Real phi, amrex::Real idx, int ilo, int ihi, T tol) {
             T hi = static_cast<T>(phi - tol);
-            bool inside;
+            bool safe;
             {
                 int i = int(Math::floor(hi - plo)*idx);
-                inside = i >= ilo && i <= ihi;
+                safe = i >= ilo && i <= ihi;
             }
-            if (inside) {
+            if (safe) {
                 return hi;
             } else {
                 // bisect the point at which the cell no longer maps to inside the domain
--- a/Src/Base/AMReX_Geometry.H
+++ b/Src/Base/AMReX_Geometry.H
@@ -73,7 +73,7 @@ public:
             T hi = static_cast<T>(phi - tol);
             bool safe;
             {
-                int i = int(Math::floor(hi - plo)*idx) + ilo;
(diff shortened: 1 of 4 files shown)


Src/Base/AMReX_Geometry.H
```diff
@@ -67,6 +67,32 @@ public:
     int coord;
 };
 
+    namespace detail {
+        template <typename T>
+        T bisect_prob_hi (amrex::Real plo, amrex::Real phi, amrex::Real idx, int ilo, int ihi, amrex::Real tol) {
+            T hi = static_cast<T>(phi - tol);
+            bool safe;
+            {
+                int i = int(Math::floor(hi - plo)*idx) + ilo;
```
WeiqunZhang: Should it be `floor((hi-plo)*idx)` instead of `floor(hi-pol)*idx`?

atmyers:
```suggestion
                int i = int(Math::floor((x - plo)*idx)) + ilo;
```

atmyers: Yes, good catch.

Src/Base/AMReX_Geometry.H
```diff
@@ -67,6 +67,32 @@ public:
     int coord;
 };
 
+    namespace detail {
+        template <typename T>
+        T bisect_prob_hi (amrex::Real plo, amrex::Real phi, amrex::Real idx, int ilo, int ihi, amrex::Real tol) {
+            T hi = static_cast<T>(phi - tol);
+            bool safe;
+            {
+                int i = int(Math::floor((x - plo)*idx)) + ilo;
```
atmyers:
```suggestion
                int i = int(Math::floor((hi - plo)*idx)) + ilo;
```

WeiqunZhang (approved):

ax3l:
I'll merge this since it's approved & tests pass :)

ax3l:
X-ref:
- https://github.com/ECP-WarpX/WarpX/issues/3155
- https://github.com/AMReX-Codes/amrex/pull/2679#issuecomment-1079283896

ax3l:
cc @PhilMiller re #2814

ax3l:
@KZhu-ME on Slack:
> It fixes issue [3155](https://github.com/ECP-WarpX/WarpX/issues/3155) for us! Thanks!

PR #2840, opened 2022-06-14
Fix a pathological case for 2d EB

WeiqunZhang:
What could happen is a cell might be cut a tiny bit at a corner such that 3
faces with an area fraction of one and one face with an area fraction of
almost one, and the volume fraction is one.  In that case, the boundary area
and centroid have been set to wrong values based on an incorrect assumption.

## Checklist

The proposed changes:


asalmgren (approved):

ax3l:
For the `1.e-15` and `1.e-30` values in this file, you might want to consider using `<limits>`: `std::numeric_limits<amrex::Real>::epsilon`?

PR #2842, opened 2022-06-15
ParamParse: Add Files at Runtime

ax3l:
## Summary

Add a new runtime API to load whole files. This was currently only possible during the first initialize.

## Questions for Reviewers

- [x] does this work with MPI or do I need to call more functions?
- [ ] where shall I add a unit test for this?

## Additional background

Needed for pyAMReX and ImpactX' Python bindings.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_ParmParse.H
+++ b/Src/Base/AMReX_ParmParse.H
@@ -552,6 +552,13 @@ public:
     //! Add a key 'name'with value 'ref' to the end of the PP table.
     void add (const char* name,
               const std::string&  val);
+
+    //! keyword for files to load
+    static std::string FileKeyword;
+
+    //! Add keys and values from a file to the end of the PP table.
+    void addfile (std::string const filename);
+
     /**
     * \brief Get the ival'th value of kth occurrence of the requested name.
     * If successful, the value is converted to an IntVect and stored
--- a/Src/Base/AMReX_ParmParse.H
+++ b/Src/Base/AMReX_ParmParse.H
@@ -557,7 +557,7 @@ public:
     static std::string FileKeyword;
 
     //! Add keys and values from a file to the end of the PP table.
-    void addfile (std::string const filename);
+    static void addfile (std::string const filename);
 
     /**
     * \brief Get the ival'th value of kth occurrence of the requested name.
@@ -1053,7 +1053,7 @@ public:
     static std::set<std::string> getEntries (const std::string& prefix = std::string());
 
     struct PP_entry;
-    typedef std::list<PP_entry> Table;
+    using Table = std::list<PP_entry>;
     static void appendTable(ParmParse::Table& tab);
     const Table& table() const {return m_table;}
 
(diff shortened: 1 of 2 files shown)


WeiqunZhang:
Yes, this works with MPI as long as all processes are calling `addfile`.

I think it's OK to not have a unit test for this.

I think it might be better to make `addfile` a static function and pass `g_table` to `addDefn`. It's `ParmParse::addfile(..)` vs. `ParmParse().addfile(..)`.  The later could be misleading because the user might have the question what happens after the temporary `ParmParse()` is gone.

WeiqunZhang:
Yes, this works with MPI as long as all processes are calling `addfile`.

I think it's OK to not have a unit test for this.

I think it might be better to make `addfile` a static function and pass `g_table` to `addDefn`. It's `ParmParse::addfile(..)` vs. `ParmParse().addfile(..)`.  The later could be misleading because the user might have the question what happens after the temporary `ParmParse()` is gone.

ax3l:
Thanks!

For MPI, I would prefer if we read from the IO processor and bcast. I have not yet found the API for that, can you pls point me there?

Static: I see. I thought this is added to some static/global state anyway, at least that how it appeared. Let's rewrite if that's clearer, thx.

WeiqunZhang:
Yes, `addDefn` will have the I/O process read and bcast.
https://github.com/AMReX-Codes/amrex/blob/6f72de283c38ef3e38ed7c5e8760a176434e6be3/Src/Base/AMReX_ParmParse.cpp#L494
https://github.com/AMReX-Codes/amrex/blob/6f72de283c38ef3e38ed7c5e8760a176434e6be3/Src/Base/AMReX_ParallelDescriptor.cpp#L1393

ax3l:
Ah perfect, so with regards to MPI there is nothing that needs to be changed and the only thing is that the new API call is a collective call. Great!

Src/Base/AMReX_ParmParse.H
```diff
@@ -552,6 +558,13 @@ public:
     //! Add a key 'name'with value 'ref' to the end of the PP table.
     void add (const char* name,
               const std::string&  val);
+
+    //! keyword for files to load
+    static std::string FileKeyword;
+
+    //! Add keys and values from a file to the end of the PP table.
+    static void addfile (std::string const filename, ParmParse::Table& tab);
```
ax3l: @WeiqunZhang oh, how can I pass a table object here? Our `table()` method is only const, so this does not work:
```C++
amrex::ParmParse pp;
pp.addfile(filename, pp.table());
```

```
error: binding reference of type ‘amrex::ParmParse::Table&’ {aka ‘std::__cxx11::list<amrex::ParmParse::PP_entry>&’} to ‘const Table’ {aka ‘const std::__cxx11::list<amrex::ParmParse::PP_entry>’} discards qualifiers
   44 |                 pp.addfile(filename, pp.table());
      |                                      ~~~~~~~~^~
```

ax3l: oh, `g_table` internally...

ax3l:
@WeiqunZhang I converted the method to be static :)

WeiqunZhang (approved):

PR #2849, opened 2022-06-23
CMake: FindDependency CUDAToolkit

ax3l:
## Summary

If we install AMReX with CUDA support using a modern CMake, we need to repopulate targets such as `CUDA::curand` from `find_dependency` for downstream.
Downstream users find us via `find_package` and that target link dependency showed up to be unpopulated in MFIX.

## Additional background

cc @jmusser304 

## Checklist

The proposed changes:


jmusser304:
I did a quick test, and this seems to fix the MFIX cuda build issues.

atmyers (approved):

PR #2856, opened 2022-06-28
Write runtime attribs to checkpoints on GPUs

atmyers:
This was being done correctly on CPU runs only.

The proposed changes:


atmyers:
@hklion reports that this fixes the issue seen in WarpX: https://github.com/ECP-WarpX/WarpX/issues/3206

WeiqunZhang (approved):

PR #2863, opened 2022-07-01
chore: Included githubactions in the dependabot config

naveensrinivasan:
This should help with keeping the GitHub actions updated on new releases. This will also help with keeping it secure.

Dependabot helps in keeping the supply chain secure https://docs.github.com/en/code-security/dependabot

GitHub actions up to date https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

https://github.com/ossf/scorecard/blob/main/docs/checks.md#dependency-update-tool
Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"


etpalmer63:
This seems like another bot PR and should probably be declined for that reason alone. We had a similar thing not too long ago with https://github.com/AMReX-Codes/amrex/pull/2828.

etpalmer63:
See, https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks#about-workflow-runs-from-public-forks

My concern is that by approving this PR now, it may make AMReX's GitHub actions more likely to be abused later, since they only need explicit approval the first time. For these type of code contributions, it's probably not worth the risk. If we like their suggestions for improving security, it's probably better to do our own original PR. 

Either way, I'm interested to know what other people think about these types of PRs.

WeiqunZhang:
I agree with @etpalmer63's concern.

PR #2874, opened 2022-07-08
Workaround to bypass issue observed at very large scale with Fujitsu MPI

lucafedeli88:
## Summary

We have observed some MPI issues at very large scale when WarpX is compiled using Fujitsu MPI (i.e., with the Fujitsu compiler). These issues seem to be related to the use of MPI Gatherv with MPI_Datatype. This PR implements a possible workaround, initially proposed by @WeiqunZhang . The idea is that, when WarpX is compiled with the Fujitsu compiler, simpler integer arrays instead of MPI_Datatype are used in the routine where the issue was observed.

## Checklist

The proposed changes:

### DIFF
--- a/Src/AmrCore/AMReX_TagBox.cpp
+++ b/Src/AmrCore/AMReX_TagBox.cpp
@@ -650,6 +650,9 @@ TagBoxArray::collate (Gpu::PinnedVector<IntVect>& TheGlobalCollateSpace) const
     const IntVect* psend = (count > 0) ? TheLocalCollateSpace.data() : nullptr;
     IntVect* precv = TheGlobalCollateSpace.data();
 
+    //Issues have been observed with the following call at very large scale when using
+    //FujitsuMPI. The issue seems to be related to the use of MPI_Datatype. We can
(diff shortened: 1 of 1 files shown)


WeiqunZhang (approved):

Src/AmrCore/AMReX_TagBox.cpp
```diff
@@ -649,7 +649,24 @@ TagBoxArray::collate (Gpu::PinnedVector<IntVect>& TheGlobalCollateSpace) const
     //
     const IntVect* psend = (count > 0) ? TheLocalCollateSpace.data() : nullptr;
     IntVect* precv = TheGlobalCollateSpace.data();
+
+    //Issues have been observed with the following call at very large scale when using
+    //FujitsuMPI. The issue seems to be related to the use of MPI_Datatype. We can
+    //bypasses the issue by exchanging simpler integer arrays.
+#ifndef __FUJITSU
```
ax3l, line 656: We found today: the new `-Nclang` Fujtsu only defines `__CLANG_FUJITSU`

PR #2875, opened 2022-07-12
Add option to derefine to AMRErrorTag

maximumcats:
## Summary

This allows a refinement field to specify *derefinement* (by setting a zone's tagging value to the clear value).

## Checklist

The proposed changes:

### DIFF
--- a/Src/AmrCore/AMReX_ErrorList.H
+++ b/Src/AmrCore/AMReX_ErrorList.H
@@ -383,6 +383,7 @@ std::ostream& operator << (std::ostream& os, const ErrorList& elst);
     Real m_min_time = std::numeric_limits<Real>::lowest();
     Real m_max_time = std::numeric_limits<Real>::max();
     int m_volume_weighting = 0;
+    int m_derefine = 0;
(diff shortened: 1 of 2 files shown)


drummerdoc (approved):
I'm just a little worried that users won't understand that this will be order-dependent (ie can detag then tag again with next criteria).  Do you maybe want to sort the criteria to make detaggers last?  Seems like there are a number of pedantic cases to get tripped up on though, maybe better to have some good documentation.

maximumcats:
That's a good point. I was going to handle that in my own inputs file by just remembering to put it last, but I agree it's better not to rely on that, I will try to implement that change.

WeiqunZhang:
@maxpkatz Do you want this to be merged first before you work on the sorting.  It seems that we will have to rely on the user to do the sorting anyway.  Maybe the user can put the AMRErrorTags for refinement and derefinement into different containers.  Or they can store them along with priority in a sortable class.

maximumcats:
Yes let's merge and I can potentially work on sorting later.

PR #2877, opened 2022-07-15
Fix the segmentation fault in selecting actual neighbor particles.

hengjiew:
## Summary
Fix the segmentation fault in selecting actual neighbor particles.

## Additional background
This bug is hidden on Summit because it allows the device to access host variables.

## Checklist

The proposed changes:


WeiqunZhang:
To avoid unnecessary synchronization, I suggest,
```
Gpu::Buffer<unsigned int> buffer({0});
auto p_np_boundary = buffer.data();

// ...

// remove Gpu::streamSynchronize after amrex_for_1d

unsigned int* hp = buffer.copyToHost();
// use *hp
```

WeiqunZhang:
More specifically, there are three synchronizations in your code: the constructor with initial value, streamSynchronize after amrex_for_1d, and Gpu::copy.  The one using Gpu::Buffer has one in copyToHost.

atmyers (approved):

PR #2881, opened 2022-07-20
EB: Add Fine Levels

WeiqunZhang:
Add a new function EB2::addFineLevels() that can be used to add more fine
levels to the existing EB IndexSpace without changing the coarse levels.
This is useful for restarting with a larger amr.max_level.  The issue is we
build EB at the finest level first and then coarsen it to the coarse levels.
If the restart run has a different finest level, the EB on the coarse levels
could be different without using this new capability.

## Checklist

The proposed changes:

### DIFF
--- a/Src/EB/AMReX_EB2.H
+++ b/Src/EB/AMReX_EB2.H
@@ -49,6 +49,7 @@ public:
     virtual const Level& getLevel (const Geometry & geom) const = 0;
     virtual const Geometry& getGeometry (const Box& domain) const = 0;
     virtual const Box& coarsestDomain () const = 0;
+    virtual void addFineLevels (int num_new_fine_levels) = 0;
 
 protected:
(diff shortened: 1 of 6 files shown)


esclapez:
@WeiqunZhang This is working nicely, thank you ! Just to double check: the added EB level indexSpaces are not usable right ? One shouldn't have EB on AMR finer than the `max_eb_level` the way you defined it in the EB/CNS example.

WeiqunZhang:
The new EB levels do have valid EB data.  If the resolution is sufficiently fine, it's probably OK to have EB on the new levels.  The issue is it might not be consistent with with the existing EB levels if the resolution is not good enough.  For example the volume fraction of a coarse cut cell may not be consistent with the volume fractions in the fine cells.

drummerdoc (approved):

PR #2893, opened 2022-07-29
Bittree updates

akashdhruv:
## Summary

This PR introduces updates to regridding using [Bittree](https://github.com/Flash-X/Bittree)

## Additional background

We have introduced a new feature to use AMR in Octree mode. This was a result of ongoing performance studies which identified that Octree mode using Bittree scales better than AMReX native patch-based implementation.

## Checklist

The proposed changes:


akashdhruv:
Currently performing tests. This PR is created to enable discussion as we tweak the code and get it ready for production.

rmrsk:
I'm curious what is meant by 'octree-mode' since octrees have a regular/leaf node separation where potentially nothing is allocated for regular nodes. But that's (almost) never the case for patch-based AMR...

akashdhruv:
In the context of this PR octree mode refers to an octree refinement using the patch based algorithm. The idea is to be able to do regridding in this configuration using `Bittree` and bypass the native `MakeNewGrids`. Bittree has shown performance improvement for PARAMESH for Flash-X and we want to see if it can improve regridding for AMReX

tklos96:
@WeiqunZhang I made various optimization changes and updated the passing of pointers as you suggested. I have one last problem before it's ready to merge - Subcycling. When I turn subcycling on for the Advection test, I run into issues after ~35 timesteps - the solution blows up and the grid layout becomes unpredictable. I am wondering if the Bittree logic doesn't properly account for the `lbase` or something. I can't tell what's happening though, any suggestions?

WeiqunZhang:
We need to introduce `USE_BITTREE` into GNUMake system and add `-DAMREX_USE_BITTREE` to the compiler flags.  Then we can use `ifdef AMREX_USE_BITTREE` on the bittree code that has extra dependence.

Src/AmrCore/AMReX_Bittree.H
```diff
@@ -0,0 +1,73 @@
...
+class btUnit {
+  // Functions used in AmrMesh
+  public:
+    static int  btRefine(BittreeAmr* const mesh,
+                         std::vector<int>& btTags,
+                         MPI_Comm comm);
+    static void btCalculateGrids(BittreeAmr* const mesh,
+                                 int lbase, Real time, int& new_finest,
+                                 Vector<BoxArray>& new_grids,
+                                 Vector<IntVect>& max_grid_size);
```
WeiqunZhang:
```suggestion
                                 Vector<IntVect> const& max_grid_size);
```

Src/AmrCore/AMReX_Bittree.H
```diff
@@ -0,0 +1,73 @@
...
+    static void btCalculateLevel(BittreeAmr* const mesh,
+                                 int lev, Real time,
+                                 BoxArray& ba,
+                                 IntVect& max_grid_size);
```
WeiqunZhang:
```suggestion
                                 IntVect const& max_grid_size);
```

Src/AmrCore/AMReX_Bittree.H
```diff
@@ -0,0 +1,73 @@
+#ifndef BL_BITTREE_H_
+#define BL_BITTREE_H_
+
+#include <AMReX_IntVect.H>
+#include <AMReX_BoxArray.H>
+#include <AMReX_DistributionMapping.H>
+
+#include <Bittree_BittreeAmr.h>
+
+using namespace bittree;
```
WeiqunZhang: Can we avoid `using namespace` in header files?

WeiqunZhang (commented):
Some comments.  I haven't finished reviewing yet.

Src/AmrCore/AMReX_Bittree.cpp
```diff
@@ -0,0 +1,343 @@
...
+
+    BoxList bl;
+
+    for(int i=id0; i<id1; ++i) {
+      //Get coordinates and morton index.
+      auto b = tree1->locate(i);
+
+      if(b.level != lev) {
+          std::string msg = "Error identifying block in btCalculateGrids";
+          //throw error?
```
WeiqunZhang: `amrex::Abort("Error identify.....");`?

Src/AmrCore/AMReX_Bittree.cpp
```diff
@@ -0,0 +1,343 @@
...
+    for(int lev=lbase; lev<=new_finest; ++lev) {
+        btCalculateLevel(mesh, lev, time, new_grids[lev],
+                         max_grid_size[lev]);
+    }
+}
+
+/** Creates a box array based on Bittree.
+  */
+void btUnit::btCalculateLevel(BittreeAmr* const mesh, int lev,
+                            Real time,
```
WeiqunZhang:
```suggestion
                            Real /*time*/,
```
The CI is very strict on unused parameter.

Src/AmrCore/AMReX_AmrMesh.cpp
```diff
@@ -762,6 +770,66 @@ AmrMesh::MakeNewGrids (int lbase, Real time, int& new_finest, Vector<BoxArray>&
...
+        // [1] Error Estimation and tagging
+        // btTags is indexed by bitid, Bittree's internal indexing scheme.
+        // For any id, btTags = 1 if should be parent, -1 if should not be parent (or not exist).
+        std::vector<int> btTags(tree0->id_upper_bound(),0);
+
+        for (int lev=max_crse; lev>=lbase; --lev) {
+
+            TagBoxArray tags(grids[lev],dmap[lev], n_error_buf[lev]);
+            ErrorEst(lev, tags, time, 0);
+            // ?? tags.buffer(n_error_buf[lev]);
```
WeiqunZhang: Yes, we need this for error buffer.  And I think we need to do something similar to the ParallelAdd part (not the mask part) of `TagBoxArray::mapPeriodicRemoveDuplicates` to parallel communicate the buffer tags to valid cells.

Src/AmrCore/AMReX_AmrMesh.cpp
```diff
@@ -762,6 +770,66 @@ AmrMesh::MakeNewGrids (int lbase, Real time, int& new_finest, Vector<BoxArray>&
...
+
+            for (MFIter mfi(tags); mfi.isValid(); ++mfi) {
+                auto const& tagbox = tags.const_array(mfi);
+                bool has_set_tags = amrex::Reduce::AnyOf(mfi.validbox(),
+                                                         [=] AMREX_GPU_DEVICE (int i, int j, int k)
+                                                         {
+                                                              return tagbox(i,j,k) == TagBox::SET;
```
WeiqunZhang:
```suggestion
                                                              return tagbox(i,j,k) != TagBox::CLEAR;
```
to include buffer cells.

WeiqunZhang (commented):
More comments

Tests/Amr/Advection_AmrCore/Exec/inputs
```diff
@@ -2,7 +2,7 @@
 # Run until nsteps == max_step or time == stop_time, 
 #     whichever comes first
 # *****************************************************************
-max_step  = 1000
+max_step  = 100
```
WeiqunZhang: Can we keep the current inputs file and add a new file for the bittree test?

WeiqunZhang:
For the failed subcycling test, is it 2d or 3d?

tklos96:
It was 2d.

akashdhruv:
> We need to introduce `USE_BITTREE` into GNUMake system and add `-DAMREX_USE_BITTREE` to the compiler flags. Then we can use `ifdef AMREX_USE_BITTREE` on the bittree code that has extra dependence.

This is done. Also created a separate directories `Src/Extern/Bittree` and `Tools/GNUmake/packages/Make.bittree` to manage source code and dependencies.

akashdhruv:
`./configure --enable-bittree=yes` is the new option to build AMReX in Bittree mode. Default is `no`. Environment variable `BITTREE_$(DIM)D_HOME` should be defined to store Bittree library path. `$(DIM)` is the dimension

akashdhruv:
Additional task: Setup a test in AMReX test suite @WeiqunZhang can you do that on your end?

Tests/Amr/Advection_AmrCore/Exec/inputs
akashdhruv: Done

Tests/Amr/Advection_AmrCore/Exec/inputs
```diff
@@ -31,15 +31,17 @@ amr.max_level       = 2       # maximum level number allowed --
 
 amr.ref_ratio       = 2 2 2 2 # refinement ratio between levels
 
+amr.use_bittree = true
+
 # *****************************************************************
 # Control of grid creation
 # *****************************************************************
 # Blocking factor for grid creation in each dimension --
 #   this ensures that every grid is coarsenable by a factor of 8 --
```
kweide: Does the comment ("every grid is coarsenable by a factor of 8") need updating?

WeiqunZhang:
Yes, once the bittree repo is public, I will set up a regression test.

Tests/Amr/Advection_AmrCore/Exec/inputs
WeiqunZhang: Good question. @akashdhruv and @tklos96 Do we still need to set blocking factors and max_grid_size in `inputs_bittree`?

Src/AmrCore/AMReX_AmrMesh.H
```diff
@@ -260,6 +264,11 @@ protected:
     Vector<DistributionMapping> dmap;
     Vector<BoxArray>            grids;
 
+#ifdef AMREX_USE_BITTREE
+    bool use_bittree;
```
WeiqunZhang, line 268: I think we should move this flag into `AmrInfo` and gives it a default value of false.  One constructor of AmrMesh does not use ParmParse at all and it relies on the parameters in `AmrInfo`.  Then for consistence, in `InitAmrMesh`, we should change from `get` to `queryAdd`.

WeiqunZhang:
The last time I tried, I could not reproduce the blow-up issue with subcycling.  After the two minor comments are addressed, I think we can merge this first, because this is a big PR.

akashdhruv:
I found an issue with a 3D setup in Flash-X. Give me a few days to resolve that and then we can merge

akashdhruv:
Performance of Paramesh vs AMReX in Flash-X

bittree

akashdhruv:
Continued in #3547

PR #2897, opened 2022-08-03
EB checkpoint files

drangara:
## Summary
The attempt here is to write the finest level embedded boundary geometry data into a checkpoint file. When reading it, the cell flags still require to be set since they are not written. The coarse levels are built the usual way from the finest level. This method could potentially save time when there are several EB iterations which can make the geometry generation step expensive. An example case added to `incflo` [here](https://github.com/AMReX-Codes/incflo/pull/59). 

## Checklist

The proposed changes:


Src/EB/AMReX_EB2.cpp
```diff
@@ -239,6 +240,22 @@ void addFineLevels (int num_new_fine_levels)
     }
 }
 
+void
+BuildFromChkptFile (std::string const& fname,
+                    const Geometry& geom, int required_coarsening_level,
+                    int max_coarsening_level, int ngrow, bool build_coarse_level_by_coarsening,
+                    bool a_extend_domain_face, int a_num_coarsen_opt)
+{
+    AMREX_ALWAYS_ASSERT_WITH_MESSAGE(a_num_coarsen_opt == 0,
```
WeiqunZhang: I think this assertion is not needed. `a_num_coarsen_opt` is used in building the finest level with the implicit function.  In the case of restarting from checkpoint, this parameter can be simply ignored.

Src/EB/AMReX_EB2_IndexSpace_chkpt_file.cpp
```diff
@@ -0,0 +1,87 @@
...
+        Box cdomain = amrex::coarsen(m_geom.back().Domain(),2);
+        Geometry cgeom = amrex::coarsen(m_geom.back(),2);
+        m_chkpt_file_level.emplace_back(this, ilev, EB2::max_grid_size, ng, cgeom, m_chkpt_file_level[ilev-1]);
+        if (!m_chkpt_file_level.back().isOK()) {
+            m_chkpt_file_level.pop_back();
+            if (ilev <= required_coarsening_level) {
+                if (build_coarse_level_by_coarsening) {
+                    amrex::Abort("Failed to build required coarse EB level "+std::to_string(ilev));
+                } else {
+                    m_chkpt_file_level.emplace_back(this, chkpt_file, cgeom, EB2::max_grid_size, ng,
```
WeiqunZhang: Since only the finest level is saved in a checkpoint file, is this line correct?

Src/EB/AMReX_EB2_Level_chkpt_file.cpp
```diff
@@ -0,0 +1,596 @@
...
+    }
+
+    for (int idim = 0; idim < AMREX_SPACEDIM; ++idim)
+    {
+        auto& edgecent = m_edgecent[idim];
+#ifdef AMREX_USE_OMP
+#pragma omp parallel if (Gpu::notInLaunchRegion())
+#endif
+        {
+            BoxArray const& covered_edge_grids = amrex::convert(m_covered_grids,
```
WeiqunZhang: BoxArray::intersections is thread safe.  We should move this outside the omp parallel region.

Src/EB/AMReX_EB_chkpt_file.cpp
```diff
@@ -0,0 +1,371 @@
...
+                                 MultiFab& levelset, int ng_gfab, const Geometry& geom,
+                                 const IntVect& ngrow_finest, bool extend_domain_face,
+                                 int max_grid_size) const
+{
+    Real prob_lo[AMREX_SPACEDIM];
+    Real prob_hi[AMREX_SPACEDIM];
+
+    std::string File(m_restart_file + "/Header");
+
+    Print() << "file=" << File << std::endl;
```
WeiqunZhang: According to xSDK, we should put all the print statements inside something like `if (amrex::Verbose())`

Src/EB/AMReX_EB_chkpt_file.cpp
```diff
@@ -0,0 +1,371 @@
...
+    {
+        Print() << "  Loading " << m_volfrac_name << std::endl;
+
+        volfrac.define(cut_grids, dmap, 1, ng_gfab);
+        volfrac.setVal(1.); // regular value
+
+        auto prefix = MultiFabFileFullPrefix(0, m_restart_file, level_prefix, m_volfrac_name);
+        MultiFab mf(The_Pinned_Arena());
+        VisMF::Read(mf, prefix);
+        volfrac.ParallelCopy(mf, 0, 0, 1, 0, ng_gfab, geom.periodicity());
```
WeiqunZhang:
```suggestion
        volfrac.define(cut_grids, dmap, 1, ng_gfab);
        auto prefix = MultiFabFileFullPrefix(0, m_restart_file, level_prefix, m_volfrac_name);
        VisMF::Read(volfrac, prefix);
```
I think we can read the data directly into volfrac.  VisMF::Read works for non-managed GPU memory too.

WeiqunZhang (changes requested):

Src/EB/AMReX_EB_chkpt_file.cpp
```diff
@@ -0,0 +1,354 @@
...
+
+    // volfrac
+    {
+        if (amrex::Verbose()) amrex::Print() << "  Loading " << m_volfrac_name << std::endl;
+
+        volfrac.define(cut_grids, dmap, 1, ng_gfab);
+        volfrac.setVal(1.); // regular value
+
+        auto prefix = MultiFabFileFullPrefix(0, m_restart_file, level_prefix, m_volfrac_name);
+        MultiFab mf(The_Pinned_Arena());
```
WeiqunZhang:
```suggestion
```

Src/EB/AMReX_EB_chkpt_file.cpp
```diff
@@ -0,0 +1,353 @@
...
+    dmap.define(cut_grids, ParallelDescriptor::NProcs());
+
+    covered_grids.define(covered_bl);
+
+    // volfrac
+    {
+        if (amrex::Verbose()) amrex::Print() << "  Loading " << m_volfrac_name << std::endl;
+
+        volfrac.define(cut_grids, dmap, 1, ng_gfab);
+        volfrac.setVal(1.); // regular value
```
WeiqunZhang: I don't think we need to call setVal on any of the MultiFabs that will be read from the checkpoint file.

WeiqunZhang (changes requested):
Once my last comment is addressed, this PR is ready for merge.  Thanks!

Src/EB/AMReX_EB_chkpt_file.cpp
drangara: Thanks for the comment. The reason I was having a `setVal` with the regular value was to handle the ghost values from invalid regions because they fall into fully regular grids. Since we have special handling for covered cells and cells outside the domain, I was thinking this is necessary for regular cells. The other option is to combine this logic into `set_invalid_ghost_data_covered`. Please let me know if this makes more sense and I can do that.

WeiqunZhang: My point is whatever is set in valfrac.setVal() will be overwritten by VisMF::Read that touches every cell.

drangara: Ah I see your point. I incorrectly thought that `setVal` will set all cells including the invalid ghost cells to the regular value and VisMF::Read will only touch the valid cells. But looks like `VisMF::Read` will overwrite everything as you said. I think I need to handle this in a separate method like `set_invalid_ghost_data_regular`.

WeiqunZhang: I might have missed something here, but I don't understand why you need to have something like `set_invalid_ghost_data_regular`.  `VisMF::Wrtie` writes out everything including ghost cells.  `VisMF::Read` restores everything in the memory.  So whatever was in the ghost cells before the checkpoint file was written will be completely restored after `VisMF::Read`.  If we didn't need `set_invalid_ghost_regular` before, we don't need it now.

drangara: When I try writing using `VisMF::Write` on a geometry which has cut, covered and regular grids and then read using `VisMF::Read`, I find that the ghost cells are NaNs on a cut grid that has a covered or a regular grid as neighbor. My understanding was that this is because we only store cut grids and optimize out the regular and covered grids. So the regular and covered grids would fall into the "invalid" region of the MultiFab and hence those ghost cells are not read. Is this not the correct behavior? thanks.

WeiqunZhang: `VisMF::Read` reads everything including ghost cells.  If there are NaNs after `VisMF::Read`, there were there already when `VisMF::Write` was called.  Do you have a test case demonstrating what you mentioned above?

Src/EB/AMReX_EB_chkpt_file.cpp
```diff
@@ -0,0 +1,353 @@
...
+    int mgs_chkptfile;
+    is >> mgs_chkptfile;
+    gotoNextLine(is);
+    AMREX_ALWAYS_ASSERT_WITH_MESSAGE(max_grid_size == mgs_chkptfile,
+                                     "EB2::ChkptFile cannot read from different max_grid_size");
+
+    BoxArray cut_grids_ba, covered_grids_ba;
+
+    if (amrex::Verbose()) amrex::Print() << "Loading cut_grids\n";
+    cut_grids_ba.readFrom(is);
```
WeiqunZhang: Also, I don't think you need to have `cut_grids_ba` and `covered_grids_ba`.  You can just read directly.
```
cut_grids.readFrom(is);
```

Src/EB/AMReX_EB_chkpt_file.cpp
drangara: I see. If that is the case I guess I shouldn't be needing `set_invalid_ghost_data_extended` or `set_invalid_ghost_data_covered` either, since they should get read from the checkpoint file. Is this right?

I tried this [case](https://ccse.lbl.gov/pub/RegressionTesting1/MFIX-Exa/2022-08-26/BENCH05-Size0008-coveredgrid.inputs.rt.covered_grid) adding `eb2.max_grid_size=10`
1. use [eb-restart](https://mfix.netl.doe.gov/gitlab/rangarad/mfix/-/tree/eb_restart) branch of mfix from my fork.
2. first run the case with `mfix.write_geom_chk=1`
3. then run the case with `mfix.geometry="chkptfile" mfix.stop_for_unused_inputs=0`

I find that the `volfrac` on cell `(-18,-18,-12)` for example seems to give nan after reading from the checkpoint file. After the call to `set_invalid_ghost_data_extended`, it becomes a valid value.

drangara: you are right though that the nan was already there when calling `VisMF::write`. I will revisit this tomorrow. Looks like when I switched from the first method of
```
MultiFab mf(The_Pinned_Arena());
VisMF::Read(mf, prefix);
levelset.ParallelCopy(mf, 0, 0, 1, 0, ng_gfab, geom.periodicity());
```
to
```
VisMF::Read(levelset, prefix);
```
it has made things more straightforward. I don't think I need the `setVal`, `set_invalid_ghost_data_extended` or `set_invalid_ghost_data_covered` any more. i will test this and update the PR again.

Thanks a lot for your help with this!

WeiqunZhang (approved):

PR #2908, opened 2022-08-09
Add Polaris to GNUMake

jmsexton03:
## Summary
This adds Polaris as an ALCF machine by copying most of the perlmutter section of Make.nersc to Make.alcf .

## Additional background
This was tested using a similar strategy to [ALCF Getting Started - gpu affinity example  ](https://github.com/argonne-lcf/GettingStarted/blob/master/Examples/Polaris/affinity_gpu/submit.sh) with amrex/Tests/GPU/Vector on 1 node of Polaris

```
module swap PrgEnv-nvhpc PrgEnv-gnu
module load nvhpc-mixed
make USE_CUDA=TRUE USE_MPI=TRUE NVCC_CCBIN=CC
NRANKS_PER_NODE=4
NDEPTH=4
NNODES=1
NTOTRANKS=$(( NNODES * NRANKS_PER_NODE ))
mpiexec -n ${NTOTRANKS} --ppn ${NRANKS_PER_NODE} --depth=${NDEPTH} --cpu-bind depth ./set_affinity_gpu_polaris.sh ./main3d.gnu.MPI.CUDA.ex inputs
```
and 
`make USE_CUDA=TRUE USE_MPI=FALSE NVCC_CCBIN=CC`

## Checklist

The proposed changes:

### DIFF
--- a/Tools/GNUMake/Make.machines
+++ b/Tools/GNUMake/Make.machines
@@ -67,14 +67,15 @@ ifdef OLCF_ROCM_ROOT
   endif
 endif
 
-ifeq ($(findstring theta, $(host_name)), theta)
-  which_site := alcf
-  which_computer := theta
-endif
-
-ifeq ($(findstring polaris, $(host_name)), polaris)
-  which_site := alcf
-  which_computer := polaris
+ifeq ($(findstring alcf.anl.gov, $(host_name)),alcf.anl.gov)
+  ifeq ($(findstring theta, $(host_name)), theta)
+    which_site := alcf
+    which_computer := theta
+  endif
+  ifeq ($(findstring polaris, $(host_name)), polaris)
+    which_site := alcf
+    which_computer := polaris
+  endif
 endif
 
 ifeq ($(findstring sierra, $(host_name)), sierra)
--- a/Tools/GNUMake/Make.machines
+++ b/Tools/GNUMake/Make.machines
@@ -72,6 +72,11 @@ ifeq ($(findstring theta, $(host_name)), theta)
   which_computer := theta
 endif
 
+ifeq ($(findstring polaris, $(host_name)), polaris)
+  which_site := alcf
+  which_computer := polaris
+endif
+
 ifeq ($(findstring sierra, $(host_name)), sierra)
   which_site := llnl
   which_computer := sierra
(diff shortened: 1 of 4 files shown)


zingale:
this works great for me on Polaris

Tools/GNUMake/Make.machines
```diff
@@ -72,6 +72,11 @@ ifeq ($(findstring theta, $(host_name)), theta)
   which_computer := theta
 endif
 
+ifeq ($(findstring polaris, $(host_name)), polaris)
```
WeiqunZhang: If possible, could we add an extra test of whether this is an ALCF machine similar to `ifdef NERSC_HOST`, `ifdef OLCF_MODULEPATH` etc.?

jmsexton03: I checked `env` for anything containing `host` `ALCF` `Argonne` as well as looked at the documentation, and I don't see anything likely to check. 

It might be possible to check for Cray variables / modules, but that's not deterministic. I could also check for the presence of various 0-size files and make assumptions based on which lustre / home file systems are available...

WeiqunZhang: What does `hostname -f` show?

jmsexton03: It ends in `polaris.alcf.anl.gov` after some node-name specific type stuff, depending on if it's a login node or a compute node. That's [`host_name := $(shell hostname -f)`
](https://github.com/AMReX-Codes/amrex/blob/development/Tools/GNUMake/Make.machines#L15) in Make.machines, right?

WeiqunZhang: Yes.  You can also use `make .... print-host_name` to find out what it is in the make.

jmsexton03: It looks like I don't have theta access, so I'm unsure about adding a generic ALCF test similar to the NERSC_HOST block since I can't test it.

WeiqunZhang: Don't worry about theta.

maximumcats: Example ThetaGPU compute node:
```
$ hostname -f
thetagpu01.mcp.alcf.anl.gov
```
Theta login node (shared between CPU and GPU nodes):
```
$ hostname -f
thetalogin4.tmi.alcf.anl.gov
```

jmsexton03: I've pushed a commit that adds a check for alcf.anl.gov

WeiqunZhang (approved):

PR #2909, opened 2022-08-09
Add: `MultiFab::sum_unique`

ax3l:
## Summary

This provides a new method to sum values in a `MultiFab`.
For non-cell-centered data, `MultiFab::sum` double counts box boundary values that are owned by multiple boxes. This provides a function that does not double count these and provides a quick way to get only the sum of physically unique values.

- [ ] test: is there a good AMReX location to call this in an existing test (or write a new one)? I'll use this in ImpactX (see below)

## Additional background

https://github.com/ECP-WarpX/impactx/pull/165

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_MultiFab.cpp
+++ b/Src/Base/AMReX_MultiFab.cpp
@@ -1589,9 +1589,14 @@ MultiFab::sum (int comp, bool local) const
 
 Real
 MultiFab::sum_unique (int comp,
+                      bool local,
                       const Periodicity& period) const
 {
     BL_PROFILE("MultiFab::sum_unique()");
+
+    // no duplicatly distributed points if cell centered
+    if (ixType().cellCentered())
+        return this->sum(comp, local);
 
     // Owner is the grid with the lowest grid number containing the data
     std::unique_ptr<iMultiFab> owner_mask = OwnerMask(period);
@@ -1627,7 +1632,9 @@ MultiFab::sum_unique (int comp,
         }
     }
 
-    ParallelAllReduce::Sum(sm, ParallelContext::CommunicatorSub());
+    if (!local) {
+        ParallelAllReduce::Sum(sm, ParallelContext::CommunicatorSub());
+    }
 
     return sm;
 }
--- a/Src/Base/AMReX_MultiFab.cpp
+++ b/Src/Base/AMReX_MultiFab.cpp
@@ -5,6 +5,7 @@
 #include <AMReX_BLProfiler.H>
 #include <AMReX_iMultiFab.H>
 #include <AMReX_FabArrayUtility.H>
+#include <AMReX_REAL.H>
 
 #ifdef AMREX_MEM_PROFILING
 #include <AMReX_MemProfiler.H>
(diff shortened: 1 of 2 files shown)


Src/Base/AMReX_MultiFab.cpp
```diff
@@ -1586,6 +1587,53 @@ MultiFab::sum (int comp, bool local) const
...
+    using namespace amrex::literals;
+
+    // Owner is the grid with the lowest grid number containing the data
+    std::unique_ptr<iMultiFab> owner_mask = OwnerMask(period);
+
+    Real sm = Real(0.0);
+#ifdef AMREX_USE_GPU
+    if (Gpu::inLaunchRegion()) {
+        auto const& ma = this->const_arrays();
+        auto const& msk = owner_mask->const_array();
```
ax3l: @WeiqunZhang I am not sure how to pass the owner mast onto GPU `ParReduce` without an `MFIter`. What am I missing?

WeiqunZhang:
```suggestion
        auto const& msk = owner_mask->const_arrays();
```

Src/Base/AMReX_MultiFab.cpp
```diff
@@ -1586,6 +1587,53 @@ MultiFab::sum (int comp, bool local) const
...
+        sm = ParReduce(TypeList<ReduceOpSum>{}, TypeList<Real>{}, *this, IntVect(0),
+        [=] AMREX_GPU_DEVICE (int box_no, int i, int j, int k) noexcept
+                       -> GpuTuple<Real>
+        {
+            return msk(i,j,k,comp) ? ma[box_no](i,j,k,comp) : 0.0_rt;
```
WeiqunZhang:
```suggestion
            return msk[box_no](i,j,k) ? ma[box_no](i,j,k,comp) : 0.0_rt;
```

Src/Base/AMReX_MultiFab.cpp
```diff
@@ -1586,6 +1587,53 @@ MultiFab::sum (int comp, bool local) const
...
+#endif
+        for (MFIter mfi(*this,true); mfi.isValid(); ++mfi)
+        {
+            Box const& bx = mfi.tilebox();
+            Array4<Real const> const& a = this->const_array(mfi);
+            Array4<int const> const& msk = owner_mask->const_array(mfi);
+            Real tmp = 0.0_rt;
+            AMREX_LOOP_3D(bx, i, j, k,
+            {
+                tmp += msk(i,j,k,comp) ? a(i,j,k,comp) : 0.0_rt;
```
WeiqunZhang:
```suggestion
                tmp += msk(i,j,k) ? a(i,j,k,comp) : 0.0_rt;
```

Src/Base/AMReX_MultiFab.cpp
```diff
@@ -1586,6 +1587,53 @@ MultiFab::sum (int comp, bool local) const
...
 }
 
+Real
+MultiFab::sum_unique (int comp,
+                      const Periodicity& period) const
+{
+    BL_PROFILE("MultiFab::sum_unique()");
+
+    using namespace amrex::literals;
+
```
WeiqunZhang:
```suggestion
```
The namespace is imported already.

WeiqunZhang (commented):
I think we should return MultiFab::sum if it's cell centered.  That is we assume there are no overlaps for cell-centered multifabs, which is not always true, but true for almost all cases.

We could also add optional `bool local=false` to give the user a chance to skip MPI communication, because they might want to aggregate the communication.

ax3l:
Thanks a lot for the review. Updated accordingly :)

WeiqunZhang (approved):

PR #2910, opened 2022-08-10
[SYCL] Remove amrex::oneapi and update deprecated device descriptors

nmnobre:
## Summary

The `amrex::oneapi` namespace is currently mapped to `sycl::ext::oneapi` for `sub_group` declarations and operations. Since both the type and the operations are also available in the standard namespace, i.e. under `cl::sycl::`, it might be advantageous to make the change.

In addition, since certain device descriptors such as `sycl::info::device::host_unified_memory` have been deprecated and replaced with device aspects such as `sycl::aspect::usm_host_allocations`, we'll eventually have to make the move.

## Additional background

The motivation for these changes is twofold:
1. Using standard features _should_ better future-proof the code even if the plan is to only ever support DPC++. 
2. It increases compatibility with other SYCL implementations opening the door to a whole new world of target devices.

I've successfully tested these changes with amrex-tutorials/ExampleCodes/Particles/ElectromagneticPIC against [intel/llvm](https://github.com/intel/llvm) and [hipSYCL](https://github.com/illuhad/hipSYCL) both while targeting a CUDA device. I've tried running against the [proprietary version of DPC++](https://www.intel.com/content/www/us/en/developer/tools/oneapi/dpc-compiler.html#gs.8swg2h) but, for some reason, I'm getting a segfault when running on the CPU (even without my suggested changes) and I don't have an Intel GPU to test this on...

Cheers,
Nuno

### DIFF
--- a/Src/Base/AMReX_GpuDevice.cpp
+++ b/Src/Base/AMReX_GpuDevice.cpp
@@ -463,8 +463,8 @@ Device::initialize_gpu ()
         device_prop.warpSize = warp_size;
         auto sgss = d.get_info<sycl::info::device::sub_group_sizes>();
         device_prop.maxMemAllocSize = d.get_info<sycl::info::device::max_mem_alloc_size>();
-        device_prop.managedMemory = d.get_info<sycl::info::device::host_unified_memory>();
-        device_prop.concurrentManagedAccess = d.get_info<sycl::info::device::usm_shared_allocations>();
+        device_prop.managedMemory = d.has(sycl::aspect::usm_host_allocations);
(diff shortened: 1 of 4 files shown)


WeiqunZhang:
Thanks for the PR!  LGTM.  As for the segfault on CPU, the sycl/oneapi build is not intended for CPU jobs.

nmnobre:
Thank you 😊 well, yeah, I guess I was being hopeful... 😝
I changed `sycl::gpu_selector` to `sycl::cpu_selector` and just went for it but no luck.

WeiqunZhang (approved):

PR #2911, opened 2022-08-10
Use 1 atomic instead of 2 per item in DenseBins::build

atmyers:
Contention in atomic operations can be the performance-limiting factor in `DenseBins::buildGPU`, especially for large bin sizes. This PR changes the implementation of this function to use one atomic operation per item instead of two. In my tests, this led to about a factor of 2 speedup for 8x8x8 bins, a ~50% speedup for 2x2x2, and 10% for 1x1x1. I tried this on both A100 and MI250X. 

I will also explore using shared memory to speed up the counting step, and make a follow-up PR if that helps.

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_DenseBins.H
+++ b/Src/Particle/AMReX_DenseBins.H
@@ -200,6 +200,7 @@ public:
 
         m_bins.resize(nitems);
         m_perm.resize(nitems);
+        m_local_offsets.resize(nitems);
 
         m_counts.resize(0);
         m_counts.resize(nbins+1, 0);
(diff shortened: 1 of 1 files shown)


PhilMiller (approved):

atmyers:
I experimented with using shared memory to accelerate the counting stage of the sort, but it did not really help. The reason is that the items are not ordered when coming in to this routine, so each block must store the entire bin space in shared memory. For realistic bin sizes (say, 2^3 to 32^3 on a 128^3 grid), this is either above the max or else it strictly limits the number of blocks that can run at a time. As a result the shared approach was usually slower, and at best a wash.

PhilMiller:
Probably not something to aim at for this PR, but it should be possible to take advantage of cases where items were previously binned and won't have moved very far to reduce the number of potential bins to which they might contribute, and thus limit the shared memory footprint.

Separately, it might be worth trying to operate on the items by assigning them to blocks and threads out of their natural order, so there's less contention.

WeiqunZhang (approved):

PR #2912, opened 2022-08-10
Open Boundary Poisson Solver

WeiqunZhang:
This adds an open boundary Poisson solver based on the James's algorithm.
To use it, the user builds an amrex:OpenBCSolver object, which can be reused
until the grids change, and then call OpenBCSolver::solve.

Currently, this is for 3D cell-centered data only. The solver works on CPU,
Nvidia GPUS, and AMD GPUs.  The SYCL version of a couple of kernels for
Intel GPUs are to be implemented.

## Checklist

The proposed changes:

### DIFF
--- a/Src/LinearSolvers/CMakeLists.txt
+++ b/Src/LinearSolvers/CMakeLists.txt
@@ -98,3 +98,15 @@ if (AMReX_HYPRE)
       MLMG/AMReX_MLNodeLaplacian_hypre.cpp
       )
 endif ()
+
+if (AMReX_SPACEDIM EQUAL 3)
+
+   target_include_directories(amrex PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/OpenBC>)
+
+   target_sources(amrex
+      PRIVATE
+      OpenBC/AMReX_OpenBC.H
+      OpenBC/AMReX_OpenBC_K.H
+      OpenBC/AMReX_OpenBC.cpp
+      )
+endif ()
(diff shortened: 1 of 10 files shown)


Src/LinearSolvers/CMakeLists.txt
```diff
@@ -98,3 +98,14 @@ if (AMReX_HYPRE)
...
       )
 endif ()
+
+if (AMReX_SPACEDIM EQUAL 3)
+
+   target_include_directories(amrex PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/OpenBC>)
+
+   target_sources(amrex
+      PRIVATE
+      OpenBC/AMReX_OpenBC.H
```
jmsexton03, line 108: Should AMReX_OpenBC_K.H also be included here?

jmsexton03 (approved):

Src/LinearSolvers/CMakeLists.txt
WeiqunZhang: We probably don't really need it because that file is only for AMReX_OpenBC.cpp, not for public.  But it's probably less likely to miss a public header if we simply include everything.  @ax3l What's your opinion?

ax3l: Yes, usually public API is what you like to add. But if you add all headers, nothing will go wrong either.

WeiqunZhang: Thanks.  I added the header.

PR #2878, opened 2022-07-15
Support SoA-Only Particles

Thierry992:
## Summary

Further updates of  #2853 where @atmyers and @ax3l can also push to.

## Upgrade Guide for C++ API Changes

These PRs show changes to keep using the *old layout*:
- example in WarpX: https://github.com/ECP-WarpX/WarpX/pull/3848
- example in HiPACE++: https://github.com/Hi-PACE/hipace/pull/921
- example in ImpactX: https://github.com/ECP-WarpX/impactx/pull/342
- example in pyAMReX: https://github.com/AMReX-Codes/pyamrex/pull/119

These PRs show changes to *use the new layout* (pure SoA):
- example in WarpX: https://github.com/ECP-WarpX/WarpX/pull/3850
- example in ImpactX: https://github.com/ECP-WarpX/impactx/pull/348
- example in pyAMReX: https://github.com/AMReX-Codes/pyamrex/pull/124

## Checklist

The proposed changes:
- [ ] fix a bug or incorrect behavior in AMReX
- [x] add new capabilities to AMReX
- [ ] changes answers in the test suite to more than roundoff level
- [ ] are likely to significantly affect the results of downstream AMReX users
- [ ] include documentation in the code and/or rst files, if appropriate

PR #2920, opened 2022-08-18
Specialize `GpuArray<T, 0>`

ax3l:
## Summary

We need an implementation of AMReX' GPU array with zero entries for AoS-free particle implementations (#2878).

This applies the same specialization as `std::array` to make `amrex::GpuArray` trivially copyable again if used with zero size.

## Additional background

The implicit copy constructors tried to assign an array member of size 1 (rhs) to a size 0 (lhs), which does not compile.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_Array.H
+++ b/Src/Base/AMReX_Array.H
@@ -136,73 +136,33 @@ namespace amrex {
         using value_type = T;
         using reference_type = T&;
 
-        /**
-         * GpuArray elements are indexed using square brackets, as with any
-         * other array.
-         */
         AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
         const T& operator [] (int) const noexcept { return *static_cast<T*>(nullptr); }
 
-        /**
-         * GpuArray elements are indexed using square brackets, as with any
-         * other array.
-         */
         AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
         T& operator [] (int) noexcept { return *static_cast<T*>(nullptr); }
 
-        /**
-         * Returns a \c const pointer to the underlying data of a GpuArray object.
-         */
         AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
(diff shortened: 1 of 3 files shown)


WeiqunZhang:
I don't see any issues you are saying.

```
        static_assert(std::is_trivially_copyable<GpuArray<int,0>>(), "");

        GpuArray<int,0> a;
        auto b = a;
        GpuArray<int,0> c;
        c = a;

        amrex::Print() << "size of a b c: " << a.size() << " " << b.size()
                       << " " << c.size() << std::endl;
```

This compiles and runs fine for me.
```
size of a b c: 0 0 0
```

WeiqunZhang:
So what does not compile?

ax3l:
I think the issue shows up in assignments of types that have a `GpuArray<T, 0>` as member type.

WeiqunZhang:
Like this?
```
template <class T>
struct S
{   
    GpuArray<T,0> x;
};
```

```
        S<int> foo;
        S<int> bar = foo;
```
works too.

ax3l:
I try to build a reproducer that has not a full `ParticleTileData` from #2878, but essentially it complains for zero sized `GpuArray` that it tries to assign a `void*` a `double *__redistrict (*)[1]`.

ax3l:
After more hours I think it's not unlikely, that the issue is somewhere else in #2878.

Fundamentally and comparing to `std::array`, I do wonder though why we want to define a `T a[1]` member at all times with `amrex::GpuArray`?

WeiqunZhang:
It was implemented that way because it was the easiest way (just a few keystrokes) to support zero-size array.  I don't think it has any issues with the implicitly defined constructors and assignment operators.  However, there is a very good argument for a specialization.  If it is specialized without a data member, its size will be just 1, whereas the current size is `sizeof(T)`.  Let's revisit this, if we want the size optimization.  We probably need to add some pragmas to disable compiler warning of casting nullptr to `T*`.

PR #2921, opened 2022-08-19
SYCL: Replace deprecated atomic types and operations

nmnobre:
## Summary

The atomic types and its operations provided by SYCL 1.2.1 used in AMReX are deprecated in SYCL 2020.
By replacing these, AMReX will, as far as I'm aware, be fully compliant with SYCL 2020.

Can the reviewer(s) please confirm we should be using `memory_scope::work_group` and not `memory_scope::device`?
On second thought it feels like it should be the latter, but I'll wait to hear from you before changing it. :-)

Cheers,
Nuno

### DIFF
--- a/Src/Base/AMReX_GpuAtomic.H
+++ b/Src/Base/AMReX_GpuAtomic.H
@@ -30,7 +30,7 @@ namespace detail {
     {
 #if defined(__SYCL_DEVICE_ONLY__)
         constexpr auto mo = sycl::memory_order::relaxed;
-        constexpr auto ms = sycl::memory_scope::work_group;
+        constexpr auto ms = sycl::memory_scope::device;
         constexpr auto as = sycl::access::address_space::global_space;
         static_assert(sizeof(R) == sizeof(I), "sizeof R != sizeof I");
         I* const add_as_I = reinterpret_cast<I*>(address);
@@ -54,7 +54,7 @@ namespace detail {
     {
 #if defined(__SYCL_DEVICE_ONLY__)
         constexpr auto mo = sycl::memory_order::relaxed;
-        constexpr auto ms = sycl::memory_scope::work_group;
+        constexpr auto ms = sycl::memory_scope::device;
         constexpr auto as = sycl::access::address_space::global_space;
         static_assert(sizeof(R) == sizeof(I), "sizeof R != sizeof I");
         I* const add_as_I = reinterpret_cast<I*>(address);
@@ -133,7 +133,7 @@ namespace detail {
         return atomicAdd(sum, value);
 #elif defined(__SYCL_DEVICE_ONLY__)
         constexpr auto mo = sycl::memory_order::relaxed;
-        constexpr auto ms = sycl::memory_scope::work_group;
+        constexpr auto ms = sycl::memory_scope::device;
         constexpr auto as = sycl::access::address_space::global_space;
         sycl::atomic_ref<T,mo,ms,as> a{*sum};
         return a.fetch_add(value);
@@ -316,7 +316,7 @@ namespace detail {
         return atomicMin(m, value);
 #elif defined(__SYCL_DEVICE_ONLY__)
         constexpr auto mo = sycl::memory_order::relaxed;
-        constexpr auto ms = sycl::memory_scope::work_group;
+        constexpr auto ms = sycl::memory_scope::device;
         constexpr auto as = sycl::access::address_space::global_space;
         sycl::atomic_ref<T,mo,ms,as> a{*m};
         return a.fetch_min(value);
@@ -377,7 +377,7 @@ namespace detail {
         return atomicMax(m, value);
 #elif defined(__SYCL_DEVICE_ONLY__)
         constexpr auto mo = sycl::memory_order::relaxed;
-        constexpr auto ms = sycl::memory_scope::work_group;
+        constexpr auto ms = sycl::memory_scope::device;
         constexpr auto as = sycl::access::address_space::global_space;
         sycl::atomic_ref<T,mo,ms,as> a{*m};
         return a.fetch_max(value);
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
Because we are using the relaxed ordering, which imposes no orders on concurrent memory accesses, maybe `memory_scope::work_item` is all we need.  I want to do some testing and think more about this.

nmnobre:
Take your time @WeiqunZhang :-)

What you said made me go read a bit more of the spec and, indeed, the section on [Memory Scope](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_memory_scope) clearly says the passed memory scope determines "the memory ordering constraints of a given atomic operation".

I was under the (wrong?) impression that it (also) determined the atomicity scope, i.e. the encompassing range of the mutex if you will... Since we are dealing with global memory, my thinking was that, in general, we'd need to guarantee atomicity across the whole device.

In DPC++, they use `spirv::AtomicIAdd` underneath, for which I found [this](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpAtomicIAdd) which adds "atomically with respect to any other atomic accesses within Scope to the same location"...
I'm a bit confused.

Thank you for your time,
-Nuno

WeiqunZhang:
@nmnobre We have reached out to Intel for help.  @bashbaug has confirmed that in the Intel implementation the memory scope is ignored for the relaxed memory order.  Here is a link to how atomics are handled in the Intel graphics compiler. https://github.com/intel/intel-graphics-compiler/blob/master/IGC/BiFModule/Implementation/atomics.cl. Because the relaxed memory order is not in PRE_OP_NEED_FENCE and POST_OP_NEEDS_FENCE,  scope checks and fence handling are skipped.

So I think we can change it to either `memory_scope::device` or `memory_scope::work_item` even though it does not matter.

Src/Base/AMReX_GpuAtomic.H
```diff
@@ -30,15 +30,16 @@ namespace detail {
     {
 #if defined(__SYCL_DEVICE_ONLY__)
         constexpr auto mo = sycl::memory_order::relaxed;
+        constexpr auto ms = sycl::memory_scope::work_group;
```
```suggestion
        constexpr auto ms = sycl::memory_scope::device;
```

Src/Base/AMReX_GpuAtomic.H
```diff
@@ -53,17 +54,18 @@ namespace detail {
...
```
```suggestion
        constexpr auto ms = sycl::memory_scope::device;
```

Src/Base/AMReX_GpuAtomic.H
```diff
@@ -131,9 +133,10 @@ namespace detail {
         return atomicAdd(sum, value);
 #elif defined(__SYCL_DEVICE_ONLY__)
         constexpr auto mo = sycl::memory_order::relaxed;
+        constexpr auto ms = sycl::memory_scope::work_group;
```
```suggestion
        constexpr auto ms = sycl::memory_scope::device;
```

Src/Base/AMReX_GpuAtomic.H
```diff
@@ -313,9 +316,10 @@ namespace detail {
         return atomicMin(m, value);
 #elif defined(__SYCL_DEVICE_ONLY__)
         constexpr auto mo = sycl::memory_order::relaxed;
+        constexpr auto ms = sycl::memory_scope::work_group;
```
```suggestion
        constexpr auto ms = sycl::memory_scope::device;
```

Src/Base/AMReX_GpuAtomic.H
```diff
@@ -373,9 +377,10 @@ namespace detail {
         return atomicMax(m, value);
 #elif defined(__SYCL_DEVICE_ONLY__)
         constexpr auto mo = sycl::memory_order::relaxed;
+        constexpr auto ms = sycl::memory_scope::work_group;
```
```suggestion
        constexpr auto ms = sycl::memory_scope::device;
```

Src/Base/AMReX_GpuAtomic.H
```diff
@@ -430,9 +435,10 @@ namespace detail {
         return atomicOr(m, value);
 #elif defined(__SYCL_DEVICE_ONLY__)
         constexpr auto mo = sycl::memory_order::relaxed;
+        constexpr auto ms = sycl::memory_scope::work_group;
```
```suggestion
        constexpr auto ms = sycl::memory_scope::device;
```

Src/Base/AMReX_GpuAtomic.H
```diff
@@ -451,9 +457,10 @@ namespace detail {
         return atomicAnd(m, value ? ~0x0 : 0);
 #elif defined(__SYCL_DEVICE_ONLY__)
         constexpr auto mo = sycl::memory_order::relaxed;
+        constexpr auto ms = sycl::memory_scope::work_group;
```
```suggestion
        constexpr auto ms = sycl::memory_scope::device;
```

Src/Base/AMReX_GpuAtomic.H
```diff
@@ -472,11 +479,12 @@ namespace detail {
     {
 #if defined(__SYCL_DEVICE_ONLY__)
         constexpr auto mo = sycl::memory_order::relaxed;
-        sycl::atomic<unsigned int,AS> a{sycl::multi_ptr<unsigned int,AS>(m)};
-        unsigned int oldi = a.load(mo), newi;
+        constexpr auto ms = sycl::memory_scope::work_group;
```
```suggestion
        constexpr auto ms = sycl::memory_scope::device;
```

Src/Base/AMReX_GpuAtomic.H
```diff
@@ -509,12 +517,13 @@ namespace detail {
         return atomicDec(m, value);
 #elif defined(__SYCL_DEVICE_ONLY__)
         constexpr auto mo = sycl::memory_order::relaxed;
+        constexpr auto ms = sycl::memory_scope::work_group;
```
```suggestion
        constexpr auto ms = sycl::memory_scope::device;
```

Src/Base/AMReX_GpuAtomic.H
```diff
@@ -535,9 +544,10 @@ namespace detail {
         return atomicExch(address, val);
 #elif defined(__SYCL_DEVICE_ONLY__)
         constexpr auto mo = sycl::memory_order::relaxed;
+        constexpr auto ms = sycl::memory_scope::work_group;
```
```suggestion
        constexpr auto ms = sycl::memory_scope::device;
```

Src/Base/AMReX_GpuAtomic.H
```diff
@@ -557,9 +567,10 @@ namespace detail {
         return atomicCAS(address, compare, val);
 #elif defined(__SYCL_DEVICE_ONLY__)
         constexpr auto mo = sycl::memory_order::relaxed;
+        constexpr auto ms = sycl::memory_scope::work_group;
```
```suggestion
        constexpr auto ms = sycl::memory_scope::device;
```

WeiqunZhang (changes requested):

WeiqunZhang (approved):

PR #2930, opened 2022-08-28
Fix Tensor Solver BC

WeiqunZhang:
This fixes some bugs in the physical domain BC of tensor linear solver.

At the corner of two no-slip walls (e.g., (0,0)), we have u(-1,0) = -u(0,0)
and u(0,-1) = -u(0,0).  It's incorrect to fill the corner ghost cell with
u(-1,-1) = u(-1,0) + u(0,-1) - u(0,0), because it will result in u(-1,-1) =
-3 * u(0,0).

In the old approach, to avoid branches in computing transverse derivatives
on cell faces, we fill the ghost cells first.  For example, to compute du/dy
at the lo-x boundary, we use the data in i = -1 and 0, just like we compute
du/dy(i) using u(i-1) and u(i) for interior faces.  The problem is the
normal velocity in the ghost cells outside a wall is filled with
extrapolation of the Dirichlet value (which is zero) and more than 1
interior cells. Because of the high-order extrapolation, u(-1) != -u(0).
This is the desired approach for computing du/dx on the wall.  However, this
produces incorrect results in dudy.

In the new approach, we explicitly handle the boundaries in the derivative
stencil.  For example, to compute transverse derivatives on an inflow face,
we use the boundary values directly.

## Checklist

The proposed changes:


WeiqunZhang:
This is still a work in progress.  It does not compile with EB.  Here are the regression test results for [IAMR on CPU](https://ccse.lbl.gov/pub/RegressionTesting1/IAMR/2022-08-28-001/index.html), [IAMR on GPU](https://ccse.lbl.gov/pub/GpuRegressionTesting/IAMR/2022-08-28-002/index.html), [incflo on CPU](https://ccse.lbl.gov/pub/RegressionTesting1/incflo/2022-08-28-001/index.html), and [incflo on GPU](https://ccse.lbl.gov/pub/GpuRegressionTesting/incflo/2022-08-28-001/index.html).

Src/LinearSolvers/MLMG/AMReX_MLTensorOp_grad.cpp
```diff
@@ -95,33 +140,36 @@ MLTensorOp::compVelGrad (int amrlev, const Array<MultiFab*,AMREX_SPACEDIM>& flux
...
+    for (MFIter mfi(sol, TilingIfNotGPU()); mfi.isValid(); ++mfi)
     {
-        Array<FArrayBox,AMREX_SPACEDIM> fluxfab_tmp;
+        Array4<Real const> const vfab = sol.const_array(mfi);
+        AMREX_D_TERM(Box const xbx = mfi.nodaltilebox(0);,
+                     Box const ybx = mfi.nodaltilebox(1);,
+                     Box const zbx = mfi.nodaltilebox(2);)
+        AMREX_D_TERM(Array4<Real> const fxfab = fluxes[0]->array(mfi);,
+                     Array4<Real> const fyfab = fluxes[1]->array(mfi);,
+                     Array4<Real> const fzfab = fluxes[1]->array(mfi);)
```
cgilet: This should be `Array4<Real> const fzfab = fluxes[2]->array(mfi);` right

WeiqunZhang: Yes.

WeiqunZhang:
```suggestion
                     Array4<Real> const fzfab = fluxes[2]->array(mfi);)
```

Src/LinearSolvers/MLMG/AMReX_MLTensor_3D_K.H
```diff
@@ -21,25 +659,22 @@ void mltensor_fill_corners (int icorner, Box const& vbox, // vbox: the valid box
...
     constexpr int ydir = 1;
     constexpr int zdir = 2;
     const auto blen = amrex::length(vbox);
     const auto vlo  = amrex::lbound(vbox);
     const auto vhi  = amrex::ubound(vbox);
-    const auto dlo  = amrex::lbound(domain);
-    const auto dhi  = amrex::ubound(domain);
     for (int icomp = 0; icomp < AMREX_SPACEDIM; ++icomp) {
         switch (icorner) {
         case 0: {
```
cgilet, line 683: Doesn't this run into basically the same problem you outlined in 2D for the case of a corner with 3 no-slip walls? If `vel(i-1,j,k) = -vel(i,j,k)` etc, then we get `vel(i-1,j-1,k-1) = -5* vel(i,j,k)`

cgilet: Nevermind, I see this is the EB case you have yet to get to

cgilet:
What I see from the IAMR regression tests looks reasonable given the code changes. Early on the differences are largest at corners & edges of domain and coarse-fine boundary.

WeiqunZhang:
The latest regression test results are here,

- IAMR https://ccse.lbl.gov/pub/RegressionTesting1/IAMR/2022-09-17-001/index.html
- incflo https://ccse.lbl.gov/pub/RegressionTesting1/incflo/2022-09-17-001/index.html

If I revert the changes on coarse/fine boundaries, the results are here,

- IAMR https://ccse.lbl.gov/pub/RegressionTesting1/IAMR/2022-09-17-002/index.html
- incflo https://ccse.lbl.gov/pub/RegressionTesting1/incflo/2022-09-17-002/index.html

Src/LinearSolvers/MLMG/AMReX_MLEBTensor_3D_K.H
```diff
@@ -170,23 +226,563 @@ void mlebtensor_cross_terms_fz (Box const& box, Array4<Real> const& fz,
...
+                    int jlom = j - flag(i,j,k-1).isConnected(0,-1,0);
+                    whi = mlebtensor_weight(jhip-jhim);
+                    wlo = mlebtensor_weight(jlop-jlom);
+                    Real dvdy = mlebtensor_dy_on_zface(i,j,k,1,vel,dyi,
+                                                       whi,wlo,jhip,jhim,jlop,jlom);
+                    Real dwdy = mlebtensor_dy_on_zface(i,j,k,2,vel,dyi,
+                                                       whi,wlo,jhip,jhim,jlop,jlom);
+                    Real divu = dudx + dvdy;
+                    Real xif = kapz(i,j,k);
+                    Real mun = 0.75*(etaz(i,j,k,2)-xif);  // restore the original eta
```
cgilet: Don't we want `Real(0.75)` here?

WeiqunZhang: For consistency, yes.

cgilet (approved):

Src/LinearSolvers/MLMG/AMReX_MLTensorOp.cpp
```diff
@@ -247,20 +254,65 @@ MLTensorOp::apply (int amrlev, int mglev, MultiFab& out, MultiFab& in, BCMode bc
                          Array4<Real> const fyfab = fluxfab_tmp[1].array();,
                          Array4<Real> const fzfab = fluxfab_tmp[2].array(););
 
-            AMREX_LAUNCH_HOST_DEVICE_LAMBDA_DIM
```
esclapez, line 250: Could `fluxfab_tmp` be allocated in The_Async_Arena ? (slightly orthogonal comment)

WeiqunZhang: Yes.

Src/LinearSolvers/MLMG/AMReX_MLTensorOp_grad.cpp
```diff
@@ -52,20 +58,59 @@ MLTensorOp::compFlux (int amrlev, const Array<MultiFab*,AMREX_SPACEDIM>& fluxes,
                          Array4<Real> const fyfab = fluxfab_tmp[1].array();,
                          Array4<Real> const fzfab = fluxfab_tmp[2].array(););
 
-            AMREX_LAUNCH_HOST_DEVICE_LAMBDA_DIM
```
esclapez, line 55: Same thing here for `fluxfab_tmp`.

Src/LinearSolvers/MLMG/AMReX_MLTensorOp.cpp
WeiqunZhang: If we use The_Async_Arena, we will have to move `FArrayBox fluxfab_tmp[AMREX_SPACEDIM` into the MFIter loop body, because `resize` does not work in this situation.  The reason that resize works with Elixir is because Elxiir takes the ownership of the memory.  But currently Elixir does not work with The_Async_Arena.  We could make it work in the future.  I think right now we can just leave it like this.  For OpenMP, we do want `FArrayBox fluxfab_tmp` to be outside the MFIter loop so that they can be reused.

esclapez: Sounds good, thanks for checking !

esclapez (approved):
LGTM. A lot of code 'just' to handle BCs !
I see differences similar to the one in IAMR and incflo in PeleLM(eX) and those are consistent in all the direction/orientation combinations.

PR #2943, opened 2022-09-08
Fix EB data inconsistency when fixing small cells and multiple cuts

WeiqunZhang:
## Summary

For consistency, we need to call the function that zeros out the level set even if that box does not have any small cells or multiple cuts.  This is because a node could exist in multiple boxes.  Furthermore, a covered cell or covered face may have a node with a level set < 0.

## Additional background

This is usually not an issue.  However, in WarpX, we use the level set to decide whether a node is an unknown in the linear system.  The inconsistency makes the solver fail in some cases.

## Checklist

The proposed changes:


WeiqunZhang:
@drangara

drangara:
LGTM

asalmgren (approved):

PR #2944, opened 2022-09-09
Fix: Make Finalize->Initialize->F->I->... Work

ax3l:
## Summary

The arenas are not set in all cases to `nullptr`s yet. This causes that repeated finalize/initialize calls fail in debug mode.

## Additional background

Seen in https://github.com/ECP-WarpX/impactx/pull/172 and https://github.com/ECP-WarpX/impactx/pull/240

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_Arena.cpp
+++ b/Src/Base/AMReX_Arena.cpp
@@ -253,6 +253,7 @@ Arena::Initialize ()
     if (initialized) return;
     initialized = true;
 
+    // see reason on allowed reuse of the default CPU BArena in Arena::Finalize
     BL_ASSERT(the_arena == nullptr || the_arena == The_BArena());
     BL_ASSERT(the_async_arena == nullptr);
     BL_ASSERT(the_device_arena == nullptr || the_arena == The_BArena());
@@ -468,6 +469,13 @@ Arena::Finalize ()
 
     initialized = false;
 
+    // we reset Arenas unless they are the default "CPU malloc/free" BArena
+    // this is because we want to allow users to free their UB objects
+    // that they forgot to destruct after amrex::Finalize():
+    //   amrex::Initialize(...);
+    //   MultiFab mf(...);  // this should be scoped in { ... }
+    //   amrex::Finalize();
+    // mf cannot be used now, but it can at least be freed without a segfault
     if (!dynamic_cast<BArena*>(the_device_arena)) {
         if (the_device_arena != the_arena) {
             delete the_device_arena;
--- a/Src/Base/AMReX_Arena.cpp
+++ b/Src/Base/AMReX_Arena.cpp
@@ -253,12 +253,12 @@ Arena::Initialize ()
     if (initialized) return;
     initialized = true;
 
-    BL_ASSERT(the_arena == nullptr);
+    BL_ASSERT(the_arena == nullptr || the_arena == The_BArena());
     BL_ASSERT(the_async_arena == nullptr);
-    BL_ASSERT(the_device_arena == nullptr);
-    BL_ASSERT(the_managed_arena == nullptr);
+    BL_ASSERT(the_device_arena == nullptr || the_arena == The_BArena());
+    BL_ASSERT(the_managed_arena == nullptr || the_arena == The_BArena());
     BL_ASSERT(the_pinned_arena == nullptr);
-    BL_ASSERT(the_cpu_arena == nullptr);
+    BL_ASSERT(the_cpu_arena == nullptr || the_arena == The_BArena());
 
 #ifdef AMREX_USE_GPU
 #ifdef AMREX_USE_DPCPP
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
The fix might be wrong.  I think it could lead to memory leak when BArena is used (which is the default for CPU builds).  Maybe we should fix the assertions instead.  Change it to `BL_ASSERT(the_arena==nullptr || the_arena == The_BArena())`.

The reason we want `the_arena` to be still valid in case of BArena after Arena::Finalize is called is the user's codes often have memory allocate by The_Arena still around after amrex::Finalize is called.  One could argue it's the user's fault.  But we have got questions about this kind of error so many times, it's really our fault to make it easy to make mistakes.  So we decided to keep the Arena alive so that the memory can still be freed.

WeiqunZhang:
Here is an example of the kind of error often reported to us.
```
amrex::Initialize(...)
MultiFab mf(...);
amrex::Finalize();
```
The dtor `~MultiFab` is called after `amrex::Finalize()`.  It could be fixed by limiting the scope of `mf`.  But it was so common that we decided to make it work without worrying about the scope.

WeiqunZhang:
Or we can just remove the assertion.  Also maybe you can add my comment to Arena::Finalize to explain why the pointers are not reset if it's a BArena.

ax3l:
The problem right now is that you cannot call `Initialize()` after `Finalize()` without triggering the assert.

Yes, that MF should be scoped. I also cannot all `MPI_Barrier` after `MPI_Finalize()`. 
One could also guard more in the dtor of `MultiFab`. Either way, most of the state of that variable will be UB after `Finalize()`.

Either way, it's an API contract and will leave some things dangling if we allow Finalize to be not a clean reset... If we `Finalize()` AMReX, isn't dangling memory management for an active MultiFab the least of the problems? If I continue working with that MultiFab, worse things will probably happen (comms, etc.)

ax3l:
Note that I will run various runs of AMReX in the same process sequentially, finalizing in between. These sequential runs might change arenas, etc.

I need a really clean slay after `Finalize()` or will (continue to) debug myself int oblivion about half-resetted globals of AMReX.

ax3l:
Looking at `Arena::Initialize`, I think if we use `BL_ASSERT(the_arena==nullptr || the_arena == The_BArena())` then we will create a memory leak between subsequent `Finalize()` and `Initialize()` calls.

WeiqunZhang:
Can you explain why they would be a memory leak?

ax3l:
I think because we would loose the handle to the existing arena and assign a fresh one here:
https://github.com/AMReX-Codes/amrex/blob/67384701a808ca973ad2c24ec86cee4c7a81fd05/Src/Base/AMReX_Arena.cpp#L314

ax3l:
> The dtor ~MultiFab is called after amrex::Finalize(). It could be fixed by limiting the scope of mf. But it was so common that we decided to make it work without worrying about the scope.

Can we throw a warning in `amrex::Finalize()`/`~Arena` if there is still allocated memory in them? That would mean people did not scope their AMReX allocs when they reach that point, no?

WeiqunZhang:
But the BArena is a static object in the The_BArena() function.  So it's the same handler.

ax3l:
True, that works.

For the original problem:
> The dtor ~MultiFab is called after amrex::Finalize(). It could be fixed by limiting the scope of mf. But it was so common that we decided to make it work without worrying about the scope.

Can we throw a warning in `amrex::Finalize()`/`~Arena` if there is still allocated memory in them? That would mean people did not scope their AMReX allocs when they reach that point, no?

That would also give them hints that they have dangling / UB variables still around (a MultiFab outliving `Finalize()` will not really work anymore besides for local read-only stuff and being deallocated, right?)

Src/Base/AMReX_Arena.cpp
```diff
@@ -253,12 +253,13 @@ Arena::Initialize ()
     if (initialized) return;
     initialized = true;
 
-    BL_ASSERT(the_arena == nullptr);
+    // see reason on allowed reuse of the default CPU BArena in Arena::Finalize
+    BL_ASSERT(the_arena == nullptr || the_arena == The_BArena());
     BL_ASSERT(the_async_arena == nullptr);
-    BL_ASSERT(the_device_arena == nullptr);
-    BL_ASSERT(the_managed_arena == nullptr);
+    BL_ASSERT(the_device_arena == nullptr || the_arena == The_BArena());
```
WeiqunZhang:
```suggestion
    BL_ASSERT(the_device_arena == nullptr || the_device_arena == The_BArena());
```

Src/Base/AMReX_Arena.cpp
```diff
@@ -253,12 +253,13 @@ Arena::Initialize ()
...
+    BL_ASSERT(the_managed_arena == nullptr || the_arena == The_BArena());
```
WeiqunZhang:
```suggestion
    BL_ASSERT(the_managed_arena == nullptr || the_managed_arena == The_BArena());
```

Src/Base/AMReX_Arena.cpp
```diff
@@ -253,12 +253,13 @@ Arena::Initialize ()
...
     BL_ASSERT(the_pinned_arena == nullptr);
-    BL_ASSERT(the_cpu_arena == nullptr);
+    BL_ASSERT(the_cpu_arena == nullptr || the_arena == The_BArena());
```
WeiqunZhang:
```suggestion
    BL_ASSERT(the_cpu_arena == nullptr || the_cpu_arena == The_BArena());
```

WeiqunZhang (changes requested):

WeiqunZhang (approved):

PR #2950, opened 2022-09-15
Add roundoff_lo corresponding to roundoff_hi for domains that don't start at 0

### LINKED ISSUE #3397: Periodic boundaries show particle depletion with DP fields, SP particles
pic5
X axis labels are cell indices, with periodicity meaning that the end points should match up.

The first plot is the run in full double precision. It shows a flat line of particle densities across the periodic boundary

The latter 3 plots are mixed precision, with `AMReX::Real = double` and `AMReX::ParticleReal = float`. As time progresses, the depth of depletion gets worse. Note the adjustment of y axis limits to suit the values in each plot.

PhilMiller:
## Summary

ECP-WarpX/WarpX#3397 shows an anomaly of particles seeming to be left outside the domain off the low side of a periodic boundary. This kind of condition was fixed for the high side of the domain by adding `roundoff_hi` in `ParticleReal` as necessary in #2839. Do the same for the low side of the domain.

Fixes ECP-WarpX/WarpX#3397

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_Geometry.H
+++ b/Src/Base/AMReX_Geometry.H
@@ -69,22 +69,22 @@ public:
 
     namespace detail {
         template <typename T>
-        T bisect_prob_lo (amrex::Real plo, amrex::Real /*phi*/, amrex::Real idx, int ilo, int ihi, amrex::Real tol) {
+        T bisect_prob_lo (amrex::Real plo, amrex::Real /*phi*/, amrex::Real dxinv, int ilo, int ihi, amrex::Real tol) {
             T lo = static_cast<T>(plo + tol);
             bool safe;
             {
-                int i = int(Math::floor((lo - plo)*idx)) + ilo;
+                int i = int(Math::floor((lo - plo)*dxinv)) + ilo;
                 safe = i >= ilo && i <= ihi;
             }
             if (safe) {
                 return lo;
             } else {
                 // bisect the point at which the cell no longer maps to inside the domain
-                T hi = static_cast<T>(plo + 0.5_rt/idx);
+                T hi = static_cast<T>(plo + 0.5_rt/dxinv);
                 T mid = bisect(lo, hi,
                                [=] AMREX_GPU_HOST_DEVICE (T x) -> T
                                {
-                                   int i = int(Math::floor((x - plo)*idx)) + ilo;
+                                   int i = int(Math::floor((x - plo)*dxinv)) + ilo;
                                    bool inside = i >= ilo && i <= ihi;
                                    return static_cast<T>(inside) - T(0.5);
                                }, static_cast<T>(tol));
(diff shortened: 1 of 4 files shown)


PhilMiller:
@atmyers Here's all the ancillary stuff around defining `roundoff_lo`

Src/Base/AMReX_Geometry.H
```diff
@@ -68,6 +68,10 @@ public:
 };
 
     namespace detail {
+        template <typename T>
+        T bisect_prob_lo (amrex::Real plo, amrex::Real phi, amrex::Real idx, int ilo, int ihi, amrex::Real tol) {
+	  return static_cast<T>(plo + tol);
```
PhilMiller: This obviously needs to be implemented.

Everything else is done, I think.

Src/Base/AMReX_Geometry.H
```diff
@@ -68,6 +68,30 @@ public:
 };
 
     namespace detail {
+        template <typename T>
+        T bisect_prob_lo (amrex::Real plo, amrex::Real /*phi*/, amrex::Real idx, int ilo, int ihi, amrex::Real tol) {
+        T lo = static_cast<T>(plo + tol);
+        bool safe;
+        {
+            int i = int(Math::floor((lo - plo)*idx)) + ilo;
```
PhilMiller: This expression, which I read as "get the cell index `i` corresponding to physical coordinate `lo`" occurs multiple times throughout these functions, and presumably in the rest of the codebase, since it's the expression that the resulting value of `roundoff_lo/hi` needs to safely satisfy. Shouldn't it be in some named function, so users don't err in its details, and for clarity?

PhilMiller: As an aside, the convention throughout the AMReX-verse of `idx` meaning "inverse dx" rather than "index" drives me nuts

Src/Particle/AMReX_ParticleUtil.H
```diff
@@ -602,7 +606,7 @@ partitionParticlesByDest (PTile& ptile, const PLocator& ploc, const ParticleBuff
                 else
                 {
                     auto p_prime = p;
-                    enforcePeriodic(p_prime, plo, phi, rhi, is_per);
+                    enforcePeriodic(p_prime, plo, phi, rhi, rlo, is_per);
```
WeiqunZhang:
```suggestion
                    enforcePeriodic(p_prime, plo, phi, rlo, rhi, is_per);
```

PhilMiller: Ooof, yeah. Thank you

Src/Particle/AMReX_ParticleUtil.H
```diff
@@ -529,7 +530,9 @@ bool enforcePeriodic (P& p,
                 p.pos(idim) -= static_cast<ParticleReal>(phi[idim] - plo[idim]);
             }
             // clamp to avoid precision issues;
-            if (p.pos(idim) < plo[idim]) p.pos(idim) = static_cast<ParticleReal>(plo[idim]);
+            if (p.pos(idim) < rlo[idim]) {
+            p.pos(idim) = rlo[idim];
+        }
```
PhilMiller: Looks like I need to fix the formatting here.

Src/Base/AMReX_Geometry.H
atmyers: I like the idea of a named function, but I'd prefer to change this globally in a separate PR.

PhilMiller:
@WeiqunZhang The ultimate review and approval should probably come from you, since Andrew and I wrote the code

WeiqunZhang:
OK

Src/Base/AMReX_Geometry.H
PhilMiller: Agreed on not doing that in this PR

And on the variable name thing, I see a mix with other sites that call it `dxi`, which is at least less of a collision. I think `invdx` or `dxinv` would ultimately be preferable.

WeiqunZhang: I think the use of `idx` for inverse dx in amrex is limited to just a few relatively new functions in Geometry.  In most other places, it does mean index.  So it might not be too late to change it.  In other places, we usually use `dxi` or `dxinv`.

PhilMiller:
There's an AMReX unit test failing:
https://github.com/AMReX-Codes/amrex/actions/runs/3062795379/jobs/4944165089
Conversely, my physics test is still very early, but it seem to look better than before

PhilMiller:
This seems to fix our observed physics issue:
image

PhilMiller:
As I told Andrew, I'm going to wait on CI, then rebase to clean up the history, and then I'll mark this Ready

WeiqunZhang (approved):

PR #2953, opened 2022-09-16
Add error message in ParallelFor for CUDA 11.6 bug

roelof-groenewald:
## Summary

Seemingly the same `nvcc` bug first detected in issue #2607 also now causes a failure during the WarpX B-field evolve step. This PR simply adds a user friendly error statement when this happens to help identify the problem.
It is still possible that the same bug would impact other AMReX functions. The location of the current error statement covers the observed issue without impacting other working cases (such as electrostatic WarpX simulations).

## Checklist

The proposed changes:


WeiqunZhang:
I don't think we should add assertion here, because the issue is caused by the particular callable objects in WarpX, not the ParallelFor function here.  So the assertion should be added in WarpX's EvolveB.

roelof-groenewald:
Oh, sorry I misunderstood then. I thought the `nvcc` bug actually made that AMReX function unusable.
I'll move the assert to a WarpX PR.

PR #2958, opened 2022-09-22
Add GPU-compatible upper bound and lower bound algorithms to AMReX_Algorithm

lucafedeli88:
## Summary

This PR adds a GPU-compatible replacement of the STL `upper_bound` and  `lower_bound` algorithms to `AMReX_Algorithm.H`. If `AMREX_DEVICE_COMPILE` is not defined, it defaults back to the STL algorithms.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_Algorithm.H
+++ b/Src/Base/AMReX_Algorithm.H
@@ -167,7 +167,7 @@ namespace amrex
             auto it = first;
             const auto step = count/2;
             it += step;
-            if (!(val<*it)){
+            if (!(val < *it)){
                 first = ++it;
                 count -= step + 1;
             }
@@ -193,7 +193,7 @@ namespace amrex
             auto it = first;
             const auto step = count/2;
             it += step;
-            if (!(val<=*it)){
+            if (*it < val){
                 first = ++it;
                 count -= step + 1;
             }
--- a/Src/Base/AMReX_Algorithm.H
+++ b/Src/Base/AMReX_Algorithm.H
@@ -162,12 +162,10 @@ namespace amrex
     ItType upper_bound (ItType first, ItType last, const ValType& val)
     {
 #if AMREX_DEVICE_COMPILE
-        ItType it;
-        std::ptrdiff_t count, step;
-        count = last-first;
+        std::ptrdiff_t count = last-first;
         while(count>0){
-            it = first;
-            step = count/2;
+            auto it = first;
+            const auto step = count/2;
             it += step;
             if (!(val<*it)){
                 first = ++it;
@@ -190,7 +188,8 @@ namespace amrex
     {
 #ifdef AMREX_DEVICE_COMPILE
         std::ptrdiff_t count = last-first;
-        do{
+        while(count>0)
+        {
             auto it = first;
             const auto step = count/2;
             it += step;
@@ -201,7 +200,7 @@ namespace amrex
             else{
                 count = step;
             }
-        } while(count>0);
+        }
 
         return first;
 #else
--- a/Src/Base/AMReX_Algorithm.H
+++ b/Src/Base/AMReX_Algorithm.H
@@ -157,6 +157,58 @@ namespace amrex
         return hi;
     }
 
+    template<typename ItType, typename ValType>
+    AMREX_GPU_DEVICE
+    ItType upper_bound (ItType first, ItType last, const ValType& val)
+    {
+#if AMREX_DEVICE_COMPILE
+        ItType it;
+        size_t count, step;
+        count = last-first;
+        while(count>0){
+            it = first;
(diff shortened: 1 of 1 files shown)


Src/Base/AMReX_Algorithm.H
```diff
@@ -157,6 +157,58 @@ namespace amrex
         return hi;
     }
 
+    template<typename ItType, typename ValType>
+    AMREX_GPU_DEVICE
```
WeiqunZhang: Could we make it `AMREX_GPU_HOST_DEVICE`?

Src/Base/AMReX_Algorithm.H
```diff
@@ -157,6 +157,58 @@ namespace amrex
...
+    ItType upper_bound (ItType first, ItType last, const ValType& val)
+    {
+#if AMREX_DEVICE_COMPILE
+        ItType it;
+        size_t count, step;
```
WeiqunZhang: I think it might not be safe to use `size_t`.  I think we should use `std::ptrdiff_t`.

Src/Base/AMReX_Algorithm.H
```diff
@@ -157,6 +157,58 @@ namespace amrex
...
+#endif
+    }
+
+    template<typename ItType, typename ValType>
+    AMREX_GPU_DEVICE
+    ItType lower_bound (ItType first, ItType last, const ValType& val)
+    {
+#ifdef AMREX_DEVICE_COMPILE
+        size_t count = last-first;
+        do{
```
WeiqunZhang: The implementation here uses `do-while`, whereas `upper_bound` uses `while`.  Maybe this is correct.  But why not using the same style as `upper_bound` (which seems the style used by gcc and llvm)?

WeiqunZhang (changes requested):

Src/Base/AMReX_Algorithm.H
lucafedeli88: sure!

Src/Base/AMReX_Algorithm.H
lucafedeli88: sure!

Src/Base/AMReX_Algorithm.H
lucafedeli88: Sure!

lucafedeli88:
@WeiqunZhang , I should have implemented all your comments

Src/Base/AMReX_Algorithm.H
```diff
@@ -157,6 +157,57 @@ namespace amrex
...
+    ItType lower_bound (ItType first, ItType last, const ValType& val)
+    {
+#ifdef AMREX_DEVICE_COMPILE
+        std::ptrdiff_t count = last-first;
+        while(count>0)
+        {
+            auto it = first;
+            const auto step = count/2;
+            it += step;
+            if (!(val<=*it)){
```
WeiqunZhang: I think this should be `*it < value`, because the algorithm is only supposed to use `<`, not `<=`.

lucafedeli88: I think that we actually need `<=` with this formulation. Otherwise with the vector  `[0,1,2,3,4,5,5,6,6,7]` passing `0` as value gives `1` instead of `0` and passing `7` gives `END` instead of `7`.

However, it could be reformulated as `if (val > *it )` . I can do that if you think that it would be better

WeiqunZhang: I am confused.  Is `if (val > *it)` mathematically equivalent to `if (*it <val)` (which I suggested) for your example?  Maybe you didn't remove `!`?

WeiqunZhang: We should only use `<`, not `>`, because a data type with `<` but without `>` works with std::lower_bound.

lucafedeli88: Oh sorry! My fault... I misread your comment: I thought that you meant `!(val<*it)` !
You are totally right!

WeiqunZhang: It won't make difference for arithmetic types.  But the user will expect this to work too.
```
struct foo {
    int i;
    bool operator< (foo const& rhs) const {
        return i < rhs.i;
    }
    // No operator> defined
};
```

lucafedeli88: I'll do the correction right now

lucafedeli88: Sorry again, I've implemented your suggestion now

WeiqunZhang: No problems.  Thanks for your hard work!

WeiqunZhang (approved):

PR #2961, opened 2022-09-23
Volume weighted sum

WeiqunZhang:
Add a new function doing volume weighted sum across AMR levels.  This may not be exactly what amrex application codes want.  But it should work for many cases.

## Checklist

The proposed changes:


zingale:
Why does it ignore rz weighting?

WeiqunZhang:
Trying to keep it simple.  Nevertheless, we could add support for rz.

WeiqunZhang:
For completeness, 1d spherical and 2d rz are supported.

asalmgren (approved):

PR #2963, opened 2022-09-26
Fix MLEBNodeFDLaplacian bottom solver

WeiqunZhang:
MLEBNodeFDLaplacian is never singular because it has Dirichlet boundary on the EB surface.  We did set the singular flag to false, but forgot about the bottom solver used a different function to query.  This fixes it by overriding the isBottomSingular function.  Note that this linear operator is used by WarpX.

## Checklist

The proposed changes:


WeiqunZhang:
https://github.com/ECP-WarpX/WarpX/issues/3429
@RemiLehe

ax3l (approved):

PR #2974, opened 2022-10-07
Runge-Kutta support for AMR

WeiqunZhang:
This adds RK2, RK3 and RK4 in a new namespace RungeKutta.  Together with the enhanced FillPatcher class, these functions can be used for RK time stepping in AMR simulations.  A new function AmrLevel::RK is added for AmrLevel based codes.  See CNS::advance in Tests/GPU/CNS/CNS_advance.cpp for an example of using the new AmrLevel::RK function.

The main motivation for this PR is that ghost cell filling for high order (> 2) RK methods at coarse/fine boundary is non-trivial when there is subcycling.

## Checklist

The proposed changes:

### DIFF
--- a/Src/AmrCore/AMReX_FillPatcher.H
+++ b/Src/AmrCore/AMReX_FillPatcher.H
@@ -207,7 +207,7 @@ private:
     int m_ncomp;
     InterpBase* m_interp;
     EB2::IndexSpace const* m_eb_index_space = nullptr;
-    FabArrayBase m_sfine;
+    MF m_sfine;
     IntVect m_ratio;
     Vector<std::pair<Real,std::unique_ptr<MF>>> m_cf_crse_data;
     std::unique_ptr<MF> m_cf_crse_data_tmp;
@@ -234,7 +234,7 @@ FillPatcher<MF>::FillPatcher (BoxArray const& fba, DistributionMapping const& fd
       m_ncomp(ncomp),
       m_interp(interp),
       m_eb_index_space(eb_index_space),
-      m_sfine(fba, fdm, 1, nghost)
+      m_sfine(fba, fdm, 1, nghost, MFInfo().SetAlloc(false))
 {
     static_assert(IsFabArray<MF>::value,
                   "FillPatcher<MF>: MF must be FabArray type");
--- a/Src/AmrCore/AMReX_FillPatcher.H
+++ b/Src/AmrCore/AMReX_FillPatcher.H
@@ -55,6 +55,17 @@ namespace amrex {
  * allowed to fill only some of the components with the fill function.
  *
  * (5) This only works for cell-centered and nodal data.
+ *
+ * This class also provides support for RungeKutta::RK3 and RungeKutta::RK4.
+ * The storeRKCoarseData function can be used to store coarse AMR level
+ * data that are needed for filling fine level data's ghost cells in this
+ * class.  The `fillRK` function can be used to fill ghost cells for fine
+ * AMR levels.  This operation at the coarse/fine boundary is non-trivial
+ * for RK orders higher than 2.  Note that it is expected that time stepping
+ * on the coarse level is perform before any fine level time stepping, and
+ * it's the user's reponsibility to properly create and destroy this object.
+ * See AmrLevel::RK for an example of using the RungeKutta functions and
+ * FillPatcher together.
  */
 
 template <class MF = MultiFab>
@@ -153,6 +164,37 @@ public:
                                  PreInterpHook const& pre_interp = {},
                                  PostInterpHook const& post_interp = {});
 
+    /**
+     * \brief Store coarse AMR level data for RK3 and RK4
+     *
+     * \tparam order RK order.  Must be 3 or 4.
+     * \param time   time at the beginning of the step
+     * \param dt     time step
+     * \param S_old  data at time
+     * \param RK_k   right-hand side at RK stages
+     */
+    template <std::size_t order>
+    void storeRKCoarseData (Real time, Real dt, MF const& S_old,
+                            Array<MF,order> const& RK_k);
+
+    /**
+     * \brief Fill ghost cells of fine AMR level for RK3 and RK4
+     *
+     * \param stage     RK stage number starting from 1
+     * \param iteration iteration number on fine level during a coarse time
+     *                  step.  For an AMR simulation with subcycling and a
+     *                  refinement ratio of 2, the number is either 1 or 2,
+     *                  denoting the first and second substep, respectively.
+     * \param ncycle    number of subcyling steps.  It's usually 2 or 4.
+     *                  Without subcycling, this will be 1.
+     * \param cbc       filling physical boundary on coarse level
+     * \param fbc       filling physical boundary on fine level
+     * \param bcs       physical BC types
+     */
+    template <typename BC>
+    void fillRK (int stage, int iteration, int ncycle, MF& mf, Real time,
+                 BC& cbc, BC& fbc, Vector<BCRec> const& bcs);
+
 private:
 
     BoxArray m_fba;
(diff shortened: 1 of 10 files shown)


WeiqunZhang:
I used this test https://github.com/WeiqunZhang/amrex-devtests/tree/main/rk34 and got expected convergence rate for max_leve =0 and 1.  I also intentionally used slightly incorrect values in the interpolation of coarse/fine boundary, and observed converge rate degradation.

Src/Base/AMReX_RungeKutta.H
```diff
@@ -0,0 +1,287 @@
...
+ * and RK4.  The function templates take the old data in FabArray/MultiFab
+ * as input, and evolve the system for one time step.  The result is stored
+ * in another FabArray/MultiFab.  These two FabArrays must have ghost cells
+ * if they are needed for evaluating the right-hand side.  The functions
+ * take three callable objects for computing the right-hand side, filling
+ * ghost cells, and optionally post-processing RK stage results.  For RK3
+ * and RK4, they also need a callable object for storing the data needed for
+ * filling coarse/fine boundaries in AMR simulations.
+ *
+ * The callable object for righ-hand side has the signature of `void(int
```
jmsexton03:
```suggestion
 * The callable object for right-hand side has the signature of `void(int
```

Src/Base/AMReX_RungeKutta.H
```diff
@@ -0,0 +1,287 @@
...
+
+    // RK4 stage 2
+    Real t_half = time + Real(0.5)*dt;
+    fillbndry(2, Unew, t_half);
+    frhs(2, rkk[1], Unew, dt/Real(3.));
+    // Unew = Uold + k2 * dt/2
+    detail::rk_update(Unew, Uold, rkk[1], Real(0.5)*dt);
+    post_stage(2, Unew);
+
+    // RK4 Step 3
```
jmsexton03:
```suggestion
    // RK4 stage 3
```

Src/Base/AMReX_RungeKutta.H
```diff
@@ -0,0 +1,287 @@
...
+    fillbndry(3, Unew, t_half);
+    frhs(3, rkk[2], Unew, dt/Real(3.));
+    // Unew = Uold + k3 * dt;
+    detail::rk_update(Unew, Uold, rkk[2], dt);
+    post_stage(3, Unew);
+
+    // Step 4 of RK4
```
jmsexton03:
```suggestion
    // RK4 stage 4
```

Src/AmrCore/AMReX_FillPatcher.H
```diff
@@ -55,6 +55,17 @@ namespace amrex {
  * allowed to fill only some of the components with the fill function.
  *
  * (5) This only works for cell-centered and nodal data.
+ *
+ * This class also provides support for RungeKutta::RK3 and RungeKutta::RK4.
+ * The storeRKCoarseData function can be used to store coarse AMR level
+ * data that are needed for filling fine level data's ghost cells in this
+ * class.  The `fillRK` function can be used to fill ghost cells for fine
+ * AMR levels.  This operation at the coarse/fine boundary is non-trivial
+ * for RK orders higher than 2.  Note that it is expected that time stepping
```
jmsexton03, line 64: Can you add a comment here or in the fillRK doxygen about how to do that trivial RK2 ghost cell fill? Is that the functionality demonstrated in the CNS test modifications in this PR? Is there a reason to not have this class support RungeKutta::RK2 as well for completeness?

Src/AmrCore/AMReX_FillPatcher.H
```diff
@@ -330,14 +385,202 @@ FillPatcher<MF>::fillCoarseFineBoundary (MF& mf, IntVect const& nghost, Real tim
...
+void FillPatcher<MF>::fillRK (int stage, int iteration, int ncycle,
+                              MF& mf, Real time, BC& cbc, BC& fbc,
+                              Vector<BCRec> const& bcs)
+{
+    int rk_order = m_cf_crse_data.size()-1;
+    if (rk_order != 3 && rk_order != 4) {
+        amrex::Abort("FillPatcher: unsupported RK order "+std::to_string(rk_order));
+        return;
+    }
+    AMREX_ASSERT(stage > 0 && stage <= rk_order);
```
jmsexton03, line 427: Is `storeRKCoarseData` templated on the order to make it easier on the compiler? It looks like `fillRK` can deduce the Runge Kutta order from FillPatcher member objects. Is this a safe assumption, and if so, can it be made in both functions for consistency?

Src/Base/AMReX_RungeKutta.H
```diff
@@ -0,0 +1,287 @@
...
+ * stage.  The FillPatcher class can be useful for implementing such a
+ * callable.  See AmrLevel::RK for an example.
+ *
+ * The callable object for post-processing stage results is optional.  It's
+ * no-op by default.  Its function signature is `void(int stage, MF& u)`,
+ * where `stage` is the RK stage number and `u` is the result of that stage.
+ *
+ * For RK3 and RK4, one must also provide a callable object with the
+ * signature of `void(Array<MF,order> const& rkk)`, where `order` is the RK
+ * order and `rkk` contains the righ-hand side at all the RK stages.  The
```
jmsexton03:
```suggestion
 * order and `rkk` contains the right-hand side at all the RK stages.  The
```

Src/Base/AMReX_RungeKutta.H
```diff
@@ -0,0 +1,287 @@
...
+
+#include <AMReX_FabArray.H>
+
+namespace amrex {
+
+/**
+ * \brief Functions for Runge-Kutta methods
+ *
+ * This namespace RungeKutta has functions for a number RK methods, RK2, RK3
+ * and RK4.  The function templates take the old data in FabArray/MultiFab
```
jmsexton03: Can you include some context/naming here (or butcher tables later?) For instance, is RK3 a specific SSPRK3 method?

Src/AmrCore/AMReX_FillPatcher.H
WeiqunZhang: `storeRKCoarseData` is templated not for the benefit of the compiler.  It it done this way trying to make the user less likely to make a mistake, because it won't compile if the user calls `RungeKutta::RK3(..., [&] (Array<MultiFab,4> const& rkk) {...});`.  If it's not a template, then the user will be asked to provide a `Vector<MultiFab>`.  Then the assertion on the size can only be done at run time.  As for `fillRK`, it's called after `storeRKCoarseData` and it does not have any parameters that depend on the order.

Src/Base/AMReX_RungeKutta.H
WeiqunZhang: Good idea.  I will include links to wikipedia.

By the way, the name SSPRK3 appeared many years after the method was presented in a paper by Shu & Osher.  So I never got used to the name.

Src/AmrCore/AMReX_FillPatcher.H
WeiqunZhang: The reason I don't want to make it support RK2 is that, for RK2, we don't need to call `storeRKCoarseData`.  So I think it might add more confusion if it supports RK2.  Also, filling ghost cells for RK2 can be done with the normal fillpath functions.  For the CNS code, the RK part of the code is actually very clean because it's based on AmrLevel.
```
    RK(rk_order, State_Type, time, dt, iteration, ncycle,
       // Given state S, compute dSdt. dtsub is needed for flux register operations                                                                                          
       [&] (int /*stage*/, MultiFab& dSdt, MultiFab const& S, Real dtsub) {
           compute_dSdt(S, dSdt, dtsub, fr_as_crse, fr_as_fine);
       },
       // Optional. In case if there is anything needed after each RK substep.                                                                                               
       [&] (int /*stage*/, MultiFab& S) { computeTemp(S,0); });
```
This works for RK2, 3 and 4.

jmsexton03 (approved):

PR #2980, opened 2022-10-11
Template average down

bsrunnels:
## Summary
Adding templating for the average_down function

## Additional background
See PR #2979

## Checklist

The proposed changes:


Src/Base/AMReX_MultiFabUtil_2D_C.H
```diff
@@ -259,7 +260,7 @@ void amrex_avgdown (Box const& bx, Array4<Real> const& crse,
         for (int i = clo.x; i <= chi.x; ++i) {
             int ii = i*facx;
             int jj = j*facy;
-            Real c = 0.;
+            T c = {};
```
bsrunnels: @WeiqunZhang 
Is there a better way to do template-safe zero initialization? There are many libraries (like `eigen`) that do not use default constructors as zero constructors.

WeiqunZhang: `auto c = T(0);`?

WeiqunZhang: Or `T c = 0;`

Src/Base/AMReX_MultiFabUtil_1D_C.H
```diff
@@ -198,7 +199,7 @@ void amrex_avgdown (Box const& bx, Array4<Real> const& crse,
     for (int n = 0; n < ncomp; ++n) {
         for (int i = clo.x; i <= chi.x; ++i) {
             int ii = i*facx;
-            Real c = 0.;
+            T c = 0.0*fine(ii,0,0,n+fcomp);
```
bsrunnels: @WeiqunZhang 
Would something like this be acceptable? I'm guessing not, but it is nice because it relies only on the existence of a `* (Real,T)` operation, which has to exist anyway.

I want to see if there is a way to implement this without using a zero initializer, because there is not a consistent way to zero initialize for most data types. For instance if `T=amrex::Real` then you can do `T c = 0`, but if `T = eigen::Matrix3d` then you would need to initialize with `T c = T::Zero()`. 

Alternatively I could implement an SFINAE solution with a specialized version for `amrex::Real` that uses `T c = 0` and a general version for other types?

WeiqunZhang: I think it should be the opposite. `T c = 0` is certainly more general than `T::Zero`.  Maybe you can have a free function that returns `T(0)` and a specialized version for the types you want to use.  Ideally the function should be constexpr.

WeiqunZhang:
@bsrunnels I actually need the changes in this PR.  So could you merge development into this branch?  I think we can get this PR merged first and then try to make it work for Matrix3d.

Src/Base/AMReX_MultiFabUtil_1D_C.H
```diff
@@ -207,30 +208,32 @@ void amrex_avgdown (Box const& bx, Array4<Real> const& crse,
...
+template <typename T>
 AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
-void amrex_avgdown (int i, int, int, int n, Array4<Real> const& crse,
-                    Array4<Real const> const& fine,
+void amrex_avgdown (int i, int, int, int n, Array4<T> const& crse,
+                    Array4<T const> const& fine,
                     int ccomp, int fcomp, IntVect const& ratio) noexcept
 {
     const int facx = ratio[0];
     const Real volfrac = 1.0/static_cast<Real>(facx);
```
WeiqunZhang: When `T` is float and `Real` is double, this will result in warning on float conversion later.  So maybe what you could do for Matrix3d is to specialize `amrex_avgdown` for Matrix3d.

Src/Base/AMReX_MultiFabUtil_2D_C.H
```diff
@@ -269,17 +270,18 @@ void amrex_avgdown (Box const& bx, Array4<Real> const& crse,
...
 AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
-void amrex_avgdown (int i, int j, int, int n, Array4<Real> const& crse,
-                    Array4<Real const> const& fine,
+void amrex_avgdown (int i, int j, int, int n, Array4<T> const& crse,
+                    Array4<T const> const& fine,
                     int ccomp, int fcomp, IntVect const& ratio) noexcept
 {
     const int facx = ratio[0];
     const int facy = ratio[1];
     const Real volfrac = static_cast<Real>(1.0)/static_cast<Real>(facx*facy);
```
WeiqunZhang: Like in 1D, this may result in float conversion warning.

Src/Base/AMReX_MultiFabUtil_3D_C.H
```diff
@@ -376,7 +378,7 @@ void amrex_avgdown (int i, int j, int k, int n, Array4<Real> const& crse,
     const int ii = i*facx;
```
WeiqunZhang, line 380: Like in 1d and 2d, `const Real volfrac = Real(1.0)/(facx*facy*facz);` could result in float conversion warning.

WeiqunZhang:
I could push the changes, if that's OK with you.

WeiqunZhang (approved):

PR #2983, opened 2022-10-11
MFIter::Finalize

ax3l:
## Summary

Add a Finalize function to MFIter.

The idea about this is, that we can call this already before destruction in Python, where `for` loops do not create scope.

This function must be robust enough to be called again in the constructor (or we need to add an extra bool to guard that it is not called again in the destructor).

## Additional background

See: https://github.com/AMReX-Codes/pyamrex/pull/30 and WarpX dev meeting notes from Oct 11th, 2022.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_MFIter.H
+++ b/Src/Base/AMReX_MFIter.H
@@ -180,6 +180,7 @@ protected:
     IndexType     typ;
 
     bool          dynamic;
+    bool          initialized;
 
     struct DeviceSync {
         DeviceSync () = default;
--- a/Src/Base/AMReX_MFIter.H
+++ b/Src/Base/AMReX_MFIter.H
@@ -201,6 +201,7 @@ protected:
     static AMREX_EXPORT int allow_multiple_mfiters;
 
     void Initialize ();
+    void Finalize ();
 };
 
 //! Is it safe to have these two MultiFabs in the same MFiter?
--- a/Src/Base/AMReX_MFIter.H
+++ b/Src/Base/AMReX_MFIter.H
@@ -180,7 +180,7 @@ protected:
     IndexType     typ;
 
     bool          dynamic;
-    bool          initialized;
+    bool          finalized = false;
 
     struct DeviceSync {
         DeviceSync () = default;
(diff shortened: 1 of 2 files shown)


ax3l:
Can we design a good test that calls `MFIter::Finalize` twice?

WeiqunZhang:
I think we should add a boolean flag.  Even if it's robust for being called repeated, there are still a lot of work in there.

Src/Base/AMReX_MFIter.H
```diff
@@ -180,6 +180,7 @@ protected:
     IndexType     typ;
 
     bool          dynamic;
+    bool          initialized;
```
WeiqunZhang: How about just
```suggestion
    bool          finalized = false;
```
Also note that this is not a static variable.  So we don't need to do anything special for OMP.  All we have to do is flip the flag in Finalize after testing the flag.

Src/Base/AMReX_MFIter.cpp
```diff
@@ -209,10 +209,19 @@ MFIter::MFIter (const FabArrayBase& fabarray_, const MFItInfo& info)
...
+void
+MFIter::Finalize ()
+{
+    if (!initialized) return;
+
 #ifdef AMREX_USE_OMP
 #pragma omp master
 #endif
     {
+        initialized = false;
```
WeiqunZhang: Unlike `depth`, this is thread private.

ax3l: Good point, fixed.

Src/Base/AMReX_MFIter.cpp
```diff
@@ -209,12 +209,14 @@ MFIter::MFIter (const FabArrayBase& fabarray_, const MFItInfo& info)
 
 MFIter::~MFIter ()
 {
-#ifdef AMREX_USE_OMP
-#pragma omp master
-#endif
-    {
-        depth = 0;
-    }
```
WeiqunZhang: Looks this is accidentally deleted.

ax3l: Ah, I was already wondering. Thx.

WeiqunZhang (approved):

PR #2985, opened 2022-10-11
MFIter: Make Finalize Public

ax3l:
## Summary

Follow-up to #2983

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_MFIter.cpp
+++ b/Src/Base/AMReX_MFIter.cpp
@@ -215,8 +215,12 @@ MFIter::~MFIter ()
 void
 MFIter::Finalize ()
 {
+    // avoid double finalize
     if (finalized) return;
     finalized = true;
+
+    // mark as invalid
+    currentIndex = endIndex;
 
 #ifdef AMREX_USE_OMP
 #pragma omp master
--- a/Src/Base/AMReX_MFIter.H
+++ b/Src/Base/AMReX_MFIter.H
@@ -164,6 +164,8 @@ public:
 
     static int allowMultipleMFIters (int allow);
 
+    void Finalize ();
+
 protected:
 
     std::unique_ptr<FabArrayBase> m_fa;  //!< This must be the first member!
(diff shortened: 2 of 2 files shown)


ax3l:
Should we do something to `nextDynamicIndex` and `currentIndex` in `MFIter::Finalize`?

That would prevent that `operator++` (and other public member functions) create an invalid state on an already finalized `MFIter`.

WeiqunZhang:
I guess we have to trust pyamrex users.

WeiqunZhang (approved):

ax3l:
> I guess we have to trust pyamrex users.

They should not be able to call this, we just need it when implementing our [wrapper iterator](https://github.com/AMReX-Codes/pyamrex/blob/82d6260d85d70ecc8f3de679b8ed1fc9d8f53758/src/Base/MultiFab.cpp#L58-L80) while we are still in C++ :D
But can we trust other C++ users not to call it? :D

Jokes aside, should I just set both to the value of `endIndex` to add some minor safety?

I would like to be at least able to call `isValid()` on it:
https://github.com/AMReX-Codes/pyamrex/blob/82d6260d85d70ecc8f3de679b8ed1fc9d8f53758/src/Base/MultiFab.cpp#L44

Src/Base/AMReX_MFIter.cpp
```diff
@@ -215,9 +215,13 @@ MFIter::~MFIter ()
 void
 MFIter::Finalize ()
 {
+    // avoid double finalize
     if (finalized) return;
     finalized = true;
 
+    // mark as invalid
+    currentIndex = endIndex;
```
WeiqunZhang, line 223: It won't hurt, although it's already invalid in normal usage, otherwise the MFIter loop won't stop.

ax3l: At least on our side, that's also the only location we would call this:
https://github.com/AMReX-Codes/pyamrex/blob/82d6260d85d70ecc8f3de679b8ed1fc9d8f53758/src/Base/MultiFab.cpp#L72-L75

I was thinking to add some safety if someone in C++ calls `mfi.Finalize()` mid loop - but why would they... maybe to abort an MFIter loop in a clean way...

WeiqunZhang: They can use `goto`.   That might be the cleanest way.  :smile:

PR #2987, opened 2022-10-13
Fourth-order interpolation from fine to coarse level

WeiqunZhang:
For fourth-order finite-difference methods with data at cell centers, we cannot use the usual averageDown function to overwrite coarse level data with fine data.  We actually need to do interpolation.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_MultiFabUtil.cpp
+++ b/Src/Base/AMReX_MultiFabUtil.cpp
@@ -1227,157 +1227,244 @@ namespace amrex
         return hv;
     }
 
-        Real volumeWeightedSum (Vector<MultiFab const*> const& mf, int icomp,
-                                Vector<Geometry> const& geom,
-                                Vector<IntVect> const& ratio,
-                                bool local)
-        {
-            ReduceOps<ReduceOpSum> reduce_op;
-            ReduceData<Real> reduce_data(reduce_op);
+    Real volumeWeightedSum (Vector<MultiFab const*> const& mf, int icomp,
+                            Vector<Geometry> const& geom,
+                            Vector<IntVect> const& ratio,
+                            bool local)
+    {
+        ReduceOps<ReduceOpSum> reduce_op;
+        ReduceData<Real> reduce_data(reduce_op);
 
 #ifdef AMREX_USE_EB
-            bool has_eb = !(mf[0]->isAllRegular());
+        bool has_eb = !(mf[0]->isAllRegular());
 #endif
 
-            int nlevels = mf.size();
-            for (int ilev = 0; ilev < nlevels-1; ++ilev) {
-                iMultiFab mask = makeFineMask(*mf[ilev], *mf[ilev+1], IntVect(0),
-                                              ratio[ilev],Periodicity::NonPeriodic(),
-                                              0, 1);
-                auto const& m = mask.const_arrays();
-                auto const& a = mf[ilev]->const_arrays();
-                auto const dx = geom[ilev].CellSizeArray();
-                Real dv = AMREX_D_TERM(dx[0],*dx[1],*dx[2]);
-#ifdef AMREX_USE_EB
-                if (has_eb) {
-                    AMREX_ASSERT(mf[ilev]->hasEBFabFactory());
-                    auto const& f = dynamic_cast<EBFArrayBoxFactory const&>
-                        (mf[ilev]->Factory());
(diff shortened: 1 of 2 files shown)


WeiqunZhang:
Note that there are some white space changes in this PR.

Src/Base/AMReX_MultiFabUtil.cpp
```diff
@@ -1227,157 +1227,244 @@ namespace amrex
...
-                        return dv*a[box_no](i,j,k,icomp);
+                        return m[box_no](i,j,k) ? Real(0.)
+                            : dv*a[box_no](i,j,k,icomp);
                     });
                 }
             }
+            Gpu::streamSynchronize();
+        }
+
+        auto const& a = mf.back()->const_arrays();
```
jmsexton03, line 1318: Does it make sense to use `nlevels-1` here (and the following `reduce_op.eval` calls)? This is primarily a readability question (assuming I understood the length of mf)
```suggestion
        auto const& a = mf[nlevels-1]->const_arrays();
```

Src/Base/AMReX_MultiFabUtil.cpp
```diff
@@ -1227,157 +1227,244 @@ namespace amrex
...
+            } else
+#endif
+            {
+                reduce_op.eval(*mf.back(), IntVect(0), reduce_data,
+                [=] AMREX_GPU_DEVICE (int box_no, int i, int j, int k) noexcept
+                {
+                    return dv*a[box_no](i,j,k,icomp);
+                });
+            }
+        }
```
jmsexton03, line 1373:
```suggestion
#ifdef AMREX_USE_EB
        if (has_eb) {
            AMREX_ASSERT(mf[nlevels-1]->hasEBFabFactory());
            auto const& f = dynamic_cast<EBFArrayBoxFactory const&>
                (mf[nlevels-1]->Factory());
            auto const& vfrac = f.getVolFrac();
            auto const& va = vfrac.const_arrays();
            reduce_op.eval(*mf[nlevels-1], IntVect(0), reduce_data,
            [=] AMREX_GPU_DEVICE (int box_no, int i, int j, int k) noexcept
                           -> Real
            {
                return dv*a[box_no](i,j,k,icomp)*va[box_no](i,j,k);
            });
        } else
#endif
        {
#if (AMREX_SPACEDIM == 1)
            if (geom[nlevels-1].IsSPHERICAL()) {
                const auto rlo = geom[nlevels-1].CellSize(0);
                reduce_op.eval(*mf[nlevels-1], IntVect(0), reduce_data,
                [=] AMREX_GPU_DEVICE (int box_no, int i, int j, int k)
                               noexcept -> Real
                {
                    constexpr Real pi = Real(3.1415926535897932);
                    Real ri = rlo + dx[0]*i;
                    Real ro = ri + dx[0];
                    return Real(4./3.)*pi*(ro-ri)*(ro*ro+ro*ri+ri*ri)
                        * a[box_no](i,j,k,icomp);
                });
            } else
#elif (AMREX_SPACEDIM == 2)
            if (geom[nlevels-1].IsRZ()) {
                const auto rlo = geom[nlevels-1].CellSize(0);
                reduce_op.eval(*mf[nlevels-1], IntVect(0), reduce_data,
                [=] AMREX_GPU_DEVICE (int box_no, int i, int j, int k)
                               noexcept -> Real
                {
                    Real ri = rlo + dx[0]*i;
                    Real ro = ri + dx[0];
                    constexpr Real pi = Real(3.1415926535897932);
                    return pi*dx[1]*dx[0]*(ro+ri)
                        * a[box_no](i,j,k,icomp);
                });
            } else
#endif
            {
                reduce_op.eval(*mf[nlevels-1], IntVect(0), reduce_data,
                [=] AMREX_GPU_DEVICE (int box_no, int i, int j, int k) noexcept
                {
                    return dv*a[box_no](i,j,k,icomp);
                });
            }
        }
```

jmsexton03 (approved):

Src/Base/AMReX_MultiFabUtil.cpp
WeiqunZhang: Sorry.  Too Late.  I enabled auto-merge.   Nevertheless, one could argue that `mf.back()` is better because one is less likely to make a mistake such as `mf[nlevels]`.

PR #2988, opened 2022-10-14
`MFIter::Finalize`: Free `m_fa`

ax3l:
## Summary

This `free` should potentially not be delayed until the destructor is called.

## Additional background

Follow-up to #2985 #2983

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_MFIter.cpp
+++ b/Src/Base/AMReX_MFIter.cpp
@@ -249,6 +249,7 @@ MFIter::Finalize ()
 #pragma omp single
 #endif
         m_fa->clearThisBD();
+        m_fa.reset(nullptr);
     }
 }
 
--- a/Src/Base/AMReX_MFIter.cpp
+++ b/Src/Base/AMReX_MFIter.cpp
@@ -249,6 +249,8 @@ MFIter::Finalize ()
 #pragma omp single
 #endif
         m_fa->clearThisBD();
+    }
+    if (m_fa) {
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
The FabArrayBase does not actually hold any data.  If you want to destroy it in Finish(), it should not be inside omp single.

Src/Base/AMReX_MFIter.cpp
```diff
@@ -249,6 +249,7 @@ MFIter::Finalize ()
 #pragma omp single
 #endif
         m_fa->clearThisBD();
+        m_fa.reset(nullptr);
     }
```
ax3l:
```suggestion
    }
    if (m_fa) {
        m_fa.reset(nullptr);
    }
```

ax3l:
> The FabArrayBase does not actually hold any data.

Ok, that explains why I did not see an Arena lifetime violation with it.

>  If you want to destroy it in Finish(), it should not be inside omp single.

Thanks, fixed! My mind blended out these macros while patching...

WeiqunZhang:
> > If you want to destroy it in Finish(), it should not be inside omp single.
> 
> Thanks, fixed! My mind blended out these macros while patching...

I was careless.  Your code was not wrong.  The pragam only applies to the next line.  Anyway, either way is fine.

WeiqunZhang (approved):

ax3l:
No worries, thanks! :)

PR #2991, opened 2022-10-15
Pre- and Post-interpolation hook interface

### LINKED ISSUE #2989: InterpHooks are incompatible between FillPatcher class and FillPatch function
The FillPatcher class requires PreInterpHook and PostInterpHook to have the signature:
`void InterpHook(amrex::MultiFab &mf, int scomp, int ncomp);`

whereas the FillPatch function requires the interpolation hooks to have the signature:
`void InterpHook(amrex::FArrayBox &fab, const amrex::Box& box, int scomp, int ncomp);`

This means that different interp hooks have to be used for each way of doing the fillpatch, which doesn't really make sense. The MultiFab version should be more efficient, since we can use the MultiFab ParallelFor, so it would be ideal if we can change the signature used in `AMReX_FillPatchUtil_I.H:548` and similar.

Problem appears here: https://github.com/BenWibking/quokka/pull/133

WeiqunZhang:
Support both Fab and MultiFab versions of pre- and post-interpolation hooks.

Close #2989.

### DIFF
--- a/Src/AmrCore/AMReX_FillPatchUtil_I.H
+++ b/Src/AmrCore/AMReX_FillPatchUtil_I.H
@@ -4,6 +4,31 @@
 
 namespace amrex {
 
+namespace detail {
+
+template <typename F, typename MF>
+auto call_interp_hook (F const& f, MF& mf, int icomp, int ncomp)
+    -> decltype(f(mf[0],Box(),icomp,ncomp))
+{
+#ifdef AMREX_USE_OMP
+#pragma omp parallel if (Gpu::notInLaunchRegion())
+#endif
+    for (MFIter mfi(mf); mfi.isValid(); ++mfi) {
+        auto& dfab = mf[mfi];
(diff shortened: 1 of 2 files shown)


WeiqunZhang:
@BenWibking  Please let us know if this works for you.

BenWibking:
Yes, this works.

BenWibking:
I've tested this out, and it seems to generate bogus values when the PreInterpHook and PostInterpHook modify the data. It worked with the same interp hooks before, so I'm not sure what is going on.

BenWibking:
I should add that when I make the interp hooks empty functions, then it works fine. Performance on an AMR simulation improves at scale by ~7%.

jmsexton03 (approved):

WeiqunZhang:
@BenWibking Could you test the latest version?

BenWibking:
> @BenWibking Could you test the latest version?

It works now :)

PR #2992, opened 2022-10-17
C++17 Transition

ax3l:
## Summary

Update AMReX to require C++17 or newer.

- [x] docs
- [x] CMake
- [x] GNUmake
- [x] CI

## Additional background

Requires a mature [C++17](https://en.wikipedia.org/wiki/C%2B%2B17) compiler, e.g., GCC 8, Clang 7, NVCC 11.0, MSVC 19.15 or newer.

Already used since 1+ year in production by downstream codes such as Castro and WarpX. Needed for modernization and new features such as #2878

## Checklist

The proposed changes:


WeiqunZhang:
Looks like the CI has some issues with GCC-8.4 on Ubuntu 20.04 because that's not the system default.  Maybe we can change those two tests to use ubuntu-22.04 that has GCC-11.2.

WeiqunZhang:
The GCC issue might be setting CXX compiler to gcc-8 instead of g++-8.

WeiqunZhang:
Anyway, converted one of the tests to use Ubuntu-22.04 and GCC-11.2.

ax3l:
@WeiqunZhang for this to become green, you will need to update the branch protection to require only the new CI entries please :)

WeiqunZhang (approved):

PR #2997, opened 2022-10-25
`FabArray::isDefined`

ax3l:
## Summary

Add a new query to `define_function_called`.

## Additional background

This is a cheaper check than `ok()` for finding out if a MultiFab has been allocated or not yet, assuming that the calling code follows the convention that `define()` is called collectively.

Update: It turns out you can also call `empty` inherited from `FabArrayBase`. The new API is quite explicit, which is ok, too.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_FabArray.H
+++ b/Src/Base/AMReX_FabArray.H
@@ -440,7 +440,10 @@ public:
 
     /** Has define() been called on this rank?
      *
-     * \return true if define has been called on this MultiFab to allocate it
+     * \return true if `define` has been called on this `FabArray`.  Note that all constructors except `FabArray ()`
+     * and `FabArray(Arena*a)` call `define`, even if the `MFInfo` argument has `alloc=false`.  One could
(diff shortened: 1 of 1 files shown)


Src/Base/AMReX_FabArray.H
```diff
@@ -438,6 +438,12 @@ public:
     */
     bool ok () const;
 
+    /** Has define() been called on this rank?
+     *
+     * \return true if define has been called on this MultiFab to allocate it
```
WeiqunZhang:
```suggestion
     * \return true if `define` has been called on this `FabArray`.  Note that all constructors except `FabArray ()`
     * and `FabArray(Arena*a)` call `define`, even if the `MFInfo` argument has `alloc=false`.  One could
     * also use `FabArrayBase::empty()` to find whether `define` is called or not, although they are not exactly
     * the same.
```

WeiqunZhang (approved):

PR #2998, opened 2022-10-26
Make The_Device_Arena non-managed

WeiqunZhang:
The_Device_Arena used to be a separate Arena.  We changed it to be an alias of The_Arena to avoid memory fragmentation.  However, the issue is we don't have an Arena that can allocate non-managed memory unless The_Arena is not managed.  Because of performance concerns, we sometimes want to allocate non-managed memory.  Therefore, we make The_Device_Arena an alias if and only if The_Arena is not managed.


ax3l:
cc @AlexanderSinn

ax3l (approved):
LGTM :+1:

PR #3001, opened 2022-10-27
Fix a bug in ReduceOps when there is no work

### LINKED ISSUE #3000: ParReduce return value is uninitialized when there are no boxes
I am refactoring my code to use ParReduce to compute the CFL timestep like this:

```
amrex::Real max_signal =
 amrex::ParReduce(amrex::TypeList<amrex::ReduceOpMin>{}, amrex::TypeList<amrex::Real>{},
                          cons_mf, amrex::IntVect(0), // no ghost cells
      [=] AMREX_GPU_DEVICE (int bx, int i, int j, int k)
          noexcept -> amrex::GpuTuple<amrex::Real>
      {
        const auto rho = cons[bx](i, j, k, HydroSystem<problem_t>::density_index);
        const auto px  = cons[bx](i, j, k, HydroSystem<problem_t>::x1Momentum_index);
        const auto py  = cons[bx](i, j, k, HydroSystem<problem_t>::x2Momentum_index);
        const auto pz  = cons[bx](i, j, k, HydroSystem<problem_t>::x3Momentum_index);
        const auto kinetic_energy = (px*px + py*py + pz*pz) / (2.0*rho);
        const double abs_vel = std::sqrt(2.0 * kinetic_energy / rho);
        double cs = NAN;

        if constexpr (is_eos_isothermal()) {
          cs = cs_iso_;
        } else {
          const auto Etot = cons[bx](i, j, k, HydroSystem<problem_t>::energy_index);
          const auto Eint = Etot - kinetic_energy;
          const auto P = Eint * (HydroSystem<problem_t>::gamma_ - 1.0);
          cs = std::sqrt(HydroSystem<problem_t>::gamma_ * P / rho);
        }
        return { cs + abs_vel };
      });
amrex::ParallelDescriptor::ReduceRealMax(max_signal);
```

This works if there are boxes on every level on every rank, but if some levels have ranks without boxes, the result is wrong (it usually returns the uninitialized values returned by ParReduce on those ranks).

WeiqunZhang:
This fixes a bug in the GPU version of ReduceOps.  If there is no work for GPU, the tuple on the device is uninitialized. It is then copied to the host and returned.  This could happen in ParReduce when a MultiFab has no boxes on a process, as reported in #3000 by @BenWibking.  The bug was introduced in April 2021.

Close #3000.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_Reduce.H
+++ b/Src/Base/AMReX_Reduce.H
@@ -274,6 +274,8 @@ class ReduceData
 public:
     using Type = GpuTuple<Ts...>;
 
+    template <typename... Ps> friend class ReduceOps;
+
     template <typename... Ps>
     explicit ReduceData (ReduceOps<Ps...>& reduce_op)
         : m_max_blocks(Gpu::Device::maxBlocksPerLaunch()),
@@ -330,6 +332,7 @@ private:
     Type* m_device_tuple = nullptr;
     GpuArray<int,AMREX_GPU_MAX_STREAMS+1> m_nblocks;
     std::function<Type()> m_fn_value;
+    bool m_device_tuple_set = false;
 };
 
 namespace Reduce { namespace detail {
@@ -450,6 +453,7 @@ public:
                 Reduce::detail::for_each_parallel<0, ReduceTuple, Ps...>(dst, r);
 #endif
             });
+            reduce_data.m_device_tuple_set = true;
         }
     }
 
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
Looks this is not the bug.  The good news is this is not an old bug that havs existed for more than a year.  The bug is only in the version of ReduceOps used by ParReduce.  The bad news is I have not figured out why.

WeiqunZhang:
I will probably reopen this even if there are no bugs, because this is slightly more efficient in case there are no works to do.

PR #3002, opened 2022-10-28
Fix EB tensor solver's boundary when EB is tilted.

WeiqunZhang:
It was incorrect to use the connectivity of valid cells to determine the stencil involving domain boundary registers.

## Checklist

The proposed changes:


jordanm304:
This seems to fix the issue we were seeing in MFIX.

asalmgren (approved):

PR #3005, opened 2022-10-31
Fix compilation for PETSc

WeiqunZhang:
We cannot include PETSc headers too early because it might redefine MPI routines as macros
(https://github.com/petsc/petsc/blob/main/include/petsclog.h#L441).  They break MPI calls like below,

    MPI_Allreduce(&tmp, &vi, 1,
                  ParallelDescriptor::Mpi_typemap<T>::type(),
                  ParallelDescriptor::Mpi_op<T,amrex::Greater<T>>(), comm);

because of the `,` in `<T,amrex::Greater<T>>`.

### DIFF
--- a/Src/Extern/PETSc/AMReX_PETSc.cpp
+++ b/Src/Extern/PETSc/AMReX_PETSc.cpp
@@ -1,13 +1,13 @@
 
-#include <petscksp.h>
-#include <AMReX_PETSc.H>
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
Let's get this merged today.

jmsexton03 (approved):

PR #3008, opened 2022-11-02
Add math functions to amrex::Math

WeiqunZhang:
Add a function to return PI and the following trig functions,

  - `cospi(x)` returns `cos(pi*x)`
  - `sinpi(x)` returns `sin(pi*x)`
  - `sincos(x)` returns `sin(x)` and `cos(x)` in `std::pair`
  - `sincospi(x)` returns `sin(pi*x)` and `cos(pi*x)` in `std::pair`

Because we use C++17 now, one could use structured binding on the `std::pair`
returned by `sincos` and `sincospi`.

    auto const [sin_x, cos_x] = amrex::Math::sincos(x);

### DIFF
--- a/Src/Base/AMReX_Math.H
+++ b/Src/Base/AMReX_Math.H
@@ -6,6 +6,8 @@
 #include <AMReX_Extension.H>
 #include <cmath>
 #include <cstdlib>
+#include <type_traits>
+#include <utility>
 
 #ifdef AMREX_USE_DPCPP
 #include <CL/sycl.hpp>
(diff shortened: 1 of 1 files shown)


Src/Base/AMReX_Math.H
```diff
@@ -57,6 +57,32 @@ using std::isinf;
 
 #endif
 
+AMREX_GPU_HOST_DEVICE inline
+void sincos (double x, double* sinx, double* cosx)
```
ax3l: Suggestion: let's make this a more C++-ish interface and return a pair or tuple?

Otherwise, we should restrict the pointers.

WeiqunZhang: Sure, we can return a pair.

WeiqunZhang:
For completeness, let's add sinpi, cospi and sincospi.

ax3l (approved):

PR #3020, opened 2022-11-05
min/max- and linear combination-preserving interpolater

BenWibking:
## Summary
This linearly interpolates onto fine grids, ensuring no new minima and maxima are created in multi-dimensions and also preserving linear combinations of components in each cell.

The current implementation is only for Cartesian geometry.

## Additional background
`CellConservativeLinear` can do each of these things separately (i.e., `CellConservativeLinear(0)` or `CellConservativeLinear(1)`), but does not preserve both of these properties at the same time.

## Checklist

The proposed changes:

### DIFF
--- a/Src/AmrCore/AMReX_MFInterp_1D_C.H
+++ b/Src/AmrCore/AMReX_MFInterp_1D_C.H
@@ -31,6 +31,8 @@ void mf_cell_cons_lin_interp_limit_minmax_llslope (int i, int, int, Array4<Real>
     for (int ns = 0; ns < ncomp; ++ns) {
         slope(i,0,0,ns) *= sfx;
     }
+
+    amrex::ignore_unused(ratio);
 }
 
 AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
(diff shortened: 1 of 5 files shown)


zingale:
it seems like the linear combination constrain applies to the entire state.  Is there a way to confine it to just a range (e.g., the mass fractions or partial density portion of a larger state)?

BenWibking:
> it seems like the linear combination constrain applies to the entire state. Is there a way to confine it to just a range (e.g., the mass fractions or partial density portion of a larger state)?

Maybe this could be handled with adding component range arguments to the constructor?

drummerdoc:
> > it seems like the linear combination constrain applies to the entire state. Is there a way to confine it to just a range (e.g., the mass fractions or partial density portion of a larger state)?
> 
> Maybe this could be handled with adding component range arguments to the constructor?

Only if the ones you want are contiguous, otherwise you'd maybe have to provide a vector of integers?  Otherwise, we'd maybe call an EOS function to overwrite the states that shouldn't be linear.

zingale:
oh nevermind.  I was thinking about this wrong.  This works for me.

BenWibking:
> oh nevermind. I was thinking about this wrong. This works for me.

Would you be able to test it on a problem in Castro? I've run it on my code, but it would be good to be sure. I don't think it's possible to cover all cases exhaustively with unit tests.

WeiqunZhang (approved):

BenWibking:
@WeiqunZhang I see you've already merged this, but it might be a good idea to add a runtime check to abort if this is used in spherical or RZ geometry, since those aren't implemented here. Let me know what you think.

PR #3024, opened 2022-11-06
Update for oneAPI 2023

WeiqunZhang:
A number of things have been deprecated and need to be updated.

  - dpcpp -> icpx

  - sycl::gpu_selector -> sycl::gpu_selector_v

  - sycl::info::device::max_work_item_sizes -> sycl::info::device::max_work_item_sizes<3>

  - sycl::accessor<...,sycl::access::target::local> -> sycl_local_accessor

  - CL/sycl.hpp -> sycl/sycl.hpp

Also remove AMREX_GPU_EXTERNAL because SYCL_EXTERNAL never worked for us.

### DIFF
--- a/.github/workflows/intel.yml
+++ b/.github/workflows/intel.yml
@@ -29,7 +29,7 @@ jobs:
             -DAMReX_PARTICLES=ON                           \
             -DAMReX_GPU_BACKEND=SYCL                       \
             -DCMAKE_C_COMPILER=$(which clang)              \
-            -DCMAKE_CXX_COMPILER=$(which dpcpp)            \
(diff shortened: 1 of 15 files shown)


WeiqunZhang:
@nmnobre I think this should work for you.

WeiqunZhang:
If there are any other issues, let me know.

nmnobre:
Thanks a lot.
Just to be clear, I'm happy with waiting until they officially release 2023.x so we can avoid the preprocessor conditional, but I'm also happy with this temporary workaround if you are.

atmyers (approved):

PR #3026, opened 2022-11-07
Fix mis-matched array type that broke mixed-precision restarts

PhilMiller:
## Summary

A mixed-precision build of WarpX (DP fields, SP particles) would fail to restart from a checkpoint with added real particle components, crashing while reading in data.

## Additional background

Isolated with this modification of a WarpX example, and this fix tested similarly with `amr.restart = blah`

```
$ diff -u ~/repos/WarpX/Examples/Physics_applications/capacitive_discharge/inputs_2d inputs_2d 
--- /home/ubuntu/repos/WarpX/Examples/Physics_applications/capacitive_discharge/inputs_2d	2022-09-07 22:02:51.200174154 +0000
+++ inputs_2d	2022-11-07 21:52:04.764307593 +0000
@@ -39,6 +39,8 @@
 electrons.momentum_distribution_type = maxwell_boltzmann
 electrons.theta = (kb*30000/(m_e*clight^2))
 
+electrons.save_previous_position = 1
+
 he_ions.species_type = helium
 he_ions.charge = q_e
 he_ions.injection_style = nuniformpercell
@@ -55,24 +57,28 @@
 
-diagnostics.diags_names = diag1
+diagnostics.diags_names = diag1 ckpt
 diag1.diag_type = Full
 diag1.intervals = 50
 diag1.intervals = 10
 diag1.fields_to_plot = rho_electrons rho_he_ions
+
+ckpt.diag_type = Full
+ckpt.intervals = 7
+ckpt.format = checkpoint
```

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_GpuContainers.H
+++ b/Src/Base/AMReX_GpuContainers.H
@@ -137,8 +137,12 @@ namespace Gpu {
     void copy (HostToDevice, InIter begin, InIter end, OutIter result) noexcept
     {
         using value_type = typename std::iterator_traits<InIter>::value_type;
+
+        using out_value_type = typename std::iterator_traits<OutIter>::value_type;
+        static_assert(std::is_same_v<value_type, out_value_type>);
         static_assert(std::is_trivially_copyable<value_type>(),
                       "Can only copy trivially copyable types");
+
         auto size = std::distance(begin, end);
         if (size == 0) return;
 #ifdef AMREX_USE_GPU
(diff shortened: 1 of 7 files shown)


PhilMiller:
It looks like `copyAsync` will accept mixed-type pointer arguments when it really oughtn't, since it's not going to do any type conversion. That would have caught the underlying error at compile time.

WeiqunZhang:
We could add `static_assert` to the copy functions.

PhilMiller:
The `static_asserts` mentioned seem to build for me locally. I'll push that once the remaining CI stuff passes

PhilMiller:
The fact that I had to make so many changes for this suggests some substantial coverage gaps in the testing. At least the `static_assert`s will catch people before they stumble into a case that would fall in those gaps.

PhilMiller:
I bet WarpX and other downstream codes are going to stumble on importing this. I'll see if I can budget the time to deal with that

PhilMiller:
I also checked that a local build with SP fields and DP particles only fails for unrelated reasons.

PhilMiller:
Just double-checked that WarpX builds in mixed-precision with the added `static_assert`s

atmyers:
Thank you for the fixes! It looks we don't have a test in AMReX itself that just checkpoints / restarts particles in the native format (although we do for the HDF5 format... ), which would have caught the issue here if run with SP particles / DP fields. I will add one.

atmyers (approved):

atmyers:
New test here: https://github.com/AMReX-Codes/amrex/pull/3027

PR #3027, opened 2022-11-08
add a new test for particle checkpoint / restart in the native format

atmyers:
This fills a gap in the test suite indicated in #3026.

The proposed changes:


atmyers:
I verified locally that the added test *fails* with SP particles DP fields, and that it is fixed if I try with the changes in #3026.

atmyers:
Note, the remaining CI test should pass once #3026 is merged.

WeiqunZhang (approved):

PR #3035, opened 2022-11-17
MLMG: template

WeiqunZhang:
MLMG and various linear operator classes use MultiFab as their data
container.  The limitations are (1) it cannot do single precision when
amrex::Real is double and (2) it cannot handle data that need to be stored
in multiple MultiFabs such as vectors on cell edges.  Recently, we have
experimented with amrex::Any that can be a type different from MultiFab.
However, this involves a lot of complicated and hacky changes.  In this PR,
we are trying to use template to make the interface more flexible.  Classes
MLMG, MLPoisson, MLALaplacian, MLABecLaplacian, MLABecLap,
MLCellLinop, MLLinOp, and MLCGSolver have been changed to templates.
More changes will follow.
    
For most users, this should not require any changes in their codes.
However, if one has classes derived from the linear operators in MLMG, some
small changes are likely needed.  For example, `getNComp()` needs to be
`this->getNComp()` due to C++ name lookup rules.


WeiqunZhang:
This PR depends on the following PRs.

- [x] #2980 
- [x] #3028 
- [x] #3030 
- [x] #3039 
- [x] #3040 
- [x] #3041 
- [x] #3042 
- [x] #3043
- [x] #3044

atmyers (approved):

PR #3054, opened 2022-12-01
Print host name in backtrace files

WeiqunZhang:
This can help identify which nodes generated the backtrace files.  Note that the MPI rank is embedded in the file name, but it does not tell us which node it is.

### DIFF
--- a/Src/Base/AMReX_BLBackTrace.cpp
+++ b/Src/Base/AMReX_BLBackTrace.cpp
@@ -187,6 +187,18 @@ BLBackTrace::print_backtrace_info (FILE* f)
 
 #ifdef __linux__
 
+    {
+        constexpr std::size_t len = 64;
+        char host_name[len];
+        host_name[len-1] = '\n';
+        // The returned buffer may not include '\n', when truncation occurs.
+        // So we insert one just in case.
+        const int ret = gethostname(host_name, len-1);
+        if (ret == 0) {
+            fprintf(f, "Host Name: %s\n", host_name);
+        }
+    }
+
     char **strings = backtrace_symbols(bt_buffer, nentries);
     if (strings != NULL) {
         int have_eu_addr2line = 0;


WeiqunZhang:
@ax3l

jmsexton03:
This is using a different path than AMReX_Machine is, should it be the same for both of them?  https://github.com/AMReX-Codes/amrex/blob/83ec5fc/Src/Base/AMReX_Machine.cpp#L346-L371

The git blame looks like it was partially written by cychan, and is more than a year old:  
https://github.com/AMReX-Codes/amrex/blame/83ec5fcfefae13264827a0724ab61d2d1b9a7eac/Src/Base/AMReX_Machine.cpp#L346-L371

WeiqunZhang:
The `HOSTNAME` environment variable is not in the POSIX standard, whereas the `gethostname` function is.  So I think it's better to use `gethostname`.  The code in AMReX_Machine also checks other environment variables like `NERSC_HOST`.  So it's not very generic.

jmsexton03:
Makes sense, it looks like the `get_machine_envs` doesn't get called anywhere specific, so it's likely not important to update that functionality to be consistent with the standard.

Is `MPI_Get_processor_name` in the MPI standard generally? would that give us any different / additionally useful information for a Backtrace?

WeiqunZhang:
https://www.mpich.org/static/docs/v3.2/www3/MPI_Get_processor_name.html
> The name returned should identify a particular piece of hardware; the exact format is implementation defined. This name may or may not be the same as might be returned by gethostname, uname, or sysinfo.

Given that it depends on the implementation, I am not sure what additional information it can provide.

jmsexton03 (approved):
Looks good, and amrex/Tools/Backtrace/parse_bt.py still gives the same output. I don't get any more info from using the mpi-based call on perlmutter. Sticking with the standard sounds great in general.

PR #3013, opened 2022-11-04
CMake: AMReX_BUILD_SHARED_LIBS

ax3l:
## Summary

Add a new variable `AMReX_BUILD_SHARED_LIBS` for per-project control of shared lib builds.

This variable is more specific (AMReX project) than `BUILD_SHARED_LIBS` (whole build, defaults) and `USE_XSDK_DEFAULTS` (whole build, changed defaults).

We need this variable so we can control the AMReX library build in superbuilds without changing all other libraries we build in other projects of the superbuild.

## Additional background

- https://github.com/ECP-WarpX/impactx/pull/123
- https://github.com/ECP-WarpX/impactx/pull/208

## Checklist

The proposed changes:
- [ ] fix a bug or incorrect behavior in AMReX
- [x] add new capabilities to AMReX
- [ ] changes answers in the test suite to more than roundoff level
- [ ] are likely to significantly affect the results of downstream AMReX users
- [ ] include documentation in the code and/or rst files, if appropriate

PR #3057, opened 2022-12-08
Fix: Shared CMake builds w/ `AMReX_BUILD_SHARED_LIBS`

ax3l:
## Summary

Forgot to honor the new option. All builds that did not set `BUILD_SHARED_LIBS` were static now.

## Additional background

Introduced in #3013

## Checklist

The proposed changes:

### DIFF
--- a/Src/CMakeLists.txt
+++ b/Src/CMakeLists.txt
@@ -5,7 +5,11 @@
 # the properties of this object, like setting the sources,
 # setting the compile definitions and so on
 #
-add_library( amrex )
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
Does it mean it's either static or shared, but it cannot be both?

ax3l:
Yes, currently/historically it is either/or.

We can add a 2nd target to build static+shared at the same time, if we want to. (Would be an independent PR and a breaking change.)

WeiqunZhang (approved):

PR #3058, opened 2022-12-08
Gpu Streams: Performance Tuning

WeiqunZhang:
* Change the number of GPU streams from 4 to 2. Recent tests on summit and crusher show that using 2 GPU streams seems to be better than 4 or 1 in most cases.  One reason is that it reduces the cost associated with stream synchronization, which could be significant for small kernels.

 * In ~MFIter and ~StreamIter, there is no need to synchronize the streams that were not used.  It appears that for HIP there is very little cost to sychronize a stream that does not have any GPU kerenels since the last synchronization, but that's not the case for CUDA.

 * Remove the dedicated stream for kernels outside MFIter, and simply use the first stream used by MFIter.

 * Remove synchronization in FabArray::define.  It was needed when we were putting FAB in managed memory, but that's no longer the case.

### DIFF
--- a/Src/Base/AMReX_MFIter.cpp
+++ b/Src/Base/AMReX_MFIter.cpp
@@ -153,7 +153,7 @@ MFIter::MFIter (const BoxArray& ba, const DistributionMapping& dm, const MFItInf
     fabArray(*m_fa),
     tile_size(info.tilesize),
     flags(info.do_tiling ? Tiling : 0),
-    streams(info.num_streams),
+    streams(std::max(1,std::min(Gpu::numGpuStreams(),info.num_streams))),
     dynamic(info.dynamic && (OpenMP::get_num_threads() > 1)),
     device_sync(info.device_sync),
     index_map(nullptr),
@@ -185,7 +185,7 @@ MFIter::MFIter (const FabArrayBase& fabarray_, const MFItInfo& info)
     fabArray(fabarray_),
     tile_size(info.tilesize),
     flags(info.do_tiling ? Tiling : 0),
-    streams(info.num_streams),
+    streams(std::max(1,std::min(Gpu::numGpuStreams(),info.num_streams))),
     dynamic(info.dynamic && (OpenMP::get_num_threads() > 1)),
     device_sync(info.device_sync),
     index_map(nullptr),
@@ -235,10 +235,14 @@ MFIter::Finalize ()
 #endif
 
 #ifdef AMREX_USE_GPU
-    if (device_sync) Gpu::streamSynchronizeAll();
-#endif
+    if (device_sync) {
+        const int nstreams = std::min(endIndex, streams);
+        for (int i = 0; i < nstreams; ++i) {
+            Gpu::Device::setStreamIndex(i);
+            Gpu::streamSynchronize();
+        }
+    }
 
-#ifdef AMREX_USE_GPU
     AMREX_GPU_ERROR_CHECK();
     Gpu::Device::resetStreamIndex();
 #endif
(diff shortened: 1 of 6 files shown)


Src/Base/AMReX_MFIter.cpp
```diff
@@ -267,15 +271,6 @@ MFIter::Initialize ()
             "Nested or multiple active MFIters is not supported by default.  This can be changed by calling MFIter::allowMultipleMFIters(true)".);
     }
 
-#ifdef AMREX_USE_GPU
-    if (device_sync) {
-#ifdef AMREX_USE_OMP
-#pragma omp single
-#endif
-        Gpu::streamSynchronize();
```
AlexanderSinn: I think this is still necessary for the case that MFIter is going to use a different stream than the stream used outside of MFIter.
Example:

1. expensive async memcpy on stream 0
2. important data async memcpy on stream 0
3. MFIter loop starts with two streams
3.1. kernel using the important data on stream 0
3.2. another kernel using the important data on stream 1
4. MFIter loop ends

The issue is that 3.2 can start before 1 is finished because there was no synchronize, however it will produce wrong results because 2 is not finished yet. Notice that if 3.2 where to use stream 0 then no synchronize in both the constructor or destructor of MFIter would be required. This is what is done in HiPACE++ which is greatly beneficial for low resolutions.

WeiqunZhang: In principle, yes, that could happen.  In practice, I have never seen that much overlap between streams that could cause issues.

We used to use the special null stream outside the kernel and we did not have the sync in the ctor of MFIter.  However, SYCL/DPC++ does not have the equivalent of the null stream.  So the behavior was not consistent before we switched to avoiding the null stream.  It was because of the concern you raised that we added a sync in the ctor MFIter.  After thinking about this for a long time, I think it's better to not do the sync in the ctor.

WeiqunZhang: For MAC projection in an amr-wind test with max_grid_size=32, this PR improves the time from 14.26 to 11.99.  The overall time went down from 56.04 to 48.92.

WeiqunZhang: Anyway, if there is data movement before MFIter, it's now up to the user to make sure that does not cause an issue.  The user could insert an explicit sync before MFIter starts or at the end of iteration 0.

WeiqunZhang: @AlexanderSinn I tested the scenario you have described. It did happen when the memcopy in (1) was very expensive.  So now the user had better make sure the memcpy finishes before MFIter starts.

AlexanderSinn: Good to know

WeiqunZhang: Let me test if I put it back, how much performance penalty there is for the amr-wind test.  Maybe all the other changes made in this PR is all that matters.

WeiqunZhang: If (1) and (2) are swapped, then I don't see any issues.  This makes sense.

AlexanderSinn: It may also be worth trying to have only one stream and no synchronize in the dtor or ctor. This should work in most cases except when code relies on the synchronize in the ctor for e.g. dtoh memcpy.

WeiqunZhang: In many cases, use multiple streams is faster than a single stream.

I am thinking of adding something that can be used to disable syncs and specify the number of streams in a given scope.

WeiqunZhang: The good news is that I do not see much difference if the sync in MFIter ctor is on in two tests with a small max grid size on summit and crusher.  So the improvement in this PR mostly comes from other changes.  I will put the sync back for safety.

WeiqunZhang: @AlexanderSinn Thanks for your constructive feedback!

atmyers (approved):

PR #3061, opened 2022-12-09
CI: Doxygen

ax3l:
## Summary

Run Doxygen and make sure it does not ~~throw warnings~~ *error while parsing* in CI. This is already run as such in downstream projects, so we need to fix our upstream errors as well.

## Checklist

The proposed changes:

### DIFF
--- a/.github/workflows/style/doxygen
+++ b/.github/workflows/style/doxygen
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+#
+# Copyright 2020 Axel Huebl
+#
+# License: BSD-3-Clause-LBNL
+
+# search recursive inside a folder if a file contains tabs
+#
+# @result 0 if no files are found, else 1
+#
+
+set -eu -o pipefail
+
+cd Docs/Doxygen
+
+# treat all warnings as errors
+echo "WARN_AS_ERROR = YES" >> doxygen.conf
+
+doxygen doxygen.conf
--- a/Docs/Doxygen/doxygen.conf
+++ b/Docs/Doxygen/doxygen.conf
@@ -2037,15 +2037,28 @@ SEARCH_INCLUDES        = YES
 # preprocessor.
 # This tag requires that the tag SEARCH_INCLUDES is set to YES.
 
-INCLUDE_PATH           = ../../Src/Amr      \
-                         ../../Src/AmrCore  \
-                         ../../Src/Base     \
-                         ../../Src/Boundary \
-                         ../../Src/EB       \
-                         ../../Src/Extern   \
-                         ../../Src/LinearSolvers \
+INCLUDE_PATH           = ../../Src/Base \
+                         ../../Src/Base/Parser \
+                         ../../Src/SDC \
+                         ../../Src/Amr \
+                         ../../Src/EB \
                          ../../Src/Particle \
-                         ../../Src/SDC
+                         ../../Src/Boundary \
+                         ../../Src/Extern \
+                         ../../Src/Extern/HYPRE \
+                         ../../Src/Extern/hpgmg \
+                         ../../Src/Extern/HDF5 \
+                         ../../Src/Extern/PETSc \
+                         ../../Src/Extern/amrdata \
+                         ../../Src/Extern/Conduit \
+                         ../../Src/Extern/SUNDIALS \
+                         ../../Src/Extern/ProfParser \
+                         ../../Src/Extern/SENSEI \
+                         ../../Src/Extern/SWFFT \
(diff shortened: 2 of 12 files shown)


.github/workflows/style/doxygen
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+#
+# Copyright 2020 Axel Huebl
+#
+# License: BSD-3-Clause-LBNL
+
+# search recursive inside a folder if a file contains tabs
```
WeiqunZhang: The comments need to be updated.

ax3l:
A few more to fix (help wanted :) ), currently stuck at
```
/home/axel/src/amrex/Src/Base/AMReX_GpuReduce.H:510: error: no matching file member found for 
AMREX_GPU_DEVICE AMREX_FORCE_INLINE int amrex::Gpu::blockReduceLogicalAnd(int source)
Possible candidates:
 'AMREX_GPU_DEVICE AMREX_FORCE_INLINE int blockReduceLogicalAnd(int source) noexcept' at line 494 of file /home/axel/src/amrex/Src/Base/AMReX_GpuReduce.H
 (warning treated as error, aborting now)
```

WeiqunZhang:
Maybe it's confused by that a function template and a non-template function have the same function signature.

Src/Base/AMReX_Random.cpp
```diff
@@ -27,7 +27,7 @@ namespace amrex {
 
 #ifdef AMREX_USE_GPU
 namespace {
-void ResizeRandomSeed (amrex::ULong gpu_seed)
+void ResizeRandomSeed (ULong gpu_seed)
```
ax3l:
```suggestion
void ResizeRandomSeed (amrex::ULong gpu_seed)
```

Docs/Doxygen/doxygen.conf
```diff
@@ -2037,7 +2037,15 @@ SEARCH_INCLUDES        = YES
 # preprocessor.
 # This tag requires that the tag SEARCH_INCLUDES is set to YES.
 
-INCLUDE_PATH           =
+INCLUDE_PATH           = ../../Src/Amr      \
```
WeiqunZhang: Could you tell me why need to do this now?  Should it also include all the directories recursively?

ax3l: Somehow it did not find some includes (already) - we just ignored it as warnings before it seems.

ax3l:
@WeiqunZhang ready to merge - the potential overflow in Random seems to be pre-existing (do we want to fix the cast separately?)

WeiqunZhang:
I think the overflow warning is false positive.  We are doing something like

WeiqunZhang:
Accidentally clicked the wrong button.

Docs/Doxygen/doxygen.conf
WeiqunZhang: > Note that the INCLUDE_PATH is not recursive, so the setting of [RECURSIVE](https://www.doxygen.nl/manual/config.html#cfg_recursive) has no effect here.

My point is we probably need to include ../../Src/Base/Parser etc. as well, if this is needed.

```
../../Src/Base \
../../Src/Base/Parser \
../../Src/SDC \
../../Src/Amr \
../../Src/EB \
../../Src/Particle \
../../Src/Boundary \
../../Src/Extern \
../../Src/Extern/HYPRE \
../../Src/Extern/hpgmg \ \
../../Src/Extern/HDF5 \
../../Src/Extern/PETSc \
../../Src/Extern/amrdata \
../../Src/Extern/Conduit \
../../Src/Extern/SUNDIALS
../../Src/Extern/ProfParser \
../../Src/Extern/SENSEI \
../../Src/Extern/SWFFT \
../../Src/AmrCore \
../../Src/LinearSolvers \
../../Src/LinearSolvers/MLMG \
../../Src/LinearSolvers/OpenBC \
../../Src/F_Interfaces \
../../Src/F_Interfaces/Octree \
../../Src/F_Interfaces/Base \
../../Src/F_Interfaces/Particle \
../../Src/F_Interfaces/AmrCore \
../../Src/F_Interfaces/LinearSolvers
```

ax3l: I didn't add it yet because I saw no futher warnings, but why not! :)

ax3l: Thanks for the list, updated accordingly :)

We excluded F_Interfaces/ so far, so I kept it that way for now.

WeiqunZhang (approved):

PR #3075, opened 2022-12-20
Docs: Sync Before Comms

kngott:
## Summary

Add a new section to the profiling section and document the sync_before_comm flag.

This section can be used for any information, flags or options that applies to both Tiny and Full profiling.

## Checklist

The proposed changes:


WeiqunZhang:
Should we add a warning that this will slow down the code.

kngott:
Good point. How's that?

WeiqunZhang (approved):

PR #3076, opened 2022-12-21
amrex::fillAsync

WeiqunZhang:
This new function can be used to fill the elements in a vector type container (e.g., Gpu::DeviceVector).  If the element type is a struct with several arithmetic types (e.g., GpuArray<Real,10>), the usual ParallelFor does not have good performance because of the memory access pattern.  The new fillAsync function will use shared memory in that case to improve the performance.

### DIFF
--- a/Src/Base/AMReX_GpuContainers.H
+++ b/Src/Base/AMReX_GpuContainers.H
@@ -402,7 +402,105 @@ namespace Gpu {
         Gpu::streamSynchronize();
     }
 
-}}
+    /**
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
I ran this test ( https://github.com/WeiqunZhang/amrex-devtests/blob/main/vector-fill/main.cpp) on crusher.  `fillAsync` is 2.5x faster than `ParallelFor`.

WeiqunZhang:
The test also checked for correctness.

ax3l (approved):

PR #3079, opened 2022-12-27
New knapsack function to reduce the cost of data movement in load balancing

WeiqunZhang:
This adds a new function DistributionMapping::KnapSackProcessorMap that makes a new DistributionMapping given an old DistributionMapping, a vector of weights and a parameter affecting the amount of the data movement that would be needed for switching from the old to the new DistributionMapping.

In this new knapsack function, some boxes will be kept on the current processes and the knapsack algorithm is applied to the rest of the boxes. This may generate a less optimal distribution.  However, the new DistributionMapping is more similar to the old one, thus reducing the cost of data movement during load balancing.

### DIFF
--- a/Src/Base/AMReX_DistributionMapping.H
+++ b/Src/Base/AMReX_DistributionMapping.H
@@ -132,17 +132,21 @@ class DistributionMapping
     //! Are the distributions different?
     bool operator!= (const DistributionMapping& rhs) const noexcept;
 
-    void SFCProcessorMap(const BoxArray& boxes, const std::vector<Long>& wgts, int nprocs,
-                         bool sort=true);
-    void SFCProcessorMap(const BoxArray& boxes, const std::vector<Long>& wgts, int nprocs,
-                         Real& efficiency, bool sort=true);
-    void KnapSackProcessorMap(const std::vector<Long>& wgts, int nprocs,
-                              Real* efficiency=0,
-                              bool do_full_knapsack=true,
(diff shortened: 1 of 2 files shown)


atmyers (approved):

kngott:
Would it be possible to include "nprocs" in the function signature of this new algorithm, like the others algorithms? It would be necessary to try a joint Knapsack/SFC load balancing algorithm with this implementation.

WeiqunZhang:
If the old DistributionMapping involves more processes than "nprocs", what is the function then supposed to do? So it may add more confusion.

kngott:
As I interpret these functions, I don't think the user is supposed to call any of the lowest level functions. i.e: the user isn't expected to manually send this DistributionMap to all ranks; that's done on a higher level of the function tree. So, I don't think these are generally user-level calls, which should minimize confusion.

And, yes, this function would require some extra consideration to do this type of thing. Maybe, you'd just ignore a certain number of ranks that fall in the middle of the sort? Or maybe based on nprocs, you'd treat every "N" bins as the same bin; just modulus-ing based on the nprocs value? (The numbering the algorithm returns would add some detail to deal with in the standard Knapsack+SFC case too. Probably need a middle-layer to reinterpret the raw DistMap result.)

I'm just asking now because I'm imagining changing the function signature later might be the bigger pain. All algorithm level functions currently have an overriden "nprocs" input, so I'm picturing needing to add it and re-adjust calls might be the bigger pain. But, it's 50/50.

PR #3085, opened 2022-12-31
GPU UUID and Arena initial size

WeiqunZhang:
Gather GPU UUIDs to find out the number of unique devices (which has already been done for CUDA).  Find out the number of processes sharing my device. Use this information, we can avoid oversubscribing GPU device memory in Arean initialization that could result in a runtime error for non-managed memory.

### DIFF
--- a/Src/Base/AMReX.cpp
+++ b/Src/Base/AMReX.cpp
@@ -427,6 +427,26 @@ amrex::Initialize (int& argc, char**& argv, bool build_parm_parse,
         pp.queryAdd("verbose", system::verbose);
     }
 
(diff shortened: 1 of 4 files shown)


WeiqunZhang:
I have tested this on perlmutter, summit, and crusher.  On summit, I also tested having one, three and six GPUs in a resource set.  On crusher, each of the two GPUs in one MI250X has a unique ID.  I also tested on a local workstation with multiple MPI processes sharing a GPU.  Cannot get MPI work on florentia.  So let's disable it for now.

atmyers (approved):

PR #3090, opened 2023-01-08
SYCL: Reinforce device compilation preprocessor conditionals

nmnobre:
Hi @WeiqunZhang,

## Summary

The changes in these two commits, although apparently redundant, guard SYCL compilations targeting CUDA or HIP devices by preventing the compiler from seeing the CUDA/HIP code already in AMReX.

## Additional background

The 'vanilla' version of LLVM, which is used by hipSYCL for example, enables `__CUDA_ARCH__` and `__HIP_DEVICE_COMPILE__` when compiling device code, for both [CUDA](https://llvm.org/docs/CompileCudaWithLLVM.html#detecting-clang-vs-nvcc-from-code) and [HIP](https://reviews.llvm.org/D45441), respectively. Intel's version of LLVM is also moving in the same direction with [CUDA](https://github.com/intel/llvm/issues/7722) and [HIP](https://github.com/intel/llvm/issues/7720), although the latter is off to a rocky start.

I couldn't decide between a more wordy implementation like this or instead defining a macro like you did for `AMREX_DEVICE_COMPILE`. I went with the former but we could just as easily switch to the latter.

Finally, I don't know if you are going to the ECP meeting next week, but it'd be great to say hi, personally thank you for your work and maybe walk you through my poster about how I've been using SYCL with AMReX.

Cheers,
Nuno

### DIFF
--- a/Src/Base/AMReX_Extension.H
+++ b/Src/Base/AMReX_Extension.H
@@ -45,7 +45,7 @@
 #if defined(AMREX_DEBUG)
 #define AMREX_PRAGMA_SIMD
 
-#elif defined(__CUDA_ARCH__)
+#elif defined(__CUDA_ARCH__) && defined(AMREX_USE_CUDA)
 #define AMREX_PRAGMA_SIMD
 
 #elif defined(__HIP_DEVICE_COMPILE__)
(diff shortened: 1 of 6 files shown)


WeiqunZhang (approved):

WeiqunZhang:
@nmnobre I am not attending the ECP meeting next week.  However, I am very interested in your poster.  Could you please send it to me by email (WeiqunZhang@lbl.gov)?  Thanks!

ax3l:
Nice! X-ref #3184

PR #3094, opened 2023-01-12
SinCos: AppleClang Fix

ax3l:
## Summary

I noticed that in some situations, e.g., when building Python sources that include Python headers, `_GNU_SOURCE` is sometimes defined. This can confuse AppleClang builds of AMReX:
```
error: no member named 'sincos' in the global namespace; did you mean '__sincos'?
    ::sincos(x, &r.first, &r.second);
```

Adding the respective guard here should allow GCC builds on macOS but ignore the GCC-ism when building with AppleClang.

Follow-up to  #3008

## Additional background

First seen in pyAMReX the other week.
Avoided with restructured includes, but ultimately a defect that CPython headers pull in.

Seen again today by @RTSandberg 

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_Math.H
+++ b/Src/Base/AMReX_Math.H
@@ -130,7 +130,7 @@ std::pair<double,double> sincos (double x)
     r.first = sycl::sincos(x, &r.second);
 #elif defined(__CUDA_ARCH__) && defined(AMREX_USE_CUDA) || \
       defined(__HIP_DEVICE_COMPILE__) && defined(AMREX_USE_HIP) || \
-      defined(_GNU_SOURCE)
+      (defined(_GNU_SOURCE) && !defined(__apple_build_version__))
     ::sincos(x, &r.first, &r.second);
 #else
     r.first  = std::sin(x);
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
The error message suggests that there is `__sincos`.  If that's always available for AppleClang, maybe we could use it.  We test for AppleClang before the block for `::sincos`.  But maybe it's not worth it.

WeiqunZhang (approved):

Src/Base/AMReX_Math.H
```diff
@@ -130,7 +130,7 @@ std::pair<double,double> sincos (double x)
     r.first = sycl::sincos(x, &r.second);
 #elif defined(__CUDA_ARCH__) && defined(AMREX_USE_CUDA) || \
       defined(__HIP_DEVICE_COMPILE__) && defined(AMREX_USE_HIP) || \
-      defined(_GNU_SOURCE)
+      (defined(_GNU_SOURCE) && !defined(__apple_build_version__))
```
RTSandberg, line 133: Ooops, also with vanilla (conda-forge) clang on macos.

We need to use `__APPLE__` or `__OSX__` here.

RTSandberg:
Axel: __sincos is available for SDKs 10.9+ as I googled yesterday, but seems a bit tricky for some implementations. Have not arrived at a conclusion if I would use it, also given the `__` marker...

ax3l:
> But maybe it's not worth it.

Exactly what my thinking is right now as well.

ax3l:
Sorry, forgot to add note that I was still testing this with @RTSandberg today.
Full fix now in: #3096

PR #3096, opened 2023-01-12
SinCos Guard: Apple Define

ax3l:
## Summary

The issue also occurs on vanilla Clang, not only AppleClang.

Thus, we exclude the specialization for all Apple platforms.

## Additional background

Follow-up to #3094 #3008

## Checklist

The proposed changes:


WeiqunZhang (approved):
Okay. I will approve this and let you merge it.

ax3l:
Confirmed to fix it (thx @RTSandberg!)

PR #3100, opened 2023-01-19
CMake clean up

WeiqunZhang:
* Clean up oneAPI compiler setup.  We no longer support early beta versions.

* Move AMREX_USE_EB macro definition from AMReXSetDefines.cmake to Src/EB/CMakeLists.txt.  The former is not supposed to set defines for optional components (as mentioned in its comments).

* AMReXParallelBackends.cmake: unset temporary variable _genex.

* Update documentation for CMAKE_CXX_STANDARD. C++14 is no longer supported.

* Update oneAPI CI test. Use icx and ifx instead of clang and gfortran for C and Fortran, respectively.


ax3l (approved):
Looks splendid, thanks a lot :tada:

PR #3101, opened 2023-01-19
ParmParse: support nan and inf

WeiqunZhang:
When the ParamParse table is dumped, if a floating point number is a NaN, it will be printed as nan.  If it's infinity, it will be printed as inf. However, if one tried to read it back, the code will crash because ParmParse did not recognize nan and inf as valid floating point numbers.  This commit adds support for nan and inf in ParmParse.

### DIFF
--- a/Src/Base/AMReX_ParmParse.cpp
+++ b/Src/Base/AMReX_ParmParse.cpp
@@ -146,7 +146,7 @@ enum PType
 
 template <class T>
 bool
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
@michaeljbrazell

michaeljbrazell:
thanks @WeiqunZhang!

atmyers (approved):

PR #3106, opened 2023-01-23
FillPatcher: Update interpolation in time

WeiqunZhang:
The previous implementation was susceptible to roundoff errors.  It was observed in 7 levels PeleC runs that the code could fail because the times in StateData were out of sync.  The updated logic is closer to that in the FillPatch function and the StateData class.


esclapez:
I confirm that this PR circumvent the out-of-sync time FillPatcher error we see in PeleC when using a deep AMR hierarchy. Thanks Weiqun !

kngott (approved):

PR #3108, opened 2023-01-23
Replace sprintf with snprintf to fix compiler warnings

WeiqunZhang:


ax3l (approved):
Splendid! Does this need the include `<stdio>` for robustness?
https://en.cppreference.com/w/cpp/io/c/fprintf

WeiqunZhang:
Since these are cpp files, I am not worried about it.  (`cstdio` has been indirectly included by including AMReX.H).

WeiqunZhang:
Anyway, it does not hurt to add cstdio.

ax3l (approved):
Perfect, thx.

PR #3111, opened 2023-01-24
OpenBCSolver: Fix reuse

WeiqunZhang:
When the open BC solver is reused, we must update the level BC.


ajnonaka (approved):
Thanks Weiqun!  This fixes the restart issue I was having.

PR #3107, opened 2023-01-23
CMake: Fix AMD Flags for ROCm 5.5

ax3l:
## First Commit: Shell-Excape Tokens

- [x] The new ROCm 5.5 flag `-Xoffload-linker --whole-archive` is a flag with an argument. CMake will deduplicate the command line if it encounters another `-Xoffload-linker` on it, which will break compilations in the future.

This marks the whole string as a token that can only be deduplicated as a whole.
Ref.: https://cmake.org/cmake/help/v3.25/command/target_link_options.html#option-de-duplication

## Second Commit: Only ROCm 5.5+

- [x] Do not add the flags `-Xoffload-linker --whole-archive` in earlier ROCm releases, avoiding compile-time warnings.

## Third Commit: Clean Up CI

- [x] Remove warning suppression introduced in #3097

## Additional background

Follow-up to @david-salinas's #3097

## Checklist

The proposed changes:
- [ ] fix a bug or incorrect behavior in AMReX
- [ ] add new capabilities to AMReX
- [ ] changes answers in the test suite to more than roundoff level
- [ ] are likely to significantly affect the results of downstream AMReX users
- [ ] include documentation in the code and/or rst files, if appropriate

PR #3112, opened 2023-01-25
CMake 3.18+: Stay OLD CUDA_ARCHITECTURES Policy

ax3l:
## Summary

For now... We removed this guard in WarpX & ImpactX a while ago. Need to revisit how we want to add the logic, probably by jumping to CMake 3.20+ requirements and ditching legacy CUDA logic.

## Additional background

Regression from switching to CMake 3.18+, which then enables new policies.
Introduced in #3107

- https://cmake.org/cmake/help/latest/policy/CMP0104.html
- https://cmake.org/cmake/help/v3.24/prop_tgt/CUDA_ARCHITECTURES.html#prop_tgt:CUDA_ARCHITECTURES

## Checklist

The proposed changes:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,6 +17,16 @@ if(POLICY CMP0127)
 endif()
 
 #
+# CMake 3.18+: CMAKE_CUDA_ARCHITECTURES
+# https://cmake.org/cmake/help/latest/policy/CMP0104.html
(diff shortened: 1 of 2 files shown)


WeiqunZhang (approved):
Works for me now.

atmyers (approved):

PR #3115, opened 2023-01-27
Rename AMReX_CUDA_MAX_THREADS to AMReX_GPU_MAX_THREADS

WeiqunZhang:
Rename cmake option AMReX_CUDA_MAX_THREADS to AMReX_GPU_MAX_THREADS.  This option can now be used for all GPU backends.

GNU Make has also been updated for this.

### DIFF
--- a/Src/Base/AMReX_GpuControl.H
+++ b/Src/Base/AMReX_GpuControl.H
@@ -7,10 +7,6 @@
 
 #include <utility>
 
-#ifndef AMREX_GPU_MAX_THREADS
-#define AMREX_GPU_MAX_THREADS 256
-#endif
-
 #if defined(AMREX_USE_CUDA) && (__CUDACC_VER_MAJOR__ > 11 || ((__CUDACC_VER_MAJOR__ == 11) && (__CUDACC_VER_MINOR__ >= 2)))
 #define AMREX_CUDA_GE_11_2 1
 #endif
--- a/Docs/sphinx_documentation/source/GPU.rst
+++ b/Docs/sphinx_documentation/source/GPU.rst
@@ -189,6 +189,12 @@ can run it and that will generate results like:
 Building with CMake
 -------------------
 
+To build AMReX with GPU support in CMake, add
+``-DAMReX_GPU_BACKEND=CUDA|HIP|SYCL`` to the ``cmake`` invocation, for CUDA,
+HIP and SYCL, respectively. By default, AMReX uses 256 threads per GPU
+block/group in most situations. This can be changed with
+``-DAMReX_GPU_MAX_THREADS=N``, where ``N`` is 128 for example.
+
 Enabling CUDA support
 ^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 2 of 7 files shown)


Src/Base/AMReX_GpuControl.H
```diff
@@ -7,8 +7,8 @@
 
 #include <utility>
 
-#ifndef AMREX_GPU_MAX_THREADS
-#define AMREX_GPU_MAX_THREADS 256
+#ifndef AMREX_USE_GPU
+#define AMREX_GPU_MAX_THREADS 0
 #endif
```
ax3l: This is a bit unusual, since `AMREX_GPU_MAX_THREADS` should be defined by `#include <AMReX_Config.H>` at this point, no?
```suggestion
```

WeiqunZhang: `AMREX_GPU_MAX_THREADS` is only defined in `AMReX_Config.H` for GPU builds.  The problem is in a few places we have things like this in the code for both gpu and cpu. https://github.com/AMReX-Codes/amrex/blob/2ad56772eac9095b9678de5ced1e5440c4e579c9/Src/Base/AMReX_CTOParallelForImpl.H#L192
For the CPU version, the template parameter is simply ignored.

WeiqunZhang: Anyway, there are multiple ways to handle this.
1. Commit your suggestion.  But it will break all CPU builds.  So we will need to fix that.
2. Revert the change to the original code,
```
#ifndef AMREX_GPU_MAX_THREADS
#define AMREX_GPU_MAX_THREADS 256
#endif
```
3. The current way in the PR.
4. Always define AMREX_GPU_MAX_THREADS in the build systems.

ax3l: I think as a quick fix to clean file-local quirks, we could go with:
> Always define AMREX_GPU_MAX_THREADS in the build systems.

in `AMReX_Config.H` and then do a cleanup is you suggest for CPU builds later.

Below this line,
https://github.com/AMReX-Codes/amrex/blob/48f5ffc629cfcded97f9214dabc551248ab11d26/Tools/CMake/AMReX_Config.H.in#L43
we could write for now:
```C++
#ifndef AMREX_GPU_MAX_THREADS
#   define AMREX_GPU_MAX_THREADS 0
#endif
```
https://cmake.org/cmake/help/latest/command/configure_file.html

ax3l (approved):
Whatever we decide, it can be a follow-up :)

Src/Base/AMReX_GpuControl.H
WeiqunZhang: Sure, we can always define AMREX_GPU_MAX_THREADS.  However, I won't insert spaces between `#` and `define`.  @zingale won't approve it.  But seriously, in this case it's better not to use any spaces, because then the words in the two lines will align perfectly.

ax3l: Haha, sure.

PR #3117, opened 2023-01-31
Remove Atomic::Inc and Dec

WeiqunZhang:
The implementation for Atomic::Inc and Dec in our SYCL backend use compare and exchange, because they do not exist in SYCL standard.  This turns out to be very inefficient.  Although both CUDA and HIP have efficient implementation of atomicInc and atomicDec, we are removing Atomic::Inc and Dec from amrex to avoid surprises.  In all cases that we have used these functions, they can be replaced with Atomic::Add.


WeiqunZhang:
Do not merge this until the next release.

atmyers (approved):

PR #3122, opened 2023-02-02
CI: NVHPC 23.1

ax3l:
## Summary

Update CI to Test against the latest NVHPC release, v23.1

## Additional background

See #2378

## Checklist

The proposed changes:
- [ ] fix a bug or incorrect behavior in AMReX
- [ ] add new capabilities to AMReX
- [ ] changes answers in the test suite to more than roundoff level
- [ ] are likely to significantly affect the results of downstream AMReX users
- [ ] include documentation in the code and/or rst files, if appropriate

PR #2378, opened 2021-10-03
NVHPC: OpenMP Atomic Capture Fixed

ax3l:
## Summary

Demonstrate `nvc++` OpenMP 4.0 issue.

CI Log: [nvhpc_omp.txt](https://github.com/AMReX-Codes/amrex/files/7279020/nvhpc_omp.txt)

Then, apply work-around only for affected NVHPC versions: <23.1

## Additional background

Work-arounded in #2365

Nvidia bug report: 3390723 (fixed in NVHPC 23.1).

- [x] try again with the NVHPC release after 21.9 in December: https://developer.nvidia.com/hpc-sdk :x:
- [x] or with 21.11 #2687 :x:
- [x] or with 23.01 #3122 :heavy_check_mark: 

## Checklist

The proposed changes:
- [ ] fix a bug or incorrect behavior in AMReX
- [ ] add new capabilities to AMReX
- [ ] changes answers in the test suite to more than roundoff level
- [ ] are likely to significantly affect the results of downstream AMReX users
- [ ] include documentation in the code and/or rst files, if appropriate

PR #3124, opened 2023-02-02
NVHPC: CUDA

ax3l:
## Summary

Ooops. In my last commit #3122, I accidentially switched NVHPC to compile for OpenMP instead of the CUDA backend.

## Additional background

But cool, that means OpenMP 4.0 works now #2378

## Checklist

The proposed changes:

### DIFF
--- a/.github/workflows/cuda.yml
+++ b/.github/workflows/cuda.yml
@@ -67,12 +67,11 @@ jobs:
             -DCMAKE_VERBOSE_MAKEFILE=ON                  \
             -DAMReX_ENABLE_TESTS=ON                      \
             -DAMReX_PARTICLES=ON                         \
-            -DAMReX_FORTRAN=OFF                          \
+            -DAMReX_FORTRAN=ON                           \
             -DAMReX_GPU_BACKEND=CUDA                     \
             -DCMAKE_C_COMPILER=$(which nvc)              \
             -DCMAKE_CXX_COMPILER=$(which nvc++)          \
             -DCMAKE_CUDA_HOST_COMPILER=$(which nvc++)    \
-            -DCMAKE_Fortran_COMPILER=$(which nvfortran)  \
             -DCMAKE_Fortran_COMPILER=$(which nvfortran)  \
             -DAMReX_CUDA_ARCH=8.0                        \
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
Oh, I thought it was intentional because you have gave up on nvhpc comping CUDA code.

ax3l:
No, NEVER giving up on CUDA support :sweat_smile: 
That we actually fixed over the last year or so already. Just NERSC did not deploy the NVHPC versions we need yet.

I just in parallel also did not give up on NVHPC becoming a working host-side OpenMP compiler #2378. I copy pasted too hard between the two PRs, that introduced the change.

.github/workflows/cuda.yml
```diff
@@ -68,10 +68,15 @@ jobs:
...
             -DAMReX_PARTICLES=ON                         \
             -DAMReX_FORTRAN=OFF                          \
-            -DAMReX_OMP=ON                               \
+            -DAMReX_GPU_BACKEND=CUDA                     \
             -DCMAKE_C_COMPILER=$(which nvc)              \
             -DCMAKE_CXX_COMPILER=$(which nvc++)          \
-            -DCMAKE_Fortran_COMPILER=$(which nvfortran)
+            -DCMAKE_CUDA_HOST_COMPILER=$(which nvc++)    \
+            -DCMAKE_Fortran_COMPILER=$(which nvfortran)  \
+            -DCMAKE_Fortran_COMPILER=$(which nvfortran)  \
```
WeiqunZhang: We can probably remove both because Fortran is off.  But at least remove one.

ax3l: Now I am curious to turn it ON :D
```suggestion
            -DCMAKE_Fortran_COMPILER=$(which nvfortran)  \
```

.github/workflows/cuda.yml
```diff
@@ -68,10 +68,15 @@ jobs:
             -DAMReX_ENABLE_TESTS=ON                      \
             -DAMReX_PARTICLES=ON                         \
             -DAMReX_FORTRAN=OFF                          \
```
ax3l:
```suggestion
            -DAMReX_FORTRAN=ON                           \
```

WeiqunZhang (approved):

ax3l:
Wow, works with Fortran, nice.

PR #3125, opened 2023-02-02
Runtime parameters for Signal handling controls

WeiqunZhang:
Add amrex.handle_sigsegv, handle_sigint, handle_sigabrt, and handle_sigfpe for more fine controls.  Disable sigsegv handling for certain Intel GPUs because it's incompatible with the managed memory.


ax3l:
Oh, maybe worth adding the same safety in the WarpX signal handler.

atmyers (approved):

PR #3126, opened 2023-02-04
GNU Make: Add support for Intel LLVM compiler in CPU builds

WeiqunZhang:
icpc: remark #10441: The Intel(R) C++ Compiler Classic (ICC) is deprecated and will be removed from product release in the second half of 2023.

### DIFF
--- a/Tools/GNUMake/comps/intel.mak
+++ b/Tools/GNUMake/comps/intel.mak
@@ -1,78 +1,10 @@
-#
-# Generic setup for using Intel
-#
-CXX = icpc
-CC  = icc
-FC  = ifort
-F90 = ifort
+CXX = icpx
+has_intel_llvm_compiler := $(shell $(CXX) -dumpversion)
 
-CXXFLAGS =
-CFLAGS   =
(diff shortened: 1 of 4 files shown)


Tools/GNUMake/comps/intel.mak
```diff
@@ -1,78 +1,10 @@
...
-########################################################################
-
-ifeq ($(DEBUG),TRUE)
-
-  CXXFLAGS += -g -O0 -traceback -Wcheck
-  CFLAGS   += -g -O0 -traceback -Wcheck
-  FFLAGS   += -g -O0 -traceback -check bounds,uninit,pointers
-  F90FLAGS += -g -O0 -traceback -check bounds,uninit,pointers
+CXX = icpx
+has_intel_llvm_compiler := $(shell $(CXX) -dumpversion)
```
kngott, line 3: This doesn't seem like a particularly robust way to do this check, but I assume it's working for our situations of interest at the moment. We should keep an eye out, though.

kngott (approved):

PR #3128, opened 2023-02-04
NeighborList: Let scan function return total sum

WeiqunZhang:
Instead of explicit memcpy, we could let Scan::ExclusiveSum return the total sum.

Also changed memcpy to memcpy_async in a number of places.  Note the scan function has to do a stream synn internally because it has to allocate some temporary space.

### DIFF
--- a/Src/Particle/AMReX_NeighborList.H
+++ b/Src/Particle/AMReX_NeighborList.H
@@ -334,8 +334,8 @@ public:
         // First pass: count the number of neighbors for each particle
         //---------------------------------------------------------------------------------------------------------
         const size_t np_size  = (num_bin_types > 1) ? np_total : np_real;
-        m_nbor_counts.resize( np_size+1, 0);
-        m_nbor_offsets.resize(np_size+1);
+        m_nbor_counts.resize( np_size);
+        m_nbor_offsets.resize(np_size);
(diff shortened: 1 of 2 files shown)


WeiqunZhang:
It seems that the NeighborList class depends on the the size of some containers being number of particles plus one.  So I am going close this PR.  Nevertheless this is not ideal, because it requires the container being resized with an initial value (with an extra gpu kernel).

Src/Particle/AMReX_NeighborList.H
```diff
@@ -334,8 +334,8 @@ public:
         // First pass: count the number of neighbors for each particle
         //---------------------------------------------------------------------------------------------------------
         const size_t np_size  = (num_bin_types > 1) ? np_total : np_real;
-        m_nbor_counts.resize( np_size+1, 0);
-        m_nbor_offsets.resize(np_size+1);
```
atmyers, line 338: I believe the iterator start / stop logic [here](https://github.com/WeiqunZhang/amrex/blob/a75649b9ced9418affd7ff89f1ab5da6bb0c060e/Src/Particle/AMReX_NeighborList.H#L158) relies on the offsets array having size `np+1` and its last element being the total size.

PR #3136, opened 2023-02-08
Initialize TinyProfiler earlier to fix segfault

### LINKED ISSUE #3135: segfault when using TINY_PROFILE=TRUE and OpenMP
Since yesterday, the combination of TINY_PROFILE=TRUE and OpenMP causes segfaults immediately when Castro is run.  See the tests here:

http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2023-02-08/index.html

AlexanderSinn:
## Summary

fix #3135

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX.cpp
+++ b/Src/Base/AMReX.cpp
@@ -524,6 +524,7 @@ amrex::Initialize (int& argc, char**& argv, bool build_parm_parse,
 
     ParallelDescriptor::Initialize();
 
+    BL_TINY_PROFILE_INITIALIZE();
     Arena::Initialize();
     amrex_mempool_init();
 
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
There was a reason we initialized tiny profiler that later (and also previously the random generator finalize was done after tiny profiler's finalize).  It was done that way to avoid profiling the initialization cost.  So I wonder if we should add a special initialize just for the memory profiling part.

AlexanderSinn:
I made a second PR #3137 where the initialization is split up

WeiqunZhang:
The bug has been fixed in #3137.

PR #3141, opened 2023-02-09
Test: Reinit AMReX

ax3l:
## Summary

Run a test that inits and finalizes AMReX multiple times.

```console
cmake -S . -B build -DAMReX_GPU_BACKEND=CUDA -DAMReX_ENABLE_TESTS=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo
cmake --build build -j 8
ctest --test-dir build --verbose -R Reinit
```
```console
build/Tests/Reinit
cuda-gdb -ex r -ex bt --args ./Test_Reinit inputs
```

## Additional background

This is often done in pytest tests, where the same process is reused for multiple tests.
Needed for pyAMReX, ImpactX, WarpX et al.

This also demonstrates a bug for GPU testing.

## Checklist

The proposed changes:
- [ ] fix a bug or incorrect behavior in AMReX
- [ ] add new capabilities to AMReX
- [ ] changes answers in the test suite to more than roundoff level
- [ ] are likely to significantly affect the results of downstream AMReX users
- [ ] include documentation in the code and/or rst files, if appropriate

PR #3143, opened 2023-02-09
Fix bug in Device::Finalize: Forgot to clear streams

WeiqunZhang:
## Summary

We forgot to clear the non-owning gpu_stream vector.  They still held the old streams after we destroyed the streams in Device::Finalize.  This bug affects codes that call amrex::Initialize and amrex::Finalize repeatedly.

Note that we have the non-owning gpu stream vector for OMP CPU thread safety.

The bug was introduced in #1064.

## Additional background

#3141 

## Checklist

The proposed changes:


ax3l (approved):
Wohoo, thanks a lot!

PR #3148, opened 2023-02-10
Add MPI support to GpuComplex

WeiqunZhang:
This creates a specialization of Mpi_typemap for GpuComplex<T>.

Needed by https://github.com/ECP-WarpX/WarpX/pull/3618/

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_ParallelDescriptor.H
+++ b/Src/Base/AMReX_ParallelDescriptor.H
@@ -10,6 +10,7 @@
 #include <AMReX_Extension.H>
 #include <AMReX_INT.H>
 #include <AMReX_REAL.H>
+#include <AMReX_GpuComplex.H>
 #include <AMReX_Array.H>
 #include <AMReX_Vector.H>
 #include <AMReX_ValLocPair.H>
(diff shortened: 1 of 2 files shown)


kngott:
Don't we have a place for the MPI_typemap objects in Base/AMReX_ccse-mpi.H? Or is this a special case?

WeiqunZhang:
This is a specialization.  We could put the new code in `AMReX_ccse-mpi.H`, but the we would need to include the GpuComplex header there.  So I think it might be better to put it in ParallelDescriptor.H, which has already included so many other headers.

kngott:
Ah.

In addition to grouping all the Mpi _typemaps together, I was picturing it the other way: to define GpuComplex Mpiobject types, the user now has to include the entire ParallelDescriptor header even if they aren't making a Comm call in the same file. But, if it's in ccse-mpi, the option exists to just include that, which seems much lighter weight.

I suppose each way has it's pluses and minuses. I'll leave the final decision to you. Otherwise, it looks great!

kngott (approved):

ax3l (approved):
Excellent, thanks a lot!

PR #3161, opened 2023-02-16
SYCL Device Info

WeiqunZhang:
Print out more SYCL device information if available.  Currently, free memory and max memory band width are not available yet.


ax3l (approved):
Yay!

atmyers (approved):

PR #3164, opened 2023-02-17
amrex::Math: No longer need to use sycl math functions

WeiqunZhang:
We introduced `amrex::Math::floor`, etc. because their std versions were not supported in device code by the Intel compiler at that time.  Otherwise, the user would have to call either sycl::floor or std::floor with the help of macros.  This is no longer the case.  Now we can switch to use the std versions.  The amrex::Math versions are kept for backward compatibility. The particle code is not updated because it's undergoing some large changes.

### DIFF
--- a/Docs/sphinx_documentation/source/GPU.rst
+++ b/Docs/sphinx_documentation/source/GPU.rst
@@ -756,7 +756,7 @@ As another example, the following function computes the max- and 1-norm of a
                    noexcept -> GpuTuple<Real,Real>
                {
                    if (mask_ma[box_no](i,j,k)) {
-                       Real a = amrex::Math::abs(data_ma[box_no](i,j,k));
+                       Real a = std::abs(data_ma[box_no](i,j,k));
                        return { a, a };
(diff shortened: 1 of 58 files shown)


j-stephan:
Hi, I just found this by accident. Do you have a source for std math functions being usable with oneAPI device code? I found nothing meaningful in the DPC++ changelogs.

WeiqunZhang:
https://github.com/intel/llvm/blob/806775fcdadf07344bf4116f889432da0c4c8114/sycl/doc/extensions/supported/C-CXX-StandardLibrary.rst

And we have been using many std cmath functions on Intel GPUs.  They work as expected, including `std::abs`.  We pay extra attention to `std::abs` because `sycl::abs` is NOT a drop-in replacement for `int std::abs(int)`.  The sycl version returns an unsigned int and one would get 4294967290 from `-sycl::abs(-6)`.

j-stephan:
Cool, thank you!

atmyers (approved):

PR #3167, opened 2023-02-18
SYCL: Optimization of small reduction

WeiqunZhang:
For small reductions, it's faster to do the second pass on CPU.  The change is for SYCL only.


WeiqunZhang:
Regression tests have passed.

atmyers (approved):

PR #3169, opened 2023-02-21
Fix race condition with TinyProfiler for memory

### LINKED ISSUE #3156: The_Pinned_Arena uses a CArena on CPU
If `AMREX_USE_GPU=FALSE`, then The_Arena, The_Device_Arena and The_Managed_Arena are all BArenas rather than CArenas like on GPU. However, The_Pinned_Arena stays a CArena but returns false when Arena::isPinned() is called. This causes it to share TinyProfiler memory diagnostics with amrex_mempool:
```
Cpu (amrex_mempool) Memory Usage:
---------------------------------------------------------------------------------------------------------------------
Name                            Nalloc  Nfree  AvgMem min  AvgMem avg  AvgMem max  MaxMem min  MaxMem avg  MaxMem max
---------------------------------------------------------------------------------------------------------------------
Hipace::Notify()                   150    150      10 MiB      10 MiB      11 MiB      12 MiB      12 MiB      12 MiB
Hipace::Wait()                     150    150     424 KiB     424 KiB     425 KiB      12 MiB      12 MiB      12 MiB
The_Pinned_Arena::Initialize()       2      2      11   B      12   B      14   B    8192 KiB    8192 KiB    8192 KiB
amrex_mempool_init()                 4      4    7219   B    7222   B    7225   B    8192 KiB    8192 KiB    8192 KiB
Hipace::Evolve()                     2      2     452 KiB     475 KiB     497 KiB     500 KiB     500 KiB     500 KiB
main()                              62     62      15   B      15   B      15   B      48   B      48   B      48   B
---------------------------------------------------------------------------------------------------------------------
```
This is an issue because the underlying map used for profiling relies on the mutex of the CArena to be thread save, but now there are two mutexes for the same map.
The following setup should maybe cause a race condition with AMREX_USE_GPU=FALSE.
```
#pragma omp parallel
{
    BL_PROFILE("Test");
    if (omp_get_thread_num() == 0) {
        // allocate from The_Pinned_Arena
    } else {
        // allocate from amrex_mempool
    } 
}
```

AlexanderSinn:
## Summary

Fix #3156 by giving every Arena its own map for profiling.  

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_CArena.H
+++ b/Src/Base/AMReX_CArena.H
@@ -173,7 +173,7 @@ protected:
     //! The amount of memory given out via alloc().
     std::size_t m_actually_used;
     //! If this arena is profiled by TinyProfiler
-    bool m_do_profiling;
+    bool m_do_profiling = false;
     //! Data structure used for profiling with TinyProfiler
     std::map<std::string, MemStat> m_profiling_stats;
 
--- a/Src/Base/AMReX_CArena.H
+++ b/Src/Base/AMReX_CArena.H
@@ -7,6 +7,7 @@
 #include <cstddef>
 #include <set>
 #include <vector>
+#include <map>
 #include <mutex>
 #include <unordered_set>
 #include <functional>
--- a/Src/Base/AMReX_CArena.H
+++ b/Src/Base/AMReX_CArena.H
@@ -60,6 +60,12 @@ public:
      */
     virtual bool hasFreeDeviceMemory (std::size_t sz) override;
 
+    /**
+     * \brief Add this Arena to the list of Arenas that are profiled by TinyProfiler.
+     * \param memory_name The name of this arena in the TinyProfiler output.
+     */
+    virtual void registerForProfiling (const std::string& memory_name) override;
+
     //! The current amount of heap space used by the CArena object.
     std::size_t heap_space_used () const noexcept;
 
@@ -165,6 +171,11 @@ protected:
     std::size_t m_used;
     //! The amount of memory given out via alloc().
     std::size_t m_actually_used;
+    //! If this arena is profiled by TinyProfiler
+    bool m_do_profiling;
+    //! Data structure used for profiling with TinyProfiler
+    std::map<std::string, MemStat> m_profiling_stats;
+
 
     std::mutex carena_mutex;
 };
--- a/Src/Base/AMReX_TinyProfiler.cpp
+++ b/Src/Base/AMReX_TinyProfiler.cpp
@@ -349,7 +349,7 @@ TinyProfiler::memory_stop () const noexcept {
 
 MemStat*
 TinyProfiler::memory_alloc (std::size_t nbytes, std::map<std::string, MemStat>& memstats) noexcept {
-    // this function is not thread save for the same memstats
+    // this function is not thread safe for the same memstats
     // the caller of this function (CArena::alloc) has a mutex
     MemStat* stat = nullptr;
 #ifdef AMREX_USE_OMP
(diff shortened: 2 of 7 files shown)


Src/Base/AMReX_CArena.H
```diff
@@ -165,6 +172,11 @@ protected:
     std::size_t m_used;
     //! The amount of memory given out via alloc().
     std::size_t m_actually_used;
+    //! If this arena is profiled by TinyProfiler
+    bool m_do_profiling;
```
WeiqunZhang: Can we give this a default value?

Src/Base/AMReX_TinyProfiler.cpp
```diff
@@ -347,21 +348,21 @@ TinyProfiler::memory_stop () const noexcept {
 }
 
 MemStat*
-TinyProfiler::memory_alloc (std::size_t nbytes, int memtype) noexcept {
-    // this function is not thread save for the same memtype
+TinyProfiler::memory_alloc (std::size_t nbytes, std::map<std::string, MemStat>& memstats) noexcept {
+    // this function is not thread save for the same memstats
```
WeiqunZhang:
```suggestion
    // this function is not thread safe for the same memstats
```

Src/Base/AMReX_TinyProfiler.cpp
```diff
@@ -374,7 +375,7 @@ TinyProfiler::memory_alloc (std::size_t nbytes, int memtype) noexcept {
 
 void
 TinyProfiler::memory_free (std::size_t nbytes, MemStat* stat) noexcept {
-    // this function is not thread save for the same memtype
+    // this function is not thread save for the same stat
```
WeiqunZhang:
```suggestion
    // this function is not thread safe for the same stat
```

WeiqunZhang:
> Now it is important that every profiled Arena is still in scope when the profile is printed.

Maybe the dtor of Arena should deregister itself.

WeiqunZhang:
Could we move the call to `BL_TINY_PROFILE_MEMORYFINALIZE` to right above the call to `amrex_mempool_finalize`?  Currently there is a block of code finalizing `TheFinalize_Function_Stack`, which may contains users' functions deallocating Arena memory.

AlexanderSinn:
I did move `BL_TINY_PROFILE_MEMORYFINALIZE`, however I think that I mixed up the implementation and `Additionally, no profiled memory can be deallocated between finalizing TinyProfiler and finalizing the Arenas now.` is not true. It should work just not show up in the profile of Nfree.
`Now it is important that every profiled Arena is still in scope when the profile is printed. ` is still somewhat true if profiled output is desired, but DeregisterArena would prevent it from crashing now.

Src/Base/AMReX_TinyProfiler.cpp
```diff
@@ -488,9 +491,33 @@ TinyProfiler::MemoryFinalize (bool bFlushing) noexcept
...
+
+void
+TinyProfiler::DeregisterArena (std::map<std::string, MemStat>& memstats) noexcept
+{
+    for (std::size_t i = 0; i < all_memstats.size(); ++i) {
+        if (all_memstats[i] == &memstats) {
+            all_memstats.erase(all_memstats.begin() + i);
+            all_memnames.erase(all_memnames.begin() + i);
+        }
+    }
```
WeiqunZhang: In practice, this is OK because there is only one element in the vector that can meet the condition (unless one mistakenly calls RegisterArena twice).  If that's the case, we could used `std::find` and then `std::distance` to figure the element index first and then remove them from `all_memestats` and `all_memnames`.  We could also add an AMREX_ASSERT() in the end to make sure there are no more matches.  I think the code will be more readable.  The issue of the current code I cannot help noticing it will not work if there two matches in a row because of `++i` and `erase`.

AlexanderSinn: Good catch, thanks! This is intended to always work.

WeiqunZhang (approved):

PR #3170, opened 2023-02-21
CMake: Parallel HDF5 Detection

WeiqunZhang:
## Summary

It turns out the words in the settings summary of HDF5 came from the words used when HDF5 was configured with CMake.  The issue is `h5pcc -showconfig` may show `Parallel HDF5: ON`, but the current versions of CMake uses `Parallel HDF5: yes` to detect Parallel HDF5.  This commit adds a more reliable way of detecting Parallel HDF5.

## Additional background

- https://github.com/spack/spack/issues/35546
- https://github.com/HDFGroup/hdf5/issues/2488
- https://gitlab.kitware.com/cmake/cmake/-/merge_requests/8234

## Checklist

The proposed changes:

### DIFF
--- a/Tools/CMake/AMReXThirdPartyLibraries.cmake
+++ b/Tools/CMake/AMReXThirdPartyLibraries.cmake
@@ -9,8 +9,9 @@ if (AMReX_HDF5)
 
     if (AMReX_MPI AND (NOT HDF5_IS_PARALLEL))
        if (CMAKE_VERSION VERSION_LESS 3.27)
-	  # The detection in earlier versions of cmake may not be reliable.
-          # So we will try to do it ourselves.
+	      # The detection in earlier versions of cmake may not be reliable.
+          # So we will try to do it ourselves. Work-around for:
+          # https://gitlab.kitware.com/cmake/cmake/-/merge_requests/8234
           execute_process(
              COMMAND ${HDF5_C_COMPILER_EXECUTABLE} -showconfig
              OUTPUT_VARIABLE amrex_hdf5_config_output
--- a/Tools/CMake/AMReXThirdPartyLibraries.cmake
+++ b/Tools/CMake/AMReXThirdPartyLibraries.cmake
@@ -6,9 +6,29 @@ if (AMReX_HDF5)
        set(HDF5_PREFER_PARALLEL TRUE)
     endif ()
     find_package(HDF5 1.10.4 REQUIRED)
+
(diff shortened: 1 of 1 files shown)


Tools/CMake/AMReXThirdPartyLibraries.cmake
```diff
@@ -6,9 +6,29 @@ if (AMReX_HDF5)
        set(HDF5_PREFER_PARALLEL TRUE)
     endif ()
     find_package(HDF5 1.10.4 REQUIRED)
+
     if (AMReX_MPI AND (NOT HDF5_IS_PARALLEL))
-       message(FATAL_ERROR "\nHDF5 library does not support parallel I/O")
+       if (CMAKE_VERSION VERSION_LESS 3.27)
+	  # The detection in earlier versions of cmake may not be reliable.
+          # So we will try to do it ourselves.
```
ax3l: Formatting and reference:
```suggestion
	      # The detection in earlier versions of cmake may not be reliable.
          # So we will try to do it ourselves. Work-around for:
          # https://gitlab.kitware.com/cmake/cmake/-/merge_requests/8234
```

Tools/CMake/AMReXThirdPartyLibraries.cmake
```diff
@@ -6,9 +6,29 @@ if (AMReX_HDF5)
        set(HDF5_PREFER_PARALLEL TRUE)
     endif ()
     find_package(HDF5 1.10.4 REQUIRED)
```
ax3l, line 8: Purely informational:

If HDF5 was build with CMake and not with its legacy (but still widely used) autotools installer, then `find_package` can use the CMake config package that it installed (based on a hint to the root path) to pick up HDF5's install config. This installation method would not be affected and is the one that the Spack package uses already.

In order to use this, we would pass `CONFIG` here as an argument (or require CMake 3.24+, more below). This is what the response here https://github.com/HDFGroup/hdf5/issues/2488#issuecomment-1442303115 refers to.

For the default `find_package` signature that we currently use, and prior to CMake 3.24 which we also currently use, the `FindHDF5.cmake` module takes precedence before checking for the Config package.
https://cmake.org/cmake/help/latest/command/find_package.html#search-modes
https://cmake.org/cmake/help/latest/command/find_package.html#search-procedure

ax3l (approved):
Thank you for digging into this, the upstream fix and work-around.

I added a tiny suggestion and some additional context on the respons we got in your HDF5 issue.

PR #3173, opened 2023-02-23
Add build option for Intel SYCL AOT GRF mode

WeiqunZhang:
By default, the Intel SYCL compiler uses up to 128 registers.  In the large register file mode, it's up to 256 registers.  There is also the auto-large mode.

For CMake, one can use `-DAMReX_SYCL_AOT_GRF_MODE=[Large|AutoLarge]` to choose the large register file or auto-large mode.  AOT must also be enabled with `-DAMReX_SYCL_AOT=TRUE`.

For GNU Make, one can use `SYCL_AOT_GRF_MODE=[Large|AutoLarge]`, and it depends on AOT being enabled by `SYCL_AOT=TRUE`.


ax3l (commented):
This is great, thanks a lot! Let's add the new options to the doc tables as well:
- https://amrex-codes.github.io/amrex/docs_html/BuildingAMReX.html
- https://amrex-codes.github.io/amrex/docs_html/BuildingAMReX.html#building-with-cmake

ax3l (approved):
:+1:

PR #3176, opened 2023-02-24
Clang-Tidy

WeiqunZhang:
Add a configuration file .clang-tidy for clang-tidy.  Add support to it in CMake.  It can be enabled by `-DAMReX_CLANG_TIDY=ON`.

Fix many warnings raised by clang-tidy.  Most of them are fixed by `clang-tidy --fix-errors`.

To bypass clang-tidy, some unfixable codes have been moved to `*.nolint.H`.

A number of issues have been found and fixed.
    
* There was use-after-move in testing OpenMP thread safety for communication.

* The MultiFab tags were not handled correctly during vector resize.

* Parentheses were missing in the definition of AMREX_REAL_LOWEST.


.clang-tidy
```diff
@@ -0,0 +1,38 @@
...
+     bugprone-*,
+     clang-analyzer-*,
+     clang-diagnostic-*,
+     modernize-*,
+     -modernize-avoid-c-arrays,
+     -modernize-use-trailing-return-type,
+     -modernize-use-using,
+     performance-*'
+
+HeaderFilterRegex: '*parser\.(tab|lex)\.cpp\.H'
```
yut23: Note: this filter is a whitelist for header files you want to see warnings from, not a blacklist. I think it's just reporting warnings from all header files right now, since the regex is invalid (leading asterisk).
There's currently no good way to have clang-tidy ignore specific files (see https://stackoverflow.com/q/71797349). It uses POSIX ERE syntax which doesn't support lookarounds, so I think the closest you can get is something along the lines of `[^.][^c][^p][^p]\.H$` to skip any files ending with `.cpp.H`.

WeiqunZhang: Thanks for your comment! @yut23 I will give it a try.

WeiqunZhang: It works!  Thanks again.

Src/Base/Make.package
```diff
@@ -72,9 +72,12 @@ C$(AMREX_BASE)_headers += AMReX_Gpu.H AMReX_GpuQualifiers.H AMReX_GpuPrint.H AMR
 C$(AMREX_BASE)_headers += AMReX_GpuKernelInfo.H
 
 C$(AMREX_BASE)_headers += AMReX_GpuLaunchMacrosG.H AMReX_GpuLaunchFunctsG.H
+C$(AMREX_BASE)_headers += AMReX_GpuLaunchMacrosG.nolint..H
```
WeiqunZhang:
```suggestion
C$(AMREX_BASE)_headers += AMReX_GpuLaunchMacrosG.nolint.H
```

jmsexton03 (approved):

.clang-tidy
WeiqunZhang: We also need to filter-in files ending with say `cqq.H`. So I think it needs to be something like `([^.][^c][^p][^p]|[^c][^p][^p]|[^p][^p]|[^p])\.H$`.

WeiqunZhang: Actually it probably should be `([^\.]...|[^c]..|[^p].|[^p])\.H$`

PR #3181, opened 2023-02-28
SYCL: Fix #3172

WeiqunZhang:
This may not be our bug.  But using sycl::reqd_work_group_size does not work for the 2d range cases.  So we have to remove that.


WeiqunZhang:
@jmsexton03 Could we get this merged ASAP?

jmsexton03 (approved):

PR #3185, opened 2023-03-01
SYCL: Tweak #3164 and switch to multi_ptr for sincos

nmnobre:
Hi @WeiqunZhang,

First, this includes a small addendum to fully get rid of the SYCL math functions.
Thanks a lot for #3164 by the way, I'm slowly phasing out the patches I once needed for Open SYCL (formerly hipSYCL).
Also, do we still need the conditional for CUDA/HIP, i.e. do we need to distinguish between `::isnan` and `std::isnan`?

Second, this makes explicit use of `multi_ptr` on the fairly new `sincos()` family of functions.
This seems to be what the [standard](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_description_of_the_built_in_types_available_for_sycl_host_and_device) asks for, see `genfloatptr`. However, having said that, it's clear DPC++ supported the previous, much simpler syntax and I don't know if that's an extension or if it's covered by the standard under some pointer conversion magic. This more cumbersome syntax does work with both DPC++ and Open SYCL. 

Lastly, I tried to slightly mitigate the clutter I created by reusing the code from `sincos()` in `sincospi()`...

Cheers
-Nuno

### DIFF
--- a/Src/Base/AMReX_Math.H
+++ b/Src/Base/AMReX_Math.H
@@ -124,7 +124,7 @@ std::pair<double,double> sincos (double x)
 {
     std::pair<double,double> r;
 #if defined(AMREX_USE_SYCL)
-    r.first = sycl::sincos(x, &r.second);
+    r.first = sycl::sincos(x, sycl::private_ptr<double>(&r.second));
 #elif defined(__CUDA_ARCH__) && defined(AMREX_USE_CUDA) || \
       defined(__HIP_DEVICE_COMPILE__) && defined(AMREX_USE_HIP) || \
       (defined(_GNU_SOURCE) && !defined(__APPLE__))
@@ -142,7 +142,7 @@ std::pair<float,float> sincos (float x)
 {
     std::pair<float,float> r;
 #if defined(AMREX_USE_SYCL)
-    r.first = sycl::sincos(x, &r.second);
+    r.first = sycl::sincos(x, sycl::private_ptr<float>(&r.second));
 #elif defined(__CUDA_ARCH__) && defined(AMREX_USE_CUDA) || \
       defined(__HIP_DEVICE_COMPILE__) && defined(AMREX_USE_HIP) || \
       (defined(_GNU_SOURCE) && !defined(__APPLE__))
@@ -160,7 +160,7 @@ std::pair<double,double> sincospi (double x)
 {
     std::pair<double,double> r;
 #if defined(AMREX_USE_SYCL)
-    r.first = sycl::sincos(pi<double>()*x, &r.second);
+    r.first = sycl::sincos(pi<double>()*x, sycl::private_ptr<double>(&r.second));
 #elif defined(__CUDA_ARCH__) && defined(AMREX_USE_CUDA) || \
       defined(__HIP_DEVICE_COMPILE__) && defined(AMREX_USE_HIP)
     ::sincospi(x, &r.first, &r.second);
@@ -177,7 +177,7 @@ std::pair<float,float> sincospi (float x)
 {
     std::pair<float,float> r;
 #if defined(AMREX_USE_SYCL)
-    r.first = sycl::sincos(pi<float>()*x, &r.second);
+    r.first = sycl::sincos(pi<float>()*x, sycl::private_ptr<float>(&r.second));
 #elif defined(__CUDA_ARCH__) && defined(AMREX_USE_CUDA) || \
       defined(__HIP_DEVICE_COMPILE__) && defined(AMREX_USE_HIP)
     ::sincospif(x, &r.first, &r.second);
--- a/Src/Base/AMReX_Math.H
+++ b/Src/Base/AMReX_Math.H
@@ -159,14 +159,11 @@ AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
 std::pair<double,double> sincospi (double x)
 {
     std::pair<double,double> r;
-#if defined(AMREX_USE_SYCL)
-    r.first = sycl::sincos(pi<double>()*x, sycl::private_ptr<double>(&r.second));
-#elif defined(__CUDA_ARCH__) && defined(AMREX_USE_CUDA) || \
-      defined(__HIP_DEVICE_COMPILE__) && defined(AMREX_USE_HIP)
+#if defined(__CUDA_ARCH__) && defined(AMREX_USE_CUDA) || \
+    defined(__HIP_DEVICE_COMPILE__) && defined(AMREX_USE_HIP)
     ::sincospi(x, &r.first, &r.second);
 #else
-    r.first  = std::sin(pi<double>()*x);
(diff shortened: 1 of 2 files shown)


Src/Base/AMReX_Math.H
```diff
@@ -109,7 +109,7 @@ std::pair<double,double> sincos (double x)
 {
     std::pair<double,double> r;
 #if defined(AMREX_USE_SYCL)
-    r.first = sycl::sincos(x, &r.second);
+    r.first = sycl::sincos(x, sycl::multi_ptr<double,sycl::access::address_space::global_space>(&r.second));
```
WeiqunZhang: Thanks for spotting this!

Looks like the ctor of multi_ptr is explicit.
```
explicit multi_ptr(typename multi_ptr<ElementType, Space, access::decorated::yes>::pointer);
```
So the conversion is not supposed to happen.  But is `global_space` the right address space?  `r` is a thread private variable likely in the register  So what does it mean to create a multi_ptr in global space from its address?

WeiqunZhang:
We need to keep using `sycl::isnan` and `sycl::isinf` because Intel compiler is very aggressive with fast-math optimization.  I just did some tests.  `std::isnan` and `std::isinf` always return false.  However, the sycl functions seem to work.

Does Open SYCL support `sycl::isnan` and `sycl::isinf`? If not, we can work around that.

WeiqunZhang:
Since `std::isnan` and `std::isinf` do not work on host either with the Intel compiler, we should use sycl::isnan and sycl::isinf for both host and device codes.

WeiqunZhang:
I tested with cuda 11.0.2 and rocm 4.1.0.  Both isnan and isinf worked.  Since amrex now requires cuda >= 11 and a recent rocm, I could use `std::isnan` and `std::isinf` now.  So I suggest we do something like
```
    template <class T>
    bool isnan (T m) noexcept
    {
#if defined(AMREX_USE_SYCL)
    return sycl::isnan(m);
#else
        return std::isnan(m);
#endif
    }
```

Src/Base/AMReX_Math.H
nmnobre: Fair point, I've done some testing, saw no performance difference but I think semantically using the local address space makes more sense like you suggested :)

nmnobre:
> We need to keep using `sycl::isnan` and `sycl::isinf` because Intel compiler is very aggressive with fast-math optimization. I just did some tests. `std::isnan` and `std::isinf` always return false. However, the sycl functions seem to work.

Okay, let me try to understand this. Say you have the following code:

```
#include <iostream>
#include <sycl/sycl.hpp>

int main()
{
    sycl::gpu_selector gpu_device_selector;
    sycl::platform platform(gpu_device_selector);
    auto const &gpu_devices = platform.get_devices();
    std::unique_ptr<sycl::device> sycl_device = std::make_unique<sycl::device>(gpu_devices[0]);
    std::unique_ptr<sycl::context> sycl_context = std::make_unique<sycl::context>(*sycl_device);
    sycl::queue* qp = new sycl::queue(*sycl_context, *sycl_device,
                                    sycl::property_list{sycl::property::queue::in_order{}});
    sycl::queue q = *qp;

    bool boo[4];
    {
        sycl::buffer<bool , 1>  booBuf{boo, sycl::range<1>{4}};

        q.submit([&](sycl::handler &h)
                 {
        sycl::accessor booResult { booBuf, h, sycl::write_only, sycl::no_init };
        h.parallel_for(sycl::range<1>(1),
        [=] (sycl::id<1> i)
        {
          booResult[0] = sycl::isnan(sqrt(-1));
          booResult[1] = std ::isnan(sqrt(-1));
          booResult[2] = sycl::isinf(1./0);
          booResult[3] = std ::isinf(1./0);
        }); });
    }

        std::cout << std::boolalpha
                  << "sycl::isnan(sqrt(-1)) = " << boo[0] << std::endl
                  << "std ::isnan(sqrt(-1)) = " << boo[1] << std::endl
                  << "sycl::isinf(1./0) = "     << boo[2] << std::endl
                  << "std ::isinf(1./0) = "     << boo[3] << std::endl;
}
```

I can see that compiling with `-O3` and `-Ofast` produces completely different results but I can't see an inconsistency between `sycl::` and `std::`. Do you?

WeiqunZhang:
Yes I do with -O3.
```
sycl::isnan(sqrt(-1)) = true
std ::isnan(sqrt(-1)) = false
sycl::isinf(1./0) = true
std ::isinf(1./0) = false
```

nmnobre:
> Yes I do with -O3.
> 
> ```
> sycl::isnan(sqrt(-1)) = true
> std ::isnan(sqrt(-1)) = false
> sycl::isinf(1./0) = true
> std ::isinf(1./0) = false
> ```

I can confirm your findings...
It works with the open-source version targeting AMD/Nvidia GPUs, but not the proprietary version (Intel oneAPI Base kit) even when targeting OpenCL on a Xeon CPU _unless_ I explicitly use `-fno-fast-math`.

I'll just apply your suggestion above since we never use `amrex::Math::isinf`, so I guess we don't have to reverse the removal of `using sycl::isinf` in #3164?

WeiqunZhang:
Yes, we should put `using sycl::isinf` and probably `sycl::isfinite` back.

nmnobre:
> Yes, we should put `using sycl::isinf` and probably `sycl::isfinite` back.

Ok, let me know if you're happy with this.
The fact they optimise these so aggressively even at `-O1` looks a bit suspicious to me, a bit too aggressive, no?
Heck, like this, if you don't specify _any_ `-O` flag, you lose the semantics, you need `-O0` or `-fno-fast-math`...

Src/Base/AMReX_Math.H
```diff
@@ -127,7 +142,7 @@ std::pair<float,float> sincos (float x)
 {
     std::pair<float,float> r;
 #if defined(AMREX_USE_SYCL)
-    r.first = sycl::sincos(x, &r.second);
+    r.first = sycl::sincos(x, sycl::multi_ptr<float,sycl::access::address_space::local_space>(&r.second));
```
WeiqunZhang: One last thing.  I think the address space should be private because variable `r` is thread private.  `local` means the work group level.  Also there is an alias for that, `sycl::private_ptr<float>`.

WeiqunZhang (changes requested):

Src/Base/AMReX_Math.H
WeiqunZhang: Private space is more appropriate.

Src/Base/AMReX_Math.H
nmnobre: Thank you, I didn't even know there was a finer address space below the thread block/work group level.
I also didn't know the spec provided aliases for the different spaces, that's convenient!

WeiqunZhang (approved):

PR #3195, opened 2023-03-10
Fix order of bins for neighbor list construction

atmyers:
Do not merge until I've confirmed the MFiX-Exa tests pass.

The proposed changes:


atmyers:
The CPU and GPU MFiX-Exa tests pass with this change. However, there is no appreciable performance difference. Still, we should probably make this change to the bin ordering convention so it looks less "surprising" to people.

Src/Particle/AMReX_NeighborList.H
```diff
@@ -371,7 +371,7 @@ public:
               for (int ii = amrex::max(ix-num_cells, 0); ii <= amrex::min(ix+num_cells, nx-1); ++ii) {
                 for (int jj = amrex::max(iy-num_cells, 0); jj <= amrex::min(iy+num_cells, ny-1); ++jj) {
                   for (int kk = amrex::max(iz-num_cells, 0); kk <= amrex::min(iz+num_cells, nz-1); ++kk) {
```
WeiqunZhang: We need to switch the order of for loops too.

atmyers:
This does seem to make a small performance improvement now...

WeiqunZhang (approved):

asalmgren:
@atmyers -- this PR is over 2 years old -- is it still relevant?

PR #3198, opened 2023-03-14
Sorting for faster current deposition

AlexanderSinn:
## Summary

For current deposition in codes such as HiPACE++ and WarpX, it is mush better to sort by `x -> y -> z -> ppc` instead of `ppc -> x -> y -> z`.

## Additional background

To use this, pass in {0,0,0} (sort by cell), {-1,-1,-1} (sort by node) or {-2, -2, -2} (sort by cell and node) as a bin_size into SortParticlesByBin.

## Checklist

The proposed changes:


atmyers:
Hi Alex,

Thank you for contributing this! Instead of adding this to `DenseBins`, what do you think about making it a separate, stand-alone function (maybe in AMReX_ParticleUtil.H) that fills a `Gpu::DeviceVector` with the permutation indices? 

Also, since as I understand it does not strictly order the particles, maybe we could make a new method of ParticleContainer called "SortParticlesForDeposition" or something like that that calls your function instead of going through DenseBins. The part that reorders the particle data according to the permutation vector could be factored out and shared between the two routines. 

Let me know if you want me to help with the refactoring.

AlexanderSinn:
Thanks, that is a great suggestion I will implement that soon.
I would like some help with the block-wide prefix sum, however. There should be more efficient algorithms, possibly using the cub library, intrinsic instructions or using the fact that the initial array consists of only 0 and 1 for each thread. Additionally, implementations for SYCL, OMP and serial might be needed although I am not sure if this method of sorting is advantageous for deposition on CPU.

WeiqunZhang:
For block-wide scan, we can use cub, rocprim and sycl.  The SYCL standard has scan functions. We also implemented our own.  For example,

- hip: https://github.com/AMReX-Codes/amrex/blob/aaab804f1438b5eeef3f613ca527c62bcabb5425/Src/Base/AMReX_Scan.H#L698
- cub: https://github.com/AMReX-Codes/amrex/blob/aaab804f1438b5eeef3f613ca527c62bcabb5425/Src/Base/AMReX_Scan.H#L833
- sycl: https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_exclusive_scan_and_inclusive_scan
- our own sycl implementation: https://github.com/AMReX-Codes/amrex/blob/aaab804f1438b5eeef3f613ca527c62bcabb5425/Src/Base/AMReX_Scan.H#L471

Maybe we should wrap our own sycl implementation into a function.  I would not be surprised if it's faster than sycl::exclusive_scan coming with the compiler, because I am not sure if it uses shared memory given how hard it is to get shared memory into sycl kernel.

atmyers (approved):

atmyers:
Thanks for doing that refactor, this looks good to me. I think sycl support and using CUB / rocprim for the scan can be put off for later.

Src/Particle/AMReX_ParticleUtil.H
```diff
@@ -758,6 +758,126 @@ void clearEmptyEntries (C& c)
...
+    Gpu::Device::streamSynchronize();
+}
+
+template <class index_type, class pos_struct>
+void PermutationForDeposition (Gpu::DeviceVector<index_type>& perm, index_type nitems,
+                               pos_struct const* v, Box bx, Geometry geom, const IntVect idx_type)
+{
+    AMREX_ALWAYS_ASSERT(idx_type.allGE(IntVect(0)) && idx_type.allLE(IntVect(2)));
+
+    const IntVect refine_vect = max(idx_type, IntVect(1)).min(IntVect(2));
```
WeiqunZhang, line 849: This may not be correct. It's doing something like this
```
IntVect tmp = amrex:max(idx_type, IntVect(1)); // This is a lexicographical comparison
const IntVect refine_vect = tmp.min(IntVect(2)); // This is an element wise operation.
```

AlexanderSinn: I want to use this function so idx_type does not change.
```
/**
 * \brief Returns the IntVect that is the component-wise maximum of two
 * argument IntVects.
 */
AMREX_GPU_HOST_DEVICE
AMREX_FORCE_INLINE
IntVect
max (const IntVect& p1, const IntVect& p2) noexcept
{
    IntVect p(p1);
    p.max(p2);
    return p;
}
```
That should be element wise

AlexanderSinn: AMReX_IntVect.H 671

WeiqunZhang: Okay.  Forgot we have a specialization for IntVect.  I thought it will pick up the function in AMReX_Algorithm.H.

atmyers:
Thanks! I will go ahead and merge.

PR #3199, opened 2023-03-14
Rework handling of roundoff domain extent

dpgrote:
## Summary
This reworks previous changes that created the round off domain which was done to avoid round off errors when locating particles in the domain, specifically when handling periodic boundaries. This PR changes how this is done so that the roundoff domain makes as small a change as possible.

## Additional background
Previous PRs #2839 and #2950 created a round off domain that has a slightly smaller extent that the actual domain. Any particles that ended up outside the round off domain but inside the actual domain were shifted to be on the boundary of the round off domain. The main purpose was to handle problems with periodic boundary with mixed precision (single precision particles and double precision grid). In that case, it is possible for a particle to have `z < lo` and `z+(hi-lo) > hi`. It is also possible for a particle to have `z < hi` but `(z-lo)*dzinv > ihi` putting it in a grid cell beyond the domain. Unfortunately, the roundoff domains was being set further inside that was needed and would affect valid particles that would not have those issues. While the problem is small, the effects can be noticeable when examining simulation properties, for example charge conservation in WarpX.

This PR changes the roundoff domain so that only problematic particles are affected. The code now uses `std::nextafter` to find the exact place where the problems occurs. At the lower end, it ensures that the casted position will be greater than ProbLo. At the upper end, it ensures that `(z-lo)*dzinv <= ihi`.

One comment is that this assumes that particle grid coordinates will always be calculated at the higher precision, with expressions like `iz = (int)(z - lo)`, without `lo` being cast to `ParticleReal`.

## Checklist

The proposed changes:


atmyers (approved):

Src/Base/AMReX_Geometry.cpp
```diff
@@ -525,36 +525,49 @@ Geometry::computeRoundoffDomain ()
...
+        // compared to (phi + plo).
+        int iters = 0;
+        while (roundoff_lo[idim] < plo && iters < 20) {
+            roundoff_lo[idim] = std::nextafter(roundoff_lo[idim], roundoff_hi[idim]);
+            iters++;
+        }
+        ParticleReal rhi = roundoff_hi[idim];
+        while (iters < 20) {
+            rhi = std::nextafter(rhi, roundoff_lo[idim]);
+            if (std::floor((rhi - plo)*dxinv) >= ihi + 1 - ilo) {
```
atmyers:
```suggestion
            if (int(std::floor((rhi - plo)*dxinv)) >= ihi + 1 - ilo) {
```

atmyers:
Thank you for this!

PR #3202, opened 2023-03-15
Clang-Tidy VI

WeiqunZhang:
* Update GNU Make.  One can use `USE_CLANG_TIDY=TRUE` to enable clang-tidy check.  For example, to enable clang-tidy check, specify the clang-tidy command (if it's the default clang-tidy), and treat warnings as errors,  one can run

      make USE_CLANG_TIDY=TRUE CLANG_TIDY=clang-tidy-12 CLANG_TIDY_WARN_ERROR=TRUE

  By default amrex/.clang-tidy is used as the config file.  One can change it with CLANG_TIDY_CONFIG_FILE=my_clang-tidy-conf.  If you want to let clang-tidy fix errors for you, you can run

      make USE_CLANG_TIDY=TRUE CLANG_TIDY="clang-tidy --fix-errors"

  To avoid race conditions, the make job should not use `-j` (unless it's `-j1`) when fixing errors.

* Add clang-tidy to GNU Make CIs.

* Fix warnings in Fortran interface, particle and plotfile tools.

### DIFF
--- a/.github/workflows/clang.yml
+++ b/.github/workflows/clang.yml
@@ -87,13 +87,16 @@ jobs:
   # Build 2D libamrex with configure
   configure-2d:
     name: Clang@7.0 NOMPI Release [configure 2D]
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
     - uses: actions/checkout@v3
     - name: Dependencies
-      run: .github/workflows/dependencies/dependencies_clang7.sh
+      run: |
+        .github/workflows/dependencies/dependencies_clang.sh 14
+        .github/workflows/dependencies/dependencies_clang-tidy.sh 14
     - name: Build & Install
       run: |
         ./configure --dim 2 --with-fortran no --comp llvm --with-mpi no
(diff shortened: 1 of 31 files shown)


zingale:
The change to `sites/Make.unknown` breaks MPI compilation for me

WeiqunZhang:
@zingale Are you using Open MPI or MPICH?

zingale:
It's mpich 3.4.2.  This only seems to be an issue on the machine with Rocky Linux 9 (which is equivalent to RHEL 9)

zingale:
I'll try debugging this more this afternoon.

WeiqunZhang:
What do you get with `mpicxx -compile_info`?  Note that there was a typo in an early version of this branch.

zingale:
oh!  I think I was using the earlier version.

zingale:
indeed, the version I was testing had `--compile_info` instead of `-compile_info`.  So I'm good now.

jmsexton03 (approved):

PR #3211, opened 2023-03-22
Warning about CPU core oversubscription

WeiqunZhang:
If the user forgets to set OMP_NUM_THREADS, OMP might be oversubscribing CPU cores. If that's the case (to the best of our knowledge) and verbosity is on, we print out a warning.

We use `std::thread::hardware_concurrency` to detect the number of cores available. Note that this is only a hint according to the C++ standard. To detect the number of MPI ranks per node, we use OMP_COMM_TYPE_NODE for Open MPI and MPI_COMM_TYPE_SHARED for others.  Since MPI_COMM_TYPE_SHARED might group processes across nodes, the return value of `ParallelDescriptor::NProcsPerNode` might also be wrong.  Nevertheless, both `std::thread::hardware_concurrency` and `ParallelDescriptor::NProcsPerNode` seem to work well.

### DIFF
--- a/Src/Base/AMReX.cpp
+++ b/Src/Base/AMReX.cpp
@@ -73,6 +73,7 @@
 #include <iomanip>
 #include <new>
 #include <stack>
+#include <thread>
 #include <limits>
 #include <vector>
 #include <algorithm>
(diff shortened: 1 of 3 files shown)


WeiqunZhang:
The CI failures have been fixed in #3202.

Src/Base/AMReX.cpp
```diff
@@ -457,6 +458,23 @@ amrex::Initialize (int& argc, char**& argv, bool build_parm_parse,
     }
 #endif
 
+#if defined(AMREX_USE_MPI) && defined(AMREX_USE_OMP)
+    if (system::verbose > 0) {
+        auto ncores = int(std::thread::hardware_concurrency());
+        if (ncores != 0 && // It might be zero according to the C++ standard.
+            ncores < omp_get_max_threads() * ParallelDescriptor::NProcsPerNode())
+        {
+            amrex::Print(amrex::ErrorStream())
```
ax3l, line 467: This is awesome!

The one nitpick that could happen is: the user manages to set `OMP_NUM_THREADS` on the originating batch node only, but forgets to forward the env variable. That way, the warning would not be raised but all but the first node would still have the issue.

`amrex::AllPrint` would check everywhere.
(If the IO rank is != the first one, then this is covered as well with `amrex::Print`.)

Admittedly, potentially academic and not worth the noise an `AllPrint` would cause.

ax3l (approved):
Thanks a lot! :rocket: :sparkles:

PR #3219, opened 2023-03-27
Speedup clang tidy

WeiqunZhang:
We make ccache save a log file, from which we find out which files need a
rebuild. Instead of running clang-tidy on every file, we now only run it on
those files that had a miss in ccache. This significantly speeds up
clang-tidy in CIs.

Note that we treat clang-tidy warnings as errors in CIs. So if there is a
hit in the ccache's cache, the file is free of clang-tidy warnings. Also
note that we set CCACHE_EXTRAFILES to the clang-tidy configuration file. So
any changes to the config file will invalidate the cache as it should.

If a CI job successfully passes the compilation stage with ccache enabled,
but fails at the clang-tidy stage, the ccache's cache has been updated. One
might think this will cause issues because we now have files that have been
cached by ccache and meanwhile have failed the clang-tidy checks. However,
this is not an issue for Github CIs. This is because ccache's cache will not
be saved as a Github cache, if a CI fails.

Fix various clang-tidy warnings in Tests/. They were not reported before
because clang-tidy was not used on the tests.

### DIFF
--- a/.github/workflows/clang.yml
+++ b/.github/workflows/clang.yml
@@ -34,6 +34,8 @@ jobs:
         export CCACHE_COMPRESS=1
         export CCACHE_COMPRESSLEVEL=10
         export CCACHE_MAXSIZE=125M
+        export CCACHE_EXTRAFILES=${{ github.workspace }}/.clang-tidy
+        export CCACHE_LOGFILE=${{ github.workspace }}/ccache.log.txt
         ccache -z
 
         mkdir build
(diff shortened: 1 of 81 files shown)


WeiqunZhang:
Depend on #3218.

WeiqunZhang:
@yut23

WeiqunZhang:
I turned off EB in a 1D test, because we do not support EB in 1D.

atmyers (approved):

PR #3220, opened 2023-03-28
Fix: queryAdd for resizing vectors

ax3l:
## Summary

I think there is a bug in `ParmParse::queryAdd` and `ParmParse::queryarr` for `std::vector` arguments.

The passed array is only up-sized if needed, but not down-sized if the key exists. This leads to bugs for long default values and shorter user-provided arrays.

Reproducer:
```C++
std::vector<amrex::ParticleReal> cos_coef = ez.default_cos_coef;  // long default vector
pp_element.queryAdd("cos_coefficients", cos_coef);                // can be shorter user input

amrex::Print() << "cos_coef.size()" << cos_coef.size() << "\n";   // size of the default vector
```

## Additional background

First seen in https://github.com/ECP-WarpX/impactx/pull/322

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_ParmParse.H
+++ b/Src/Base/AMReX_ParmParse.H
@@ -966,8 +966,8 @@ public:
      * added to the ParmParse database.  The return value indicates if it
      * existed previously.
      *
-     * If `name` is found, then the ref argument will be resized
-     * (shrunk or extended) according to the number of values in the inputs.
+     * If `name` is found, then the ref argument will be reallocated
+     * (and resized) according to the number of values in the inputs.
      */
     template <typename T>
     int queryAdd (const char* name, std::vector<T>& ref) {
--- a/Src/Base/AMReX_ParmParse.H
+++ b/Src/Base/AMReX_ParmParse.H
@@ -965,10 +965,18 @@ public:
      * be stored in the `ref` argument.  If not, the value in `ref` will be
      * added to the ParmParse database.  The return value indicates if it
      * existed previously.
+     *
+     * If `name` is found, then the ref argument will be resized
+     * (shrunk or extended) according to the number of values in the inputs.
      */
     template <typename T>
     int queryAdd (const char* name, std::vector<T>& ref) {
-        int exist = this->queryarr(name, ref);
+        std::vector<T> empty;
+        int exist = this->queryarr(name, empty);
+        if (exist) {
+            ref.resize(empty.size());
+            this->queryarr(name, ref);
+        }
         if (!exist && !ref.empty()) {
             this->addarr(name, ref);
         }
--- a/Src/Base/AMReX_ParmParse.H
+++ b/Src/Base/AMReX_ParmParse.H
@@ -974,8 +974,7 @@ public:
         std::vector<T> empty;
         int exist = this->queryarr(name, empty);
         if (exist) {
-            ref.resize(empty.size());
-            this->queryarr(name, ref);
+            ref = std::move(empty);
         }
         if (!exist && !ref.empty()) {
             this->addarr(name, ref);
--- a/Src/Base/AMReX_ParmParse.cpp
+++ b/Src/Base/AMReX_ParmParse.cpp
@@ -867,7 +867,7 @@ squeryarr (const ParmParse::Table& table,
     if ( num_val == 0 ) return true;
 
     int stop_ix = start_ix + num_val - 1;
-    if ( static_cast<int>(ref.size()) <= stop_ix )
+    if ( static_cast<int>(ref.size()) != stop_ix + 1 )  // grow or shrink to fit
     {
         ref.resize(stop_ix + 1);
     }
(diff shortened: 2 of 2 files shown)


Src/Base/AMReX_ParmParse.cpp
```diff
@@ -867,7 +867,7 @@ squeryarr (const ParmParse::Table& table,
     if ( num_val == 0 ) return true;
 
     int stop_ix = start_ix + num_val - 1;
-    if ( static_cast<int>(ref.size()) <= stop_ix )
+    if ( static_cast<int>(ptr.size()) != stop_ix + 1 )  // grow or shrink to fit
```
ax3l, line 870: I am not yet 100% happy with the fix, since it might resize a vector where the API might allow to only partially read values from the params....?

Not 100% sure about this API contract, as the changes on the passed vector are only partly documented in the header file.

If needed, we could also apply a resize fix one level up in `queryAdd` for `std::vector`...?

Src/Base/AMReX_ParmParse.cpp
```diff
@@ -867,7 +867,7 @@ squeryarr (const ParmParse::Table& table,
     if ( num_val == 0 ) return true;
 
     int stop_ix = start_ix + num_val - 1;
-    if ( static_cast<int>(ref.size()) <= stop_ix )
+    if ( static_cast<int>(ref.size()) != stop_ix + 1 )  // grow or shrink to fit
```
ax3l: I am not yet 100% happy with the fix, since it might resize a vector where the API might allow to only partially read values from the params....?

I am not sure I am not missing something about this API contract: the behavior on the _passed vector_ argument is not documented in the doxygen string in the header file.

If needed, we could also apply a resize fix one level up in `queryAdd` for `std::vector` as in https://github.com/ECP-WarpX/impactx/pull/327 ?

WeiqunZhang: The doxygen string (for `getktharr`) does say "ref is resized (if necessary) to hold num_val_values".  The doxygen string for `queryarr` says "Same as ...".

On one hand, it seems a good idea to resize the vector so that the size of the vector tells us how many values ParmParse has.  Otherwise, we have to call `countval` to find it out. On the other hand, it's possible it might break some people's code. So I don't know.

It seems that the doxygen string for `getktharr` is wrong on what `start_ix` is.  It says
```
The value in the list indexed by start_ix is copied into std::vector<int>[0],
std::vector<int>[1] holds start_ix+1, etc.
```
But it seems to me the code puts the value in the list indexed by `start_ix` into `std::vector<int>[start_ix]`, not `[0]`.

Anyway, I don't think I ever used the two optional parameters, `start_ix` and `num_val`.

WeiqunZhang: If we do change it to always resize, we don't need to do `if ( static_cast<int>(ref.size()) != stop_ix + 1 )  // grow or shrink to fit`.  Just `resize` will work for all cases.

WeiqunZhang: If we have some codes that actually rely on the no resize unless necessary behavior, @asalmgren and @drummerdoc might know.

ax3l: Let me push the same high-level fix as I used in ImpactX as a work-around.
It applies only to one overload of `queryAdd` where it is definitely safe to resize.

 -> Done.

WeiqunZhang:
Let's do a rebase on development to get the clang-tidy speedup.

Src/Base/AMReX_ParmParse.H
```diff
@@ -965,10 +965,18 @@ public:
...
+     * (shrunk or extended) according to the number of values in the inputs.
      */
     template <typename T>
     int queryAdd (const char* name, std::vector<T>& ref) {
-        int exist = this->queryarr(name, ref);
+        std::vector<T> empty;
+        int exist = this->queryarr(name, empty);
+        if (exist) {
+            ref.resize(empty.size());
+            this->queryarr(name, ref);
```
WeiqunZhang:
```suggestion
            ref = std::move(empty);
```

ax3l: Nice, good idea.

Src/Base/AMReX_ParmParse.H
```diff
@@ -965,10 +965,17 @@ public:
      * be stored in the `ref` argument.  If not, the value in `ref` will be
      * added to the ParmParse database.  The return value indicates if it
      * existed previously.
+     *
+     * If `name` is found, then the ref argument will be resized
+     * (shrunk or extended) according to the number of values in the inputs.
```
ax3l:
```suggestion
     * If `name` is found, then the ref argument will be reallocated
     * (and resized) according to the number of values in the inputs.
```

WeiqunZhang (approved):

PR #3225, opened 2023-03-28
Support complete elliptic integrals of 1st and 2nd type in Parser

aveksler1:
## Summary
We want to use complete elliptic integrals from `cmath`, `comp_ellint_1` and `comp_ellint_2` in the parser in order to make it easier to load in magnetic fields from coils in WarpX. I added these to the parser, and tested against the same functions in scipy - the results match. 

Since clang does not support special functions in C++17 (https://en.cppreference.com/w/cpp/compiler_support/17#C.2B.2B17_library_features) I implemented it for GCC compilers only, following the structure of how `jn` is implemented in the parser.

## Checklist

The proposed changes:


ax3l:
@aveksler1 I think you might need to add GPU paths here as well. At least CUDA has a lot of special math functions, so you could check if they have what you need.

An example of bessel functions we added earlier is in https://github.com/ECP-WarpX/WarpX/pull/1854
(Note: this PR is before we moved the parser upstream to AMReX.)

WeiqunZhang:
The CIs failed because we treat warnings as errors.  Those warnings seem to come from Bison changing the integer type in various places from int to size_t. We need to suppress those warnings.

aveksler1:
I added a separate GPU path and have suppressed the sign comparison warnings.

WeiqunZhang:
@aveksler1 Could you merge development into this so that the Intel CI can pass?

WeiqunZhang (approved):

PR #3226, opened 2023-03-29
Github Cache Eviction Policy

WeiqunZhang:
After a workflow finishes, delete all old caches associated with it.

Add an action to delete caches associated with a PR upon the close of the PR.

Use variables in the github context as part of the cache key names instead of manually repeating.

This is how cache works now. There is a cache file for the default branch (i.e., development). When
a PR is submitted, the main cache file is checked out to be used by the PR. When the PR finishes
the CIs, a copy of the cache including any updates will be saved. This cache has a scope associated
with the PR. Another PR will be be able to access it. If the author pushes an update to the BR branch,
The new CIs will run using the latest PR cache and it will evict old cache in the PR. When the PR is
merged, its cache will be deleted. The development branch's workflow will generate the latest main
cache.

### DIFF
--- a/.github/workflows/clang.yml
+++ b/.github/workflows/clang.yml
@@ -14,15 +14,30 @@ jobs:
   library_clang:
     name: Clang@7.0 C++17 SP NOMPI Debug [lib]
     runs-on: ubuntu-20.04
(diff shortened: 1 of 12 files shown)


WeiqunZhang:
This won't work because of the lack of write permission.

ax3l:
Argh

WeiqunZhang:
This is now implemented in #3229 with a different approach.

PR #3229, opened 2023-03-30
GitHub Cache Management

WeiqunZhang:
The total size of all caches in a GitHub repo is limited to 10 GB. So we need to proactively manage the caches to avoid undesired cache eviction.

GitHub cache has its scope associated with branch. When a new PR is opened, the last used cache of the default branch (i.e., development for AMReX) is restored for the PR's actions. After the actions finish, the caches will be saved under a unique name every time and only that PR has access to it. (Presumably this is for security reason to avoid cache poison.) If the actions are run many times, a large number of caches will be accumulated. This uses a lot of space, even though we only need to keep the most recent cache. In this PR, we implement the following approach to keep only the last used cache for each job and remove all caches associated with a PR when it's closed.

We have added a workflow that is triggered by the events of completed workflows (e.g., LinuxClang workflow that has a number of jobs using cache.) This new workflow runs on the default branch and it's therefore safe to be granted write permission. Each triggering workflow saves the PR number as an artifact, if it's a pull request run. This is a way to pass the PR number to the cache cleanup workflow, which will clean up old caches associated with the jobs in the triggering workflow. When a PR is closed or merged, a post PR workflow is trigger, and it will in turn trigger another workflow that cleans up all the caches associated with the PR. After the merge, the actions will be run on the default branch. The newly added workflows will also clean up the old caches in the default branch.

### DIFF
--- a/.github/workflows/cleanup-cache.yml
+++ b/.github/workflows/cleanup-cache.yml
@@ -1,0 +1,64 @@
+name: CleanUpCache
+
+on:
+  workflow_run:
+    workflows: [LinuxClang, cuda, LinuxGcc, hip, Hypre, intel, macos, PETSc, SUNDIALS, windows]
+    types:
+      - completed
+
+concurrency:
+  group: ${{ github.ref }}-${{ github.head_ref }}-${{ github.event.workflow_run.name }}
+  cancel-in-progress: true
+
+jobs:
+  CleanUpCcacheCache:
+    name: Clean Up Ccahe Cache for ${{ github.event.workflow_run.name }}
+    runs-on: ubuntu-latest
+    permissions:
+      actions: write
+      contents: read
+    env:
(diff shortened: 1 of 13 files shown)


.github/workflows/cleanup-cache.yml
```diff
@@ -0,0 +1,64 @@
...
+            pr_number=`cat pr_number.txt`
+            BRANCH=refs/pull/${pr_number}/merge
+          fi
+
+          # Setting this to not fail the workflow while deleting cache keys.
+          set +e
+
+          # In our cache keys, substring after `-git-` is git hash, substring
+          # before that is a unique id for jobs (e.g., `ccache-LinuxClang-configure-2d`).
+          # Th goal is to keep the last used key of each job and delete all otheres.
```
jmsexton03, line 49:
```suggestion
          # The goal is to keep the last used key of each job and delete all others.
```

jmsexton03 (approved):

PR #3233, opened 2023-03-31
CodeQL: Add query filters

WeiqunZhang:
This should reduce the number of warnings and fix the SARIF error in CI.


WeiqunZhang:
This does fix the CodeQL issue. We will deal with nvhpc in another PR.

atmyers (approved):

PR #3234, opened 2023-03-31
Ccache: Use depend mode for Intel SYCL compiler

WeiqunZhang:
The Intel SYCL compiler generates temporary files during compilations that result in cache misses for ccache. The ccache depend mode forgoes the preprocessed file information, and thus works around the issue.


WeiqunZhang:
It has been verified that ccache now works for SYCL CIs.

jmsexton03 (approved):

PR #3243, opened 2023-04-07
Fix periodic boundary bug in #3199

atmyers:
PR #3199 broke the Nyx regression tests. The issue is we require the particle positions to be in [plo, phi). This won't happen if rhi is exactly equal to phi. The fix is to make sure rhi is always slightly inside the domain. 

EDIT - fixed this in a different way. Now, when applying periodic boundaries, the particle positions will be set to be strictly less than rhi.

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_ParticleUtil.H
+++ b/Src/Particle/AMReX_ParticleUtil.H
@@ -569,8 +569,8 @@ bool enforcePeriodic (P& p,
     for (int idim = 0; idim < AMREX_SPACEDIM; ++idim)
     {
         if (! is_per[idim]) continue;
-        if (p.pos(idim) >= phi[idim]) {
-            while (p.pos(idim) >= phi[idim]) {
+        if (p.pos(idim) >= rhi[idim]) {
+            while (p.pos(idim) >= rhi[idim]) {
                 p.pos(idim) -= static_cast<ParticleReal>(phi[idim] - plo[idim]);
             }
             // clamp to avoid precision issues;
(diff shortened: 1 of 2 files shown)


dpgrote:
I had setup `roundoff_hi` to mirror `phi` intentionally so that the two were consistent. A particle at either `roundoff_hi` or `phi` should be considered out of the domain. See in `outsideRoundoffDomain` that the check for outside is `x >= roundoff_hi[0]`.

WeiqunZhang:
Maybe `enforcePeriodic` in AMReX_ParticleUtil.H needs to be updated. For example, `if (p.pos(idim) >= phi[idim])` should be `if (p.pos(idim) >= rhi[idim])`.

atmyers:
In that case, it seems that we need to change the code in `enforcePeriodic` to make sure the particle position is < rhi after applying periodic boundaries here: https://github.com/AMReX-Codes/amrex/blob/development/Src/Particle/AMReX_ParticleUtil.H#L586

EDIT - crossed in the air with Weiqun.

Src/Particle/AMReX_ParticleUtil.H
```diff
@@ -584,8 +584,8 @@ bool enforcePeriodic (P& p,
                 p.pos(idim) += static_cast<ParticleReal>(phi[idim] - plo[idim]);
             }
             // clamp to avoid precision issues;
-            if (p.pos(idim) > rhi[idim]) {
-                p.pos(idim) = rhi[idim];
+            if (p.pos(idim) >= rhi[idim]) {
+                p.pos(idim) = std::nextafter(p.pos(idim), rlo[idim]);
```
dpgrote, line 588: Yes, this looks like the right thing to do.

jmsexton03 (approved):

ax3l:
Hi, I think we have a similar test in pyAMReX that breaks with GNU 10.1 now.
When you have a chance, let me know what to update pls :)
https://github.com/AMReX-Codes/pyamrex/pull/120

ax3l:
There seem to be also a series of regressions in WarpX:
https://github.com/ECP-WarpX/WarpX/pull/3822

PR #3247, opened 2023-04-10
Fix Roundoff Domain

WeiqunZhang:
There was a flaw in amrex::Geometry::computeRoundoffDomain. If probhi is close to zero, std::nextafter towards negative infinity will be a very small number (e.g., 1.e-300). So the function will be able to find the lowest value of roundoff_hi that's outside the domain within a reasonable number of iterations.

In the new implementation, we use bisection to find the lowest value of roundoff_lo that's inside the domain and the highest value fo roundoff_hi that's inside the domain, up to a tolerance.

X-ref:
- #3243
- #3199

### DIFF
--- a/Src/Base/AMReX_Geometry.cpp
+++ b/Src/Base/AMReX_Geometry.cpp
@@ -518,10 +518,14 @@ Geometry::computeRoundoffDomain ()
         inv_dx[k] = 1.0_rt/dx[k];
     }
 
+    constexpr int maxiters = 200;
+    int numiters = 0;
+
     for (int idim = 0; idim < AMREX_SPACEDIM; ++idim)
     {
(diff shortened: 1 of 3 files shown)


WeiqunZhang:
A test is here. https://github.com/WeiqunZhang/amrex-devtests/blob/main/roundoff_domain/main.cpp It takes up to 53 iterations.

Need to test with different compiler optimization flags and fp model flags.

Also need to test single precision and mixed precision.

WeiqunZhang:
Using `std::nextafter` seems very fragile. Switch to a tolerance based on std::numeric_limits::epsilon now. The max number of iterations in the test is now 17.

atmyers (approved):

PR #3248, opened 2023-04-10
Add Tests/RoundoffDomain

WeiqunZhang:
The new test exposes an issue with Intel compiler. So we have to rewrite the computeRoundofDomain function in a slightly different way to avoid roundoff difference between `auto r = f(x-y)` and `z = x-y; auto r = f(z)`.

Also fix an assertion issue in #3247.


ax3l (approved):
Thanks a lot! :rocket: :sparkles:

PR #3251, opened 2023-04-12
Add powi function in AMReX_Math.H

lucafedeli88:
## Summary
This PR adds a replacement of `std::pow` (`powi`)  that is intended to be used when the exponent is integer and known at compile time. If fast math optimizations are not enabled, `powi` can be one order of magnitude faster than `std::pow`.

## Additional background
Adding this function to `AMReX_Math.H` was discussed with @WeiqunZhang and @ax3l in https://github.com/ECP-WarpX/WarpX/pull/3824

## Checklist

The proposed changes:


WeiqunZhang:
Your original implementation had the data type (e.g., double or float) as a template parameter. Are there any reasons for changing to the current approach? It also seems that we could support integral types too (but make sure the exponent is >= 0).

lucafedeli88:
> Your original implementation had the data type (e.g., double or float) as a template parameter. Are there any reasons for changing to the current approach? It also seems that we could support integral types too (but make sure the exponent is >= 0).

The reason is that with `float` and `double` we can support also negative exponents, but this is not possible with integers. So the implementation forces the use of either a `double` or a `float`. 

Would it be possible to use  `std::is_floating_point` in a GPU kernel? If this is the case I can re-write the function using a template parameter for the data type (accepting also integral types).

WeiqunZhang:
You can use enable_if with `!std::is_integral<T>() || Power>=0`. In principle, complex numbers should also work.

WeiqunZhang:
> Would it be possible to use std::is_floating_point in a GPU kernel?

Yes

lucafedeli88:
@WeiqunZhang , I've followed your advice. Now the function accepts also integer types, but only if Power >= 0

WeiqunZhang (approved):

PR #3270, opened 2023-04-24
Doc: Real/ParticleReal Doxygen

ax3l:
## Summary

Add missing doxygen string, so users can click through on important AMReX types when they look at API documentation.

## Additional background

First seen in missing links in https://impactx.readthedocs.io/en/latest/_static/doxyhtml/classimpactx_1_1_impact_x_particle_container.html#a8cabfd7912cd98a8871bc75734474472

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_REAL.H
+++ b/Src/Base/AMReX_REAL.H
@@ -66,9 +66,28 @@ typedef double amrex_particle_real;
 
 #ifdef __cplusplus
 namespace amrex {
+    /** Floating Point Type for Fields
+     *
+     * This is a floating point type used by default for fields such as
+     * FabArray and MultiFab classes. This can be changed at compile time to
+     * run double, single or experimental floating point precisions for field
+     * computations and storage.
+     */
     using Real = amrex_real;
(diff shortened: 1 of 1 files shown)


Src/Base/AMReX_REAL.H
```diff
@@ -66,9 +66,28 @@ typedef double amrex_particle_real;
 
 #ifdef __cplusplus
 namespace amrex {
+    /** Floating Point Type for Fields
+     *
+     * This is a floating point type used by default for fields such as
+     * FabArray and MultiFab classes. This can be changed at compile time to
```
WeiqunZhang: FabArray does not have a default type. Maybe change it to FArrayBox. Also, what's the experimental floating point precision you mention below?

ax3l: Thanks, typo fixed!

I meant to write: this is the location where you would play with FP16, bfloat16 or custom floating point types from external libs :)

WeiqunZhang (approved):

PR #3277, opened 2023-04-26
Update SortParticlesForDeposition for pure SoA

atmyers:
I also fixed a typo in the SoA particle `pos` routine that returned a `RealVect`.

Follow-up to #2878.

The proposed changes:


ax3l (approved):
Thank you :)

PR #3278, opened 2023-04-26
Implement increment for pure SoA

atmyers:
Follow-up to #2878.

The proposed changes:
- [x] fix a bug or incorrect behavior in AMReX
- [x] add new capabilities to AMReX
- [ ] changes answers in the test suite to more than roundoff level
- [ ] are likely to significantly affect the results of downstream AMReX users
- [ ] include documentation in the code and/or rst files, if appropriate

PR #3281, opened 2023-04-27
MakeParticle SoA: Const Only

ax3l:
## Summary

Since we return particles by value, `make_particle` for SoA should better only return the const variant.

## Additional background

~~Related to, but~~ does ~~not yet~~ fix https://github.com/AMReX-Codes/amrex/pull/3278#issuecomment-1524490087

Follow-up to #2878.

## Checklist

The proposed changes:


atmyers:
This seems to have issues, but I believe with #3283 and #3284 we should be good.

PR #3286, opened 2023-04-30
FillBoundary for Spherical Coordinates

WeiqunZhang:
This adds a FillBoundary function for spherical coordinates (theta, phi, r). The function can also be used to implement other cases

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,6 +1,7 @@
 Checks: >
     '-*,
      bugprone-*,
+       -bugprone-branch-clone,
        -bugprone-easily-swappable-parameters,
        -bugprone-exception-escape,
        -bugprone-implicit-widening-of-multiplication-result,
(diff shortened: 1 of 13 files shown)


zingale:
we're interested in spherical coords for Castro -- is there other interest here?

WeiqunZhang:
This is a feature requested a CarpetX user @lwji. Justin Shafner has also expressed interest in using this for his simulations in cylindrical coordinates. @zingale, remember the special periodic boundary (e.g., x-lo is rotated to y-lo) you were talking about? It can be done with this.

zingale:
I'm no longer interested in that boundary.  I finished the sim just doing the whole thing.

But now I want to do 2-d spherical (r-theta) with phi axisymmetry in Castro.  So we'll want to get that geometry support into AMReX.  This will allow us to model a spherical shell in a star.

jmsexton03 (approved):

PR #3288, opened 2023-05-01
Github Cache: Path and Hypre

WeiqunZhang:
Use ~/.cache/ccache instead of ~/.cache as path, because the system might store all kinds of caches in ~/.cache, but we only want to save ccache's cache. (Currently, this is not an issue, but it might be in the future.)

Cache hypre cuda build.


ax3l (approved):
Oh, good to know!

PR #3280, opened 2023-04-27
SoA: InitRandom/OnePerCell

ax3l:
## Summary

Support more init methods on SoA particle containers.

- [x] `ParticleInitData`
- [x] `InitOnePerCell`
- [x] `InitRandom` #3325
- [ ] `InitRandomPerBox`

## Additional background

Follow-up to #2878.
Needed for https://github.com/AMReX-Codes/pyamrex/pull/124

## Checklist

The proposed changes:
- [ ] fix a bug or incorrect behavior in AMReX
- [x] add new capabilities to AMReX
- [ ] changes answers in the test suite to more than roundoff level
- [ ] are likely to significantly affect the results of downstream AMReX users
- [ ] include documentation in the code and/or rst files, if appropriate

PR #3290, opened 2023-05-02
PureSoA: Disable AoS Access

AlexanderSinn:
## Summary

When using a PureSoA particle container, trying to access the AoS components will now result in a compile-time error. 

Previously it still compiled, and I suspect it even allocated memory for AoS.
New error message:
```
/hipace/src/Hipace.cpp:1499:39: error: ‘struct amrex::ThisParticleTileHasNoAoS’ has no member named ‘begin’
 1499 |         const auto& pos_structs = aos.begin() + nreal;
      |                                       ^~~~~
```
Using this I found more places that needed to be updated to account for PureSoA.

PureSoA still has the following issues:
- [x] I believe it’s not possible to use the binning functions (in `DenseBins.H`) as they require a particle pointer.
- [ ] The `ParticleTileData` struct that is used to access SoA data can be very large, over 300 bytes in some cases. I recommend changing the array-of-pointers to a full 2D array. This way ParticleTileData could be as small as 24 bytes (real pointer, int pointer, stride).

## Additional background

Follow-up to #2878.

## Checklist

The proposed changes:


Src/Particle/AMReX_ParIter.H
```diff
@@ -88,11 +88,11 @@ public:
 
     [[nodiscard]] SoARef GetStructOfArrays () const { return GetParticleTile().GetStructOfArrays(); }
 
-    [[nodiscard]] int numParticles () const { return GetArrayOfStructs().numParticles(); }
+    [[nodiscard]] int numParticles () const { return GetParticleTile().numParticles(); }
```
ax3l: added via #3303

atmyers (approved):

ax3l (approved):

ax3l:
For the errors in
```
detected during instantiation of "void amrex::ParticleContainer_impl<T_ParticleType, T_NArrayReal, T_NArrayInt, Allocator>::InitRandom
```
I started porting those in #3280

ax3l (approved):

PR #3291, opened 2023-05-02
[WIP] Fix: RedistributeCPU for SoA

ax3l:
## Summary

This is a first draft and needs to be generalized.

## Additional background

Follow-up to #2878.
ImpactX w/ OMP now: https://github.com/ECP-WarpX/impactx/pull/348
- [x] not crashing with space charge anymore
- [ ] results still off expectations: we loose particles now.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_ParticleContainerI.H
+++ b/Src/Particle/AMReX_ParticleContainerI.H
@@ -7,13 +7,13 @@ void
 ParticleContainer_impl<ParticleType, NArrayReal, NArrayInt, Allocator>::SetParticleSize ()
 {
     num_real_comm_comps  = 0;
-    int comm_comps_start = AMREX_SPACEDIM + NStructReal;
+    int comm_comps_start = ParticleType::is_soa_particle ? 0 : AMREX_SPACEDIM + NStructReal;
     for (int i = comm_comps_start; i < comm_comps_start + NumRealComps(); ++i) {
         if (h_redistribute_real_comp[i]) {++num_real_comm_comps;}
     }
 
     num_int_comm_comps = 0;
-    comm_comps_start   = 2 + NStructInt;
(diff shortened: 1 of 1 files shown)


Src/Particle/AMReX_ParticleContainerI.H
```diff
@@ -1676,7 +1676,7 @@ ParticleContainer_impl<ParticleType, NArrayReal, NArrayInt, Allocator>
                           // We own it but must shift it to another place.
                           auto index = std::make_pair(pld.m_grid, pld.m_tile);
                           AMREX_ASSERT(tmp_local[pld.m_lev][index].size() == num_threads);
-                          tmp_local[pld.m_lev][index][thread_num].push_back(p);
+                          //tmp_local[pld.m_lev][index][thread_num].push_back(p);
```
ax3l, line 1679: What is this for a pure SoA particle?
Does the later
```C++
aos.insert(aos.end(), aos_tmp[i].begin(), aos_tmp[i].end());
aos_tmp[i].erase(aos_tmp[i].begin(), aos_tmp[i].end());
```
work as we expect?

Let's add more inline comments for the individual code blocks, so we understand the in-transit data structure and logic each of the many loops/copies/moves executes :)

Not clear to me if in-transit it is still an SoA and when what is overwritten. I see that later AoS attributes are not communicated correctly in ImpactX (e.g., momenta) but cannot say why yet.

ax3l:
Underlying bug fixed via #3296.

We keep this open since we want to check if sections like
```
tmp_local[pld.m_lev][index][thread_num].push_back(p);
```
are well-defined/sensible for pure SoA particles or if we need additional branching there.

ax3l:
Continued in #3299

PR #3296, opened 2023-05-04
Implement StructOfArray::empty() and ParticleTile::empty() for pure SoA

atmyers:
The implementation of `ParticleTile::empty()` was wrong for pure SoA, which led to problems in ImpactX: https://github.com/ECP-WarpX/impactx/pull/348/

Follow-up to #2878.

The proposed changes:


ax3l (approved):
Brilliant! Thank you so much!

PR #3311, opened 2023-05-12
HIP Debug Build

WeiqunZhang:
Use -O1 instead of -O0, because the HIP compiler has various issues with the latter. For example, it segfaults when compiling WarpX. It fails with errors of size exceeding limit at link time when compiling Tests/Amr/Advection_AmrCore/Exec. Now both codes work (with GNU Make).

Add -munsafe-fp-atomics in debug mode too.

The changes are for GNU Make only. Even with `-O1`, CMake still does not work.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -82,6 +82,12 @@ jobs:
           ccache -s
           du -hs ~/.cache/ccache
 
+          # Make sure CodeQL has something to do
+          touch Src/Base/AMReX.cpp
+          export CCACHE_DISABLE=1
+          cd build
+          make -j 2
+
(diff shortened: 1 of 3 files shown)


jmsexton03 (approved):

esclapez:
@WeiqunZhang do you want to relax this too ?
https://github.com/AMReX-Codes/amrex/blob/ab567b81cbf1382fc46f8f1c239d89478f72995c/Tools/GNUMake/Make.defs#L59

WeiqunZhang:
Yes, thanks!

ax3l:
Oops, we chatted about this work-around but forgot to put it into CMake. Will push a PR.

PR #3315, opened 2023-05-18
GNU Make: use CLANG_TIDY_CONFIG_FILE if it's set

yut23:
## Summary

If a downstream project defined `CLANG_TIDY_CONFIG_FILE` with their own config, it didn't actually get added to the clang-tidy arguments. This PR makes `CLANG_TIDY_CONFIG_FILE` default to the amrex config file instead.

## Checklist

The proposed changes:


WeiqunZhang:
Maybe we can do `CLANG_TIDY_CONFIG_FILE ?= $(AMREX_HOME)/.clang-tidy`.

yut23:
Yes, I think that's cleaner

WeiqunZhang (approved):

PR #3321, opened 2023-05-20
Add a new CI for multi-d builds

WeiqunZhang:
In this CI, we also run installation tests and the spack smoke test.

This is a follow-up on #3309.


ax3l (approved):
Awesome, thanks a lot :)

PR #3324, opened 2023-05-22
Fix: AMReXConfig.cmake Multi-Dim

ax3l:
## Summary

Fix two regressions from #3309 in the installed `AMReXConfig.cmake`:
```
CMake Error at lib/cmake/AMReX/AMReXConfig.cmake:257 (add_library):
  add_library cannot create ALIAS target "AMReX::amrex" because another
  target with the same name already exists.
```

and setting properties for newer CUDA.

## Additional background

Regression to #3309, spotted by @drangara, @ajnonaka, wdf et al. 

## Checklist

The proposed changes:

### DIFF
--- a/Tools/CMake/AMReXConfig.cmake.in
+++ b/Tools/CMake/AMReXConfig.cmake.in
@@ -236,11 +236,13 @@ endif ()
 # CMake targets
 include( "${CMAKE_CURRENT_LIST_DIR}/AMReXTargets.cmake" )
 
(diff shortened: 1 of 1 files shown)


ajnonaka (approved):

ajnonaka:
This fixes the regression tests on biollante.  Thanks @ax3l !!!

ax3l:
Thank you for the quick & detailed report :pray: :tada:

PR #3327, opened 2023-05-24
cleaning and fixes for PureSoA ParticleTile

AlexanderSinn:
## Summary

* fix return by value/reference mismatch of `id()` and `cpu()` with `decltype(auto)`
* remove dangerous functions `data` and `dataPtr` form `ArrayOfStructs` (they should have returned ParticleType* but returned the position array of the first particle)
* update `packParticleData` and `unpackParticleData` for PureSoA
* change `getSuperParticle` and add `setSuperParticle` for PureSoA. Remaining issue: super particle has redundant/unused fields for positions, id and cpu
* update `ConstParticleTileData` for PureSoA
* don’t call `getParticleTileData` in `ParticleTile::id()` etc. as this could be very slow if used in a for loop
* use if constexpr instead if SFINAE in a number of places
* change PODVector constructor so that `getParticleTileData()` won't allocate memory if no runtime components are used

## Additional background

Follow-up to #2878.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_ArrayOfStructs.H
+++ b/Src/Particle/AMReX_ArrayOfStructs.H
@@ -62,15 +62,9 @@ public:
         resize(nrp + num_neighbors);
     }
 
-    [[nodiscard]] int getNumNeighbors () { return m_num_neighbor_particles; }
+    [[nodiscard]] int getNumNeighbors () const { return m_num_neighbor_particles; }
 
     [[nodiscard]] bool empty () const { return m_data.empty(); }
-
-    [[nodiscard]] const RealType* data () const { return &(m_data[0].m_pos[0]); }
-    [[nodiscard]]       RealType* data ()       { return &(m_data[0].m_pos[0]); }
-
-    [[nodiscard]] const RealType* dataPtr () const { return data(); }
-    [[nodiscard]]       RealType* dataPtr ()       { return data(); }
 
     [[nodiscard]] std::pair<int,int> dataShape () const {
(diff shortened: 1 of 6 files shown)


ax3l:
Epic changes all along. Thanks a lot! :rocket: :sparkles:

ax3l (approved):

Src/Particle/AMReX_ArrayOfStructs.H
```diff
@@ -62,16 +62,10 @@ public:
         resize(nrp + num_neighbors);
     }
 
-    [[nodiscard]] int getNumNeighbors () { return m_num_neighbor_particles; }
+    [[nodiscard]] int getNumNeighbors () const { return m_num_neighbor_particles; }
 
     [[nodiscard]] bool empty () const { return m_data.empty(); }
 
-    [[nodiscard]] const RealType* data () const { return &(m_data[0].m_pos[0]); }
```
atmyers, line 69: I agree that these are weird, but unfortunately they are needed for backwards compatibility with codes that still pass particle structs into Fortran subroutines.

ax3l: Should we potentially make the `void*` pointers to reflect what they do?

(Can be a separate PR.)

ax3l: Ah I see, changed to `ParticleType*` is also good.

atmyers (approved):

PR #3184, opened 2023-03-01
CI: Clang CUDA

ax3l:
## Summary

Try to compile with vanilla Clang/LLVM's CUDA C++ frontend & its LLVM PTX backend.

Vanilla LLVM is used in a variety of tools that we want to use for research.

## Additional background

ping @ludgerpaehler

## Checklist

The proposed changes:
- [ ] fix a bug or incorrect behavior in AMReX
- [x] add new capabilities to AMReX
- [ ] changes answers in the test suite to more than roundoff level
- [ ] are likely to significantly affect the results of downstream AMReX users
- [ ] include documentation in the code and/or rst files, if appropriate

PR #3331, opened 2023-05-26
Reinforce __CUDACC__ conditionals w/ AMREX_USE_CUDA

nmnobre:
## Summary

Same motivation as #3090.
LLVM, the workhorse of most SYCL compilers, doesn't define `__NVCC__` but does define `__CUDACC__`, see [here](https://llvm.org/docs/CompileCudaWithLLVM.html#detecting-clang-vs-nvcc-from-code).
Even though [these aren't exactly the same](https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#nvcc-identification-macro), it feels like for AMReX's purposes they should be fairly equivalent and this is a cleaner patch than checking for `AMREX_USE_CUDA`.

Cheers,
-Nuno

### DIFF
--- a/Src/Base/AMReX_CudaGraph.H
+++ b/Src/Base/AMReX_CudaGraph.H
@@ -2,7 +2,7 @@
 #define AMREX_CUDA_GRAPH_H_
 #include <AMReX_Config.H>
 
-#if defined(__CUDACC__)
+#if defined(__CUDACC__) && defined(AMREX_USE_CUDA)
 
 #include <AMReX.H>
 #include <AMReX_Array.H>
--- a/Src/Base/AMReX_FBI.H
+++ b/Src/Base/AMReX_FBI.H
@@ -385,7 +385,7 @@ FabArray<FAB>::CMD_remote_setVal_gpu (typename FabArray<FAB>::value_type x,
     });
 }
 
-#if defined(__CUDACC__)
+#if defined(__CUDACC__) && defined (AMREX_USE_CUDA)
 template <class FAB>
 void
 FabArray<FAB>::FB_local_copy_cuda_graph_1 (const FB& TheFB, int scomp, int ncomp)
@@ -551,7 +551,7 @@ FabArray<FAB>::FB_local_copy_cuda_graph_n (const FB& TheFB, int scomp, int ncomp
 
 #ifdef AMREX_USE_GPU
 
-#if defined(__CUDACC__)
+#if defined(__CUDACC__) && defined(AMREX_USE_CUDA)
 
 template <class FAB>
 void
--- a/Src/Base/AMReX_FabArrayBase.H
+++ b/Src/Base/AMReX_FabArrayBase.H
@@ -492,7 +492,7 @@ public:
         Long         m_nuse{0};
         bool         m_multi_ghost = false;
         //
-#if defined(__CUDACC__)
+#if defined(__CUDACC__) && defined (AMREX_USE_CUDA)
         CudaGraph<CopyMemory> m_localCopy;
         CudaGraph<CopyMemory> m_copyToBuffer;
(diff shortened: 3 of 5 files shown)


ax3l:
Nice! X-ref #3184

ax3l:
Awesome!

Would LLVM CUDA still work with that, too? #3090 :)

nmnobre:
> Awesome!
> 
> Would LLVM CUDA still work with that, too? #3090 :)

It depends. Do we want to allow the code under `__CUDACC__` to be seen by LLVM CUDA compilations?
I think it's mostly CUDA graph stuff, I'm not sure what that's used for?...

If yes, then I better use `AMREX_USE_CUDA` like in #3090. I must admit I overlooked #3184.

WeiqunZhang:
It appears that nvc++ defines `__CUDACC__`, but not `__NVCC__`. So I don't think we should make the changes proposed here.

nmnobre:
> It appears that nvc++ defines `__CUDACC__`, but not `__NVCC__`. So I don't think we should make the changes proposed here.

Ah, that's unfortunate, but very well noted, thank you. cmake for the NVHPC workflow still identifies nvcc as the CUDA compiler, so I was a bit confused.

In any case, would checking for `AMREX_USE_CUDA` like in #3090 work for you? :)

WeiqunZhang:
Yes, checking `AMREX_USE_CUDA` or `! defined(AMREX_USE_SYCL)` is probably the way to do. Also could we limit the changes to the places that actually break the SYCL compiler for your (not the Windows stuff assuming you are not compiling it on Windows)?

ax3l:
> It depends. Do we want to allow the code under __CUDACC__ to be seen by LLVM CUDA compilations?

Yes.
LLVM CUDA is another popular CUDA compiler and used in many research tool.

> I think it's mostly CUDA graph stuff, I'm not sure what that's used for?...

Usually it is also used to separate device-only code, e.g., in math abstractions.

nmnobre:
> It appears that nvc++ defines __CUDACC__, but not __NVCC__. So I don't think we should make the changes proposed here.

Do we intend to use nvc++ to compile CUDA code eventually? I've been advised to not rely on nvc++ for CUDA code, at least in its present state. :-)

> > It depends. Do we want to allow the code under **CUDACC** to be seen by LLVM CUDA compilations?
> 
> Yes. LLVM CUDA is another popular CUDA compiler and used in many research tool.
 
Absolutely, I was rather asking if CUDA graphs or the other features under `defined(__CUDACC__)` were supported by LLVM. I didn't find anything to suggest otherwise, so I'll assume they are. :-) This means I'm now using `defined(AMREX_USE_CUDA)` to reinforce the conditionals rather than the previously suggested `defined(__NVCC__)`.

> > I think it's mostly CUDA graph stuff, I'm not sure what that's used for?...
> 
> Usually it is also used to separate device-only code, e.g., in math abstractions.

Thanks!

 > Also could we limit the changes to the places that actually break the SYCL compiler for your (not the Windows stuff assuming you are not compiling it on Windows)?

Of course, I've now changed only the absolutely necessary preprocessor conditionals. This means that, for instance, the definition for `FB_local_copy_cuda_graph_1()` is shielded with `defined(AMREX_USE_CUDA)` whereas its declaration isn't, and that the conditional for the definition of `emptyKernel()` remains unchanged, is that okay?

ax3l:
CUDA Graphs: yes, I do think that should work with LLVM Clang as well, since it's a runtime feature. cc @kngott

ax3l (approved):
This looks safe to me now :)

WeiqunZhang:
Not sure why some CIs are not running. I am going to rebase this branch to retrigger the CIs.

WeiqunZhang:
Looks like I don't have the permission to push to your branch. I will try to manually do it.

PR #3335, opened 2023-05-30
CMake: Link option when HIP is a language

WeiqunZhang:
## Summary

Although HIP is treated as C++ in AMReX CMake, we should add the link options `-Xoffload-linker --whole-archive -fgpu-rdc` when the link language is HIP, not just CXX. Otherwise, an AMReX application code that treats HIP as a CMake supported language will not get those flags.

## Additional background

https://github.com/xsdk-project/xsdk-examples/issues/50

## Checklist

The proposed changes:


ax3l (approved):
Thank you, looks good as discussed on Slack :+1: 

Technically, the HIP targets should pull `whole-archive` in from AMD, but it will not hurt if set twice. Likely this is not propagated yet, because we are not yet building HIP targets ourselves. The RDC flags were definitely missing on our side.

PR #3338, opened 2023-06-01
Particle Ids: Add int Constants

ax3l:
## Summary

Ideally we do not need them, because 2B particles
generated per GPU is very low, but we do need this temporarily until we find a way to generalize large id values again for pure SoA particle layouts.

cc @AlexanderSinn 

## Additional background

Follow-up to #2878.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_ParticleTile.H
+++ b/Src/Particle/AMReX_ParticleTile.H
@@ -347,7 +347,7 @@ struct SoAParticle : SoAParticleBase
     {
     }
 
-    static Long the_next_id;
+    static int the_next_id;
 
     //functions to get id and cpu in the SOA data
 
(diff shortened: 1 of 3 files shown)


Src/Particle/AMReX_ParticleTile.H
```diff
@@ -421,25 +421,25 @@ SoAParticle<NArrayReal, NArrayInt>::NextID ()
 #endif
     next = the_next_id++;
 
-    if (next > LastParticleID)
+    if (next > IntParticleIds::LastParticleID)
```
AlexanderSinn, line 424: If the user adds particles in large batches, then this check could miss an overflow after which next would be negative.

ax3l: Yes, fully agree. I am doubtful about a lot of these checks here.
The same problem is present for the (signed) `Long` ids.

I guess the idea about those is that one catches at least some problem if assigning ids contiguously. One could improve the checks here (and for the `Long`s) slightly by checking hat a `next` id is also not negative. But probably a separate PR.

ax3l:
ping @atmyers @AlexanderSinn ready to merge? :)

WeiqunZhang (approved):

PR #3347, opened 2023-06-06
Fix Warning in #3338

WeiqunZhang:
Explicitly convert amrex::Long to int.


ax3l (approved):
Ah, I cannot be trusted after all 😅

PR #3351, opened 2023-06-06
CI: oneAPI SYCL for Nvidia GPUs

nmnobre:
Hi @WeiqunZhang, @ax3l,

This is a spin-off from #3341 that targets Nvidia GPUs only and which downloads the required plug-in as part of the workflow's dependencies. A few notes/questions:

1) I'm currently using a (public) token with only 10 allowed daily usages. Apparently there's no IP check, it just works. I've asked Codeplay to generate one for us with a couple hundred allowed usages per day and I think then we should follow @ax3l's suggestion of hiding it from the public as a repo secret.

2) Can we use `ubuntu-latest` instead of `ubuntu-20.04`?

3) How do you decide on the size of the ccache?

4) Can I compile with `AMReX_EB` and `AMReX_Fortran` both `ON` on a single job?  The existing SYCL workflow has two separate jobs, but I couldn't understand why?

5) Even though the default sub-group size is 32, I chose to explicitly pass it to CMake for clarity. Do you think this is sensible or should I just remove the explicit variable?

6) Is there any documentation tweaks that need to (or should) be made? Or would you rather keep this fairly hidden for the time being?

Cheers,
-Nuno

### DIFF
--- a/.github/workflows/dependencies/dependencies_dpcpp.sh
+++ b/.github/workflows/dependencies/dependencies_dpcpp.sh
@@ -35,3 +35,8 @@ do
         || { sleep 10; }
 done
 if [[ ${status} -ne 0 ]]; then exit 1; fi
+
+TOKEN=NDI2ODkwOWI1MjljZDI3ZmM1NThjMzE3ZN7wAAAuqxCriVveW7v4ZivecXw_ZUOqa34K5I_bXu6LG13SzmrOsy0GTZtMi2E3hpORpeDxCejBb-V70nwUqjOCPysp
+curl -o oneapi_nvidia.sh -L "https://developer.codeplay.com/api/v1/products/download?product=oneapi&variant=nvidia&aat=${TOKEN}"
+chmod +x oneapi_nvidia.sh
+sudo ./oneapi_nvidia.sh --yes
--- a/.github/workflows/intel.yml
+++ b/.github/workflows/intel.yml
@@ -98,6 +98,54 @@ jobs:
         ccache -s
         du -hs ~/.cache/ccache
 
+  tests-oneapi-sycl-eb-nvidia:
+    name: oneAPI SYCL for Nvidia GPUs [tests w/ EB]
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Dependencies
+      run: |
+        .github/workflows/dependencies/dependencies_nvcc12.sh
+        .github/workflows/dependencies/dependencies_dpcpp.sh
+        .github/workflows/dependencies/dependencies_ccache.sh
+    - name: Set Up Cache
+      uses: actions/cache@v3
+      with:
+        path: ~/.cache/ccache
+        key: ccache-${{ github.workflow }}-${{ github.job }}-git-${{ github.sha }}
+        restore-keys: |
+             ccache-${{ github.workflow }}-${{ github.job }}-git-
+    - name: Build & Install
+      # mkl/rng/device/detail/mrg32k3a_impl.hpp has a number of sign-compare error
+      # mkl/rng/device/detail/mrg32k3a_impl.hpp has missing braces in array-array initalization
+      # clang currently supports CUDA up to version 11.5 and a warning is issued with newer versions
+      env: {CXXFLAGS: "-fsycl -fsycl-targets=nvptx64-nvidia-cuda -fno-operator-names -Werror -Wall -Wextra -Wpedantic -Wnull-dereference -Wfloat-conversion -Wshadow -Woverloaded-virtual -Wextra-semi -Wunreachable-code -Wnon-virtual-dtor -Wno-sign-compare -Wno-missing-braces -Wno-unknown-cuda-version"}
+      run: |
+        export CCACHE_COMPRESS=1
+        export CCACHE_COMPRESSLEVEL=10
+        export CCACHE_MAXSIZE=55M
+        export CCACHE_DEPEND=1
+        ccache -z
+
+        set +e
+        source /opt/intel/oneapi/setvars.sh --include-intel-llvm
+        set -e
+        cmake -S . -B build                                \
+            -DCMAKE_VERBOSE_MAKEFILE=ON                    \
+            -DAMReX_EB=ON                                  \
+            -DAMReX_ENABLE_TESTS=ON                        \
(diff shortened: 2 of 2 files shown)


WeiqunZhang:
>     2. Can we use `ubuntu-latest` instead of `ubuntu-20.04`?
Yes

>     3. How do you decide on the size of the ccache?
GitHub only provides 10 GB of cache space. So we are trying not to set the size too big. My rule of thumb is 1.5x of the size of a fresh run.

>     4. Can I compile with `AMReX_EB` and `AMReX_Fortran` both `ON` on a single job?  The existing SYCL workflow has two separate jobs, but I couldn't understand why?
Yes. There are no deep reasons other than A=ON and B=ON does not mean other combinations will always work. It's also impossible to test all possible combinations. So the choices are somewhat random.

>     5. Even though the default sub-group size is 32, I chose to explicitly pass it to CMake for clarity. Do you think this is sensible or should I just remove the explicit variable?
For CI, it does not matter. However, I think we should make the valid choices of sub-group size depend on the GPU target.

>     6. Is there any documentation tweaks that need to (or should) be made? Or would you rather keep this fairly hidden for the time being?
Let's not worry about it for now.

nmnobre:
> There are no deep reasons other than A=ON and B=ON does not mean other combinations will always work.

In the case of EB and Fortran though, is it not (approximately) the case that the compiled code is a superset of the code compiled when these features are disabled?

WeiqunZhang:
Maybe for Fortran, but certainly not the case for EB.

nmnobre:
> > 2. Can we use `ubuntu-latest` instead of `ubuntu-20.04`?
> 
> Yes

Done. See also #3354 for consistency.

> > 3. How do you decide on the size of the ccache?
>
> GitHub only provides 10 GB of cache space. So we are trying not to set the size too big. My rule of thumb is 1.5x of the size of a fresh run.

I've tried gauging the appropriate size by checking the size of `build/Src/CMakeFiles` at the end of a compilation on my local workstation for the new job as well as the old jobs (taking into account the max size you set for those). Hopefully it's fine, even though in this last commit, the hit rate is 0%.

> > 4. Can I compile with `AMReX_EB` and `AMReX_Fortran` both `ON` on a single job?  The existing SYCL workflow has two separate jobs, but I couldn't understand why?
> 
> Yes. There are no deep reasons other than A=ON and B=ON does not mean other combinations will always work. It's also impossible to test all possible combinations. So the choices are somewhat random.

Done, enabled both.

> > 5. Even though the default sub-group size is 32, I chose to explicitly pass it to CMake for clarity. Do you think this is sensible or should I just remove the explicit variable?
>
> For CI, it does not matter. However, I think we should make the valid choices of sub-group size depend on the GPU target.

I've removed the explicit CMake variable.
I agree with you on the necessity to constrain the valid sub-group sizes depending on the target.
I thought a bit about how to go about this, but I think it's best if we agree on an approach before I implement anything.
We could:
1) Change `AMReX_INTEL_ARCH` to `AMReX_SYCL_ARCH` (or replace `AMReX_CUDA_ARCH` with `AMReX_NVIDIA_ARCH`) and then, based on its value, add the appropriate compilation flags and constrain the sub-group sizes appropriately?
2) Do some sort of very simple regex check on `CXXFLAGS`, to see if it includes `-fsycl-targets=nvptx64-nvidia-cuda` and then define the possible sub-group sizes from there?

> > 6. Is there any documentation tweaks that need to (or should) be made? Or would you rather keep this fairly hidden for the time being?
> 
> Let's not worry about it for now.

Okay. :)

WeiqunZhang:
The size is about cache size, nothing to do with CMakeFiles. More specifically, `~/.cache/ccache` is being stored in Github cache.

As for the zero hit rate, maybe ccache does not work in this case.

Yes, we need to think about the CMake flags. We don't need to do anything right now. Maybe we should keep AMReX_INTEL_ARCH and add AMReX_SYCL_TARGET?

nmnobre:
> The size is about cache size, nothing to do with CMakeFiles. More specifically, `~/.cache/ccache` is being stored in Github cache.

Oh, perhaps I didn't explain myself properly. `build/Src/CMakeFiles/amrex_3d.dir` is where CMake stores the object files as it compiles AMReX. I was using this to estimate (perhaps an upper bound on) the size of the cache. I would think this should be related? At least the sizes on my machine are very close to the cache sizes you chose. :)
But I understand now you were rather suggesting looking at the output of `du -hs ~/.cache/ccache`, yes I also looked at that. It fluctuates a bit, but I think the 55MB I wrote should work.

> As for the zero hit rate, maybe ccache does not work in this case.
> 

Thanks for rerunning the workflow. It actually makes a lot of sense we were seeing a 0% hit rate, it was the first run with EB enabled. :P Now, even with the cache it's still taking 100min... so let me know if you'd like me to disable EB again to make it around 40min instead.

> Yes, we need to think about the CMake flags. We don't need to do anything right now. Maybe we should keep AMReX_INTEL_ARCH and add AMReX_SYCL_TARGET?

Yes, when I said "replace", I probably should've said "add" with the view to perhaps replace in the future if need be, I understand the fear of breaking the workflow of so many people who work with AMReX.
Currently, we have `AMReX_GPU_BACKEND=CUDA|HIP|SYCL` and we define `AMReX_CUDA_ARCH` but then also `AMReX_AMD_ARCH` (and not `AMReX_HIP_ARCH`) and `AMReX_INTEL_ARCH` (and not `AMReX_SYCL_ARCH`).
I find this slightly counterintuitive because, of the three, the one I actually see the most tied to a vendor is CUDA, but it's for the others we use the vendor name to specify the architecture.

I don't know if you noticed, but in #3341 I was compiling for the A100 (sm_80) just because when using the AMD/HIP backend DPC++ forces you to specify an architecture and so, for consistency, I did the same with the Nvidia/CUDA path. Here, I'm not specifying the architecture, and by default DPC++ will compile for sm_50 (old Maxwell cards).
So there's probably an argument that, for now, we could simply add `AMReX_SYCL_TARGET` for AOT device compilations as you suggest, the default being SPIR-V, a second valid option being NVPTX/CUDA. But, eventually, we might need `AMReX_SYCL_ARCH` or was your idea to simply use `AMReX_CUDA_ARCH` once we know `AMReX_SYCL_TARGET=CUDA`? That could also work. :)

WeiqunZhang:
Close this because of the token.

PR #3354, opened 2023-06-07
CI: Tweak Intel compiler workflow

nmnobre:


WeiqunZhang:
I don't think we need to worry about the Intel Classic Compiler. According to the compiler's warning message, it will be removed in the second half of 2023. At that time, we will switch to icpx.

nmnobre:
> I don't think we need to worry about the Intel Classic Compiler. According to the compiler's warning message, it will be removed in the second half of 2023. At that time, we will switch to icpx.

You are right, I shouldn't have touched `dependencies_dpcpp.sh`.
I suppose that was the issue?
We could forget and then once it was removed, the package installation could start failing?
I've avoided that, and so now I just literally avoid the redundancy and use the same syntax/options as in the other jobs, the changes are now self-contained. :)

.github/workflows/intel.yml
```diff
@@ -124,27 +116,27 @@ jobs:
...
         restore-keys: |
              ccache-${{ github.workflow }}-${{ github.job }}-git-
-    - name: build
+    - name: Build & Install
       env: {CXXFLAGS: "-Werror"}
       run: |
         export CCACHE_COMPRESS=1
         export CCACHE_COMPRESSLEVEL=10
         export CCACHE_MAXSIZE=40M
+        export CCACHE_DEPEND=1
```
WeiqunZhang: Usually, CCACHE_DEPEND=1 should not be used. We use it in the sycl builds as a workaround because otherwise `icpx -fsycl` does not work with ccache.

nmnobre: Done.

WeiqunZhang (approved):

WeiqunZhang:
There are conflicts must be resolved.

WeiqunZhang (approved):

PR #3356, opened 2023-06-07
CI: Linux Mold Linker

ax3l:
## Summary

Check if using the [mold linker](https://github.com/rui314/mold) can speed up our Linux [linking in CI](https://github.com/rui314/setup-mold).

## Checklist

The proposed changes:


ax3l:
Discussed on Slack:
- It does not seem to be faster using mold.
- We are concerned about not testing the standard linkers. Mold may have bugs. And we may have issues with the standard linkers that we want to cover in CI.

ax3l:
cc @jrood-nrel fyi

PR #3357, opened 2023-06-07
CMake: AMReX_TEST_TYPE

WeiqunZhang:
Add a new option, AMReX_TEST_TYPE=[All|Small]. Be default, all tests are built when AMReX_ENABLE_TEST is on. When the type is small, only four tests are enabled. If a CI job takes too long to compile and run all the tests, we could switch to the small test type.

Re-enable tests and change the test type to Small for the nvhpc CI. We had to disable it due to space limit in. It should no longer be a problem for the small set of tests.

### DIFF
--- a/.github/workflows/cuda.yml
+++ b/.github/workflows/cuda.yml
@@ -149,12 +149,12 @@ jobs:
         nvfortran --version
         cmake --version
 
-        # Disable Fortran and Tests due to space limit
         cmake -S . -B build                              \
             -DCMAKE_VERBOSE_MAKEFILE=ON                  \
-            -DAMReX_ENABLE_TESTS=OFF                     \
+            -DAMReX_ENABLE_TESTS=ON                      \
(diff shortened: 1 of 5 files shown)


.github/workflows/cuda.yml
```diff
@@ -148,11 +148,12 @@ jobs:
...
         cmake --version
 
-        # Disable Fortran and Tests due to space limit
         cmake -S . -B build                              \
             -DCMAKE_VERBOSE_MAKEFILE=ON                  \
-            -DAMReX_ENABLE_TESTS=OFF                     \
-            -DAMReX_FORTRAN=OFF                          \
+            -DAMReX_ENABLE_TESTS=ON                      \
+            -DAMReX_TEST_TYPE=Small                      \
+            -DAMReX_PARTICLES=ON                         \
```
WeiqunZhang:
```suggestion
```

atmyers (approved):

PR #3360, opened 2023-06-08
Add workflow for oneAPI SYCL on Nvidia GPUs

WeiqunZhang:
The Codeplay token used in this workflow is saved as a secret. Because of how GitHub secrets work, the workflow will only run when a PR is merged, not when it's trigger by a PR from an outside fork.

### DIFF
--- a/.github/workflows/dependencies/dependencies_codeplay.sh
+++ b/.github/workflows/dependencies/dependencies_codeplay.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+#
+# Copyright 2020 The AMReX Community
+#
+# License: BSD-3-Clause-LBNL
+
+set -eu -o pipefail
+
+curl -o oneapi_nvidia.sh -LJO "https://developer.codeplay.com/api/v1/products/download?product=oneapi&variant=nvidia&filters[]=linux&aat=$1"
+chmod +x oneapi_nvidia.sh
+sudo ./oneapi_nvidia.sh --yes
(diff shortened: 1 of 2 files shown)


.github/workflows/dependencies/dependencies_codeplay.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+#
+# Copyright 2020 The AMReX Community
+#
+# License: BSD-3-Clause-LBNL
+
+set -eu -o pipefail
+
+curl -o oneapi_nvidia.sh -LJO "https://developer.codeplay.com/api/v1/products/download?product=oneapi&variant=nvidia&filters[]=linux&aat=$1"
```
nmnobre: Options `-O` and `-J` ask the output file name to be given by the remote server and shouldn't be specified when using `-o`. On my system, this syntax issues the following warning: `Warning: Got more output options than URLs`,  so I'd recommend removing those two options. :)

WeiqunZhang: I don't see any warnings on Ubuntu 20.04. But I trust you.

WeiqunZhang:
```suggestion
curl -o oneapi_nvidia.sh -L "https://developer.codeplay.com/api/v1/products/download?product=oneapi&variant=nvidia&filters[]=linux&aat=$1"
```

WeiqunZhang:
This is a rebirth of #3351.

.github/workflows/dependencies/dependencies_codeplay.sh
nmnobre: I think it's only available on curl 7.81.0 or newer :)
https://github.com/curl/curl/commit/ee2ca5826bea7c061f3525cfc6b542c35d41fa50

atmyers (approved):

PR #3361, opened 2023-06-10
Coverity Scan

WeiqunZhang:
## Summary

Fix issues found by Coverity Scan, including infinite loop bugs in the kernels of MLALaplacian.

The scan will be done nightly on a local workstation.

Add coverity scan and license badges to README.md. Remove the cmake README.md because it's broken.

## Checklist

The proposed changes:

### DIFF
--- a/Tools/GNUMake/Make.defs
+++ b/Tools/GNUMake/Make.defs
@@ -991,6 +991,10 @@ ifeq ($(USE_CLANG_TIDY),TRUE)
   include        $(AMREX_HOME)/Tools/GNUMake/tools/Make.clang-tidy
 endif
 
+ifeq ($(USE_COVERITY),TRUE)
+  DEFINES += -DAMREX_USE_COVERITY
+endif
+
 # place holder
 F90CACHE =
 
--- a/README.md
+++ b/README.md
@@ -8,7 +8,9 @@
   <a href="https://doi.org/10.5281/zenodo.2555438">
   <img src="https://zenodo.org/badge/DOI/10.5281/zenodo.2555438.svg" alt="DOI">
   </a>
-  <img src="https://github.com/AMReX-codes/amrex/workflows/cmake/badge.svg?branch=development" alt="CI: CMake on Development">
+  <a href="https://scan.coverity.com/projects/amrex-codes-amrex">
+  <img alt="Coverity Scan Build Status" src="https://scan.coverity.com/projects/28563/badge.svg"/>
+  </a>
 </p>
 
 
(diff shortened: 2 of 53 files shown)


Tools/GNUMake/Make.defs
```diff
@@ -991,6 +991,10 @@ ifeq ($(USE_CLANG_TIDY),TRUE)
   include        $(AMREX_HOME)/Tools/GNUMake/tools/Make.clang-tidy
 endif
 
+ifeq ($(USE_COVERITY),TRUE)
+  DEFINES += -DAMREX_USE_COVERITY
+endif
```
ax3l, line 996: We can add  this also to CMake, but if your build logic on Coverity is GNUmake only then this is fine, too.

ax3l (approved):
Looks great :+1:

ax3l:
@WeiqunZhang just for sake of documentation, where did you implement the nightly sync-to-the-coverity-branch logic?

WeiqunZhang:
I have a cron job running on this. We can only run 3 scans per day on average.
```
export PATH=$HOME/opt/cov-analysis/bin:$PATH

cd $HOME/coverity/amrex
git pull

rm -rf cov-int amrex.tgz

make distclean

./configure --with-mpi yes \
            --with-omp no \
            --with-fortran no \
            --enable-fortran-api no \
            --enable-eb yes \
            --enable-linear-solver yes \
            --enable-particle yes \
            --enable-tiny-profile yes

cov-build --dir cov-int make -j4 CXX=g++ USE_ASSERTION=TRUE USE_COVERITY=TRUE

tar cvfz amrex.tgz cov-int

AMREX_VERSION=$(git describe --abbrev=12 --dirty --always --tags)

curl --form token=${TOKEN} \
  --form email=weiqunzhang@lbl.gov \
  --form file=@amrex.tgz \
  --form version="AMReX-${AMREX_VERSION}" \
  --form description="AMReX" \
  https://scan.coverity.com/builds?project=AMReX-Codes%2Famrex
```

ax3l:
Ah I see, got it.

In the past, I merged to a branch, which kicked off the upload via a plugin. Curling works as well.

If you like, you can also define cron-jobs in GH actions instead of locally:  https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule

WeiqunZhang:
Good suggestion. We could do it.

WeiqunZhang:
But then we have to download the coverity tools everytime.

WeiqunZhang:
The tarball is quite big. More than 1 GB.

ax3l:
Oh, I see. Yeah, it's on the size of the CUDA toolkit or oneAPI then :sweat_smile:

PR #3348, opened 2023-06-06
Add AMReX Apps to CI

WeiqunZhang:
Castro and WarpX are added in this commit. We can add more in the future.

This provides more coverage. The issues in #3342 could have been avoided.

PR #3365, opened 2023-06-13
CI: Simplify App Checkout

ax3l:
## Summary

Based on API in `actions/checkout`:
https://github.com/actions/checkout

## Additional background

Follow-up to #3348

## Checklist

The proposed changes:

### DIFF
--- a/.github/workflows/apps.yml
+++ b/.github/workflows/apps.yml
@@ -12,6 +12,14 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
+    - name: Download Castro
+      uses: actions/checkout@v3
+      repository: 'AMReX-Astro/Castro'
+      path: 'Castro'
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
Sure we can use checkout action. But this CI was intended to test the latest release of amrex apps not the latest development branch. So we still need to get the latest tag. Then we could use
```
    # The branch, tag or SHA to checkout. When checking out the repository that
    # triggered a workflow, this defaults to the reference or SHA for that event.
    # Otherwise, uses the default branch.
    ref: ''
```
It does not look like there is an easy way to get the latest relase tag. So we may still need to do something like
```
tag=$(wget https://github.com/AMReX-Astro/Castro/releases/latest 2>&1 | grep Location: | awk '{print $2}' | awk -F/ '{print $NF}')
```

WeiqunZhang (approved):

PR #3370, opened 2023-06-15
EB 2D: Fix levelset on nodes adjacent to covered edges

WeiqunZhang:
## Summary

This fixes inconsistency in 2D EB data. When an edge is covered, its edge length is set to zero. If the level set value on its nodes is not set to less than zero, the EB solver LEBNodeFDLaplacian will treat them as normal nodes and that will cause floating point issues.

Note that we have already been doing this kind of fix in 3D.

## Additional background

This issue was reported by @RemiLehe.

## Checklist

The proposed changes:


atmyers (approved):

RemiLehe:
Thanks a lot for this PR!
It did fix the issue that I was seeing.

PR #3382, opened 2023-06-22
GPU Assignment

WeiqunZhang:
When there are multiple GPU devices visible to an MPI process, we need to make a decision on which device to use. In the current development branch, we create a local sub-communicator using something like MPI_COMM_TYPE_SHARED and assign the devices in a round-robin way based on the local rank. This however does not always work. In this WarpX issue
(https://github.com/ECP-WarpX/WarpX/issues/3967), it was reported that all processes in a node used the same GPU because the sub-communicator had only one process in it. In this PR, we use MPI_Get_processor_name as an alternative way of obtaining the ranks within an actual node and use that for GPU assignment. If both approaches fail, we will fall back to use the rank in the global communicator modulo the number of devices.

In this PR, we have also removed AMREX_GPUS_PER_SOCKET and AMREX_GPUS_PER_NODE. They have not been used for some time now.


ax3l (approved):
Thank you, LGTM :+1:

PR #3384, opened 2023-06-23
GPU: Always deallocate gpu_rand_state in Finalize()

### LINKED ISSUE #3383: `gpu_rand_state` is not deallocated in `Finalize()` when using HIP
`DeallocateRandomSeedDevArray()` is only called when using CUDA (see [here](https://github.com/mirenradia/amrex/blob/239d4d89933488e1d8f6e39ceae211ab690aa855/Src/Base/AMReX.cpp#L696-L698)).

I think this should also happen in the HIP case.

mirenradia:
## Summary
This fixes #3383.

## Additional background
I noticed this problem when I tried to build an application with multiple `Initialize()` and `Finalize()` calls with HIP e.g. something like:
```cpp
amrex::Initialize(MPI_COMM_WORLD);
<some AMReX code>
amrex::Finalize();
amrex::Initialize(MPI_COMM_WORLD);
<some AMReX code>
amrex::Finalize();
```
I'm not sure if this officially supported but it works fine with CUDA (and with MPI so long as you only initialize and finalize this once outside of the above code) but not with HIP which fails at the second initialization [here](https://github.com/AMReX-Codes/amrex/blob/239d4d89933488e1d8f6e39ceae211ab690aa855/Src/Base/AMReX_Random.cpp#L36) with the error message
```
amrex::Abort::0::CArena::free: unknown pointer !!!
SIGABRT
```

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX.cpp
+++ b/Src/Base/AMReX.cpp
@@ -693,7 +693,7 @@ amrex::Finalize (amrex::AMReX* pamrex)
     BL_TINY_PROFILE_FINALIZE();
     BL_PROFILE_FINALIZE();
 
-#ifdef AMREX_USE_CUDA
+#ifdef AMREX_USE_GPU
     amrex::DeallocateRandomSeedDevArray();
 #endif
 


Src/Base/AMReX.cpp
```diff
@@ -693,7 +693,7 @@ amrex::Finalize (amrex::AMReX* pamrex)
     BL_TINY_PROFILE_FINALIZE();
     BL_PROFILE_FINALIZE();
 
-#ifdef AMREX_USE_CUDA
+#if defined(AMREX_USE_CUDA) || defined(AMREX_USE_HIP)
```
WeiqunZhang:
```suggestion
#if defined(AMREX_USE_GPU)
```
Let's change this to `AMREX_USE_GPU`.

WeiqunZhang (changes requested):

WeiqunZhang (approved):

PR #3389, opened 2023-06-28
Allow users to change the default vector growth stategy

atmyers:
This allows users to set at runtime an alternate to the default vector growth factor of 1.5. This could help save memory in simulations that use a lot of particles and are close to the capacity of the GPU device.

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_PODVector.cpp
+++ b/Src/Base/AMReX_PODVector.cpp
@@ -9,5 +9,13 @@ namespace amrex::VectorGrowthStrategy
     void Initialize () {
         ParmParse pp("amrex");
         pp.queryAdd("vector_growth_factor", growth_factor);
+
+        // sanity checks
+        auto eps = std::numeric_limits<Real>::epsilon();
+        auto huge = 1000._rt;  // huge enough for our purposes...
+        AMREX_ALWAYS_ASSERT_WITH_MESSAGE(growth_factor - 1.0_rt >= eps,
+            "User-specified vector growth factor is too small.");
+        AMREX_ALWAYS_ASSERT_WITH_MESSAGE(growth_factor < huge,
+            "User-specified vector growth factor is too large.");
     }
 }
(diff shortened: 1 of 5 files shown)


Src/Base/AMReX_PODVector.cpp
```diff
@@ -0,0 +1,13 @@
...
+#include <AMReX_ParmParse.H>
+#include <AMReX_REAL.H>
+
+namespace amrex::VectorGrowthStrategy
+{
+    Real growth_factor = 1.5_rt;
+
+    void Initialize () {
+        ParmParse pp("amrex");
+        pp.queryAdd("vector_growth_factor", growth_factor);
```
WeiqunZhang, line 11: We probably should check it's not less (1+epsilon) and not greater a huge number.

atmyers: Good idea.

Src/Base/AMReX_PODVector.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+        // clamp user input to reasonable values
+        constexpr Real min_factor = 1.05_rt;
+        constexpr Real max_factor = 4._rt;
+
+        if (growth_factor < min_factor) {
+            if (Verbose()) {
+                amrex::Print() << "Warning: user-provided vector growth factor is to small."
```
WeiqunZhang:
```suggestion
                amrex::Print() << "Warning: user-provided vector growth factor is too small."
```

Src/Base/AMReX_PODVector.cpp
```diff
@@ -0,0 +1,33 @@
...
+                               << " Clamping to " << min_factor << ". \n";
+            }
+            growth_factor = min_factor;
+        }
+
+        if (growth_factor > max_factor) {
+            if (Verbose()) {
+                amrex::Print() << "Warning: user-provided vector growth factor is to large."
```
WeiqunZhang:
```suggestion
                amrex::Print() << "Warning: user-provided vector growth factor is too large."
```

WeiqunZhang (approved):

PR #3391, opened 2023-06-29
Fix allocateSlice for EB

WeiqunZhang:
The slice data has a different BoxArray, so we cannot use the original Factory.

## Checklist

The proposed changes:


WeiqunZhang:
@dpgrote

WeiqunZhang:
@dpgrote reported that this fixed the issue he had.

asalmgren (approved):

PR #3394, opened 2023-06-30
Add function for setting growth factor

atmyers:
## Additional background
Add the ability to set the custom vector growth factor through a function, instead of only as a runtime input.

Also reducing the min allowed factor, as there seems to be demand for that.

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_PODVector.cpp
+++ b/Src/Base/AMReX_PODVector.cpp
@@ -6,28 +6,40 @@ namespace amrex::VectorGrowthStrategy
 {
     Real growth_factor = 1.5_rt;
 
+    // clamp user input to reasonable values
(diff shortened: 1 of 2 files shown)


Src/Base/AMReX_PODVector.cpp
```diff
@@ -6,28 +6,40 @@ namespace amrex::VectorGrowthStrategy
...
     }
+
+    void Initialize () {
+        ParmParse pp("amrex");
+        pp.queryAdd("vector_growth_factor", growth_factor);
+
+        detail::ValidateUserInput();
+    }
+
+    void SetGrowthFactor (Real a_factor) {
```
WeiqunZhang, line 41: Could we also have to a function to get the current growth factor?

WeiqunZhang (approved):

PR #3400, opened 2023-07-04
SYCL: Add table of GNU Make config vars to docs

mirenradia:
## Summary
~This PR allows the user to parallelize the AOT device compilation when building with the SYCL backend using GNU Make.~
This PR adds a table of GNU make config vars to the docs.

## Additional background
~When building with AOT device compilation on the SYCL backend, the device compilation takes place at the linking step and so is not currently parallelized (even when leveraging multiple GNU Make jobs e.g. by using the `-j` flag). It can therefore be quite slow. This PR introduces the `SYCL_MAX_PARALLEL_LINK_JOBS` makefile variable which is passed as the argument to the [`-fsycl-max-parallel-link-jobs`](https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/developer-guide-reference/2023-1/fsycl-max-parallel-link-jobs.html) flag of `icpx`. Although the Intel documentation does state~

> ~It is not guaranteed that spawning device link processes can be safely combined with the build system-level parallelization.~

~I think it's probably fine for the case of building an AMReX application where there is only 1 link step at the end after compiling source files.~

To accompany this change, I have added a table of GNU make variables that are relevant to SYCL.

~I guess it would make sense to add a similar option to the CMake build system but I am not very familiar with it.~

## Checklist

The proposed changes:

### DIFF
--- a/Docs/sphinx_documentation/source/GPU.rst
+++ b/Docs/sphinx_documentation/source/GPU.rst
@@ -187,6 +187,43 @@ can run it and that will generate results like:
    [The  Pinned Arena] space (MB): 8
    AMReX (19.06-404-g0455b168b69c-dirty) finalized
 
+SYCL configuration variables
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+When building with ``USE_SYCL=TRUE``, one can set the following makefile
+variables to configure the build
+
+.. raw:: latex
+
+   \begin{center}
+
+.. _tab:gnumakesyclvar:
+
+.. table:: AMReX SYCL-specific GNU Make build options
+
+   +------------------------------+-------------------------------------------------+-------------+-----------------+
+   | Variable Name                | Description                                     | Default     | Possible values |
+   +==============================+=================================================+=============+=================+
+   | SYCL_AOT                     | Enable SYCL ahead-of-time compilation           | FALSE       | TRUE, FALSE     |
+   +------------------------------+-------------------------------------------------+-------------+-----------------+
+   | SYCL_AOT_GRF_MODE            | Specify AOT register file mode                  | Default     | Default, Large, |
+   |                              |                                                 |             | AutoLarge       |
+   +------------------------------+-------------------------------------------------+-------------+-----------------+
+   | AMREX_INTEL_ARCH             | Specify target if AOT is enabled                | None        | pvc, etc.       |
+   +------------------------------+-------------------------------------------------+-------------+-----------------+
+   | SYCL_SPLIT_KERNEL            | Enable SYCL kernel splitting                    | FALSE       | TRUE, FALSE     |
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
Thanks! I will do follow-up on CMake.

Tools/GNUMake/comps/dpcpp.mak
```diff
@@ -146,6 +146,13 @@ ifeq ($(SYCL_AOT),TRUE)
       $(error SYCL_AOT_GRF_MODE ($(SYCL_AOT_GRF_MODE)) must be either Default, Large, or AutoLarge)
     endif
   endif
+  ifneq ($(SYCL_SPLIT_KERNEL),FALSE)
+    # There's no point in linking in parallel if the device code is not split
+    # sufficiently as it's limited by the number of device code modules
+    ifdef SYCL_MAX_PARALLEL_LINK_JOBS
+      LDFLAGS += -fsycl-max-parallel-link-jobs=$(SYCL_MAX_PARALLEL_LINK_JOBS)
+    endif
+  endif
```
WeiqunZhang:
```suggestion
```
The change here has appeared in #3498.

Docs/sphinx_documentation/source/GPU.rst
```diff
@@ -186,6 +186,43 @@ can run it and that will generate results like:
...
+   +------------------------------+-------------------------------------------------+-------------+-----------------+
+   | AMREX_INTEL_ARCH             | Specify target if AOT is enabled                | None        | pvc, etc.       |
+   +------------------------------+-------------------------------------------------+-------------+-----------------+
+   | SYCL_SPLIT_KERNEL            | Enable SYCL kernel splitting                    | FALSE       | TRUE, FALSE     |
+   +------------------------------+-------------------------------------------------+-------------+-----------------+
+   | USE_ONEDPL                   | Enable SYCL's oneDPL algorithms                 | NO          | YES, NO         |
+   +------------------------------+-------------------------------------------------+-------------+-----------------+
+   | SYCL_SUB_GROUP_SIZE          | Specify subgroup size                           | 32          | 64, 32, 16      |
+   +------------------------------+-------------------------------------------------+-------------+-----------------+
+   | SYCL_MAX_PARALLEL_LINK_JOBS  | Number of parallel AOT device compilations      | 1           | 1, 2, 3, etc.   |
```
```suggestion
   | SYCL_MAX_PARALLEL_LINK_JOBS  | Number of parallel jobs in device link          | 1           | 1, 2, 3, etc.   |
```

Tools/GNUMake/comps/dpcpp.mak
mirenradia: Oh ok.

mirenradia:
I made the changes you requested and updated the name and original description of the PR.

WeiqunZhang (approved):

PR #3411, opened 2023-07-07
Fix Fortran module installation

### LINKED ISSUE #3410: Fortran modules missing
No .mod files are installed.

Sha `bacaa10a76366ed40b90b422cf5c7da43565281a` had it. The current version doesn't work.

I'm specifically interested in amrex_flash_fluxregister_module.mod @houjun

WeiqunZhang:
It was broken in #3309, in which multi-D support was added.


WeiqunZhang:
This works if only one value for dimension is specified. It's still a problem if multiple dimension values are specified. Maybe we should add a check making sure there is only one dimension value when Fortran is ON. @ax3l

asalmgren (approved):

PR #3414, opened 2023-07-09
New feature in EB: Add regular coarse levels

WeiqunZhang:
## Summary

This feature could be useful when the EB is so thin that it is only resolved at the finest level and the usual way of generate coarse levels fails with multi-cut or multi-value cells. When the EB is fully refined to the finest level and there is no subcycling, no EB information is really needed at coarse levels. ~~Note that tagging cut cells for refinement no longer works if that level has all regular EB.~~ amrex::TagCutCells has been updated to work with the new feature.

## Checklist

The proposed changes:

### DIFF
--- a/Src/EB/AMReX_EB2.H
+++ b/Src/EB/AMReX_EB2.H
@@ -56,6 +56,7 @@ public:
     [[nodiscard]] virtual const Geometry& getGeometry (const Box& domain) const = 0;
     [[nodiscard]] virtual const Box& coarsestDomain () const = 0;
     virtual void addFineLevels (int num_new_fine_levels) = 0;
+    virtual void addRegularCoarseLevels (int num_new_coarse_levels) = 0;
 
 protected:
     static AMREX_EXPORT Vector<std::unique_ptr<IndexSpace> > m_instance;
@@ -88,6 +89,7 @@ public:
         return m_geom.back().Domain();
     }
     void addFineLevels (int num_new_fine_levels) final;
+    void addRegularCoarseLevels (int num_new_coarse_levels) final;
 
     using F = typename G::FunctionType;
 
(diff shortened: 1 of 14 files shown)


drummerdoc (changes requested):
I think this means that we need to implement some other approach to keep the fine grid on the EB or we'll lose it on the first regrid.  Is there somewhere the capability to tag IntVects on the coarse level where there are cut cells on the fine?

WeiqunZhang:
Are you using `amrex::TagCutCells`? We can probably make it work for this case.

drummerdoc:
> Are you using `amrex::TagCutCells`? We can probably make it work for this case.

Yeah, using that.

WeiqunZhang:
amrex::TagCutCell has been updated. I tested with EB_CNS. Regrid appears to work.
```
diff --git a/Tests/EB_CNS/Source/CNS_init_eb2.cpp b/Tests/EB_CNS/Source/CNS_init_eb2.cpp
index c64663a8bb..a729aebedf 100644
--- a/Tests/EB_CNS/Source/CNS_init_eb2.cpp
+++ b/Tests/EB_CNS/Source/CNS_init_eb2.cpp
@@ -69,6 +69,8 @@ initialize_EB2 (const Geometry& geom, int /*required_coarsening_level*/,
     else
 #endif
     {
-        EB2::Build(geom, max_coarsening_level, max_coarsening_level, 4);
+//        EB2::Build(geom, max_coarsening_level, max_coarsening_level, 4);
+        EB2::Build(geom, 0, 0, 4);
+        EB2::addRegularCoarseLevels(max_coarsening_level);
     }
 }
```

drummerdoc (approved):
Excellent!  Thanks Weiqun.  This looks great.  Can't wait to try it out on a thin body problem!

WeiqunZhang:
Hopefully, the fully covered cells will not cause any issues for your code.

drummerdoc:
> Hopefully, the fully covered cells will not cause any issues for your code.

What's the concern?

WeiqunZhang:
Like averageDown triggers floating point exceptions.

PR #3415, opened 2023-07-10
generalize interpolaters to not need or use extra values in a directi…

asalmgren:
…on for which the refinement ratio is 1

## Summary
The previous version of the interpolation routines in AmrCore would grow the "CoarseBox" in all directions even if one of the directions had refinement ratio 1.  This PR changes that behavior so the CoarseBox is not grown, and slopes are not computed, in any direction for which the refinement ratio is 1.

## Checklist

The proposed changes:
- [X ] add new capabilities to AMReX


WeiqunZhang:
We may need to update all the interpolation kernels including `do_linear_limiting=true` and RZ.

WeiqunZhang (approved):

PR #3427, opened 2023-07-18
Memory Efficient AoS Particle Sorting

WeiqunZhang:
Because of the size of AoS particle, creating a temporary vector for all the AoS data during reording might use too much memory. In fact, this is usually the memory usage bottleneck in WarpX simulations. In this new approach, we divide the AoS into smaller chunks and reorder them one at a time, thus significantly reducing the peak memory usage.

Thanks @AlexanderSinn for the suggestion.

### DIFF
--- a/Src/Particle/AMReX_ParticleContainerI.H
+++ b/Src/Particle/AMReX_ParticleContainerI.H
@@ -1097,31 +1097,41 @@ ParticleContainer_impl<ParticleType, NArrayReal, NArrayInt, Allocator>
 
     if (memEfficientSort) {
         if constexpr(!ParticleType::is_soa_particle) {
-            ParticleVector tmp_particles(np_total);
-            auto src = ptile.getParticleTileData();
-            ParticleType* dst = tmp_particles.data();
-
-            AMREX_HOST_DEVICE_FOR_1D( np_total, i,
-            {
-                dst[i] = i < np ? src.m_aos[permutations[i]] : src.m_aos[i];
-            });
-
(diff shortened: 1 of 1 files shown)


AlexanderSinn:
Interesting use of `tmp_t`. I guess if this keeps on using too much memory in the future, we will eventually be reordering one byte or one bit at a time.

Src/Particle/AMReX_ParticleContainerI.H
```diff
@@ -1097,31 +1097,41 @@ ParticleContainer_impl<ParticleType, NArrayReal, NArrayInt, Allocator>
...
-            AMREX_HOST_DEVICE_FOR_1D( np_total, i,
-            {
-                dst[i] = i < np ? src.m_aos[permutations[i]] : src.m_aos[i];
-            });
-
+            static_assert(sizeof(ParticleType)%4 == 0 && sizeof(uint32_t) == 4);
+            using tmp_t = std::conditional_t<sizeof(ParticleType)%8 == 0,
+                                             uint64_t, uint32_t>;
+            constexpr std::size_t nchunks = sizeof(ParticleType) / sizeof(tmp_t);
+            Gpu::DeviceVector<tmp_t> tmp(np);
```
WeiqunZhang, line 1104: There is a bug here, the size is too small.

WeiqunZhang: nvm

atmyers: Looks right to me.

atmyers (approved):

PR #3429, opened 2023-07-19
SYCL RelWithDebInfo Build: Add -g1

WeiqunZhang:
Add `-g1` to RelWithDebInfo build with oneAPI SYCL compiler (i.e., IntelLLVM). This is because with `-g` the link step can be very slow and the final executable can be order of magnitude bigger than that generated with `-g1`.

This actually passes `-g -g1` to the compiler. The first flag `-g` comes from CMake with the RelWithDebInfo build type. Here, the order of the two flags matter, because the latter flag `-g1` overrides `-g`.


ax3l (approved):
Thank you!

PR #3438, opened 2023-07-21
Disable Managed Memory for The_Arena by default.

WeiqunZhang:
It used to be that The_Arena was managed for CUDA and SYCL, but not for HIP. The users can still turn it on with
`amrex.the_arena_is_managed=1`. They can also use The_Managed_Arena explicitly.


ax3l (approved):
:shipit:

PR #3440, opened 2023-07-21
ParmParse::addfile needs Init

ax3l:
## Summary

`ParmParse` can be used before AMReX is initialized for most functionality, which is greatly helpful for us. One function that cannot be used before init is reading from a file, because it needs an MPI context to broadcast.

Add a clean error message instead of a segfault.

## Additional background

cc @ezoni @RevathiJambunathan 

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_ParmParse.cpp
+++ b/Src/Base/AMReX_ParmParse.cpp
@@ -17,6 +17,7 @@
 #include <list>
 #include <numeric>
 #include <regex>
+#include <stdexcept>
 #include <set>
 #include <sstream>
 #include <string>
@@ -538,6 +539,13 @@ read_file (const char* fname, std::list<ParmParse::PP_entry>& tab)
     //
     if ( fname != nullptr && fname[0] != 0 )
     {
+#ifdef AMREX_USE_MPI
+        if (ParallelDescriptor::Communicator() == MPI_COMM_NULL)
+        {
+            throw std::runtime_error("read_file: AMReX must be initialized");
+        }
+#endif
+
         Vector<char> fileCharPtr;
         std::string filename = fname;
         ParallelDescriptor::ReadAndBcastFile(filename, fileCharPtr);
@@ -1069,6 +1077,13 @@ ParmParse::prefixedName (const std::string& str) const
 
 void
 ParmParse::addfile (std::string const& filename) {
+#ifdef AMREX_USE_MPI
+    if (ParallelDescriptor::Communicator() == MPI_COMM_NULL)
+    {
+        throw std::runtime_error("ParmParse::addfile: AMReX must be initialized");
+    }
+#endif
+
     auto l = std::list<std::string>{filename};
     auto file = FileKeyword;
     addDefn(file,


Src/Base/AMReX_ParmParse.cpp
```diff
@@ -538,6 +538,11 @@ read_file (const char* fname, std::list<ParmParse::PP_entry>& tab)
...
     if ( fname != nullptr && fname[0] != 0 )
     {
+#ifdef AMREX_USE_MPI
+        if (!amrex::Initialized())
+            amrex::Error("read_file: AMReX must be initialized");
+#endif
+
         Vector<char> fileCharPtr;
         std::string filename = fname;
         ParallelDescriptor::ReadAndBcastFile(filename, fileCharPtr);
```
ax3l, line 551: The `ParallelDescriptor::ReadAndBcastFile` with MPI needs `amrex::Initialized()` but should work if build w/o MPI.

WeiqunZhang:
This does not work because amrex::Initialized() does not work during amrex::Initialize(). I think the way to do it is
```
#ifdef AMREX_USE_MPI
    AMREX_ALWAYS_ASSERT_WITH_MESSAGE(ParallelDescriptor::Communicator() != MPI_COMM_NULL, "error message goes here");
#endif
```

ax3l:
Agreed on the solution - but isn't the issue due to `amrex::Error` not working if AMReX is not initialized?

Src/Base/AMReX_ParmParse.cpp
```diff
@@ -538,6 +538,11 @@ read_file (const char* fname, std::list<ParmParse::PP_entry>& tab)
     //
     if ( fname != nullptr && fname[0] != 0 )
     {
+#ifdef AMREX_USE_MPI
+        if (!amrex::Initialized())
+            amrex::Error("read_file: AMReX must be initialized");
```
ax3l:
```suggestion
        AMREX_ALWAYS_ASSERT_WITH_MESSAGE(amrex::Initialized(), "read_file: AMReX must be initialized");
```

Src/Base/AMReX_ParmParse.cpp
```diff
@@ -1068,6 +1073,11 @@ ParmParse::prefixedName (const std::string& str) const
 
 void
 ParmParse::addfile (std::string const& filename) {
+#ifdef AMREX_USE_MPI
+    if (!amrex::Initialized())
+        amrex::Error("ParmParse::addfile: AMReX must be initialized");
```
ax3l:
```suggestion
        AMREX_ALWAYS_ASSERT_WITH_MESSAGE(amrex::Initialized(), "ParmParse::addfile: AMReX must be initialized");
```

Src/Base/AMReX_ParmParse.cpp
```diff
@@ -1068,6 +1072,10 @@ ParmParse::prefixedName (const std::string& str) const
 
 void
 ParmParse::addfile (std::string const& filename) {
+#ifdef AMREX_USE_MPI
+        AMREX_ALWAYS_ASSERT_WITH_MESSAGE(amrex::Initialized(), "ParmParse::addfile: AMReX must be initialized");
```
ax3l:
```suggestion
    AMREX_ALWAYS_ASSERT_WITH_MESSAGE(amrex::Initialized(), "ParmParse::addfile: AMReX must be initialized");
```

Src/Base/AMReX_ParmParse.cpp
```diff
@@ -538,6 +538,10 @@ read_file (const char* fname, std::list<ParmParse::PP_entry>& tab)
     //
     if ( fname != nullptr && fname[0] != 0 )
     {
+#ifdef AMREX_USE_MPI
+        AMREX_ALWAYS_ASSERT_WITH_MESSAGE(amrex::Initialized(), "read_file: AMReX must be initialized");
```
ax3l: if CI fails:
```suggestion
        AMREX_ALWAYS_ASSERT_WITH_MESSAGE(ParallelDescriptor::Communicator() != MPI_COMM_NULL, "read_file: AMReX must be initialized");
```

Src/Base/AMReX_ParmParse.cpp
```diff
@@ -1068,6 +1072,10 @@ ParmParse::prefixedName (const std::string& str) const
 
 void
 ParmParse::addfile (std::string const& filename) {
+#ifdef AMREX_USE_MPI
+    AMREX_ALWAYS_ASSERT_WITH_MESSAGE(amrex::Initialized(), "ParmParse::addfile: AMReX must be initialized");
```
ax3l: if CI fails:
```suggestion
    AMREX_ALWAYS_ASSERT_WITH_MESSAGE(ParallelDescriptor::Communicator() != MPI_COMM_NULL, "ParmParse::addfile: AMReX must be initialized");
```

WeiqunZhang:
That's right. So there are two issues. We cannot call `amrex::Initialized()` and `ParallelDescriptor::Abort()` here. So maybe we can do
```
#ifdef AMREX_USE_MPI
    if (ParallelDescriptor::Communicator() == MPI_COMM_NULL) {
        std::cerr << "......" << std::endl;
        std::abort();
    }
#endif

ax3l:
Yes, but I would rather throw an exception.
This error is not fatal and one can recover in interactive contexts if we throw.

WeiqunZhang (approved):

PR #3446, opened 2023-07-24
PlotFileUtil: Add Direct Includes

ax3l:
## Summary

Adds the used direct includes to `AMReX_PlotFileUtil.H`.

## Additional background

https://github.com/AMReX-Codes/pyamrex/pull/153

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_PlotFileUtil.H
+++ b/Src/Base/AMReX_PlotFileUtil.H
@@ -2,14 +2,20 @@
 #define AMREX_PLOTFILEUTIL_H_
 #include <AMReX_Config.H>
 
+#include <AMReX_Array.H>
+#include <AMReX_Box.H>
 #include <AMReX_Geometry.H>
+#include <AMReX_IntVect.H>
 #include <AMReX_MultiFab.H>
 #include <AMReX_PlotFileDataImpl.H>
+#include <AMReX_REAL.H>
+#include <AMReX_Vector.H>
 
 #ifdef AMREX_USE_HDF5
 #include <AMReX_PlotFileUtilHDF5.H>
 #endif
 
+#include <ostream>
 #include <string>
 #include <memory>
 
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
Correctness here is subjective. I think all those headers have been indirectly included. Are you seeing real issues?

ax3l:
Correctness in the sense of include stability:
This will make it long-term stable if transitive headers change.
https://include-what-you-use.org

Changing the label appropriately to `cleaning`.

WeiqunZhang:
I tried include-what-you-use in the past. It is too much work to make it happy because we have a lot of build options and we need to add a lot of IWYU pragmas. If we were using it, we would have marked AMReX_MultiFab.H as exporting AMReX_Box.H, AMReX_REAL.H etc.

Src/Base/AMReX_PlotFileUtil.H
```diff
@@ -2,14 +2,20 @@
...
 #include <AMReX_Config.H>
 
+#include <AMReX_Array.H>
+#include <AMReX_Box.H>
 #include <AMReX_Geometry.H>
+#include <AMReX_IntVect.H>
 #include <AMReX_MultiFab.H>
 #include <AMReX_PlotFileDataImpl.H>
+#include <AMReX_REAL.H>
+#include <AMReX_Vector.H>
```
WeiqunZhang:
```suggestion
#include <AMReX_Array.H>
#include <AMReX_Geometry.H>
#include <AMReX_MultiFab.H>
#include <AMReX_PlotFileDataImpl.H>
#include <AMReX_Vector.H>
```
It think it's safe to assume they will be "exported" by the MultiFab header.

WeiqunZhang: To be more specific, "they" means Box, IntVect, and Real.

ax3l: Ok :+1:

ax3l:
> I tried include-what-you-use in the past.

Agreed (for the tool). I was more referring to the concept :)

WeiqunZhang:
I agree with the concept too. :)

WeiqunZhang (approved):

PR #3447, opened 2023-07-25
Mac Arm64 Architecture fpe handling

johnson452:
## Summary
Handles floating pointing exception when compiled on arch64 and Apple (for M1 and M2 chips). Only works for overflow currently, there is an apparent limitation to these architectures for catching division by zero.

## Additional background
SIGILL instead of SIGFPE is needed to catch arch64 signals, see ref: https://developer.apple.com/forums/thread/689159

## Checklist

The proposed changes:

### DIFF
--- a/Docs/sphinx_documentation/source/Debugging.rst
+++ b/Docs/sphinx_documentation/source/Debugging.rst
@@ -27,7 +27,7 @@ use of uninitialized values, AMReX also initializes ``FArrayBox``\ s in
 ``MulitFab``\ s and arrays allocated by ``bl_allocate`` to signaling NaNs when it is compiled
 with ``TEST=TRUE`` or ``DEBUG=TRUE`` in GNU make, or with ``-DCMAKE_BUILD_TYPE=Debug`` in CMake.
 One can also control the setting for ``FArrayBox`` using the runtime parameter, ``fab.init_snan``.
-Note for Macs, M1 and M2 chips using Arm64 architecture are not able to trap division by zero.  
+Note for Macs, M1 and M2 chips using Arm64 architecture are not able to trap division by zero.
 
 One can get more information than the backtrace of the call stack by
 instrumenting the code.  Here is an example.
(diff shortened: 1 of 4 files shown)


WeiqunZhang:
We need to restore the previous signal handler for SIGILL in amrex::Finalize(). We could also update `BLBackTrace::handler` to print out a message about the type of signal.

Docs/sphinx_documentation/source/Debugging.rst
```diff
@@ -27,6 +27,7 @@ use of uninitialized values, AMReX also initializes ``FArrayBox``\ s in
 ``MulitFab``\ s and arrays allocated by ``bl_allocate`` to signaling NaNs when it is compiled
 with ``TEST=TRUE`` or ``DEBUG=TRUE`` in GNU make, or with ``-DCMAKE_BUILD_TYPE=Debug`` in CMake.
 One can also control the setting for ``FArrayBox`` using the runtime parameter, ``fab.init_snan``.
+Note for Macs, M1 and M2 chips using Arm64 architecture are not able to trap division by zero.
```
atmyers:
```suggestion
Note for Macs, M1 and M2 chips using Arm64 architecture are not able to trap division by zero.
```

That should placate the CI.

atmyers:
This worked for me on my M1 mac.

Docs/sphinx_documentation/source/Debugging.rst
johnson452: Ah, thank you

atmyers (approved):

Src/Base/AMReX.cpp
```diff
@@ -554,6 +558,16 @@ amrex::Initialize (int& argc, char**& argv, bool build_parm_parse,
...
+#elif defined(__APPLE__) && defined(__aarch64__)
+                fenv_t env;
+                fegetenv(&env);
+                if (invalid)   env.__fpcr |= __fpcr_trap_invalid;
+                if (divbyzero) env.__fpcr |= __fpcr_trap_divbyzero;
+                if (overflow)  env.__fpcr |= __fpcr_trap_overflow;
+                fesetenv(&env);
+                // SIGILL ref: https://developer.apple.com/forums/thread/689159
+                prev_handler_sigfpe = std::signal(SIGILL,  BLBackTrace::handler);
+
```
WeiqunZhang:
```suggestion
```
Don't think we want to call `prev_handler_* = std::singal(SIGILL,..)` twice, otherwise the previous handler will be our handler not the one before amrex::Initialize is called.

PR #3452, opened 2023-07-26
call_f refactor

atmyers:
Using `if constexpr` in these functions allows us to substantially reduce the number of overloads needed.

The proposed changes:


Src/Particle/AMReX_DenseBins.H
```diff
@@ -93,18 +93,12 @@ private:
...
-        return f(v,index);
-    }
-
-    template <typename F, typename I>
-    AMREX_GPU_HOST_DEVICE
-    static auto call_f(F const& f, const_pointer_input_type v, I& index)
-        noexcept -> decltype(f(v[index]))
-    {
-        return f(v[index]);
+    static auto call_f(F const& f, const_pointer_input_type v, I& index) {
```
atmyers:
```suggestion
    static auto call_f (F const& f, const_pointer_input_type v, I& index) {
```

WeiqunZhang (approved):

PR #3458, opened 2023-07-29
Fix New Coverity Scan Warning

WeiqunZhang:
Coverity Scan warned about logically dead code, because `index >= 0 && index < Nreal` can never be true if `NReal` is 0.


WeiqunZhang:
The change in this PR has appeared in #3476.

PR #3490, opened 2023-08-12
fgradient: A new plotfile tool for computing gradient

WeiqunZhang:


zingale:
this looks good to me and works as expected.

I just don't understand what this line is doing:

```
 196         Vector<std::string> args(std::istream_iterator<std::string>{is},                                                                                                
 197                                  std::istream_iterator<std::string>{  });
```

what's the need for the second string here?

zingale (approved):

WeiqunZhang:
It creates a vector of strings by splitting the original argument string by space. `std::istream_iterator<std::string>{is}` is like `begin` and `std::istream_iterator<std::string>{}` is like `end`. https://en.cppreference.com/w/cpp/iterator/istream_iterator
>The default-constructed std::istream_iterator is known as the end-of-stream iterator. When a valid std::istream_iterator reaches the end of the underlying stream, it becomes equal to the end-of-stream iterator.

So the code is similar to
```
std::vector<string> a_orig(....);
std::vector<string> a_copy(std::begin(a_orig),std::end(a_orig))
```

PR #3498, opened 2023-08-17
Update gpu elixir and async array synchronization for sycl

cbauinge:
## Summary

Updated the way the GpuElixir and GpuAsyncArray synchronize. Instead of a Gpu::streamSynchronize(); it now uses a host task by default since it was implemented in SYCL 2020. Removed the codeplay-extension version.

In addition, the option SYCL_PARALLEL_LINK_JOBS was added to dpcpp.mak. Setting it to a positive integer <n> then allows linking the SYCL code with <n> threads, speeding up the whole compilation process, especially when using AOT in SYCL.

## Checklist

The proposed changes:


WeiqunZhang:
@cbauinge Why the change from `-O3` to `-O2`?

WeiqunZhang:
Could you also tell us more about `-g1` vs. `-gline-tables-only -fdebug-info-for-profiling`?

cbauinge:
> @cbauinge Why the change from `-O3` to `-O2`?

I found an issue where the ahead of time compilation did not work with O3. It failed in the last linking step. Unfortunately, I am still trying to reproduce that on a smaller example. I still need to verify all the oneAPI version which suffer from this. For the most part, you can keep it to O3. Just be aware this could come up.

cbauinge:
> Could you also tell us more about `-g1` vs. `-gline-tables-only -fdebug-info-for-profiling`?

-gline-tables-only -fdebug-info-for-profiling is the default to enable GPU profiling (following the Vtune user guide here). I assumed that is what you want. But I wasn't sure. Please let me know if I should roll that back.

WeiqunZhang:
Thanks for the explanation! Let's roll back the `-O` change, but keep the debug info change. I will do some testing to see if that's enough debug info for backtracing on the host.

WeiqunZhang:
The debug info change is OK. The information is sufficient for our backtrace to show line numbers. The new flags produce a slightly bigger executable, but both the new flags and `-g1` produce executables that are order of magnitude smaller than `-g` does.

cbauinge:
> Thanks for the explanation! Let's roll back the `-O` change, but keep the debug info change. I will do some testing to see if that's enough debug info for backtracing on the host.

Done. Please let me know if there is anything else!

WeiqunZhang (approved):

PR #3500, opened 2023-08-18
FillRandom & FillRandomNormal

WeiqunZhang:
## Summary

Add new functions that fill a MultiFab or a 1d vector with random numbers from a uniform or normal distribution.

The functions use CUDA and HIP's host APIs for random number generations. Compared with our existing capability of using device APIs, the new functions are slightly (about 15%) faster on Nvidia GPUs, whereas they are 4x faster on the AMD GPU we tested. For Intel GPUs, we are still using the device APIs, because we have not figured out how to use its host APIs. More specially, we are missing some LDFLAGS at link time.

The functions work for CPU builds too, and for normal distribution they are about 2x faster than using `amrex::RandomNormal` because the latter only uses one of two random numbers generated by the Box-Muller transform.

The functions work for MultiFabs with any index types (e.g., nodal). However, if the numbers on shared nodes need to be consistent across boxes and MPI processes, it's the user's responsibility to call OverrideSync. (This gives the user a chance to cache the mask used by OverrideSync.) It's also the user's responsibility to fix data consistency between overlapping ghost and valid points, if that is needed.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_GpuError.H
+++ b/Src/Base/AMReX_GpuError.H
@@ -79,6 +79,11 @@ namespace Gpu {
                            + ": " + cudaGetErrorString(amrex_i_err)); \
         amrex::Abort(errStr); \
     }}
+
+#define AMREX_CURAND_SAFE_CALL(x) do { if((x)!=CURAND_STATUS_SUCCESS) { \
+    std::string errStr(std::string("CURAND error in file ") + __FILE__  \
+                       + " line " + std::to_string(__LINE__));          \
+    amrex::Abort(errStr); }} while(0)
 #endif
 
 #ifdef AMREX_USE_HIP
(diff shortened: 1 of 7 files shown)


WeiqunZhang:
@isriva

WeiqunZhang:
Unfortunately, there are some issues with the cuda packages in GitHub actions. However, my branch did pass the cuda CIs before the issue. https://github.com/WeiqunZhang/amrex/actions/runs/5844059557 All the changes since then were in the Intel part.

BenWibking:
The fact that the range of `FillRandom` is different on different devices is a bit wild. Is it possible to use `std::nexttoward` on device to make CUDA and HIP the same as CPU and SYCL?

kngott (approved):

WeiqunZhang:
In `amrex::Random()`, what we do is to return `1.0-curand_uniform_double()`. Here, we could launch another kernel doing that, but maybe the user does not care about it. Even if the user needs it, they could handle it more efficiently in their code when they use it without launching an extra kernel. So we probably should not pay the price unless it's really needed.

PR #3510, opened 2023-08-23
Simplify filterParticles Kernel

ax3l:
## Summary

On Summit, generation of this kernel shows compiler issues with `nvcc` 11.3.109: it compiles without warnings but leads to a segmentation fault at runtime.

The fix for the compiler bug is to implement the trivial lambda that is passed to `copyParticles` w/o capture: `[]`.

We also add explicit capture to a few other lambdas, to simplify compiler intake complexity.

~~First commit: wow, this reliably triggereds the compiler bug (invalid device function) in step 1.~~ ([see reason below](https://github.com/AMReX-Codes/amrex/pull/3510#issuecomment-1692773745))

This simplifies the kernel generation, which also solves the issue seen with WarpX [for this line](https://github.com/ECP-WarpX/WarpX/blob/43d2ac7b54546c87d2cb540df7a2b3cb57592b84/Source/Diagnostics/WarpXOpenPMD.cpp#L585).

## Checklist

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_ParticleContainerI.H
+++ b/Src/Particle/AMReX_ParticleContainerI.H
@@ -995,7 +995,7 @@ ParticleContainer_impl<ParticleType, NArrayReal, NArrayInt, Allocator>::
 copyParticles (const PCType& other, bool local)
 {
     using PData = ConstParticleTileData<typename ParticleType::StorageParticleType, NArrayReal, NArrayInt>;
-    copyParticles(other, [=] AMREX_GPU_HOST_DEVICE (const PData& /*data*/, int /*i*/) { return 1; }, local);
+    copyParticles(other, [] AMREX_GPU_HOST_DEVICE (const PData& /*data*/, int /*i*/) { return 1; }, local);
 }
 
 template <typename ParticleType, int NArrayReal, int NArrayInt,
@@ -1006,7 +1006,7 @@ ParticleContainer_impl<ParticleType, NArrayReal, NArrayInt, Allocator>::
 addParticles (const PCType& other, bool local)
 {
     using PData = ConstParticleTileData<typename ParticleType::StorageParticleType, NArrayReal, NArrayInt>;
-    addParticles(other, [=] AMREX_GPU_HOST_DEVICE (const PData& /*data*/, int /*i*/) { return 1; }, local);
+    addParticles(other, [] AMREX_GPU_HOST_DEVICE (const PData& /*data*/, int /*i*/) { return 1; }, local);
 }
 
 template <typename ParticleType, int NArrayReal, int NArrayInt,
--- a/Src/Particle/AMReX_ParticleTransformation.H
+++ b/Src/Particle/AMReX_ParticleTransformation.H
@@ -577,7 +577,7 @@ int filterAndTransformParticles (DstTile1& dst1, DstTile2& dst2, const SrcTile& 
     const auto src_data = src.getConstParticleTileData();
 
     amrex::ParallelForRNG(np,
-    [=] AMREX_GPU_DEVICE (int i, amrex::RandomEngine const& engine) noexcept
+    [p, p_mask, src_data] AMREX_GPU_DEVICE (int i, amrex::RandomEngine const& engine) noexcept
     {
         amrex::ignore_unused(p, p_mask, src_data, engine);
         if constexpr (IsCallable<Pred,decltype(src_data),int,RandomEngine>::value) {
@@ -620,7 +620,7 @@ Index filterAndTransformParticles (DstTile& dst, const SrcTile& src, Pred&& p, F
     const auto src_data = src.getConstParticleTileData();
 
     amrex::ParallelForRNG(np,
-    [=] AMREX_GPU_DEVICE (int i, amrex::RandomEngine const& engine) noexcept
+    [p, p_mask, src_data, src_start] AMREX_GPU_DEVICE (int i, amrex::RandomEngine const& engine) noexcept
     {
         amrex::ignore_unused(p, p_mask, src_data, src_start, engine);
         if constexpr (IsCallable<Pred,decltype(src_data),Index,RandomEngine>::value) {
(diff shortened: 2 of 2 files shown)


Src/Particle/AMReX_ParticleTransformation.H
```diff
@@ -403,16 +403,19 @@ Index filterParticles (DstTile& dst, const SrcTile& src, Pred&& p,
     auto* p_mask = mask.dataPtr();
     const auto src_data = src.getConstParticleTileData();
 
-    amrex::ParallelForRNG(n,
-    [=] AMREX_GPU_DEVICE (int i, amrex::RandomEngine const& engine) noexcept
-    {
-        amrex::ignore_unused(p, p_mask, src_data, src_start, engine);
-        if constexpr (IsCallable<Pred,decltype(src_data),Index,RandomEngine>::value) {
+    if constexpr (IsCallable<Pred,decltype(src_data),Index,RandomEngine>::value) {
```
AlexanderSinn:
```suggestion
    amrex::ParallelForRNG(n,
    [p, p_mask, src_data, src_start] AMREX_GPU_DEVICE (int i, amrex::RandomEngine const& engine) noexcept
    {
        amrex::ignore_unused(p, p_mask, src_data, src_start, engine);
        if constexpr (IsCallable<Pred,decltype(src_data),Index,RandomEngine>::value) {
            p_mask[i] = p(src_data, src_start+i, engine);
        } else {
            p_mask[i] = p(src_data, src_start+i);
        }
    });
```

Explicitly specifying the capture group has worked for me in the past. Putting the if constexpr outside of the kernel would result in more nvcc issues.

Src/Particle/AMReX_ParticleTransformation.H
```diff
@@ -413,6 +413,18 @@ Index filterParticles (DstTile& dst, const SrcTile& src, Pred&& p,
...
+        {
+            p_mask[i] = p(src_data, src_start+i, engine);
+        });
+    } else {
+        amrex::ParallelFor(n,
+        [=] AMREX_GPU_DEVICE (int i) noexcept
+        {
+            p_mask[i] = p(src_data, src_start+i);
+        });
+    }
```
ax3l: cleaning.
```suggestion
```

Src/Particle/AMReX_ParticleTransformation.H
WeiqunZhang: Using `amrex::ignore_unused` should have the same effect as explicit capture.

Some background. A WarpX run crashes on summit in this function. So we are trying to understand what's going on. I think it's most likely a compiler bug.  CMake plus `-O2` works. CMake plus `-O3` does not. GNU Make works with `-O3`.

WeiqunZhang: The same run also works on perlmutter. And we have run it on CPU with valgrind and fsanitizer.

WeiqunZhang:
For CUDA, `amrex::IsCallable` and `std::is_invocable` work in device code. They also work in host code if the type is a functor (even if its operator() is device only) or a host and device lambda (i.e., AMREX_GPU_HOST_DEVICE). Unfortunately, they do not work in host code on device lambda (i.e., AMREX_GPU_DEVICE). A device lambda has a type of `__nv_dl_wrapper_t` on the host. It seems that its operator() can take any types such that `amrex::IsCallable` and `std::is_invocable` are always true on them. `cuda::std::is_invocable` has the same limitation.

The limitation is CUDA only. HIP and SYCL work.

ax3l:
@jmsexton03 asked on slack
> [for amrex::ParallelForRNG] is there a reason to use AMREX_GPU_DEVICE rather than AMREX_GPU_HOST_DEVICE? in terms of consitency, is one of these a better pattern?


@WeiqunZhang was your comment in response to that or did I miss a connection?

ax3l:
I tried the latest version of this PR with the explicit capture by @AlexanderSinn .
Unfortunately this still fails on Summit with CUDA 11.3

WeiqunZhang:
I didn't know @jmsexton03's comment on slack. My comment was trying to explain why the first version of this draft PR does not work. It's because `if constexpr` is always true in that case.
> First commit: wow, this now reliably triggers the compiler bug (invalid device function) in step 1.

Src/Particle/AMReX_ParticleContainerI.H
```diff
@@ -995,7 +995,7 @@ ParticleContainer_impl<ParticleType, NArrayReal, NArrayInt, Allocator>::
 copyParticles (const PCType& other, bool local)
 {
     using PData = ConstParticleTileData<typename ParticleType::StorageParticleType, NArrayReal, NArrayInt>;
-    copyParticles(other, [=] AMREX_GPU_HOST_DEVICE (const PData& /*data*/, int /*i*/) { return 1; }, local);
+    copyParticles(other, [] AMREX_GPU_HOST_DEVICE (const PData& /*data*/, int /*i*/) { return 1; }, local);
```
ax3l, line 998: @WeiqunZhang should those be only `AMREX_GPU_DEVICE`?

WeiqunZhang: It should not matter. Both should work. If we can go back in time, we should provide AMREX_GPU_HOST_DEVICE only so that we can always detect extended lambda's function signature.

ax3l:
THIS FIXES IT!!! (lol)

Src/Particle/AMReX_ParticleContainerI.H
ax3l: not tried yet, but seems not needed

WeiqunZhang:
`[]` or `AMREX_GPU_DEVICE`?

ax3l:
no capture: `[]`

ax3l:
The `bug` label here stands for: "produces a bug with a third-party compiler" not "is an actual bug in our code".

WeiqunZhang (approved):

PR #3517, opened 2023-08-26
Empty level fix

### LINKED ISSUE #3516: Coarsenable assertion fails on empty level in AMREX_MLLinOp.H
When using the linear solver interface in amrex, I encounter an issue when the number of levels is reduced during the simulation due to no part of the domain being flagged.

This manifests as an assertion error which occurs in Src/LinearSolvers/MLMG/AMReX_MLLinOp.H which checks that the levels are coarsenable. This fails if the level is empty.

sw561:
## Summary

Closes #3516 . This works for me in practice, and is more convenient than filtering the levels on the caller's side.

## Additional background

Idea is to simply not consider the empty levels in the linear solver.

## Checklist

The proposed changes:


WeiqunZhang:
One is expected to use the following AmrMesh functions to get objects for the linear solver.
```
Vector<Geometry> Geom (int a_coarsest_lev, int a_finest_lev) const
Vector<BoxArray> boxArray (int a_coarsest_lev, int a_finest_lev) const
Vector<DistributionMapping> DistributionMap (int a_coarsest_lev, int a_finest_lev) const
```
And if the coarsest level for the solver is not AMR level 0, the versions without arguments that get the objects starting from AMR level 0 will always be wrong.

I am not sure if we should try to guess the user's intention or add an assertion making sure BoxArrays are not empty. Let me think about this.

WeiqunZhang (approved):

PR #3520, opened 2023-08-28
UniqueString tweak

WeiqunZhang:
# Summary

Previously UniqueString used MPI_Wtime, which could have a very low resolution (1e-5s) on some machines. In this PR, we switch to use amrex::second(), which uses the highest resolution steady clock (usually 1.e-9s). We have also changed a hardwired number according to resolution of the clock so that the digits are not wasted on trailing zeros.

## Additional background

https://github.com/ECP-WarpX/impactx/pull/424

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_Utility.cpp
+++ b/Src/Base/AMReX_Utility.cpp
@@ -183,10 +183,17 @@ amrex::FileExists(const std::string &filename)
 std::string
 amrex::UniqueString()
 {
-  std::stringstream tempstring;
-  tempstring << std::setprecision(11) << std::fixed << ParallelDescriptor::second();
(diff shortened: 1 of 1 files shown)


ax3l (approved):
Thanks! That improves the precision we saw on macOS w/ MPI :)

WeiqunZhang:
Somehow `wget` has stopped working. Maybe we should switch to using the checkout action. Will do it later.

ax3l:
I saw some generally connection splits on CI runners today, maybe that was a temporary fluke. I see still some issues on https://www.githubstatus.com/

But yes, `checkout` might have higher limits due to token usage.

PR #3522, opened 2023-08-30
Constness in buildInfo: num_modules

ax3l:
## Summary

Clang-Tidy:
```
warning: variable 'num_modules' of type 'int' can be declared 'const' [misc-const-correctness]
```

## Checklist

The proposed changes:

### DIFF
--- a/Tools/C_scripts/makebuildinfo_C.py
+++ b/Tools/C_scripts/makebuildinfo_C.py
@@ -110,7 +110,7 @@ const char* buildInfoGetAux(int i) {
 }
 
 int buildInfoGetNumModules() {
-  // int num_modules = X;
+  // int const num_modules = X;
   @@NUM_MODULES@@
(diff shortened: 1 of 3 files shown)


zingale (approved):

zingale:
I think we can change line 369 of `make_buildinfo_C.py` to add a `const` there too.

Tools/C_scripts/makebuildinfo_C.py
```diff
@@ -110,7 +110,7 @@
 }
 
 int buildInfoGetNumModules() {
-  // int num_modules = X;
+  // int const num_modules = X;
```
WeiqunZhang, line 113: @zingale's comment is correct. This is just a comment, not the real code.

ax3l:
Thanks! Overlooked that script in my grep 🙏

PR #3523, opened 2023-08-30
Add ability for GCC 8 in CMake to build fgradient which uses std::filesystem

jrood-nrel:
## Summary

Since GCC 8 is still a base compiler on several Linux distributions, I think it should probably continue to be supported. This allows `fgradient` to link to the std::filesystem library necessary for GCC < 9 and succeed in building.

## Checklist

The proposed changes:


ax3l (approved):
Thank you for this.
Brave! :D

PR #3535, opened 2023-09-08
Support multiple CUDA architectures at compilation

philip-blakely:
## Summary

This allows users to specify multiple (space-separated) values for CUDA_ARCH in Makefiles. This may be useful for users on heterogeneous networks who want to have a single executable that can run on multiple CUDA hardware types.

## Additional background

The changes here are done for Makefiles only, and have only been tested fully for the nvcc compiler. The correct compiler-command syntax has also been implemented for PGI and NVHPC compilers, but not tested (beyond manual checking of the command-line) because I don't have access to these.
One CI test is updated to compile for two different CUDA architectures simultaneously and passes. Trying this in another case resulted in a too-large executable for the available disk.

## Checklist

The proposed changes:


WeiqunZhang:
Unfortunately, the nvhpc test now runs out of disk space. Maybe we can test it with `configure-3d-cuda`, if that one works.

philip-blakely:
OK, I've tested it in configure-3d-cuda instead, and all the checks pass.

WeiqunZhang (approved):

PR #3545, opened 2023-09-15
ReduceOps: reset result readiness flag

WeiqunZhang:
In #3421, a change was made to avoid a common mistake of calling ReduceData::value multiple times. Unfortunately, it breaks codes that using one ReduceOps with multiple ReduceData objects. This PR fixes both cases.

## Checklist

The proposed changes:


WeiqunZhang:
https://github.com/AMReX-FHD/FHDeX/pull/141

jgalenwang:
Hi Weiqun,

It's working now on my test case. Now all three directions are getting non-zero rhs. Thank you.

ajnonaka (approved):

PR #3553, opened 2023-09-17
Add Fortran interface for average_down_faces

WeiqunZhang:


akashdhruv:
`fgeom` is missing

WeiqunZhang:
The fine level geometry is not needed. The data on the fine level is constant. On the coarse level we need to make sure that the data on periodic faces are consistent. Thus we need the coarse level geometry.

akashdhruv:
Divergence is still not zero at the boundary. I will add a figure for reference to #327

akashdhruv:
> The fine level geometry is not needed. The data on the fine level is constant. On the coarse level we need to make sure that the data on periodic faces are consistent. Thus we need the coarse level geometry.

Ok. Thank you.

WeiqunZhang:
It could be something else. But first, we need to make sure this function is being used. In the test you set up, it is not. Then we can print out the divergence before and after fillpatch to see if the divergence comes from fillpath or other parts of the code.

akashdhruv:
I have tested this PR on Flash-X side and the test suite is passing. The only issue is with divergence which needs to be investigated on my end. I will be performing the investigations here: 

https://github.com/Lab-Notebooks/AMReX-Divfree

However, it shouldn't hold this PR back. Waiting on this PR to merge another Flash-X PR.

akashdhruv:
@atmyers Can you let me know when you can merge this?

atmyers (approved):

PR #3547, opened 2023-09-15
Bittree sept 2023

akashdhruv:
## Checklist

The proposed changes:
- [ ] fix a bug or incorrect behavior in AMReX
- [x] add new capabilities to AMReX
- [ ] changes answers in the test suite to more than roundoff level
- [ ] are likely to significantly affect the results of downstream AMReX users
- [x] include documentation in the code and/or rst files, if appropriate

PR #3555, opened 2023-09-18
Adding bittree interface to improve regridding performance in octree mode

akashdhruv:
## Summary

This PR introduces dependency on [Bittree library](https://github.com/Flash-X/Bittree) to improve regridding performance in octree mode.

## Additional background

Testing and development of this feature is done in sync with Flash-X and is recorded in this [reproducibility capsule](https://github.com/Lab-Notebooks/AMReX-Bittree-Performance). This PR is primarily created to consolidate development work and avoid creating multiple branches. At present using AMReX+Bittree improves regridding performance [by a factor of 2](https://github.com/Lab-Notebooks/AMReX-Bittree-Performance/blob/14faa2212c4e5dba7fd99a6526c6937414f9c109/analysis/Performance.ipynb) at > 20000 ranks. We hope to improve performance further using a bittree-based distribution mapping and therefore adding a new function `AmrMesh::MakeDistributionMap` as a place-holder.

Continuation of #2893 and #3547

## Checklist

The proposed changes:


Src/AmrCore/AMReX_AmrMesh.H
```diff
@@ -253,13 +257,20 @@ public:
...
+
 protected:
 
     int finest_level;    //!< Current finest level.
     Vector<Geometry>            geom;
     Vector<DistributionMapping> dmap;
     Vector<BoxArray>            grids;
 
+#ifdef AMREX_USE_BITTREE
+    bool use_bittree;
```
WeiqunZhang:
```suggestion
    bool use_bittree = false;
```
We can have a default value. Then `use_bittree` will not longer be a required parameter.

Src/AmrCore/AMReX_AmrMesh.H
```diff
@@ -253,13 +257,20 @@ public:
...
+    std::shared_ptr<bittree::BittreeAmr> btmesh;
```
WeiqunZhang:
```suggestion
    std::unique_ptr<bittree::BittreeAmr> btmesh;
```
We could use unique_ptr instead.

Src/AmrCore/AMReX_AmrMesh.cpp
```diff
@@ -783,11 +899,48 @@ AmrMesh::MakeNewGrids (Real time)
...
+            for(int i=0; i<AMREX_SPACEDIM; ++i) {
+                top[i] = ncells[i] / max_grid_size[0][i];
+            }
+
+            // includes = boolean to check each coarsest level grid exists
+            // (Bittree supports having "holes" in the mesh)
+            int ngrids = AMREX_D_TERM(top[0],*top[1],*top[2]);
+            std::vector<int> includes(ngrids,1);
+
+            btmesh = std::make_shared<bittree::BittreeAmr>(top.data(),includes.data());
```
WeiqunZhang:
```suggestion
            btmesh = std::make_unique<bittree::BittreeAmr>(top.data(),includes.data());
```

Src/AmrCore/AMReX_AmrMesh.cpp
```diff
@@ -376,6 +383,10 @@ AmrMesh::InitAmrMesh (int max_level_in, const Vector<int>& n_cell_in,
 
     finest_level = -1;
 
+#ifdef AMREX_USE_BITTREE
+    pp.get("use_bittree",use_bittree);
```
WeiqunZhang:
```suggestion
    pp.queryAdd("use_bittree",use_bittree);
```

Src/AmrCore/AMReX_AmrMesh.cpp
```diff
@@ -437,6 +448,41 @@ AmrMesh::LevelDefined (int lev) noexcept
...
 }
 
+DistributionMapping
+AmrMesh::MakeDistributionMap (int lev, BoxArray const& ba)
+{
+
+    BL_PROFILE("AmrMesh::MakeDistributionMap()");
+
+    // useful print statement to notify that a new distribution map is being created at lev
+    amrex::Print() << "creating new distribution map on level: " << lev + 1 << "\n";
```
WeiqunZhang:
```suggestion
    if (verbose) {
        amrex::Print() << "Creating new distribution map on level: " << lev << "\n";
    }
```
Don't want to print this if verbose is 0. Also let's use the C++ convention that level numbering starts with 0.

Src/AmrCore/AMReX_AmrMesh.cpp
```diff
@@ -437,6 +448,41 @@ AmrMesh::LevelDefined (int lev) noexcept
...
+#ifdef AMREX_USE_BITTREE
+    // Bittree version
+    if(use_bittree)
+    {
+        // create distribution mapping
+        dm.define(ba);
+    }
+#endif
+
+    return dm;
```
WeiqunZhang:
```suggestion
#ifdef AMREX_USE_BITTREE
    // if (use_bittree) {
    //     return DistributionMapping(ba);
    // } else
#endif
    {
        return DistributionMapping(ba);
    }
```

WeiqunZhang (changes requested):
I will add a CI after this a merged.

WeiqunZhang:
I have run the test at `amrex/Tests/Amr/Advection_AmrCore/Exec`. It works as expected. However, the grids generated are not very efficient.

phi_plt00080_bittree
This is a picture from a run using bittree.

phi_plt00080_native
This is from a run using the native *non-tree* mode. The area covered by level 1 grids is much smaller.

akashdhruv:
> I have run the test at `amrex/Tests/Amr/Advection_AmrCore/Exec`. It works as expected. However, the grids generated are not very efficient.
>
> The area covered by level 1 grids is much smaller.

I think the refine-derefine criteria maybe too strict. I can look into this after the merge and create a follow-up PR

Src/Extern/Bittree/AMReX_Bittree.cpp
```diff
@@ -0,0 +1,377 @@
...
+                }
+            }}}
+        }
+    }}}
+
+    // Return false otherwise
+    return false;
+}
+
+/** Calculate integer coordinates of neighbors, taking into acount BCs.
```
WeiqunZhang:
```suggestion
/** Calculate integer coordinates of neighbors, taking into account BCs.
```

kweide:
> the grids generated are not very efficient.

I think we cannot do much better, given the octtree-mimicking parameters that we are using (and have been using way before Bittree) in Flash-X for the AMR/Amrex Grid implementation:
```Fortran
  ! Setup AMReX in octree mode
  call pp_amr%add   ("max_grid_size_x",     NXB)
  call pp_amr%add   ("max_grid_size_y",     NYB)
  call pp_amr%add   ("max_grid_size_z",     NZB)
  call pp_amr%add   ("blocking_factor_x", 2*NXB)
  call pp_amr%add   ("blocking_factor_y", 2*NYB)
  call pp_amr%add   ("blocking_factor_z", 2*NZB)
  call pp_amr%add   ("refine_grid_layout", 0)
  call pp_amr%add   ("grid_eff",  1.0)

  ! According to Weiqun n_proper=1 is an appropriate setting that will result in
  ! correct nesting.
  call pp_amr%add("n_proper", 1)
```
This is from our `gr_amrexInit.F90`. Maybe not all the lines are relevant here, maybe only the first 6 are. 

(I think at one point at least I understood how the `max_grid_size_` and `blocking_factor_` AMReX parameters work, and why they have to be set this way in order to limit refinement patters to the octtree-ish ones that can be generated by PARAMESH.)

WeiqunZhang (approved):

akashdhruv:
> phi_plt00080_bittree This is a picture from a run using bittree.
> 

@WeiqunZhang Can you create a time evolution so we can see how derefinement logic is working? It maybe possible that derefinement logic is not very strict.

PR #3558, opened 2023-09-19
Type issue

FR13ndSDP:
## Checklist

The proposed changes:

- The type of `use_gpu_aware_mpi` is not consistent with documentation. 
- If HIP version is lower than 5.2, provide an option to assign the value of `num_device_partners` directly.

### DIFF
--- a/Src/Base/AMReX_ParallelDescriptor.H
+++ b/Src/Base/AMReX_ParallelDescriptor.H
@@ -107,7 +107,7 @@ while ( false )
     void Initialize ();
     void Finalize ();
 
-    extern AMREX_EXPORT int use_gpu_aware_mpi;
+    extern AMREX_EXPORT bool use_gpu_aware_mpi;
     inline bool UseGpuAwareMpi () { return use_gpu_aware_mpi; }
 
     //! Split the process pool into teams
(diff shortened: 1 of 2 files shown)


Src/Base/AMReX_ParallelDescriptor.H
```diff
@@ -107,7 +107,7 @@ while ( false )
     void Initialize ();
     void Finalize ();
 
-    extern AMREX_EXPORT int use_gpu_aware_mpi;
+    extern AMREX_EXPORT bool use_gpu_aware_mpi;
```
WeiqunZhang, line 110: Casting from int to bool is well defined. So I don't think we need to make any changes here. Is your compiler complaining about it?

FR13ndSDP: Actually, there is no problem compiling the code, but if I use "amrex.use_gpu_aware_mpi=False" according to the docs, there will be a parmparse error.

FR13ndSDP: I've recently tested the GPU-aware MPI feature on a platform that uses Hygon DCU (AMD arch gfx906), and the result is not very satisfying.

WeiqunZhang: The documentation is incorrect. It should be 0 or 1.

WeiqunZhang: Even if the type is bool, the flag needs to be `false`, not `False`. So I will fix the documentation. If we change the type here to bool, we need to do that to its definition in AMReX_ParallelDescriptor.cpp too.

WeiqunZhang: #3562

WeiqunZhang: By the way, using `1` or `0` is always right for ParmParse boolean type parameters whether the C++ type is `int` or `bool`.

FR13ndSDP: Yeah, ok. So we just need to correct the documentation.

FR13ndSDP:
I will close this draft because more consideration is needed for HIP <= 5

PR #3566, opened 2023-09-27
Disable CCache in Windows CIs

WeiqunZhang:
CCache has stopped working on Windows since 2023-09-20 for unknown reasons. After a couple of hours of investigation, I still could not get it to work. So let's disable it for now.


ax3l (approved):
Very odd. For some reason it cannot find the executable anymore.

PR #3568, opened 2023-09-28
Fix Boundary Centroid in a Corner Case in 2D

WeiqunZhang:
Due to roundoff errors, It's possible to have a cell with area fraction of 1-epsilon, 1, 1, and 1 at its four faces. In this corner case, the previous way of computing boundary centroid was not correct. In this PR, it's fixed by using the level set.

## Checklist

The proposed changes:


WeiqunZhang:
@drangara

WeiqunZhang:
Do not merge before the 23.10 release.

cgilet (approved):
Yeah, unfortunately Deepak is right that we still also need a fix within the nodal solver. Thanks for this part of the fix Weiqun.

PR #3571, opened 2023-09-29
simplify how 2d surface integrals are computed

drangara:
## Summary
Previously, the 2D surface integrals were generated after calculating the intersection points of the EB on each of the cell faces. With this change, the intersection points are determined directly from `bcent`, `bnorm` and `barea`. This  approach seems to simplify the need to handle special cases.

## Checklist

The proposed changes:


drangara:
@cgilet

cgilet (approved):
This completely fixes the issue I was seeing. Thanks Deepak

PR #3574, opened 2023-09-29
Free Disk Space in Github CIs

WeiqunZhang:
Add a new script that removes some packages we do not need. It can free up about 7 GB of space on Ubuntu.

Use it in nvhpc and sycl/amd CIs. These CIs often run out of space. We don't want to use it in every CI, because it takes a few minutes to run the script.


ax3l:
I was always wondering if that helps, because docker essentially does a diff to what it started with...

But makes sense, I think `df` usage is what counts here for quotas...

ax3l (approved):

PR #3577, opened 2023-10-01
Add Bittree CI

WeiqunZhang:
Also fix some clang-tidy warnings.

Note that the 3D test does not actually work as expected. It has zero cells. We will try to fix in the future, because it's out of the scope of this PR adding CI.


WeiqunZhang:
This PR has a conflict with #3566. If one is merged first, we will update the other.

asalmgren (approved):

PR #3581, opened 2023-10-05
BCType::ext_dir_cc

### LINKED ISSUE #3580: Location of ext_dir
I just looked more carefully into this. When we need to fill a fine cell from spatial interpolation and it is near the physical boundary, we may need to use the physical boundary conditions stored in coarse level ghost cells to compute slopes. The BC type bug does not affect how ghost cells outside physical boundary are filled, because it calls a user provided function that does not use the BC type we set. For Flash-X, this function is `gr_fillPhysicalBC`. What the BC type bug affects is how slopes are computed. See https://github.com/AMReX-Codes/amrex/blob/87556458d867cb9cb2f148b4bcfc920c67bc918a/Src/AmrCore/AMReX_MFInterp_C.H#L13 for an example. You may wonder whether we treat `ext_dir` and `hoextrap` differently. This is because historically BoxLib codes assumed that the boundary location of `ext_dir` and `hoextrap` is exactly on the face, not half cell outside. So this is actually a known defect for codes that assume the location is at half cell outside. Maybe we should introduce a new type called `ext_dir_cc`. So I assume some Flash-X tests have `ext_dir` BC. If so, in the interpolation function, it was treated as being at cell centered before that bug fix and now as being on the domain face.

_Originally posted by @WeiqunZhang in https://github.com/AMReX-Codes/amrex/discussions/3578#discussioncomment-7190999_

WeiqunZhang:
Add a new BC type `BCType::ext_dir_cc` to address the issue in #3580. For cell conservation interpolation, the boundary location for `BCType::ext_dir` is on the domain face. Applications assuming boundary location at cell centers can now use `BCType::ext_dir_cc`.

## Checklist

The proposed changes:


WeiqunZhang:
@kweide

kweide:
I assume this means that basically everywhere in the Flash-X code, `ext_dir` should be replaced by `ext_dir_cc`?

WeiqunZhang:
Yes, I think that's correct.

kweide:
I just verified that, in the Flash-X test `Comparison_DustCollapse_AMR_2dCyl_AMReX` that I've been talking about for this issue, with the AMReX code in WeiqunZhang:ext_dir_cc (eb54ee13ccbc):

- with Flash-X as it was, the previous "new" (after 6eb91be02) results are reproduced.
- with a change `s/ext_dir/ext_dir_cc/g` in Flash-X code, the previous "old" (before 6eb91be02) results are reproduced.

So yeah, the older test results were better - even though I had convinced myself that the newer ones **looked** slightly better. :)

asalmgren (approved):

PR #3583, opened 2023-10-05
Fix ICC CI by Freeing up Disk Space

WeiqunZhang:
The ICC CI has been failing in the last few days with following message, `System.IO.IOException: No space left on device`. So we need to free up more disk space.


ax3l (approved):
Thank you! :+1:

PR #3585, opened 2023-10-06
Pure SoA Particle: Separate Array for IdCPU

### LINKED ISSUE #3569: Pure SoA Particles: cpuid as Separate Array
For pure SoA particles, we noticed slight regressions in kernels that depend on ID operatins, e.g., for sorting, etc.
There reason for that lies in the now required double-load of two int32 arrays (cpu and ids) instead of a single load of a int64 array (cpuid) in AoS.

As a discussed solution to this with @atmyers, we should add a third default array to SoA particles:
- 64bit cpuid (new)
- Real (default: SPACEDIM members for positions)
- int (default: no members)

cc @AlexanderSinn

ax3l:
## Summary

This addresses a regression we see when moving to pure SoA particles:
- slightly slower read/write to Ids when needed, e.g., for sorting
- issues going up to the full 64bit range

## Additional background

Once finished, this will close #3569.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_ParticleTile.H
+++ b/Src/Particle/AMReX_ParticleTile.H
@@ -43,6 +43,7 @@ struct ParticleTileData
 
     ParticleType* AMREX_RESTRICT m_aos;
 
+    uint64_t* m_idcpu;
     GpuArray<ParticleReal*, NAR> m_rdata;
     GpuArray<int*, NAI> m_idata;
 
@@ -67,7 +68,7 @@ struct ParticleTileData
         if constexpr(!ParticleType::is_soa_particle) {
             return this->m_aos[index].id();
         } else {
-            return this->m_idata[0][index];
+            return this->m_idcpu[index];  // TODO unwrap
         }
     }
 
@@ -77,7 +78,17 @@ struct ParticleTileData
         if constexpr(!ParticleType::is_soa_particle) {
             return this->m_aos[index].cpu();
         } else {
-            return this->m_idata[1][index];
+            return this->m_idcpu[index];  // TODO unwrap
+        }
+    }
+
+    [[nodiscard]] AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+    decltype(auto) idcpu (const int index) const &
+    {
+        if constexpr(ParticleType::is_soa_particle) {
+            return this->m_idcpu[index];
+        } else {
+            amrex::Abort("not implemented");
         }
     }
 
@@ -231,9 +242,8 @@ struct ParticleTileData
     {
         AMREX_ASSERT(index < m_size);
         SuperParticleType sp;
+        sp.m_idcpu = m_idcpu[index];
         for (int i = 0; i < AMREX_SPACEDIM; ++i) {sp.pos(i) = m_rdata[i][index];}
-        sp.id()  = m_idata[0][index];
-        sp.cpu() = m_idata[1][index];
         for (int i = 0; i < NAR; ++i) {
             sp.rdata(i) = m_rdata[i][index];
         }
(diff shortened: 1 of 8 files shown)


AlexanderSinn:
It would be nice if there was some sort of option to disable the 64-bit ID to save memory in the cases where a unique id for each particle is not required (i.e., 1 bit per particle in some other component is sufficient to track valid/invalid). It might be possible to “hack” this outside of amrex by setting the allocator of the ID PODVector to a null allocator.

ax3l:
Hi @AlexanderSinn, thank you for the comment.

On first thought, I am not sure this small optimization is worth the additional compile time that this will cause in pyAMReX, ABLASTR, AMReX and HiPACE++ for differentially specialized PC template classes.

What performance benefits do you see from this? Are the species that *never* need tracking very large and will significantly save memory without it?

AlexanderSinn:
For the plasma in HiPACE++ there is no per-particle diagnostics as it would be too much data, so there is no possibility to use the id for particle tracking. Besides the id the plasma has two other integer components so we could put those inside a 64 bit id.

The beam in HiPACE++ does have per particle diagnostics and can be interesting for particle tracking, however only if the per particle diagnostics is actually used. It is possible to have multiple beams that are each a few terabytes in size of which the 64bit id would be 12.5%. This is too much data for diagnostics/tracking, but the id still has to be communicated. Technically because the communication in HiPACE++ is custom, we don’t have to communicate every component that is in the particle container if we add a special case.

ax3l:
Thank you for the details. I am not yet fully convinced if this is not a premature optimization, because for background plasmas (in WarpX), filtering diagnostics can be essential to statistically analyze representative particles. The same can be true in the future for HiPACE++ for specific research tasks. But let's chat more tomorrow in the dev meeting about it.

In general, if we can easily make this a **runtime decision**, then I am all for it. @atmyers takes the lead on this, what do you think?

I just want to reduce the compile-time variations for particle containers at all cost, instead of driving them up further. For instance, the allocator being a compile-time option in particles (vs. a runtime one in MultiFabs - which is better) is another design (as in: maintainability, compile-time, link-time, extensibility vs. performance benefit) issue we should address another time (off-topic) #3643

AlexanderSinn:
Using amrex::PolymorphicArenaAllocator the allocator can be changed at runtime. That way only some components could have a different allocator.

Src/Particle/AMReX_StructOfArrays.H
```diff
@@ -119,7 +124,10 @@ struct StructOfArrays {
     */
     [[nodiscard]] std::size_t size () const
     {
-        if constexpr (NReal > 0) {
+        // TODO: if pure SoA:
```
atmyers:
```suggestion
```

Src/Particle/AMReX_ParticleTile.H
```diff
@@ -663,7 +699,10 @@ struct ParticleTile
         ArrayOfStructs<ParticleType, Allocator>>::type;
     //using ParticleVector = typename AoS::ParticleVector;
 
-    using SoA = StructOfArrays<NArrayReal, NArrayInt, Allocator>;
+    using SoA = typename std::conditional<
+        ParticleType::is_soa_particle,
+        StructOfArrays<NArrayReal, NArrayInt, Allocator, true>,
+        StructOfArrays<NArrayReal, NArrayInt, Allocator, true>>::type;
```
ax3l: @atmyers isn't this in both cases `true` now - so essentially we can remove the extra template argument in `StructOfArrays`?

ax3l (commented):
Thank you Andrew for your help!

I pushed a minor cleanup and inline comment and added a quick question on the new template argument in `StructOfArrays`.

Src/Particle/AMReX_ParticleTile.H
atmyers: This is a typo, I believe the second should be `false`...

atmyers (approved):

PR #3591, opened 2023-10-10
replace AMREX_DEVICE_COMPILE with AMREX_IF_ON_DEVICE and AMREX_IF_ON_HOST

### LINKED ISSUE #3586: [CUDA] replace __CUDA_ARCH__ with portable NV_IF_TARGET macros
In order to use the NVHPC unified compilation mode (`nvc++ -cuda`), it is necessary to avoid the `CUDA_ARCH` preprocessor macro and use `if target` instead (https://twitter.com/blelbach/status/1382041660859686912).

However, `nvcc` does not support `if target`. In order to fix this situation, libcu++ provides a new set of macros that provide a portable replacement for `__CUDA_ARCH__` that will work with either compiler:
```
  NV_IF_TARGET(query, true, ...)        | Queries compilation mode and emits code if true                |
  NV_IF_ELSE_TARGET(query, true, false) | As above, but can also emit different code when false          |
  NV_DISPATCH_TARGET(...)               | Similar to a switch statement emitting code for multiple modes |
```
(From https://github.com/NVIDIA/cccl/blob/265b574df68760ff81e2a68e373f09123d15ff55/libcudacxx/examples/nv_target.cu)

Additional usage examples in libcu++ itself: https://github.com/NVIDIA/libcudacxx/pull/432/files

A limitation of these macros is that since `if target` can only be used in function scope, `NV_IF_TARGET` can likewise only be used in function scope.

BenWibking:
## Summary
This adds the macros `AMREX_IF_ON_DEVICE((code_for_device))` and `AMREX_IF_ON_HOST((code_for_host))` that are compatible with single-pass host/device compilation (as used by `nvc++ -cuda`), as well as backward compatible with all other compilers.

This also replaces all uses of `AMREX_DEVICE_COMPILE` with these macros.

Fixes https://github.com/AMReX-Codes/amrex/issues/3586.

## Additional background
Single-pass compilation evalutes the preprocessor macros once for each source file. This means that preprocessor conditionals cannot be used to choose between host and device code. In particular, NVHPC with `-cuda` does not support `__CUDA_ARCH__`, instead requiring the use of the `if target` construct. This creates portable macros that work for either single-pass or two-pass compilation, but requires restructuring of any code that uses AMREX_DEVICE_COMPILE so that the code appears as a macro argument.

This PR will allow using NVHPC with `-cuda` as the unified host/device compiler for AMReX. In the future, single-pass compilers for other backends may be available, e.g., SYCL (https://dl.acm.org/doi/abs/10.1145/3585341.3585351).

AMReX can be configured to build with `nvc++ -cuda` using CMake:
```
cmake .. -DAMReX_GPU_BACKEND=CUDA -DCMAKE_C_COMPILER=nvc -DCMAKE_CXX_COMPILER=nvc++ -DCMAKE_CUDA_COMPILER=nvc++ -DCMAKE_CUDA_COMPILER_ID=NVCXX -DCMAKE_CUDA_ARCHITECTURES=80 -DCMAKE_CUDA_COMPILER_FORCED=ON -DCMAKE_CUDA_COMPILE_FEATURES=cuda_std_17 -DAMReX_GPU_RDC=OFF -DCMAKE_CXX_FLAGS="-cuda --gcc-toolchain=$(which gcc)" -DCMAKE_CUDA_FLAGS="-cuda --gcc-toolchain=$(which gcc)" -DAMReX_ENABLE_TESTS=ON -DCMAKE_CUDA_HOST_LINK_LAUNCHER=nvc++ -DCMAKE_CUDA_LINK_EXECUTABLE="<CMAKE_CUDA_HOST_LINK_LAUNCHER> <FLAGS> <LINK_FLAGS> <OBJECTS> -o <TARGET> <LINK_LIBRARIES>"
```
CMake hacks (https://github.com/NVIDIA/cub/blob/0fc3c3701632a4be906765b73be20a9ad0da603d/cmake/CubCompilerHacks.cmake) are tested with CMake 3.22.1 and NVHPC 23.5, 23.7, and 23.9 (earlier versions do not work). However, it currently fails to link the executables for the tests due to a [compiler/linker bug](https://forums.developer.nvidia.com/t/nvc-cuda-fails-to-link-code-when-using-device-curand-functions/270401/5).

(Note that by default, `nvcc` preserves denormals, whereas `nvc++` does not. Also, `nvc++` generates relocatable device code by default, whereas `nvcc` does not.)

## Checklist

The proposed changes:


BenWibking:
There is an odd failure for one use of `AMREX_IF_ON_HOST` that happens only when OpenMP is enabled: https://github.com/AMReX-Codes/amrex/actions/runs/6519867645/job/17706863160?pr=3591#step:7:186

BenWibking:
The OpenMP builds fail due to a strange quirk of how `#pragma omp atomic` is implemented:
```
#pragma omp atomic update
        *sum += value;
```
is considered valid by GCC, whereas
```
#pragma omp atomic update
        { *sum += value; }
```
is considered invalid.

The above is the preprocessed output for the host compile from this code in AMReX_GpuAtomic.H:
```
#ifdef AMREX_USE_OMP
#pragma omp atomic update
#endif
        AMREX_IF_ON_HOST((
            *sum += value;
        ))
```

Strangely, replacing the above with:
```
#ifdef AMREX_USE_OMP
        AMREX_IF_ON_HOST((
#pragma omp atomic update
            *sum += value;
        ))
#else
        AMREX_IF_ON_HOST((*sum += value;))
#endif
```
still expands to:
```
#pragma omp atomic update
        { *sum += value; }
```

I'm not sure how to work around this.

BenWibking:
This is the only portable workaround I could find:
```
namespace HostDevice::Atomic {

    template <class T>
    AMREX_FORCE_INLINE
    void Add_Host (T* const sum, T const value) noexcept
    {
#ifdef AMREX_USE_OMP
#pragma omp atomic update
#endif
        *sum += value;
    }

    template <class T>
    void Add (T* const sum, T const value) noexcept
    {
        AMREX_IF_ON_DEVICE((Gpu::Atomic::AddNoRet(sum,value);))
        AMREX_IF_ON_HOST((Add_Host(sum,value);))
    }

}
```

BenWibking:
In order to workaround the macro limitations, I had to replace the CLZ functions with, e.g.:
```
int clz (std::uint64_t x) noexcept
{
#if defined(AMREX_USE_CUDA) // all supported cuda versions have __clz
    AMREX_IF_ON_DEVICE((return detail::clz_wrapper(detail::clz_tag{}, x);))
    AMREX_IF_ON_HOST((return clz_generic(x);))
#elif AMREX_HAS_BUILTIN_CLZ
    AMREX_IF_ON_HOST((return detail::builtin_clz_wrapper(detail::clz_tag{}, x);))
    AMREX_IF_ON_DEVICE((return clz_generic(x);))
#else
    return clz_generic(x);
#endif
}
```

BenWibking:
When building with NVHPC, I get several errors about missing return statements:
```
"amrex/Src/Base/AMReX_GpuRange.H", line 40: warning: missing return statement at end of non-void function "amrex::Gpu::range_detail::size(const amrex::Box &) noexcept" [implicit_return_from_non_void_function]
  }
  ^
```
However, these appear to be spurious, due to conditional constructs like this:
```
AMREX_GPU_HOST_DEVICE
AMREX_FORCE_INLINE Long size (Box const& b) noexcept
{
    AMREX_IF_ON_DEVICE((return b.numPts();))
    AMREX_IF_ON_HOST((
        amrex::ignore_unused(b);
        return 1;
    ))
} // AMReX_GpuRange.H:40
```

BenWibking:
When building the tests with NVHPC, there are multiple definition linker errors for every binary for some cuRand symbols (they are the same for every binary):
```
[ 66%] Linking CUDA executable 3d/Test_Amr_Advection_AmrCore_3d
nvlink error   : Multiple definition of 'precalc_xorwow_matrix' in 'CMakeFiles/Test_Amr_Advection_AmrCore_3d.dir/Source/AdvancePhiAtLevel.cpp.o', first defined in 'CMakeFiles/Test_Amr_Advection_AmrCore_3d.dir/Source/AdvancePhiAllLevels.cpp.o'
nvlink error   : Multiple definition of 'precalc_xorwow_offset_matrix' in 'CMakeFiles/Test_Amr_Advection_AmrCore_3d.dir/Source/AdvancePhiAtLevel.cpp.o', first defined in 'CMakeFiles/Test_Amr_Advection_AmrCore_3d.dir/Source/AdvancePhiAllLevels.cpp.o'
nvlink error   : Multiple definition of 'mrg32k3aM1' in 'CMakeFiles/Test_Amr_Advection_AmrCore_3d.dir/Source/AdvancePhiAtLevel.cpp.o', first defined in 'CMakeFiles/Test_Amr_Advection_AmrCore_3d.dir/Source/AdvancePhiAllLevels.cpp.o'
nvlink error   : Multiple definition of 'mrg32k3aM2' in 'CMakeFiles/Test_Amr_Advection_AmrCore_3d.dir/Source/AdvancePhiAtLevel.cpp.o', first defined in 'CMakeFiles/Test_Amr_Advection_AmrCore_3d.dir/Source/AdvancePhiAllLevels.cpp.o'
nvlink error   : Multiple definition of 'mrg32k3aM1SubSeq' in 'CMakeFiles/Test_Amr_Advection_AmrCore_3d.dir/Source/AdvancePhiAtLevel.cpp.o', first defined in 'CMakeFiles/Test_Amr_Advection_AmrCore_3d.dir/Source/AdvancePhiAllLevels.cpp.o'
nvlink error   : Multiple definition of 'mrg32k3aM2SubSeq' in 'CMakeFiles/Test_Amr_Advection_AmrCore_3d.dir/Source/AdvancePhiAtLevel.cpp.o', first defined in 'CMakeFiles/Test_Amr_Advection_AmrCore_3d.dir/Source/AdvancePhiAllLevels.cpp.o'
nvlink error   : Multiple definition of 'mrg32k3aM1Seq' in 'CMakeFiles/Test_Amr_Advection_AmrCore_3d.dir/Source/AdvancePhiAtLevel.cpp.o', first defined in 'CMakeFiles/Test_Amr_Advection_AmrCore_3d.dir/Source/AdvancePhiAllLevels.cpp.o'
nvlink error   : Multiple definition of 'mrg32k3aM2Seq' in 'CMakeFiles/Test_Amr_Advection_AmrCore_3d.dir/Source/AdvancePhiAtLevel.cpp.o', first defined in 'CMakeFiles/Test_Amr_Advection_AmrCore_3d.dir/Source/AdvancePhiAllLevels.cpp.o'
nvlink error   : Multiple definition of '__cr_lgamma_table' in 'CMakeFiles/Test_Amr_Advection_AmrCore_3d.dir/Source/AdvancePhiAtLevel.cpp.o', first defined in 'CMakeFiles/Test_Amr_Advection_AmrCore_3d.dir/Source/AdvancePhiAllLevels.cpp.o'
nvlink fatal   : merge_elf failed
pgacclnk: child process exit status 2: /mnt/ufs18/home-208/wibkingb/spack/opt/spack/linux-centos7-zen/gcc-6.4.0/nvhpc-23.9-xysfmihvcig2jao55qhodg36ybrsxkde/Linux_x86_64/23.9/compilers/bin/tools/nvdd
make[2]: *** [Tests/Amr/Advection_AmrCore/3d/Test_Amr_Advection_AmrCore_3d] Error 2
make[1]: *** [Tests/Amr/Advection_AmrCore/CMakeFiles/Test_Amr_Advection_AmrCore_3d.dir/all] Error 2
```
@ax3l do you have any idea how this is happening?

EDIT: For device cuRAND functions, it might be necessary to compile with RDC enabled: https://gitlab.com/nvidia/headers/cuda-individual/curand/-/blob/main/curand_poisson.h?ref_type=heads#L182. Otherwise, we run into this issue: https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#device-code-in-libraries

Src/Base/AMReX_GpuQualifiers.H
```diff
@@ -8,6 +8,12 @@
 #include <hip/hip_runtime.h>
 #endif
 
+#if defined(AMREX_USE_CUDA) && defined(AMREX_CXX_PGI)
```
WeiqunZhang: Does this work?
```suggestion
#if defined(AMREX_USE_CUDA) && (defined(AMREX_CXX_PGI) || defined(AMREX_CXX_NVHPC))
```
In case, in the future nvc++ no longer defines `__PGI`.

Src/Base/AMReX_Algorithm.H
```diff
@@ -260,59 +261,91 @@ int clz_wrapper (clzll_tag, T x) noexcept
...
-int clz (std::uint16_t x) noexcept
+int clz (std::uint8_t x) noexcept
 {
-#if (AMREX_DEVICE_COMPILE && defined(AMREX_USE_CUDA)) // all supported cuda versions have __clz
-    return detail::clz_wrapper(detail::clz_tag{}, x);
-#elif (!AMREX_DEVICE_COMPILE && AMREX_HAS_BUILTIN_CLZ)
-    return detail::builtin_clz_wrapper(detail::clz_tag{}, x);
+#if defined(AMREX_USE_CUDA) // all supported cuda versions have __clz
+    AMREX_IF_ON_DEVICE((return detail::clz_wrapper(detail::clz_tag{}, x);))
+    AMREX_IF_ON_HOST((return clz_generic(x);))
```
WeiqunZhang: If AMREX_USE_CUDA is defined, the host code will always call `clz_generic`. So it's not the same as before. Would something like this work?
```
#if defined AMREX_USE_CUDA

int clz (std::uint8_t x) noexcept
{
    AMREX_IF_ON_DEVICE((return detail::clz_wrapper(detail::clz_tag{}, x);))
#if AMREX_HAS_BUILTIN_CLZ
    AMREX_IF_ON_HOST((return detail::builtin_clz_wrapper(detail::clz_tag{}, x);))
#else
    AMREX_IF_ON_HOST((return clz_generic(x);))
#endif
}

#else

int clz (std::uint8_t x) noexcept
{
#if (!AMREX_DEVICE_COMPILE && AMREX_HAS_BUILTIN_CLZ)
    return detail::builtin_clz_wrapper(detail::clz_tag{}, x);
#else
    return clz_generic(x);
#endif
}

#endif
```
I assume nvhpc does support builtin_clz and it defines `__clang__` or `__GNUC__`.

Src/Base/AMReX_GpuAtomic.H
```diff
@@ -386,8 +387,11 @@ namespace detail {
         sycl::atomic_ref<T,mo,ms,as> a{*m};
         return a.fetch_max(value);
 #else
-        amrex::ignore_unused(m,value);
-        return T(); // should never get here, but have to return something
+        AMREX_IF_ON_DEVICE(( return atomixMax(m, value); ))
```
```suggestion
        AMREX_IF_ON_DEVICE(( return atomicMax(m, value); ))
```

This and `Min_device` can probably be written as
```
  ...
#else
    AMREX_IF_ON_DEVICE(...);
    AMREX_IF_ON_HOST(...);
#endif
```

WeiqunZhang: and other functions like that.

Src/Base/AMReX_GpuRange.H
```diff
@@ -98,7 +98,12 @@ struct range_impl
 #elif defined (__SYCL_DEVICE_ONLY__)
         return iterator(m_b, m_gid, m_grange);
 #else
-        return iterator(m_b,0,1);
+        AMREX_IF_ON_DEVICE((
```
WeiqunZhang, line 98: Similar to Max_device, there is no reason to repeat this in two places.

Src/Base/AMReX_Random.H
```diff
@@ -99,8 +121,13 @@ namespace amrex
         mkl::rng::device::poisson<unsigned int> distr(lambda);
         return mkl::rng::device::generate(distr, *random_engine.engine);
 #else
-        amrex::ignore_unused(random_engine);
-        return RandomPoisson(lambda);
+        AMREX_IF_ON_DEVICE((
```
WeiqunZhang, line 108: This is also similar. Our assumption is only cuda device code can get here and has a non-empty AMREX_IF_ON_DEVICE. So we don't really need the `if defined __CUDA_ARCH__`. In fact, we can probably remove `__CUDA_ARCH__` entirely from amrex, except where we define AMREX_IF_ON_DEVIOCE.

Src/Base/AMReX_RealVect.H
```diff
@@ -521,13 +521,13 @@ public:
   /**
      This is a RealVect all of whose components are equal to zero.
   */
-  static AMREX_EXPORT const RealVect Zero;
```
WeiqunZhang: Why is `AMREX_EXPORT` removed? I think we need it for Windows. @ax3l

WeiqunZhang:
@BenWibking Thanks for the PR! I have done a first pass. Looks great overall. But I want to read the changes more carefully and do more testing on various architectures. Maybe we can target merging this at the beginning of next month.

WeiqunZhang:
A downside of these macros is it makes debugging harder. Everything becomes a single line.
```
#include <iostream>

#if defined (__PGI)

#include <nv/target>
#define AMREX_IF_ON_HOST(CODE) NV_IF_TARGET(NV_IS_HOST, CODE)

#else

#define AMREX_IMPL_STRIP_PARENS(X) AMREX_IMPL_ESC(AMREX_IMPL_ISH X)
#define AMREX_IMPL_ISH(...) AMREX_IMPL_ISH __VA_ARGS__
#define AMREX_IMPL_ESC(...) AMREX_IMPL_ESC_(__VA_ARGS__)
#define AMREX_IMPL_ESC_(...) AMREX_IMPL_VAN_##__VA_ARGS__
#define AMREX_IMPL_VAN_AMREX_IMPL_ISH

#define AMREX_IF_ON_HOST(CODE) \
  { AMREX_IMPL_STRIP_PARENS(CODE) }

#endif

int main (int argc, char* argv[])
{   
    AMREX_IF_ON_HOST((
        int a = 3, b = 4;
        std::cout << " line number " << __LINE__ << std::endl;
        // some comments 
        // more comments 
         std::cout << "Hello World, a = " << a << ", b = " << b
                  << " line number " << __LINE__ << std::endl;
    ))
}

$ nvc++ -cuda -E foo.cpp
...
int main (int argc, char* argv[])
{
    if target (nv::target::is_host) { int a = 3, b = 4; std::cout << " line number " << 30 << std::endl; std::cout << "Hello World, a = " << a << ", b = " << b << " line num
ber " << 30 << std::endl; } else { }
# 31 "foo.cpp"
}

$ g++ -E foo.cpp
...
# 21 "foo.cpp"
int main (int argc, char* argv[])
{
    { int a = 3, b = 4; std::cout << " line number " << 30 << std::endl; std::cout << "Hello World, a = " << a << ", b = " << b << " line number " << 30 << std::endl; }







}
```

:disappointed: nvhpc :-1:

BenWibking:
> A downside of these macros is it makes debugging harder. Everything becomes a single line.
> 

That's very unfortunate.

It does appear that `cpp` considers this behavior a bug (https://gcc.gnu.org/onlinedocs/cpp/Newlines-in-Arguments.html) but I doubt it will change "in the near future" since it has been "considered a bug" since at least 2002 (https://web.archive.org/web/20020222070034/https://gcc.gnu.org/onlinedocs/cpp/Newlines-in-Arguments.html) :(

WeiqunZhang:
Maybe we should do something like this
```
#if defined(AMREX_USE_NVCPP) /* nvc++ is used to compile cuda code */
#  define AMREX_DEVICE_COMPILE 1
#  define AMREX_HOST_COMPILE 1
#  define AMREX_IF_NV_TARGET_IS_DEVICE if target (nv::target::is_device)
#  define AMREX_IF_NV_TARGET_IS_HOST if target (nv::target::is_host)
#else
#  define AMREX_DEVICE_COMPILE (__CUDA_ARCH__ || __HIP_DEVICE_COMPILE__ || __SYCL_DEVICE_ONLY__)
#  define AMREX_HOST_COMPILE (!AMREX_DEVICE_COMPILE)
#  define AMREX_IF_NV_TARGET_IS_DEVICE
#  define AMREX_IF_NV_TARGET_IS_HOST
#endif


#if AMREX_HOST_COMPILE
    AMREX_IF_NV_TARGET_IS_HOST
    {
        std::cout << "host code\n";
    }
#endif
#if AMREX_DEVICE_COMPILE
    AMREX_IF_NV_TARGET_IS_DEVICE
    {
        printf("device code\n");
    }
#endif
```

BenWibking:
> Maybe we should do something like this
> 
> ```
> #if defined(AMREX_USE_NVCPP) /* nvc++ is used to compile cuda code */
> #  define AMREX_DEVICE_COMPILE 1
> #  define AMREX_HOST_COMPILE 1
> #  define AMREX_IF_NV_TARGET_IS_DEVICE if target (nv::target::is_device)
> #  define AMREX_IF_NV_TARGET_IS_HOST if target (nv::target::is_host)
> #else
> #  define AMREX_DEVICE_COMPILE (__CUDA_ARCH__ || __HIP_DEVICE_COMPILE__ || __SYCL_DEVICE_ONLY__)
> #  define AMREX_HOST_COMPILE (!AMREX_DEVICE_COMPILE)
> #  define AMREX_IF_NV_TARGET_IS_DEVICE
> #  define AMREX_IF_NV_TARGET_IS_HOST
> #endif
> 
> 
> #if AMREX_HOST_COMPILE
>     AMREX_IF_NV_TARGET_IS_HOST
>     {
>         std::cout << "host code\n";
>     }
> #endif
> #if AMREX_DEVICE_COMPILE
>     AMREX_IF_NV_TARGET_IS_DEVICE
>     {
>         printf("device code\n");
>     }
> #endif
> ```

If there is CI to build and test with NVCPP in single-pass mode, then I guess I am ok with this.

I am hestiant because this solution makes it possible to accidentally omit either the outer `#if AMREX_DEVICE COMPILE` or the inner `AMREX_IF_NV_TARGET_IS_DEVICE` in which case it will work fine on one compiler but not the other.

WeiqunZhang:
If we make `AMREX_IF_NV_TARGET_IS_HOST` and `AMREX_IF_NV_TARGET_IS_DEVICE` include a `{`, it would make those hard to miss. But not sure about `AMREX_DEVICE_COMPILE` and `AMREX_HOST_COMPILE`. The main issue is probably nvcc, which does not support `if constexpr` well.

Src/Base/AMReX_Algorithm.H
BenWibking: Fixed.

Src/Base/AMReX_GpuAtomic.H
BenWibking: Fixed.

Src/Base/AMReX_GpuRange.H
BenWibking: Fixed.

Src/Base/AMReX_Random.H
BenWibking: Fixed. Also fixed in AMReX_Math.H.

Remaining uses of `__CUDA_ARCH__` are:
- AMReX_Extension.H (used to conditionally define `AMREX_PRAGMA_SIMD` and `AMREX_FORCE_INLINE`)
- AMReX_GpuAtomic.H (one use of `#if defined(AMREX_USE_CUDA) && (__CUDA_ARCH__ < 600)`)
- AMReX_Parser_Y.H (used for elliptic integrals that are only supported on host GCC)

Src/Base/AMReX_RealVect.H
BenWibking: Fixed.

BenWibking:
> When building the tests with NVHPC, there are multiple definition linker errors for every binary for some cuRand symbols (they are the same for every binary):
> 
> ```
> [ 66%] Linking CUDA executable 3d/Test_Amr_Advection_AmrCore_3d
> nvlink error   : Multiple definition of 'precalc_xorwow_matrix' in 'CMakeFiles/Test_Amr_Advection_AmrCore_3d.dir/Source/AdvancePhiAtLevel.cpp.o', first defined in 'CMakeFiles/Test_Amr_Advection_AmrCore_3d.dir/Source/AdvancePhiAllLevels.cpp.o'
> nvlink error   : Multiple definition of 'precalc_xorwow_offset_matrix' in 'CMakeFiles/Test_Amr_Advection_AmrCore_3d.dir/Source/AdvancePhiAtLevel.cpp.o', first defined in 'CMakeFiles/Test_Amr_Advection_AmrCore_3d.dir/Source/AdvancePhiAllLevels.cpp.o'
> nvlink error   : Multiple definition of 'mrg32k3aM1' in 'CMakeFiles/Test_Amr_Advection_AmrCore_3d.dir/Source/AdvancePhiAtLevel.cpp.o', first defined in 'CMakeFiles/Test_Amr_Advection_AmrCore_3d.dir/Source/AdvancePhiAllLevels.cpp.o'
> nvlink error   : Multiple definition of 'mrg32k3aM2' in 'CMakeFiles/Test_Amr_Advection_AmrCore_3d.dir/Source/AdvancePhiAtLevel.cpp.o', first defined in 'CMakeFiles/Test_Amr_Advection_AmrCore_3d.dir/Source/AdvancePhiAllLevels.cpp.o'
> nvlink error   : Multiple definition of 'mrg32k3aM1SubSeq' in 'CMakeFiles/Test_Amr_Advection_AmrCore_3d.dir/Source/AdvancePhiAtLevel.cpp.o', first defined in 'CMakeFiles/Test_Amr_Advection_AmrCore_3d.dir/Source/AdvancePhiAllLevels.cpp.o'
> nvlink error   : Multiple definition of 'mrg32k3aM2SubSeq' in 'CMakeFiles/Test_Amr_Advection_AmrCore_3d.dir/Source/AdvancePhiAtLevel.cpp.o', first defined in 'CMakeFiles/Test_Amr_Advection_AmrCore_3d.dir/Source/AdvancePhiAllLevels.cpp.o'
> nvlink error   : Multiple definition of 'mrg32k3aM1Seq' in 'CMakeFiles/Test_Amr_Advection_AmrCore_3d.dir/Source/AdvancePhiAtLevel.cpp.o', first defined in 'CMakeFiles/Test_Amr_Advection_AmrCore_3d.dir/Source/AdvancePhiAllLevels.cpp.o'
> nvlink error   : Multiple definition of 'mrg32k3aM2Seq' in 'CMakeFiles/Test_Amr_Advection_AmrCore_3d.dir/Source/AdvancePhiAtLevel.cpp.o', first defined in 'CMakeFiles/Test_Amr_Advection_AmrCore_3d.dir/Source/AdvancePhiAllLevels.cpp.o'
> nvlink error   : Multiple definition of '__cr_lgamma_table' in 'CMakeFiles/Test_Amr_Advection_AmrCore_3d.dir/Source/AdvancePhiAtLevel.cpp.o', first defined in 'CMakeFiles/Test_Amr_Advection_AmrCore_3d.dir/Source/AdvancePhiAllLevels.cpp.o'
> nvlink fatal   : merge_elf failed
> pgacclnk: child process exit status 2: /mnt/ufs18/home-208/wibkingb/spack/opt/spack/linux-centos7-zen/gcc-6.4.0/nvhpc-23.9-xysfmihvcig2jao55qhodg36ybrsxkde/Linux_x86_64/23.9/compilers/bin/tools/nvdd
> make[2]: *** [Tests/Amr/Advection_AmrCore/3d/Test_Amr_Advection_AmrCore_3d] Error 2
> make[1]: *** [Tests/Amr/Advection_AmrCore/CMakeFiles/Test_Amr_Advection_AmrCore_3d.dir/all] Error 2
> ```
> 
> @ax3l do you have any idea how this is happening?
> 
> EDIT: For device cuRAND functions, it might be necessary to compile with RDC enabled: https://gitlab.com/nvidia/headers/cuda-individual/curand/-/blob/main/curand_poisson.h?ref_type=heads#L182. Otherwise, we run into this issue: https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#device-code-in-libraries

Even if I modify the cuRAND headers to unconditionally define these variables as `inline __device__` instead of `static __device__`, the linker error persists. I don't understand what is going on in the device linking process here.

@WeiqunZhang @ax3l any ideas? (If I try `-DAMReX_GPU_RDC=ON`, I get compiler errors before even getting to the linking step.)

WeiqunZhang:
I tried it on perlmutter and got lots of
```
CMake Error at Tests/CMakeLists.txt:26 (add_executable):
  Target "Test_LinearSolvers_ABecLaplacian_C_3d" links to target
  "CUDA::curand" but the target was not found.  Perhaps a find_package() call
  is missing for an IMPORTED target, or an ALIAS target is missing?
Call Stack (most recent call first):
  Tests/LinearSolvers/ABecLaplacian_C/CMakeLists.txt:16 (setup_test)


CMake Error at Tests/CMakeLists.txt:26 (add_executable):
  Target "Test_LinearSolvers_NodalPoisson_3d" links to target "CUDA::curand"
  but the target was not found.  Perhaps a find_package() call is missing for
  an IMPORTED target, or an ALIAS target is missing?
Call Stack (most recent call first):
  Tests/LinearSolvers/NodalPoisson/CMakeLists.txt:10 (setup_test)
```

BenWibking:
Were you using nvhpc/23.9 on Perlmutter?

Perhaps NVIDIA technical support will respond here: https://forums.developer.nvidia.com/t/nvc-cuda-fails-to-link-code-when-using-device-curand-functions/270401

WeiqunZhang:
That was a cmake thing. I have ot switch to a more recent version of cmake.

WeiqunZhang:
If I manually compile AMReX_Random.cpp
```
nvc++ -O3 -fast -gopt -std=c++17  -cuda -gpu=cc80 -gpu=rdc \
      -DBL_NO_FORT -DAMREX_USE_CUDA -DAMREX_USE_GPU -DBL_COALESCE_FABS -DAMREX_GPU_MAX_THREADS=256 -DAMREX_USE_GPU_RDC -DBL_SPACEDIM=2 -DAMREX_SPACEDIM=2 \
      -DBL_FORT_USE_UNDERSCORE -DAMREX_FORT_USE_UNDERSCORE -DBL_Linux -DAMREX_Linux -DAMREX_PARTICLES -DNDEBUG \
      -Itmp_build_dir/s/2d.nvhpc.CUDA.EXE -I. -I. -I..//Source -I..//Source/Src_K -I../../../../Src/Base \
      -I../../../../Src/Base/Parser -I../../../../Src/Boundary -I../../../../Src/AmrCore -I../../../../Src/Particle \
      -c ../../../../Src/Base/AMReX_Random.cpp -o tmp_build_dir/o/2d.nvhpc.CUDA.EXE/AMReX_Random.o
```
I get
```
NVC++-S-1207-Long double in GPU code is unsupported (../../../../Src/Base/AMReX_Random.cpp: 1785)
NVC++-F-0155-Compiler failed to translate accelerator region (see -Minfo messages): Long double in GPU code is unsupported (../../../../Src/Base/AMReX_Random.cpp: 1)
NVC++/x86-64 Linux 22.7-0: compilation aborted
```

BenWibking:
> I get
> 
> ```
> NVC++-S-1207-Long double in GPU code is unsupported (../../../../Src/Base/AMReX_Random.cpp: 1785)
> NVC++-F-0155-Compiler failed to translate accelerator region (see -Minfo messages): Long double in GPU code is unsupported (../../../../Src/Base/AMReX_Random.cpp: 1)
> NVC++/x86-64 Linux 22.7-0: compilation aborted
> ```

Is there a newer version of NVC++ on Perlmutter? I installed nvhpc 23.9 via `spack install nvhpc`.

WeiqunZhang:
There is 23.1. But same error. `-m64` does not help.

BenWibking:
Ok, it turns out 23.5 is the first version that can compile it succesfully. Still fails to link with the same error as 23.9.

WeiqunZhang:
I installed the latest nvhpc and modified the gnu makefile. I now get the same multiple definition error as you at link time.

WeiqunZhang:
https://github.com/WeiqunZhang/nvhpc_curand

BenWibking:
I tried to switch to your version of the macros in order to preserve line numbers, but I made several syntax errors that were somehow not picked up by the compiler. (The tests failed, however.) With this in mind, I'm not sure if it's a good idea to try to avoid the function-like macros.

Src/Base/AMReX_Random.H
BenWibking: What do you want to do about the `__CUDA_ARCH__` in AMReX_Extension.H? I don't think the latter two files really matter.

WeiqunZhang: The use of `__CUDA_ARCH__` in AMReX_Extenstion.H is fine.

In AMReX_GpuAtomic.H, I think we need to change from `#if defined(AMREX_USE_CUDA) && (__CUDA_ARCH__ < 600)` to `#if defined(AMREX_USE_CUDA) && defined(__CUDA_ARCH__) && (__CUDA_ARCH__ < 600)`, because `__CUDA_ARCH__` would be become 0 in the comparison if it's not define.

As for AMReX_Parser_Y.H, `nvc++ -E -dM - < /dev/null` shows `#if defined(__GNUC__) && !defined(__clang__) && !defined(__CUDA_ARCH__)` is true. So we need to update it unless `std::comp_ellint_1` and `_2` work on device with nvhpc .

Src/Base/Parser/AMReX_Parser_Y.H
```diff
@@ -350,7 +350,7 @@ template <typename T>
 AMREX_GPU_HOST_DEVICE AMREX_NO_INLINE
 T parser_math_comp_ellint_1 (T a)
 {
-#if defined(__GNUC__) && !defined(__clang__) && !defined(__CUDA_ARCH__)
+#if defined(__GNUC__) && !defined(__clang__) && !defined(__CUDA_ARCH__) && !defined(__NVCOMPILER)
```
BenWibking, line 353: @WeiqunZhang Does this work?

Src/Base/AMReX_GpuAtomic.H
```diff
@@ -175,7 +175,7 @@ namespace detail {
 
 #endif
 
-#if defined(AMREX_USE_CUDA) && (__CUDA_ARCH__ < 600)
+#if defined(AMREX_USE_CUDA) && defined(__CUDA_ARCH__) && (__CUDA_ARCH__ < 600)
```
BenWibking, line 178: Fixed.

Src/Base/Parser/AMReX_Parser_Y.H
WeiqunZhang: sounds good

WeiqunZhang:
The multiple definition issue is a nvhpc compiler issue. We can ignore it for now. I will take one more pass within a couple of days before merging this.

Src/Base/AMReX_Math.H
```diff
@@ -66,55 +66,47 @@ constexpr std::enable_if_t<std::is_floating_point<T>::value,T> pi ()
 AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
 double cospi (double x)
 {
-#if defined(AMREX_USE_SYCL)
+#if defined(__SYCL_DEVICE_ONLY__)
```
WeiqunZhang:
```suggestion
#if defined(AMREX_USE_SYCL)
```

Src/Base/AMReX_Math.H
```diff
@@ -66,55 +66,47 @@ constexpr std::enable_if_t<std::is_floating_point<T>::value,T> pi ()
...
+    AMREX_IF_ON_HOST(( return std::cos(pi<double>()*x); ))
 #endif
 }
 
 //! Return cos(x*pi) given x
 AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
 float cospi (float x)
 {
-#if defined(AMREX_USE_SYCL)
+#if defined(__SYCL_DEVICE_ONLY__)
```
WeiqunZhang:
```suggestion
#if defined(AMREX_USE_SYCL)
```

Src/Base/AMReX_Math.H
```diff
@@ -66,55 +66,47 @@ constexpr std::enable_if_t<std::is_floating_point<T>::value,T> pi ()
...
+    AMREX_IF_ON_HOST(( return std::cos(pi<float>()*x); ))
 #endif
 }
 
 //! Return sin(x*pi) given x
 AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
 double sinpi (double x)
 {
-#if defined(AMREX_USE_SYCL)
+#if defined(__SYCL_DEVICE_ONLY__)
```
WeiqunZhang:
```suggestion
#if defined(AMREX_USE_SYCL)
```

Src/Base/AMReX_Math.H
```diff
@@ -66,55 +66,47 @@ constexpr std::enable_if_t<std::is_floating_point<T>::value,T> pi ()
...
+    AMREX_IF_ON_HOST(( return std::sin(pi<double>()*x); ))
 #endif
 }
 
 //! Return sin(x*pi) given x
 AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
 float sinpi (float x)
 {
-#if defined(AMREX_USE_SYCL)
+#if defined(__SYCL_DEVICE_ONLY__)
```
WeiqunZhang:
```suggestion
#if defined(AMREX_USE_SYCL)
```

Src/Base/AMReX_Math.H
```diff
@@ -123,15 +115,14 @@ AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
 std::pair<double,double> sincos (double x)
 {
     std::pair<double,double> r;
-#if defined(AMREX_USE_SYCL)
+#if defined(__SYCL_DEVICE_ONLY__)
```
WeiqunZhang:
```suggestion
#if defined(AMREX_USE_SYCL)
```

Src/Base/AMReX_Math.H
```diff
@@ -123,15 +115,14 @@ AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
...
     r.first = sycl::sincos(x, sycl::private_ptr<double>(&r.second));
-#elif defined(__CUDA_ARCH__) && defined(AMREX_USE_CUDA) || \
-      defined(__HIP_DEVICE_COMPILE__) && defined(AMREX_USE_HIP) || \
-      (defined(_GNU_SOURCE) && !defined(__APPLE__))
-    ::sincos(x, &r.first, &r.second);
 #else
-    r.first  = std::sin(x);
-    r.second = std::cos(x);
+    AMREX_IF_ON_DEVICE(( ::sincos(x, &r.first, &r.second); ))
+    AMREX_IF_ON_HOST((
```
WeiqunZhang: Need to fix this for GNU source.

Src/Base/AMReX_Math.H
```diff
@@ -141,15 +132,14 @@ AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
 std::pair<float,float> sincos (float x)
 {
     std::pair<float,float> r;
-#if defined(AMREX_USE_SYCL)
+#if defined(__SYCL_DEVICE_ONLY__)
```
WeiqunZhang:
```suggestion
#if defined(AMREX_USE_SYCL)
```

Src/Base/AMReX_Math.H
```diff
@@ -141,15 +132,14 @@ AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
...
     r.first = sycl::sincos(x, sycl::private_ptr<float>(&r.second));
-#elif defined(__CUDA_ARCH__) && defined(AMREX_USE_CUDA) || \
-      defined(__HIP_DEVICE_COMPILE__) && defined(AMREX_USE_HIP) || \
-      (defined(_GNU_SOURCE) && !defined(__APPLE__))
-    ::sincosf(x, &r.first, &r.second);
 #else
-    r.first  = std::sin(x);
-    r.second = std::cos(x);
+    AMREX_IF_ON_DEVICE(( ::sincosf(x, &r.first, &r.second); ))
+    AMREX_IF_ON_HOST((
```
WeiqunZhang: Fix for GNU source

Src/Base/AMReX_Math.H
```diff
@@ -159,11 +149,11 @@ AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
 std::pair<double,double> sincospi (double x)
 {
     std::pair<double,double> r;
-#if defined(__CUDA_ARCH__) && defined(AMREX_USE_CUDA) || \
-    defined(__HIP_DEVICE_COMPILE__) && defined(AMREX_USE_HIP)
-    ::sincospi(x, &r.first, &r.second);
-#else
+#if defined(__SYCL_DEVICE_ONLY__)
```
WeiqunZhang:
```suggestion
#if defined(AMREX_USE_SYCL)
```

Src/Base/AMReX_Math.H
```diff
@@ -173,11 +163,11 @@ AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
 std::pair<float,float> sincospi (float x)
 {
     std::pair<float,float> r;
-#if defined(__CUDA_ARCH__) && defined(AMREX_USE_CUDA) || \
-    defined(__HIP_DEVICE_COMPILE__) && defined(AMREX_USE_HIP)
-    ::sincospif(x, &r.first, &r.second);
-#else
+#if defined(__SYCL_DEVICE_ONLY__)
```
WeiqunZhang:
```suggestion
#if defined(AMREX_USE_SYCL)
```

Src/Base/AMReX_RealVect.H
```diff
@@ -527,7 +527,7 @@ public:
   /**
      This is a RealVect all of whose components are equal to one.
   */
-  static AMREX_EXPORT const RealVect Unit;
+  static const RealVect Unit;
```
WeiqunZhang:
```suggestion
  static AMREX_EXPORT const RealVect Unit;
```

Src/Base/AMReX_Algorithm.H
```diff
@@ -252,7 +252,7 @@ int clz (T x) noexcept;
 AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
 int clz_generic (std::uint8_t x) noexcept
 {
-    constexpr int clz_lookup[16] = { 4, 3, 2, 2, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0 };
+    static constexpr int clz_lookup[16] = { 4, 3, 2, 2, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0 };
```
BenWibking, line 255: For some reason, nvc++ won't compile it with `static`.

WeiqunZhang: Maybe you can file a bug report. I suspect it should pretty easy to write a small reproducer.

Since we need to wait for the inline bug to be fixed anyway, we probably don't need to worry about this issue for now. If we need to work around this for nvc++, we could use `if !defined(__NVHPC)`. If we don't use `static`, the compiler might waste stack space unnecessarily.

WeiqunZhang (approved):

BenWibking:
FYI- The nvhpc linker bug should be fixed in NVHPC 24.3: https://forums.developer.nvidia.com/t/nvc-cuda-fails-to-link-code-when-using-device-curand-functions/270401/7.

BenWibking:
With the following patch:
```
diff --git a/Src/Base/AMReX_Algorithm.H b/Src/Base/AMReX_Algorithm.H
index 4718606b7..0fc26df05 100644
--- a/Src/Base/AMReX_Algorithm.H
+++ b/Src/Base/AMReX_Algorithm.H
@@ -289,7 +289,11 @@ int clz (T x) noexcept;
 int clz_generic (std::uint8_t x) noexcept
 {
+#if !defined(__NVCOMPILER)
     static constexpr int clz_lookup[16] = { 4, 3, 2, 2, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0 };
+#else
+    constexpr int clz_lookup[16] = { 4, 3, 2, 2, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0 };
+#endif
     auto upper = x >> 4;
     auto lower = x & 0xF;
     return upper ? clz_lookup[upper] : 4 + clz_lookup[lower];
diff --git a/Src/Base/AMReX_INT.H b/Src/Base/AMReX_INT.H
index f54852cb6..4356c70f1 100644
--- a/Src/Base/AMReX_INT.H
+++ b/Src/Base/AMReX_INT.H
@@ -31,7 +31,7 @@ namespace amrex {
 }
 #endif

-#if (defined(__x86_64) || defined (__aarch64__)) && !defined(_WIN32) && (defined(__GNUC__) || defined(__clang__))
+#if (defined(__x86_64) || defined (__aarch64__)) && !defined(_WIN32) && (defined(__GNUC__) || defined(__clang__)) && !defined(__NVCOMPILER)

 #define AMREX_INT128_SUPPORTED 1
```
CMake configures succesfully:
```
cmake .. -DAMReX_GPU_BACKEND=CUDA -DCMAKE_C_COMPILER=nvc -DCMAKE_CXX_COMPILER=nvc++ -DCMAKE_CUDA_COMPILER=nvc++ -DCMAKE_CUDA_COMPILER_ID=NVCXX -DCMAKE_CUDA_ARCHITECTURES=80 -DCMAKE_CUDA_COMPILER_FORCED=ON -DCMAKE_CUDA_COMPILE_FEATURES=cuda_std_17 -DAMReX_GPU_RDC=OFF -DCMAKE_CXX_FLAGS="-cuda --gcc-toolchain=$(which gcc)" -DCMAKE_CUDA_FLAGS="-cuda --gcc-toolchain=$(which gcc)" -DAMReX_ENABLE_TESTS=ON -DCMAKE_CUDA_HOST_LINK_LAUNCHER=nvc++ -DCMAKE_CUDA_LINK_EXECUTABLE="<CMAKE_CUDA_HOST_LINK_LAUNCHER> <FLAGS> <LINK_FLAGS> <OBJECTS> -o <TARGET> <LINK_LIBRARIES>"
```
Then NVHPC 24.3 crashes with the error:
```
[bwibking@dt-login02 Src]$ /u/bwibking/hpc_sdk/Linux_x86_64/24.3/compilers/bin/nvc++  -DAMREX_SPACEDIM=3 -I/u/bwibking/amrex/Src/Base -I/u/bwibking/amrex/Src/Base/Parser -I/u/bwibking/amrex/Src/Boundary -I/u/bwibking/amrex/Src/AmrCore -I/u/bwibking/amrex/Src/Amr -I/u/bwibking/amrex/Src/LinearSolvers -I/u/bwibking/amrex/Src/LinearSolvers/MLMG -I/u/bwibking/amrex/Src/LinearSolvers/OpenBC -I/u/bwibking/amrex/Src/Particle -I/u/bwibking/amrex/build -I/sw/spack/deltas11-2023-03/apps/linux-rhel8-zen3/gcc-11.4.0/openmpi-4.1.5-ejhvyny/include -I/sw/spack/deltas11-2023-03/apps/linux-rhel8-zen3/gcc-11.4.0/cuda-11.8.0-vfixfmc/include -cuda --gcc-toolchain=/sw/spack/deltas11-2023-03/apps/linux-rhel8-x86_64/gcc-8.5.0/gcc-11.4.0-yycklku/bin/gcc -pthread  -c /u/bwibking/amrex/Src/Base/AMReX_BlockMutex.cpp -o CMakeFiles/amrex_3d.dir/Base/AMReX_BlockMutex.cpp.o
nvdd-Fatal-/sw/spack/deltas11-2023-03/apps/linux-rhel8-zen3/gcc-11.4.0/cuda-11.8.0-vfixfmc/bin/ptxas TERMINATED by signal 11
NVC++-F-0155-Compiler failed to translate accelerator region (see -Minfo messages): Device compiler exited with error status code (/u/bwibking/amrex/Src/Base/AMReX_BlockMutex.cpp: 14)
NVC++/x86-64 Linux 24.3-0: compilation aborted
```

PR #3595, opened 2023-10-16
Temporary Runtime Components in Particle Container

ax3l:
## Summary

For diagnostics purposes in WarpX, we would like to temporarily add particle runtime components to a particle container, and deallocate them timely again to avoid large memory spikes in I/O.

This adds new functions to remove runtime components.

### To Do

- [ ] doc issues in existing `AddReal/IntComp`: no doxygen, no mention of [API contract for allocation](https://amrex-codes.github.io/amrex/docs_html/Particle.html#adding-particle-components-at-runtime)
  - [ ] for an already allocated species, how does one add-and-allocate *only the newly added runtime attributes*?
- [ ] deallocation: what does a user need to do after `RemoveRealComp`/`RemoveIntComp` to deallocate the memory?

## Checklist

The proposed changes:


ax3l:
@atmyers let's revisit this soon?

atmyers:
This doesn't currently deallocate anything, for that we need to loop over the particle tiles and remove the old component form the struct of arrays.

asalmgren:
@ax3l -- this PR is over 2 years old -- is it still relevant?

ax3l:
Yes, we should refactor it once we support all-runtime PCs

PR #3600, opened 2023-10-18
[Breaking] Prefix `amrex_` to each plotfile Tool

ax3l:
## Summary

Packaging AMReX for Conda, we received feedback that our plotfile tool names are too generic and can collide with other packages.

Thus, we propose to rename them with a common prefix.

- [x] decide prefix or suffix (suffix: `<TAB>` completion will work; prefix: easier for new users)
- [x] decide name: `plt`/`amrex`/`amr`/...
  - Weiqun and Axel prefer `amrex_` so far
- [ ] also change for GNUmake
  - [ ] update [test harness](https://github.com/AMReX-Codes/regression_testing)
- [ ] update user-facing documentation for tools
- [x] ~update [tutorials](https://github.com/AMReX-Codes/amrex-tutorials)?~

## Additional background

https://github.com/conda-forge/staged-recipes/pull/24294

## Checklist

The proposed changes:

### DIFF
--- a/Tools/Plotfile/CMakeLists.txt
+++ b/Tools/Plotfile/CMakeLists.txt
@@ -27,6 +27,11 @@ foreach( _exe IN LISTS _exe_names)
    if (AMReX_CUDA)
       set_source_files_properties(${_exe}.cpp PROPERTIES LANGUAGE CUDA)
    endif()
+
+    # Add prefix to each tool's name to make them unique when installed.
+    # This avoids potential collisions of names on user systems, e.g., in
+    # software packages (Spack/Conda/Debian/...).
+    set_target_properties(${_exe} PROPERTIES OUTPUT_NAME "amrex_${_exe}")
 endforeach()
 
 
--- a/.github/workflows/clang.yml
+++ b/.github/workflows/clang.yml
@@ -64,7 +64,7 @@ jobs:
             CLANG_TIDY_ARGS="--config-file=${{github.workspace}}/.clang-tidy --warnings-as-errors=*"
 
         export PATH=/tmp/my-amrex/bin:$PATH
-        which fcompare
+        which amrex_fcompare
 
         ctest --output-on-failure
 
--- a/.github/workflows/gcc.yml
+++ b/.github/workflows/gcc.yml
@@ -60,7 +60,7 @@ jobs:
             CLANG_TIDY_ARGS="--config-file=${{github.workspace}}/.clang-tidy --warnings-as-errors=*"
 
         export PATH=/tmp/my-amrex/bin:$PATH
-        which fcompare
+        which amrex_fcompare
 
         ctest --output-on-failure
 


Tools/Plotfile/CMakeLists.txt
```diff
@@ -27,6 +27,11 @@ foreach( _exe IN LISTS _exe_names)
    if (AMReX_CUDA)
       set_source_files_properties(${_exe}.cpp PROPERTIES LANGUAGE CUDA)
    endif()
+
+    # Add suffix to each tool's name to make them unique when installed.
+    # This avoids potential collisions of names on user sytems, e.g., in
```
ax3l:
```suggestion
    # This avoids potential collisions of names on user systems, e.g., in
```

Tools/Plotfile/CMakeLists.txt
```diff
@@ -27,6 +27,11 @@ foreach( _exe IN LISTS _exe_names)
    if (AMReX_CUDA)
       set_source_files_properties(${_exe}.cpp PROPERTIES LANGUAGE CUDA)
    endif()
+
+    # Add suffix to each tool's name to make them unique when installed.
+    # This avoids potential collisions of names on user systems, e.g., in
+    # software packages (Spack/Conda/Debian/...).
+    set_target_properties(${_exe} PROPERTIES OUTPUT_NAME "${_exe}_amrex")
```
isuruf: Wouldn't this change the executable name from `foo.exe` to `foo_amrex`? i.e. drop the `.exe` suffix?

WeiqunZhang: The old name does not have `.exe` either.

WeiqunZhang: I think using prefix here is better because all the amrex files will then be grouped together by default (by `ls` for example).

ax3l: The output name is before the executable suffix in cmake. It will keep .exe on Windows and none on Unix.

Refs.:
- https://cmake.org/cmake/help/v3.28/prop_tgt/OUTPUT_NAME.html#output-name
- https://cmake.org/cmake/help/v3.28/prop_tgt/SUFFIX.html#suffix

ax3l:
This change might break workflows, in particular:
- tests
- tutorials
- Python test harness by @zingale

and coordinated updates are welcome after/with merge :)

WeiqunZhang:
We do not need to change the GNU Make build. Currently the names are like `fcompare.gnu.ex` for GNU Make. So they have never consistent.

As for documentation, we only used gnumake as an example. https://amrex-codes.github.io/amrex/docs_html/Post_Processing.html#fextract Maybe we can add a paragraph about building these tools with cmake. We have already mentioned the build option here https://amrex-codes.github.io/amrex/docs_html/BuildingAMReX.html#id2. Maybe we can say something about they will be installed at `<amrex_installation_directdory>/bin/` and what the names are.

WeiqunZhang:
I did a case insensitive grep of `plotfile_tools`, `tools/plotfile` and `fcompare` in amrex-tutorials. Nothing showed up.

WeiqunZhang:
Re: regression_testing, even when cmake is used to build the application code and amrex, we still use gnumake built plotfile tools. See https://ccse.lbl.gov/pub/RegressionTesting1/MFIX-Exa/2023-10-19/BENCH01-Size0001.html.
```
Compilation command:
cmake_3.20 --build /scratch/AMReX_RegTesting/mfix/builddir -j 8 -- mfix

...

/scratch/AMReX_RegTesting/amrex/Tools/Plotfile//fcompare.gnu.ex ...
```

WeiqunZhang:
@jrood-nrel amr-wind and pele need to be updated after this is merged.

WeiqunZhang:
@ax3l if we do not change gnu make, this PR should be ready for merge.

WeiqunZhang (approved):

maximumcats:
I support leaving GNU make alone.

ax3l:
I think producing different artifacts for the different build systems - and documenting the tools with inconsistent names - will cause a lot of confusion. Especially since we build with CMake for all user-facing package managers (e.g., Spack, Conda).

https://github.com/AMReX-Codes/amrex/pull/3600#issuecomment-1771599522
That sounds good. The tools will actually be in `PATH` after install and generally available to users.
If we change the docs to (prominently/primarily? because users) refer to the CMake names that people will encounter, and leave GNUmake for hard-core devs with legacy names, I would feel better as well. But my primary concern is users getting confused.

In the end, we can avoid this though and future confusion by synchronizing the tool names...

WeiqunZhang:
I don't think the potential confusion is an issue if we clearly document it, especially since they end up in very different directories. The names have always been inconsistent. I think should just leave GNU make alone.

In any case, we need to update the documentation. Do you want to do it in this PR or a separate PR?

ax3l:
Currently on an airplane. Let's merge this and have a follow-up PR on docs? I can do that end of the week.

PR #3601, opened 2023-10-19
Revert "Add ability for GCC 8 in CMake to build fgradient which uses std::filesystem"

jrood-nrel:
Reverts AMReX-Codes/amrex#3523

PR #3602, opened 2023-10-19
Avoid std::filesystem

WeiqunZhang:
It has caused a lot of issues on systems with a default gcc <= 8.  PR #3523 was trying to the issue. But unfortunately it does not always work. So we are going to simply avoid std::filesystem.


ax3l (commented):
LGTM. Want to remove the build logic around it as well with the PR?

ax3l (approved):
Ah, I see #3601 now

PR #3599, opened 2023-10-18
Install Move Tools to `shared/amrex`

ax3l:
## Summary

Packaging AMReX for Conda, we realized that we break with common conventions on Unix systems by installing a top-level `Tools/` directory.

This moves its content in both the build directory and the install destination to `share/amrex/` and `<CMakePkgRoot>/AMReXCMakeModules/`, respectively.

## Additional background

https://github.com/conda-forge/staged-recipes/pull/24294

## Checklist

The proposed changes:
- [x] fix a bug or incorrect behavior in AMReX
- [ ] add new capabilities to AMReX
- [ ] changes answers in the test suite to more than roundoff level
- [ ] are likely to significantly affect the results of downstream AMReX users
- [ ] include documentation in the code and/or rst files, if appropriate

PR #3609, opened 2023-10-26
`AMReXBuildInfo.cmake`: AMReX_DIR

ax3l:
## Summary

The `AMReX_DIR` points if set to the CMake module path root.

The old logic did not work for me in a situation (ImpactX) where:
- AMReX is pre-installed but
- found through a superbuild of another transient lib (ABLASTR)

## Additional background

Follow-up to #3599

## Checklist

The proposed changes:

### DIFF
--- a/Tools/CMake/AMReXBuildInfo.cmake
+++ b/Tools/CMake/AMReXBuildInfo.cmake
@@ -38,7 +38,7 @@ include(AMReXTargetHelpers)
 #
 # Set paths
 #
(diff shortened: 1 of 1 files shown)


ax3l:
@WeiqunZhang Tested with WarpX and ImpactX. Does this also still work with MFIX?

WeiqunZhang:
Yes, it still works with MFIX's super and standalone builds.

WeiqunZhang (approved):

PR #3610, opened 2023-10-26
use hipPointerAttribute_t.type as HIP is removing hipPointerAttribute_t.memoryType

rakesroy:
## Summary
This replaces hipPointerAttribute_t.memoryType with hipPointerAttribute_t.type.

## Additional background
In ROCm6.0 hipPointerAttribute_t.memoryType will be removed from HIP.
Instead hipPointerAttribute_t.type to be used.
This is causing build failure in https://github.com/Exawind/amr-wind.git.
hipPointerAttribute_t.type has been existing since ROCm5.5, so this change will be backward compatible till ROCm5.5.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_GpuUtility.H
+++ b/Src/Base/AMReX_GpuUtility.H
@@ -63,7 +63,11 @@ namespace Gpu {
 #if defined(AMREX_USE_HIP)
         hipPointerAttribute_t attrib;
         hipError_t r = hipPointerGetAttributes(&attrib, p);
+#if defined(HIP_VERSION_MAJOR) && (HIP_VERSION_MAJOR < 6)
         return r == hipSuccess && attrib.memoryType == hipMemoryTypeDevice;
+#else
+        return r == hipSuccess && attrib.type == hipMemoryTypeDevice;
+#endif // (HIP_VERSION_MAJOR) && (HIP_VERSION_MAJOR < 6)
 #elif defined(AMREX_USE_CUDA)
         CUpointer_attribute attrib = CU_POINTER_ATTRIBUTE_MEMORY_TYPE;
         CUmemorytype mem_type = static_cast<CUmemorytype>(0);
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
Could we use the `HIP_VERSION_MAJOR` macro to maintain backward compatibility?

rakesroy:
Yes, HIP_VERSION_MAJOR  can be used to maintain backward compatibility.
I see existing AMREX code uses like `#if defined(AMREX_USE_HIP) && defined(HIP_VERSION_MAJOR) && (HIP_VERSION_MAJOR < 5)`.
Do you want me to add in similar way?

WeiqunZhang:
Yes, something like that. In this case it's already in HIP specific code, so we don't need to test AMREX_USE_HIp..

rakesroy:
I have updated this PR to to use .type when (HIP_VERSION_MAJOR >= 6), otherwise to use .memoryType for backwards compatibility.

Src/Base/AMReX_GpuUtility.H
```diff
@@ -63,7 +63,11 @@ namespace Gpu {
 #if defined(AMREX_USE_HIP)
         hipPointerAttribute_t attrib;
         hipError_t r = hipPointerGetAttributes(&attrib, p);
+#if defined(HIP_VERSION_MAJOR) && (HIP_VERSION_MAJOR >= 6)
+        return r == hipSuccess && attrib.type == hipMemoryTypeDevice;
+#elif defined(HIP_VERSION_MAJOR) && (HIP_VERSION_MAJOR < 6)
         return r == hipSuccess && attrib.memoryType == hipMemoryTypeDevice;
+#endif // (HIP_VERSION_MAJOR) && (HIP_VERSION_MAJOR >= 6)
```
WeiqunZhang:
```suggestion
#if defined(HIP_VERSION_MAJOR) && (HIP_VERSION_MAJOR < 6)
        return r == hipSuccess && attrib.memoryType == hipMemoryTypeDevice;
#else
        return r == hipSuccess && attrib.type == hipMemoryTypeDevice;
#endif // (HIP_VERSION_MAJOR) && (HIP_VERSION_MAJOR < 6)
```
I wonder if we should write it in a way that it will still work if in the future HIP_VERSION_MAJOR is not defined.

rakesroy: Updated the PR as per this suggestion.

WeiqunZhang (approved):

PR #3611, opened 2023-10-27
[WIP] FabArraySet class

eebasso:
## Summary
A container class for a set of FabArray's. Intended for use as a single object for input/output to linear operator classes that inherit from MLLinOpT. Useful when one has different components of a field living on slightly different grids due to differing IndexType's. One example is the Yee staggered grid, with the electric field being edge-centered and the magnetic field being face-centered.

## Additional background

Also adds the MLLinOpSet class as an example of what a linear operator acting FabArraySet's would look like.

The class labeled TEST are currently being used for debugging and not intended to be kept.


WeiqunZhang:
We probably don't need this anymore for linear solvers one staggered grids. See #3682

WeiqunZhang:
No longer needed.

PR #3618, opened 2023-11-08
solve_bicgstab: avoid iter==1 check

eebasso:
## Summary

This PR restructures solve_bicgstab such that the iter==1 check in the iter loop is avoided.

## Additional background

This is one of the split off changes to the CG and BiCGSTAB implementations in MLCGSolver.H


WeiqunZhang:
Could you please explain why the change is made? To prepare for more changes or just to avoid the `iter == 1` test? But now `rho = dotxy(rh,r);` appears twice.

eebasso:
The main reason is to avoid the `iter==1` test, but it will also help prepare for more changes related to the larger PR suggesting improvements to the BiCGSTAB implementation. It's true that now `rho = dotxy(rh,r)` appears twice, but I currently can't think of a way to eliminate that while also avoiding the `iter==1` test. One way to look at it is that everything above the `iter` loop is essentially the `iter=0` initialization of the variables, including `sol`, `r`, `rh`, `p`, and `rho`.

WeiqunZhang:
I think `if (iter == 1)` is much less an evil than duplicating `rho = dotxy(rh,r);`. Even if the latter were slightly better, that would not be a good reason to make the change, because any changes have the potential of introducing bugs.

eebasso:
Okay, it was been rewritten so that `rho = dotxy(rh,r)` is not duplicated. The unnecessary initializations of `alpha`, `omega`, and `rho_1` have also been eliminated. As far as bugs are concerned, what do the tests show?

PR #3619, opened 2023-11-08
solve_bicgstab: use linop.make instead of MF constructor

eebasso:
## Summary

This PR replaces the explicit use of MF constructors in ```MLCGSolverT<MF>::solve_bicgstab``` with calls to the `make` method of the linear operator associated with the MLCGSolverT object.

## Additional background

The use of `MLLinOpT<MF>::make` allows for inheritance of MLCGSolverT without an override of `solve_bicgstab` even if the MF class lacks a constructor with the same arguments as those MultiFab. For the MLMG template classes, `make` should generally be used instead of explicit MF constructors. Another PR to change this in `solve_cg` will follow once this is fully vetted and approved.


WeiqunZhang (approved):

WeiqunZhang:
Need to fix some warnings.

PR #3620, opened 2023-11-08
solve_bicgstab: initialize RT variables outside of iter loop

eebasso:
## Summary

A small PR that avoids initializing the RT variables for each iteration of the for loop in `solve_bicgstab`. This PR also makes all divide-by-0 checks use the same `if` structure with an explicit cast of RT(0.0).


WeiqunZhang:
To be honest, I don't like the changes.
```
for () {
    double xxx = ...;
    ...
}
```
is way better than
```
double xxx;
for () {
    xxx = ...;
    ...
}
```

eebasso:
Noted

PR #3624, opened 2023-11-13
solve_bicgstab: eliminate use of rho_1

eebasso:
## Summary

Eliminate use of `rho_1` and cut one division operation from the calculation of `beta`.


WeiqunZhang:
This division is outside any loops over cells. Its cost is tiny compared to all the MultiFab level works we have to do in the solver. I am not sure a tiny optimization like this is worth the time we have to spend on making sure the code does not break.

PR #3625, opened 2023-11-13
solve_bicgstab: consistent RT casting

eebasso:
## Summary

This PR makes sure that every instance in solve_bicgstab of setting an RT variable to 0 or checking if it is equal to 0 uses the explicit RT(0.0) cast.


WeiqunZhang:
Why do we need to change from
```
if (rhTv != RT(0.0))
{
    alpha = rho/rhTv;
}
else
{
    ret = 2; break;
}
```

to

```
if (rhTv == RT(0.0))
{
   ret = 2; break;
}
alpha = rho/rhTv;
```
I did not write the original code. Either way is fine with me. But why does it matter?

eebasso:
This change takes up less lines of code, and makes this zero check consistent in structure with the `rho == RT(0.0)` and `omega == RT(0.0)` checks. In my humble opinion, it looks more elegant this way.

asalmgren:
This is a small change for the sake of change;  in the best interest of the developers' time we would like contributors to focus their efforts on fixing erroneous code (aka bugs) and contributing new functionality.

PR #3628, opened 2023-11-14
CMake: AMReX_PARALLEL_LINK_JOBS

WeiqunZhang:
Add AMReX_PARALLEL_LINK_JOBS option to the CMake build system and use it for SYCL builds.

Adjust SYCL's RelWithDebInfo mode to the recommendation of Intel VTune.

This is a follow-up on #3498 that made similar changes to GNU Make.

## Checklist

The proposed changes:


.github/workflows/intel.yml
```diff
@@ -44,7 +44,8 @@ jobs:
             -DCMAKE_C_COMPILER=$(which icx)                \
             -DCMAKE_CXX_COMPILER=$(which icpx)             \
             -DCMAKE_Fortran_COMPILER=$(which ifx)          \
-            -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
+            -DCMAKE_CXX_COMPILER_LAUNCHER=ccache           \
+            -DAMReX_SYCL_PARALLEL_LINK_JOBS=2
```
ax3l: We have to be careful not to oversubscribe CI:
if we build more than one AMReX lib (1D, 2D, 3D), then CMake will with `--parallel 2` already run two linkers.

This is not the case here, so we are good :+1:

ax3l (approved):
Great option!

I think there is a general option now for vanilla LLVM and GCC's LTO as well that supports threading... Maybe we can unify them into a general name instead of calling it SYCL?

`AMReX_SYCL_PARALLEL_LINK_JOBS` -> `AMReX_PARALLEL_LINK_JOBS`?

WeiqunZhang:
> `AMReX_SYCL_PARALLEL_LINK_JOBS` -> `AMReX_PARALLEL_LINK_JOBS`?

Sounds good. I will make the change. But will limit its use to SYCL for now.

PR #3630, opened 2023-11-14
SYCL: Use get_multi_ptr instead of get_pointer

WeiqunZhang:
The latter has been deprecated in SYCL 2020.


WeiqunZhang:
Looks like the ICC CI fails to find `ifort`. So I disabled Fortran in that CI. Since ICC has been deprecated, we should convert that CI for non-SYCL code to use the new LLVM based compiler in the future.

atmyers (approved):

PR #3631, opened 2023-11-15
Fix typo in `MLMGT<MF>::getGradSolution` when `MF` is different from `AMF`

eebasso:
## Summary

This PR fixes a small typo in `MLMGT<MF>::getGradSolution` that causes an error when `AMR` is distinct from `MF`. A test demonstrating the fix is added for the case where `MF`=`MultiFab` and `AMF`=`iMultiFab`.

## Additional background

This typo was noticed when working on generalizing the `MLMG` template to the WIP class `FabArraySet`.

## Checklist

The proposed changes:


WeiqunZhang:
Thanks for fixing the bug! Could you merge development into this so that the Intel CIs can pass? Could you also use `fMultiFab` (i.e., `FabArray<BaseFab<float>>`) instead of `iMultiFab` in the test, because the data are floating point numbers?

eebasso:
You're welcome, and will do. 😀

WeiqunZhang (approved):

PR #3638, opened 2023-11-17
solve_cg: consistent RT casting

eebasso:
## Summary

This PR makes sure that every instance in `solve_cg` of setting an RT variable to 0 or checking if it is equal to 0 uses the explicit RT(0.0) cast.

## Additional background

This is similar to the PR made for `solve_bicgstab`.

### DIFF
--- a/Src/LinearSolvers/MLMG/AMReX_MLCGSolver.H
+++ b/Src/LinearSolvers/MLMG/AMReX_MLCGSolver.H
@@ -294,7 +294,7 @@ MLCGSolverT<MF>::solve_cg (MF& sol, const MF& rhs, RT eps_rel, RT eps_abs)
 
     for (; iter <= maxiter; ++iter)
     {
-        RT rho = dotxy(r,r);
+        const RT rho = dotxy(r,r);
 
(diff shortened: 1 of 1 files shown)


Src/LinearSolvers/MLMG/AMReX_MLCGSolver.H
```diff
@@ -306,21 +306,17 @@ MLCGSolverT<MF>::solve_cg (MF& sol, const MF& rhs, RT eps_rel, RT eps_abs)
...
-        {
-            alpha = rho/pw;
-        }
-        else
+        const RT pw = dotxy(p,q);
+        if ( pw == RT(0.0) )
         {
             ret = 1; break;
         }
+        const RT alpha = rho/pw;
```
eebasso, line 319: This restructuring of the divide by zero check allows `alpha` to be defined and declared in one line as a `const RT` instead of the current `RT` declaration.

asalmgren:
This is a small change for the sake of change;  in the best interest of the developers' time we would like contributors to focus their efforts on fixing erroneous code (aka bugs) and contributing new functionality.

PR #3644, opened 2023-11-29
Rename scalar constant variables to `ascalar` and `bscalar` in linear solvers to avoid confusion

eebasso:
## Summary

The PR renames the scalar constant variables towards the convention neutral names `ascalar` and `bscalar` as opposed to either `A` and `B` or `alpha` and `beta`.

## Additional background

Precedence for using the names `ascalar` and `bscalar` can be found in MLABecLaplacian.H, MLALaplacian.H, MLEBABecLap.H and various MyTest.H files. This naming scheme compliments that of the coefficients, often of the form `acoeffs`, `a_coeffs`, etc., as well as the functions named `setAScalar()`, `setACoeffs`, etc. For example, look at the usage of `acf` and `bcf` in MLNodeABecLap_3D_K.H.

Hopefully this change can be considered independently of whether the `A \alpha`  or `\alpha A` convention is preferred in the future. This is related to PR #3640.


WeiqunZhang:
Let's not worry about this.

PR #3648, opened 2023-11-29
GMRES

WeiqunZhang:
Add a GMRES solver class template

This is still work in progress. So the interface may not be stable yet.

### DIFF
--- a/Tests/LinearSolvers/ABecLaplacian_C/MyTest.cpp
+++ b/Tests/LinearSolvers/ABecLaplacian_C/MyTest.cpp
@@ -1,5 +1,7 @@
 #include "MyTest.H"
 
+#include <AMReX_GMRES.H>
+#include <AMReX_GMRES_MLMG.H>
 #include <AMReX_MLNodeABecLaplacian.H>
 #include <AMReX_MLABecLaplacian.H>
 #include <AMReX_MLPoisson.H>
@@ -20,7 +22,11 @@ MyTest::solve ()
     if (prob_type == 1) {
         solvePoisson();
     } else if (prob_type == 2) {
-        solveABecLaplacian();
+        if (use_gmres) {
+            solveABecLaplacianGMRES();
+        } else {
+            solveABecLaplacian();
+        }
     } else if (prob_type == 3) {
         solveABecLaplacianInhomNeumann();
     } else if (prob_type == 4) {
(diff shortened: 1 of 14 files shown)


Tests/LinearSolvers/ABecLaplacian_C/MyTest.cpp
```diff
@@ -460,6 +466,86 @@ MyTest::solveNodeABecLaplacian ()
...
+
+        mlmg.apply({&res}, {&solution[ilev]}); // res = L(sol)
+        MultiFab::Subtract(res, rhs[ilev], 0, 0, 1, 0); // now res = L(sol) - rhs
+        if (verbose) {
+            amrex::Print() << "Final residual = " << res.norminf(0)
+                           << " " << res.norm1(0) << " " << res.norm2(0) << std::endl;
+        }
+    }
+}
+
```
eebasso, line 548: Is some generalization of this loop eventually intended to be added to either `GRESMLMG` or `GMRES`?

Could we perhaps generalize this to replace `gmres` with other solver types that have thus far been used only as bottom solvers? Some examples would include the `CG` and `BiCGStab` methods. I imagine it could be any class that implements `solve` and `apply` methods and gets initialized by some MLLinOpT object.

As is, `MLMG` seems to force the user to always use the smoother for the non-bottom steps, which uses up the overwhelming majority of computational time. It would be nice if one could make that an option. I imagine the boundary conditions would have to handled more carefully in the case of the non-smoother methods at the finer levels.

It would also be nice if one could select `GMRES` as an option in `MLMG` as the bottom solver.

WeiqunZhang: Thanks for your comments. This PR is just first the step of adding GMRES support. Yes, eventually we would like to be able to use GMRES as a bottom solver in MLMG and use MLMG as a multigrid preconditioner for GMRES. The GMRES class will probably be more or less the same, but we need to make some changes to MLMG and the interface class GMRESMLMG.

> As is, MLMG seems to force the user to always use the smoother for the non-bottom steps,

MLMG calls `linop.smooth`. So it's up to the linear operator class on what to do. Nevertheless, I think the name `smooth` is appropriate, because that's how multigrid works. The relaxation step is supposed to smooth out the high-frequency modes of the error. I don't know what is in your mind as not a "smoother". I suspect whatever that is, it needs to behave like a smoother in terms of eliminating high-frequency errors for it to be effective.

atmyers (approved):

PR #3650, opened 2023-11-30
Robustify the Cache Cleanup Scripts

WeiqunZhang:
Make changes to handle the cases where the workflow names contain spaces. Note that none of the workflow names in amrex has spaces.


asalmgren (approved):

ax3l (commented):
Thank you!

PR #3667, opened 2023-12-13
Clang-Tidy in CI: Keep Going after Errors

WeiqunZhang:
Add `-k` to the make command running clang-tidy. With that, the jobs will keep going and show all the clang-tidy check errors instead of stopping on the first error.


ax3l (approved):
:+1:

PR #3671, opened 2023-12-14
PureSoA IdCpu fixes

AlexanderSinn:
## Summary

I noticed a few issues in AMReX while trying to update HiPACE++ for #3585.

Additionally, I would like to point out that `ParticleTile` has `push_back_real` and `push_back_int` functions but for PureSoA there is no `push_back_idcpu`, however this is not added in this PR.

## Checklist

The proposed changes:


atmyers (approved):

ax3l:
Thank you, @AlexanderSinn 👍

PR #3674, opened 2023-12-17
Add a for loop that is unrolled at compile time

maximumcats:
## Summary

The constexpr_for function is fully unrolled at compile time. This is useful for relatively short loops where some of the functions inside the loop are known to possible to evaluate at compile time and may be relatively expensive, so evaluating them at compile time rather than at runtime may be beneficial for performance reasons.

## Additional background

This has been used in AMReX-Astro/Microphysics successfully in the context of evaluating some nuclear reaction network quantities at compile time.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_Loop.H
+++ b/Src/Base/AMReX_Loop.H
@@ -211,6 +211,32 @@ void LoopConcurrentOnCpu (Box const& bx, int ncomp, F&& f) noexcept
     }}}}
 }
 
+// Implementation of "constexpr for" based on
+// https://artificial-mind.net/blog/2020/10/31/constexpr-for
+//
+// Approximates what one would get from a compile-time
+// unrolling of the loop
+// for (int i = 0; i < N; ++i) {
+//    f(i);
+// }
+//
+// The mechanism is recursive: we evaluate f(i) at the current
+// i and then call the for loop at i+1. f() is a lambda function
+// that provides the body of the loop and takes only an integer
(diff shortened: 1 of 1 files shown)


Src/Base/AMReX_Loop.H
```diff
@@ -211,6 +211,32 @@ void LoopConcurrentOnCpu (Box const& bx, int ncomp, F&& f) noexcept
...
+// unrolling of the loop
+// for (int i = 0; i < N; ++i) {
+//    f(i);
+// }
+//
+// The mechanism is recursive: we evaluate f(i) at the current
+// i and then call the for loop at i+1. f() is a lambda function
+// that provides the body of the loop and takes only an integer
+// i as its argument. It is assumed that at the loop call site
+// the lambda is declared [&] so that it can operate on the data
```
WeiqunZhang: I am not sure how what this comment about `[&]` means. If the loop body uses a pointer (a raw pointer or Array4), I don't think you need `[&]` and `[=]` would also work. (I am not saying = is better.) (Edited to add `not` in my last sentence. I unfortunately have the tendency of having "not" in my mind and not typing it.)

maximumcats: That was there because of its specific usage in Microphysics. You are right that it is not helpful when copied to AMReX for general usage. I've removed that comment just to avoid confusion.

WeiqunZhang (approved):

PR #3682, opened 2023-12-22
New Linear Solver: Curl of Curl

WeiqunZhang:
Add a new linear solver for curl (alpha curl E) + beta E = rhs in 2D & 3D,
where E is an array of 3 MultiFabs on edges, and alpha and beta are
scalars. It supports periodic, homogeneous Dirichlet, and symmetry
boundaries. At the symmetry boundary, the normal component changes the sign,
whereas the transverse components do not.

### DIFF
--- a/Src/LinearSolvers/MLMG/AMReX_MLCurlCurl.H
+++ b/Src/LinearSolvers/MLMG/AMReX_MLCurlCurl.H
@@ -77,24 +77,12 @@ public:
                  MF& res, const MF& crse_sol, const MF& crse_rhs,
                  MF& fine_res, MF& fine_sol, const MF& fine_rhs) const override;
 
-    void compFlux (int amrlev, const Array<MF*,AMREX_SPACEDIM>& fluxes,
-                   MF& sol, Location loc) const override;
-
-    void compGrad (int amrlev, const Array<MF*,AMREX_SPACEDIM>& grad,
-                   MF& sol, Location loc) const override;
-
-    void applyMetricTerm (int /*amrlev*/, int /*mglev*/, MF& /*rhs*/) const override {}
-
-    void unapplyMetricTerm (int /*amrlev*/, int /*mglev*/, MF& /*rhs*/) const override {}
-
     void prepareForSolve () override;
 
     [[nodiscard]] bool isSingular (int /*amrlev*/) const override { return false; }
     [[nodiscard]] bool isBottomSingular () const override { return false; }
 
     RT xdoty (int amrlev, int mglev, const MF& x, const MF& y, bool local) const override;
-
-    std::unique_ptr<MLLinOpT<MF>> makeNLinOp (int /*grid_size*/) const override { return nullptr; }
 
     [[nodiscard]] RT normInf (int amrlev, MF const& mf, bool local) const override;
 
--- a/Src/LinearSolvers/MLMG/AMReX_MLCurlCurl.H
+++ b/Src/LinearSolvers/MLMG/AMReX_MLCurlCurl.H
@@ -1,0 +1,155 @@
+#ifndef AMREX_ML_CURL_CURL_H_
+#define AMREX_ML_CURL_CURL_H_
+#include <AMReX_Config.H>
+
+#include <AMReX_MLLinOp.H>
+
+namespace amrex {
+
+/**
+ * \brief curl (alpha curl E) + beta E = rhs
+ *
+ * Here E is an Array of 3 MultiFabs on staggered grid, alpha is a positive
+ * scalar, and beta is a non-negative scalar.
+ *
+ * TODO: If beta is zero, the system could be singular.
+ * TODO: Try different restriction & interpolation strategies.
+ */
+class MLCurlCurl
+    : public MLLinOpT<Array<MultiFab,3> >
+{
+public:
+    using MF        = Array<MultiFab,3>;
+    using RT        = typename MLLinOpT<MF>::RT;
+    using BCType    = typename MLLinOpT<MF>::BCType;
+    using BCMode    = typename MLLinOpT<MF>::BCMode;
+    using StateMode = typename MLLinOpT<MF>::StateMode;
+    using Location  = typename MLLinOpT<MF>::Location;
+
+    MLCurlCurl () = default;
+    MLCurlCurl (const Vector<Geometry>& a_geom,
+                const Vector<BoxArray>& a_grids,
+                const Vector<DistributionMapping>& a_dmap,
+                const LPInfo& a_info = LPInfo());
+
+    void define (const Vector<Geometry>& a_geom,
+                 const Vector<BoxArray>& a_grids,
+                 const Vector<DistributionMapping>& a_dmap,
+                 const LPInfo& a_info = LPInfo());
+
+    void setScalars (RT a_alpha, RT a_beta) noexcept;
+
+    [[nodiscard]] std::string name () const override {
+        return std::string("curl of curl");
+    }
+
+    void setLevelBC (int amrlev, const MF* levelbcdata,
+                     const MF* robinbc_a = nullptr,
+                     const MF* robinbc_b = nullptr,
+                     const MF* robinbc_f = nullptr) override;
+
+    void restriction (int amrlev, int cmglev, MF& crse, MF& fine) const override;
+
+    void interpolation (int amrlev, int fmglev, MF& fine, const MF& crse) const override;
+
+    void interpAssign (int amrlev, int fmglev, MF& fine, MF& crse) const override;
+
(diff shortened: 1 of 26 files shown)


eebasso:
I’m happy to see this PR. Can we *please* swap the labeling of alpha and beta to match the existing conventions with the ABec solvers? 🙏

i.e., make it

alpha E + curl (beta curl E) = rhs

The beta coefficient has always been for the derivative term and alpha for the non-derivative term.

WeiqunZhang:
Will discuss this with Justin who gave me the equation.

Src/LinearSolvers/MLMG/AMReX_MLCurlCurl_K.H
```diff
@@ -0,0 +1,285 @@
...
+ * E_x        : (0,1,1)
+ * E_y        : (1,0,1)
+ * E_z        : (1,1,0)
+ * (curl E)_x : (1,0,0)
+ * (curl E)_y : (0,1,0)
+ * (curl E)_z : (0,0,1)
+ */
+
+AMREX_GPU_DEVICE AMREX_FORCE_INLINE
+void mlcurlcurl_adotx_x (int i, int j, int k, Array4<Real> const& Ax,
```
eebasso, line 19: I feel like this "adotx" language is confusing given the multiple components. If I understand correctly, "adotx" comes from thinking about the operator as `A`, the input as `x` and the output as `Ax`. This can be confusing when `x` itself has x, y, and z components. Why not replace "adotx" with "apply" or something else?

WeiqunZhang: I use `A dot x` because linear equations are often written as `A x = b`. For example, https://www.mathworks.com/help/matlab/ref/mldivide.html

WeiqunZhang:
This is ready for review. We still need to try different restriction and smoothing strategies. But that could be left to the future. More specifically, in stead of simple injection, I want to try weighted averaging like what we do for the nodal solver. That makes the restriction operator more consistent with prolongation.

ax3l:
I'll take a look.

Attn for review if you like: @dpgrote @JustinRayAngus @debog @RemiLehe :)

Src/LinearSolvers/MLMG/AMReX_MLCurlCurl.H
```diff
@@ -0,0 +1,155 @@
...
+public:
+    using MF        = Array<MultiFab,3>;
+    using RT        = typename MLLinOpT<MF>::RT;
+    using BCType    = typename MLLinOpT<MF>::BCType;
+    using BCMode    = typename MLLinOpT<MF>::BCMode;
+    using StateMode = typename MLLinOpT<MF>::StateMode;
+    using Location  = typename MLLinOpT<MF>::Location;
+
+    MLCurlCurl () = default;
+    MLCurlCurl (const Vector<Geometry>& a_geom,
```
ax3l, line 30: Does it make sense to add more Doxygen strings to APIs here (or complete what might be missing and duplicate in `MLLinOpT`?

Src/LinearSolvers/MLMG/AMReX_MLCurlCurl.H
```diff
@@ -0,0 +1,155 @@
...
+
+    void compFlux (int amrlev, const Array<MF*,AMREX_SPACEDIM>& fluxes,
+                   MF& sol, Location loc) const override;
+
+    void compGrad (int amrlev, const Array<MF*,AMREX_SPACEDIM>& grad,
+                   MF& sol, Location loc) const override;
+
+    void applyMetricTerm (int /*amrlev*/, int /*mglev*/, MF& /*rhs*/) const override {}
+
+    void unapplyMetricTerm (int /*amrlev*/, int /*mglev*/, MF& /*rhs*/) const override {}
```
ax3l: Does it make sense to runtime asserts in these implementations?

Src/LinearSolvers/MLMG/AMReX_MLCurlCurl.cpp
```diff
@@ -0,0 +1,552 @@
+
+#include <AMReX_MLCurlCurl.H>
+#include <AMReX_MLCurlCurl_K.H>
+
```
ax3l, line 4: Maybe?
```suggestion
#include <AMReX_MLCurlCurl.H>
#include <AMReX_MLCurlCurl_K.H>


```

Src/LinearSolvers/MLMG/AMReX_MLCurlCurl_K.H
```diff
@@ -0,0 +1,362 @@
...
+AMREX_GPU_DEVICE AMREX_FORCE_INLINE
+void mlcurlcurl_gsrb_x (int i, int j, int k,
+                        Array4<Real      > const& ex,
+                        Array4<Real const> const& ey,
+                        Array4<Real const> const& ez,
+                        Array4<Real const> const& rhs,
+                        Real alpha, Real beta,
+                        GpuArray<Real,AMREX_SPACEDIM> const& dxinv, int redblack)
+{
+    constexpr Real omega = Real(1.15);
```
ax3l, line 148: But `using namespace amrex::literals;` :D

Tests/LinearSolvers/CurlCurl/MyTest.cpp
```diff
@@ -0,0 +1,112 @@
+#include <AMReX_MLCurlCurl.H>
+
+#include "MyTest.H"
+
+#include <AMReX_MLMG.H>
```
ax3l, line 5:
```suggestion
#include "MyTest.H"

#include <AMReX_MLCurlCurl.H>
#include <AMReX_MLMG.H>
```

Tests/LinearSolvers/CurlCurl/initProb.cpp
```diff
@@ -0,0 +1,32 @@
+
+#include "MyTest.H"
+#include "initProb_K.H"
```
ax3l:
```suggestion
#include "initProb_K.H"

#include "MyTest.H"
```

Tests/LinearSolvers/CurlCurl/initProb_K.H
```diff
@@ -0,0 +1,107 @@
...
+
+AMREX_GPU_DEVICE AMREX_FORCE_INLINE
+void actual_init_prob (int i, int j, int k,
+                       amrex::GpuArray<amrex::Array4<amrex::Real>,3> const& rhs,
+                       amrex::GpuArray<amrex::Array4<amrex::Real>,3> const& sol,
+                       amrex::GpuArray<amrex::Real,AMREX_SPACEDIM> const& problo,
+                       amrex::GpuArray<amrex::Real,AMREX_SPACEDIM> const& dx,
+                       amrex::Real alpha, amrex::Real beta)
+{
+    using namespace amrex;
```
ax3l, line 14: To simplify a few of the `Real(...)` with `..._rt`:
```suggestion
    using namespace amrex;
    using namespace amrex::literals;
```

Tests/LinearSolvers/CurlCurl/main.cpp
```diff
@@ -0,0 +1,17 @@
+
+#include <AMReX.H>
+#include <AMReX_BLProfiler.H>
+#include "MyTest.H"
```
ax3l:
```suggestion
#include "MyTest.H"

#include <AMReX.H>
#include <AMReX_BLProfiler.H>
```

ax3l (approved):
This looks great and very clean. I added a few inline comments on:
- doc strings
- include orders
- using literals for compactness

Tests/LinearSolvers/CurlCurl/MyTest.cpp
WeiqunZhang: This was actually intentional for testing AMReX_MLCurlCurl.H.

Tests/LinearSolvers/CurlCurl/initProb_K.H
WeiqunZhang: I found `Real` to be easier for manipulation with regex.

WeiqunZhang: So I usually prefer Real to _rt.

Tests/LinearSolvers/CurlCurl/MyTest.cpp
WeiqunZhang: That is the product here is AMReX_MLCurlCurl.H, not the test.

Src/LinearSolvers/MLMG/AMReX_MLCurlCurl.H
WeiqunZhang: These functions are being called. I guess one could argue that these should not be pure virtual in the base class.

PR #3685, opened 2023-12-29
Disable optimization for computeRoundOffDomain

WeiqunZhang:
It appears dangerous to have optimization on when compiling the Geometry::computeRoundOffDomain function. In the past, we had to change the source code for Intel compilers. Recently, there appears to be issues with ROCm 6.0 when optimization is on. Thus, we disable optimization for that function.

Inline functions outsideRoundoffDomain and insideRoundoffDomain.


WeiqunZhang:
The issue appears to be integer overflow not really this function.

PR #3688, opened 2024-01-05
Add special named flag for invalid particles

atmyers:
This lets you do 

`p.m_idcpu = LongParticleIDs::InvalidParticleID;` 

instead of 

`amrex::ParticleIDWrapper{p.m_idcpu} = -1;`

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_Particle.H
+++ b/Src/Particle/AMReX_Particle.H
@@ -22,7 +22,6 @@ namespace
         constexpr Long LastParticleID = GhostParticleID - 2;
         constexpr Long DoSplitParticleID = GhostParticleID - 3;
         constexpr Long NoSplitParticleID = GhostParticleID - 4;
-        constexpr Long InvalidParticleID = 16777216;  // corresponds to id = -1, cpu = 0
     }
 
     using namespace LongParticleIds;
--- a/Src/Particle/AMReX_Particle.H
+++ b/Src/Particle/AMReX_Particle.H
@@ -15,13 +15,14 @@ namespace amrex {
 
 namespace
 {
-    /** Used for 64bit Long particle Ids as in AoS layout */
+    /** Special flags used for 64-bit Long particle Ids */
     namespace LongParticleIds {
         constexpr Long GhostParticleID = 549755813887L; // 2**39-1
         constexpr Long VirtualParticleID = GhostParticleID - 1;
         constexpr Long LastParticleID = GhostParticleID - 2;
         constexpr Long DoSplitParticleID = GhostParticleID - 3;
         constexpr Long NoSplitParticleID = GhostParticleID - 4;
+        constexpr Long InvalidParticleID = 16777216;
     }
 
     using namespace LongParticleIds;
--- a/Src/Particle/AMReX_Particle.H
+++ b/Src/Particle/AMReX_Particle.H
@@ -22,7 +22,7 @@ namespace
         constexpr Long LastParticleID = GhostParticleID - 2;
         constexpr Long DoSplitParticleID = GhostParticleID - 3;
         constexpr Long NoSplitParticleID = GhostParticleID - 4;
-        constexpr Long InvalidParticleID = 16777216;
+        constexpr Long InvalidParticleID = 16777216;  // corresponds to id = -1, cpu = 0
     }
 
     using namespace LongParticleIds;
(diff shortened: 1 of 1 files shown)


Src/Particle/AMReX_Particle.H
```diff
@@ -15,13 +15,14 @@ namespace amrex {
 
 namespace
 {
-    /** Used for 64bit Long particle Ids as in AoS layout */
+    /** Special flags used for 64-bit Long particle Ids */
     namespace LongParticleIds {
```
atmyers, line 19: wouldn't it be better for these to be `uint64_t` instead of `amrex::Long`, to match `m_idcpu`?

Src/Particle/AMReX_Particle.H
```diff
@@ -15,13 +15,14 @@ namespace amrex {
...
         constexpr Long GhostParticleID = 549755813887L; // 2**39-1
         constexpr Long VirtualParticleID = GhostParticleID - 1;
         constexpr Long LastParticleID = GhostParticleID - 2;
         constexpr Long DoSplitParticleID = GhostParticleID - 3;
         constexpr Long NoSplitParticleID = GhostParticleID - 4;
+        constexpr Long InvalidParticleID = 16777216;
```
atmyers:
```suggestion
        constexpr Long InvalidParticleID = 16777216;  // corresponds to id = -1, cpu = 0
```

Src/Particle/AMReX_Particle.H
ax3l: Good point, yes.

@AlexanderSinn, do you agree or do we overlook something?

Src/Particle/AMReX_Particle.H
```diff
@@ -15,13 +15,14 @@ namespace amrex {
 
 namespace
 {
-    /** Used for 64bit Long particle Ids as in AoS layout */
+    /** Special flags used for 64-bit Long particle Ids */
```
ax3l, line 18:
```suggestion
    /** Special flags used for 64-bit particle idcpu */
```

Src/Particle/AMReX_Particle.H
```diff
@@ -15,13 +15,14 @@ namespace amrex {
...
     namespace LongParticleIds {
         constexpr Long GhostParticleID = 549755813887L; // 2**39-1
```
ax3l, line 20: Would then potentially have a different suffix
```suggestion
        constexpr std::uint64_t GhostParticleID = 549755813887LL; // 2**39-1
```
or
```suggestion
        constexpr std::uint64_t GhostParticleID = 549755813887U; // 2**39-1
```
https://en.cppreference.com/w/cpp/language/integer_literal

Src/Particle/AMReX_Particle.H
AlexanderSinn: Since `ptd.idcpu()` is supposed to be used with `InvalidParticleID` (type uint64_t) and `ptd.id()` should be used with the other special ids (type amrex::Long), I think `InvalidParticleID` should be in a separate namespace with type `uint64_t` while the other special IDs remain how they are. Maybe it can be shortened to `ParticleIDCPUs::Invalid`.

Once all locations that make particles invalid are updated to this, it would allow for more efficient checks if a particle is invalid `if (ptd.idcpu(ip) == ParticleIDCPUs::Invalid)` and the use of a 40-bit id instead of a 39-bit one.

ax3l: Not sure about the last part, because what counts is memory alignment for transfer, which will be 64bit, but agree on everything else. Thanks!

Src/Particle/AMReX_Particle.H
```diff
@@ -25,6 +25,13 @@ namespace
     }
 
     using namespace LongParticleIds;
+
+    /** Flags used to set the entire uint64_t idcpu
+        to special values at once.
+     */
+    namespace ParticleIdCpus {
+        constexpr std::uint64_t Invalid = 16777216; // corresponds to id = -1, cpu = 0
+    }
```
ax3l, line 34: Adding this as well?
```suggestion
    }

    using namespace ParticleIdCpus;
```

AlexanderSinn (approved):

atmyers:
Note that doing 

`p.m_idcpu *= -1;`

is still okay, because if the original `p.id()` was positive, the new one will be negative, as long as the id is less than `LongParticleIds::LastParticleID`. Also, doing

`p.m_idcpu *= -1;`
`p.m_idcpu *= -1;`

will restore the original `p.id()` and `p.cpu()`.

PR #3696, opened 2024-01-11
BaseFab::lockAdd: Faster version of BaseFab::atomicAdd for OpenMP

WeiqunZhang:
For WarpX's Gordon Bell runs on Fugaku, Stephan Jaure of ATOS optimized atomicAdd using pthread spin locks. This commit implements Stephan's approach using OpenMP.

In AMReX::Initialize, we create a number of OMP locks. When BaseFab::lockAdd is called, we loop over planes in z-direction and try to acquire a lock with omp_test_lock. If it's successful, we can access the data in that z-plane without worrying about race conditions. This allows us to use simd instructions instead of omp atomic adds. If it's not successful, we will try a different z-plane. The process stops till all planes are processed.

### DIFF
--- a/Src/Base/AMReX_BaseFab.H
+++ b/Src/Base/AMReX_BaseFab.H
@@ -20,10 +20,8 @@
 #include <AMReX_Reduce.H>
 #include <AMReX_Gpu.H>
 #include <AMReX_Math.H>
-
-#ifdef AMREX_USE_OMP
-#include <omp.h>
-#endif
+#include <AMReX_OpenMP.H>
+#include <AMReX_MemPool.H>
 
 #include <cmath>
 #include <cstdlib>
@@ -978,6 +976,19 @@ public:
 #endif
     BaseFab<T>& atomicAdd (const BaseFab<T>& src, const Box& srcbox, const Box& destbox,
                            int srccomp, int destcomp, int numcomp=1) noexcept;
+
+    /**
+    * \brief Atomically add srcbox region of src FAB to destbox region of this FAB.
+    * The srcbox and destbox must be same size. When OMP is on, this uses OMP locks
+    * in the implementation and it's usually faster than atomicAdd.
+    */
+#if defined(AMREX_USE_GPU)
+    template <RunOn run_on>
+#else
+    template <RunOn run_on=RunOn::Host>
+#endif
+    BaseFab<T>& lockAdd (const BaseFab<T>& src, const Box& srcbox, const Box& destbox,
+                         int srccomp, int destcomp, int numcomp) noexcept;
 
     //! FAB SAXPY (y[i] <- y[i] + a * x[i]), in place.
 #if defined(AMREX_USE_GPU)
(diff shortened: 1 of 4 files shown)


WeiqunZhang:
I have used a test at https://github.com/WeiqunZhang/amrex-devtests/blob/main/fab_atomicAdd/main.cpp on my desktop machine and Frontier. The result look good. The new version is up to 10x faster, and its performance is similar to non-atomic add (which of course does not produce correct results).

atmyers (approved):

ax3l:
X-ref a Windows symbol linker regression in the 24.02 release:
https://github.com/conda-forge/warpx-feedstock/pull/80#issuecomment-1925448705

Src/Base/AMReX_BaseFab.H
```diff
@@ -3300,6 +3311,98 @@ BaseFab<T>::atomicAdd (const BaseFab<T>& src, const Box& srcbox, const Box& dest
...
+        for (int ip = 0; ip < nplanes; ++ip) {
+            mask[ip] = false;
+        }
+
+        int mm = 0;
+        int planes_left = nplanes;
+        while (planes_left > 0) {
+            AMREX_ASSERT(mm < nplanes);
+            auto const m = mm + dlo[AMREX_SPACEDIM-1];
+            int ilock = m % OpenMP::nlocks;
```
ax3l, line 3352: Should there be a runtime assert in this function if the active OMP threads exceed `OpenMP::nlocks`?

Src/Base/AMReX_OpenMP.H
```diff
@@ -13,7 +15,11 @@ namespace amrex::OpenMP {
     inline int in_parallel     () { return omp_in_parallel();     }
     inline void set_num_threads (int num) { omp_set_num_threads(num); }
 
-    void init_threads ();
+    void Initialize ();
+    void Finalize ();
+
+    static constexpr int nlocks = 128;
```
ax3l, line 21: A short inline comment what the relation between `nlocks` and number of OMP threads is (constraints or assumptions or performance impact) would be good here, I think.

ax3l:
X-ref a Windows symbol linker regression in the 24.02 release:
https://github.com/conda-forge/warpx-feedstock/pull/80#issuecomment-1925448705

I think we need to either write
```C++
// .H
extern AMREX_EXPORT std::array<omp_lock_t,nlocks> omp_locks;

// .cpp
AMREX_EXPORT std::array<omp_lock_t,nlocks> omp_locks;
```
or
```C++
// .H
AMREX_EXPORT std::array<omp_lock_t,nlocks> omp_locks;

// .cpp
std::array<omp_lock_t,nlocks> omp_locks;
```
or
```C++
// .H
// not sure that works...
extern "C" AMREX_EXPORT std::array<omp_lock_t,nlocks> omp_locks;

// .cpp
std::array<omp_lock_t,nlocks> omp_locks;
```
because this involves the STL in the type...

Src/Base/AMReX_BaseFab.H
WeiqunZhang: The threads are competing for locks. Even if OpenMP::nlocks is 1, it would still work with any number of threads.

ax3l:
To understand this better: why is this global variable declared `extern` and not `static` (or no annotation)? Do we need to change it from across TUs?

WeiqunZhang:
If it's static, we will have a copy for every TU. It would be a waste and it would be tricky (though possible) to initialize them properly.

ax3l:
Got it, thx!

I wonder if we might need to use a plain C array here instead of `std::array`?

WeiqunZhang:
`std::array` is probably fine. We have `Src/Base/AMReX.H:        extern AMREX_EXPORT std::string exename;` and that was not an issue. So we just need to add `AMREX_EXPORT`. I don't think we need it in cpp file.

WeiqunZhang:
We need to add a CI to check the windows link issue.

ax3l:
> So we just need to add AMREX_EXPORT. I don't think we need it in cpp file.

I wonder if I miss it: where would you add `AMREX_EXPORT` in this patch? (Don't we declare it already in the .H file?)

ax3l:
> We need to add a CI to check the windows link issue.

I am a bit surprised we did not catch it in WarpX. We have a CI entry "Clang C++17 w/ OMP w/o MPI" with ClangCl that builds shared and with OpenMP, just like in conda-forge...

WeiqunZhang:
Oh, I missed it. I though the issue was I forgot about adding `AMREX_EXPORT`.

ax3l:
I wish, but this patch looks so clean, I cannot spot it yet :D

Could also be a missing include, but they look clean here as well.

WeiqunZhang:
Maybe amrex was not built with openmp support?

ax3l:
Yes, or the compiler or OpenMP backend implementation were mismatched. I cannot see that yet, I checked compiler options and looks identical:
https://github.com/conda-forge/warpx-feedstock/pull/80#issuecomment-1925460484

PR #3698, opened 2024-01-15
Add a linspace-like function in AMReX_Algorithm.H

lucafedeli88:
## Summary

I would like to propose to add a  linspace-like function in AMReX_Algorithm.H , since filling a container with equally spaced numbers is a task that occurs rather frequently. This PR proposes a possible implementation.
The container is modified in place only if it has at least 2 elements. 

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_Algorithm.H
+++ b/Src/Base/AMReX_Algorithm.H
@@ -208,7 +208,9 @@ namespace amrex
 #endif
     }
 
-    template<typename ItType, typename ValType>
+    template<typename ItType, typename ValType,
+        typename std::enable_if<std::is_floating_point<typename ItType::value_type>::value,int>::type = 0,
+        typename std::enable_if<std::is_floating_point<ValType>::value,int>::type = 0>
     AMREX_GPU_HOST_DEVICE
     void linspace (ItType first, const ItType& last, const ValType& start, const ValType& stop)
     {
--- a/Src/Base/AMReX_Algorithm.H
+++ b/Src/Base/AMReX_Algorithm.H
@@ -209,7 +209,10 @@ namespace amrex
     }
 
     template<typename ItType, typename ValType,
-        typename std::enable_if<std::is_floating_point<typename ItType::value_type>::value,int>::type = 0,
+        typename std::enable_if<
+            std::is_floating_point<typename std::iterator_traits<ItType>::value_type>::value &&
+            std::is_floating_point<ValType>::value,
+        int>::type = 0>
         typename std::enable_if<std::is_floating_point<ValType>::value,int>::type = 0>
     AMREX_GPU_HOST_DEVICE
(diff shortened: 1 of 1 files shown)


lucafedeli88:
Hello everyone. Do you think that it could be interesting to offer this function in AMReX ?
It would be quite useful in a radiation module that we are currently developing for the WarpX code.

WeiqunZhang:
I think such a function is useful. I wonder if we should make sure that both `typename std::iterator_traits<ItType>::value_type` and `ValType` are floating point types. Otherwise calls like `linspace(..., 0, 3)` will not work as expected.

Src/Base/AMReX_Algorithm.H
```diff
@@ -208,6 +208,20 @@ namespace amrex
...
     }
 
+    template<typename ItType, typename ValType>
+    AMREX_GPU_HOST_DEVICE
+    void linspace (ItType first, const ItType& last, const ValType& start, const ValType& stop)
+    {
+        const std::ptrdiff_t count = last-first;
+        if (count >= 2){
+            const auto delta = (stop - start)/(count - 1);
+            for (int i = 0; i < count-1; ++i){
```
WeiqunZhang: Could use `std::ptrdiff_t` instead of `int`.

lucafedeli88: It's a good idea. I've replaced `int` with `std::ptrdiff_t`

lucafedeli88:
> I think such a function is useful. I wonder if we should make sure that both `typename std::iterator_traits<ItType>::value_type` and `ValType` are floating point types. Otherwise calls like `linspace(..., 0, 3)` will not work as expected.

I agree! 
Is it OK like that?
```
    template<typename ItType, typename ValType,
        typename std::enable_if<std::is_floating_point<typename ItType::value_type>::value,int>::type = 0,
        typename std::enable_if<std::is_floating_point<ValType>::value,int>::type = 0>
```
It seems to work, although the syntax is quite verbose!

Src/Base/AMReX_Algorithm.H
```diff
@@ -208,6 +208,22 @@ namespace amrex
 #endif
     }
 
+    template<typename ItType, typename ValType,
+        typename std::enable_if<std::is_floating_point<typename ItType::value_type>::value,int>::type = 0,
```
WeiqunZhang: I think we need to use `iterator_traits` otherwise raw pointers would not work as `ItType`.

We could also combine the two enable_ifs into one.

WeiqunZhang: https://en.cppreference.com/w/cpp/iterator/iterator_traits

lucafedeli88: Thanks for the link. This version uses `iterator_traits` and combines the two `enable_if`

WeiqunZhang (approved):

PR #3700, opened 2024-01-16
lockAdd: case of 2D plane in 3D

WeiqunZhang:
## Summary

In HiPACE++, atomicAdd is used on 2d x & y planes even though AMREX_SPACEDIM is 3. In that case, we would have all threads competing for a single lock in the previous implementation of lockAdd. This PR fixes this use case by having locks associated with the y-direction when the number of cells in the z-direction is 1.

## Additional background

https://github.com/Hi-PACE/hipace/pull/1059

## Checklist

The proposed changes:


AlexanderSinn (approved):
I tested HiPACE++ running on two 48-core CPUs with `4095*4095*100` cells and a tile size of `256^2`. While the lockAdd version from this PR is faster, particularly at lower thread counts, it does not solve the bad scaling past 48 threads.
image

PR #3701, opened 2024-01-16
Add macro for loop unrolling across compilers

atmyers:
The proposed changes:

### DIFF
--- a/Src/Base/AMReX_Extension.H
+++ b/Src/Base/AMReX_Extension.H
@@ -137,10 +137,10 @@
 #define AMREX_TO_STRING_HELPER(X)   #X
 #define AMREX_TO_STRING(X)          AMREX_TO_STRING_HELPER(X)
 
-#if defined(__GNUC__)
-#define AMREX_UNROLL_LOOP(n)  _Pragma(AMREX_TO_STRING(GCC unroll (n)))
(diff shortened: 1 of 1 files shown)


Src/Base/AMReX_Extension.H
```diff
@@ -133,6 +133,21 @@
 #define AMREX_NO_INLINE
 #endif
 
+// unroll loop
+#define AMREX_TO_STRING_HELPER(X)   #X
+#define AMREX_TO_STRING(X)          AMREX_TO_STRING_HELPER(X)
+
+#if defined(__GNUC__)
+#define AMREX_UNROLL_LOOP(n)  _Pragma(AMREX_TO_STRING(GCC unroll (n)))
+#elif defined(__clang__) || defined(__CUDACC__) || defined(__HIP__) || defined(__INTEL_CLANG_COMPILER))
```
WeiqunZhang: I think the test for `__clang__` should be done before `__GNUC__` before clang defines `__GNUC__` too.

WeiqunZhang: For example,
```
$ clang++-12 -dM -E - < /dev/null | grep "GNUC"
#define __GNUC_MINOR__ 2
#define __GNUC_PATCHLEVEL__ 1
#define __GNUC_STDC_INLINE__ 1
#define __GNUC__ 4
```

WeiqunZhang (approved):

PR #3702, opened 2024-01-17
Follow-on to #3701

atmyers:
The parenthesis are not needed and some compilers warn about them.

The proposed changes:


WeiqunZhang:
Does `#pragma unroll n+1` work? At least for cuda `#pragma unroll (n+1)` should work.

WeiqunZhang:
How about `n   +   1` etc.? I am wondering if having `()` provides more protection.

WeiqunZhang (approved):

PR #3711, opened 2024-01-19
GNU Make: Link flags

WeiqunZhang:
The link information returned by mpich may include some compiler flags like `-fallow-argument-mismatch -g`. On Mac, this causes an issue for the clang based linker because `-fallow-argument-mismatch` is a gfortran flag. So we need to filter them out.


WeiqunZhang:
@RevathiJambunathan Could you give this a try?

RevathiJambunathan:
@WeiqunZhang Thank you !! 
This works!!!!

asalmgren (approved):

PR #3713, opened 2024-01-21
Remove various deprecated stuff not prefixed by amrex or bl

WeiqunZhang:
Most codes should not be affected by the removal. Some old codes that used to use BoxLib might need updates.

AMReX_ArrayLim.H contains macros that are useful for calling Fortran subroutines from C++. The old macros that do not start with `AMREX_` or `BL_` have been removed.

Some macros defining BC types in AMReX_BC_TYPES.H have been removed. One should use the C++ enum types (e.g., amrex::BCType::reflect_odd) or Fortran module amrex_bc_types_module.

Macros defined in AMReX_CONSTANTS.H for Fortran have been removed. One should use amrex_constants_module instead.

Fortran subroutines filcc and filccn have been removed. One should use amrex_filcc and amrex_filccn instead.

The REAL_T macro for Fortran has been removed. One should use amrex_real in amrex_fort_module instead.

The blutilrand function has been removed. If one needs random numbers in Fortran, one should amrex_random in amrex_fort_module. Apparently this function is very old. Recently, we had a multiple definition problem when we tried to use amrex in a code containing this function that was copied from BoxLib many years ago.

Macros like D_DECL have been removed from AMReX_SPACE.H. One should use the AMREX_ version instead.

Macros like LO_DIRICHLET have been removed from AMReX_LO_BCTYPES.H. One should use the AMREX_ version instead.


zingale:
Castro and MAESTROeX are ready for this change now

zingale:
The Castro failure is because it is testing the last tag and not development

asalmgren (approved):

JhonCordova:
Hello, months ago I used this code to compare energy spectra that had been produced in[ Motheau's work](https://arxiv.org/pdf/1902.06665). Now, I'm having trouble with AMReX to replicate the spectra and compare other turbulence models. Does anyone have any recommendations on what I should change or modify to reuse this [code](https://github.com/AMReX-Astro/AmrDeriveSpectrum)?

WeiqunZhang:
Maybe you can use an old version of AMReX. Or rewrite that code. It should be much easier now to read a plotfile into MultiFab and we support FFT. Something like
```
PlotFileData plotfile("name");
int level = 0;
MultiFab mf = plotfile.get(level);
FFT::R2C<Real,FFT::Direction::forward> r2c(plotfile.probDomain(level));
auto const& [cba, cdm] = r2c.getSpectralDataLayout();
cMultiFab spectral_data(cba, cdm, 1, 0);
int icomp = 3; // component 3
MultiFab real_data(mf, amrex::make_alias, icomp, 1);
r2c.forward(read_data, spectral_data);
// spectral_data is now the result of real-to-complex forward FFT of the level 0 component 3 data.
```

zingale:
yes, we should rewrite this code.  This should be discussed in that repo.  I'll add an issue there.

PR #3717, opened 2024-01-23
relax constraint that real_comp_names.size() == pc.NumRealComps() + NStructReal for pure SoA plotfiles

atmyers:
The proposed changes:


ax3l (approved):
Thank you, yay! :rocket: :sparkles:

PR #3724, opened 2024-01-24
Provide portable Gpu::Atomic::Multiply and Gpu::Atomic::Divide implemented with CAS.

atmyers:
The proposed changes:

### DIFF
--- a/Src/Base/AMReX_GpuAtomic.H
+++ b/Src/Base/AMReX_GpuAtomic.H
@@ -17,7 +17,7 @@ namespace Gpu::Atomic {
 // For Exch and CAS, the data type is generic.
 // All these functions are non-atomic in host code!!!
 // If one needs them to be atomic in host code, use HostDevice::Atomic::*.  Currently only
-// HostDevice::Atomic is supported.  We could certainly add more.
+// HostDevice::Atomic::Add is supported.  We could certainly add more.
 
 namespace detail {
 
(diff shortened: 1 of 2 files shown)


Src/Base/AMReX_GpuAtomic.H
```diff
@@ -526,6 +526,136 @@ namespace detail {
...
+//  Multiply
+////////////////////////////////////////////////////////////////////////
+
+#ifdef AMREX_USE_GPU
+
+    template<class T>
+    AMREX_GPU_DEVICE AMREX_FORCE_INLINE
+    T Multiply_device (T* const /*prod*/, T const /*value*/) noexcept
+    {
+        amrex::Abort("Gpu::Atomic::Multiply is not implemented for this data type.");
```
WeiqunZhang: It might be better to remove this function template. It's better to have a compile time error than a run time error.

WeiqunZhang: If we want, we could also implement this in a different way to support user defined types in case they have overloaded operators. Something like
```
template <typename T, std::enable_if_t<sizeof(T) == sizeof(int), int> = 0>
T Multiply_device (T* const prod, T const value) noexcept
{
    return detail::atomic_op<T, int>(prod,value,amrex::Multiplies<T>());
}

template <typename T, std::enable_if_t<sizeof(T) == sizeof(unsigned long long), int> = 0>
T Multiply_device (T* const prod, T const value) noexcept
{   
    return detail::atomic_op<T, unsigned long long>(prod,value,amrex::Multiplies<T>());
}
```

Src/Base/AMReX_GpuAtomic.H
```diff
@@ -526,6 +528,78 @@ namespace detail {
...
+    template <typename T, std::enable_if_t<sizeof(T) == sizeof(int), int> = 0>
+    AMREX_GPU_DEVICE AMREX_FORCE_INLINE
+    T Divide_device (T* const quot, T const value) noexcept
+    {
+        return detail::atomic_op<T, int>(quot,value,amrex::Divides<T>());
+    }
+
+    template <typename T, std::enable_if_t<sizeof(T) == sizeof(unsigned long long), int> = 0>
+    AMREX_GPU_DEVICE AMREX_FORCE_INLINE
+    T Divides_device (T* const quot, T const value) noexcept
```
```suggestion
    T Divide_device (T* const quot, T const value) noexcept
```

WeiqunZhang (approved):

PR #3726, opened 2024-01-28
Disable SYCL on Nvidia and AMD CIs

WeiqunZhang:
The Codeplay APT has stopped working for a few weeks now. It fails with error messages like

    The following packages have unmet dependencies:
     oneapi-nvidia-12.0 : Depends: intel-basekit (>= 2024.0.2) but it is not going to be installed
    E: Unable to correct problems, you have held broken packages.
    Error: Process completed with exit code 1.

To avoid confusion, let's disable them for now.


ax3l:
cc @nmnobre just FYI, we will disable this for now due to breakage. Can be reintroduced via PR anytime :)

ax3l (approved):

PR #3727, opened 2024-01-29
Fix: `MPI_Comm_free(MPI_COMM_NULL)`

ax3l:
## Summary

Avoid that a minimal cycle of MPI and AMReX init and finalize segfaults.

Situation: compiled with MPI, but run serially.

## Additional background

First seen in https://github.com/ECP-WarpX/impactx/pull/510

for the `test_impactx_module` pytest.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_ParallelContext.H
+++ b/Src/Base/AMReX_ParallelContext.H
@@ -106,7 +106,7 @@ inline void set_last_frame_ofs (const std::string & filename) {
 }
 //! Note that it's the user's responsibility to free the MPI_Comm
 inline void pop () {
-    if (frames.size() > 0) {
+    if (!frames.empty()) {
         frames.pop_back();
     }
 }
--- a/Src/Base/AMReX_ParallelContext.H
+++ b/Src/Base/AMReX_ParallelContext.H
@@ -106,7 +106,7 @@ inline void set_last_frame_ofs (const std::string & filename) {
 }
 //! Note that it's the user's responsibility to free the MPI_Comm
 inline void pop () {
-    if (frames.size() > 0u) {
+    if (frames.size() > 0) {
         frames.pop_back();
     }
 }
--- a/Src/Base/AMReX_ParallelContext.H
+++ b/Src/Base/AMReX_ParallelContext.H
@@ -105,7 +105,11 @@ inline void set_last_frame_ofs (const std::string & filename) {
     frames.back().set_ofs_name(filename);
 }
 //! Note that it's the user's responsibility to free the MPI_Comm
-inline void pop () { frames.pop_back(); }
+inline void pop () {
+    if (frames.size() > 0u) {
+        frames.pop_back();
+    }
+}
 
 }
 
--- a/Src/Base/AMReX_ParallelDescriptor.cpp
+++ b/Src/Base/AMReX_ParallelDescriptor.cpp
@@ -428,7 +428,7 @@ EndParallel ()
         m_mpi_ops.clear();
     }
 
-    if (!call_mpi_finalize) {
+    if (!call_mpi_finalize && m_comm != MPI_COMM_NULL) {
         BL_MPI_REQUIRE( MPI_Comm_free(&m_comm) );
     }
     m_comm = MPI_COMM_NULL;


Src/Base/AMReX_ParallelContext.H
```diff
@@ -105,7 +105,11 @@ inline void set_last_frame_ofs (const std::string & filename) {
     frames.back().set_ofs_name(filename);
 }
 //! Note that it's the user's responsibility to free the MPI_Comm
-inline void pop () { frames.pop_back(); }
+inline void pop () {
+    if (frames.size() > 0u) {
```
WeiqunZhang: This might just hide errors somewhere else. One is not supposed to pop an empty container. For example, `std::vector::pop_back()` on empty container is undefined behavior.

Src/Base/AMReX_ParallelDescriptor.cpp
```diff
@@ -428,7 +428,7 @@ EndParallel ()
         m_mpi_ops.clear();
     }
 
-    if (!call_mpi_finalize) {
+    if (!call_mpi_finalize && m_comm != MPI_COMM_NULL) {
```
WeiqunZhang, line 431: I think we had better understand why this is needed. It does not hurt to add the additional test. But how did it get into the situation?

WeiqunZhang:
> Situation: compiled with MPI, but run serially.

What does that mean exactly?

ax3l:
>> Situation: compiled with MPI, but run serially.
>
> What does that mean exactly?

This test:
https://github.com/ECP-WarpX/impactx/blob/4e2a6e5e227b90a82db4aec9315fa8ab94222bf5/tests/python/test_impactx.py#L19-L24

runs:
https://github.com/ECP-WarpX/impactx/blob/4e2a6e5e227b90a82db4aec9315fa8ab94222bf5/tests/python/conftest.py#L11-L41
- `MPI_Init()` (triggered by `from mpi4py import MPI`)
- `amrex::Initialize(...)` ([no passed MPI comm: defaulted](https://github.com/AMReX-Codes/pyamrex/blob/24.01/src/Base/AMReX.cpp#L100))
- `amrex::Finalize()`  - crashes here
- `MPI_Finalize()` (does not get there)

Src/Base/AMReX_ParallelContext.H
ax3l:
```suggestion
    if (frames.size() > 0) {
```

ax3l: I saw this being called in particular in `ParallelContext::EndParallel()`:
https://github.com/AMReX-Codes/amrex/blob/ab99ea69089f4fffbfd727ed965d5ceb3d905baa/Src/Base/AMReX_ParallelDescriptor.cpp#L436

Src/Base/AMReX_ParallelDescriptor.cpp
ax3l: I think the situation is:
- MPI is already initialized externally.
- But I do not pass the MPI communicator to `amrex::Initialize()` and rely on the default argument `MPI_COMM_WORLD` instead.

Then this part should not be called:
https://github.com/AMReX-Codes/amrex/blob/ab99ea69089f4fffbfd727ed965d5ceb3d905baa/Src/Base/AMReX_ParallelDescriptor.cpp#L293-L309

and instead the default of `m_comm` is pushed back onto the stack:
https://github.com/AMReX-Codes/amrex/blob/ab99ea69089f4fffbfd727ed965d5ceb3d905baa/Src/Base/AMReX_ParallelDescriptor.cpp#L332

the default value is `MPI_COMM_NULL`:
https://github.com/AMReX-Codes/amrex/blob/ab99ea69089f4fffbfd727ed965d5ceb3d905baa/Src/Base/AMReX_ParallelDescriptor.cpp#L332

ax3l: So I agree, should fix `StartParallel` instead.

ax3l: Huh, there is an 
```C++
else {
    MPI_Comm_dup(a_mpi_comm, &m_comm);
    call_mpi_finalize = 0;
}
```
Why does that not work ... o.0

Src/Base/AMReX_ParallelContext.H
```diff
@@ -105,7 +105,11 @@ inline void set_last_frame_ofs (const std::string & filename) {
     frames.back().set_ofs_name(filename);
 }
 //! Note that it's the user's responsibility to free the MPI_Comm
-inline void pop () { frames.pop_back(); }
+inline void pop () {
+    if (frames.size() > 0) {
```
ax3l:
```suggestion
    if (!frames.empty()) {
```

WeiqunZhang:
Does not make sense. Maybe `MPI_Comm_dup` failed. In that case, we should check its error code. Could we check the return value of https://github.com/AMReX-Codes/amrex/blob/ab99ea69089f4fffbfd727ed965d5ceb3d905baa/Src/Base/AMReX_ParallelDescriptor.cpp#L307 But even if it does fail, the ParallelContext's stack should not be empty unless that version of ParallelDescriptor::StartParallel was never called.

ax3l:
Yes, a bit confusing.
Another part that might cause this: we also call init-finalize multiple times in a row.

WeiqunZhang:
> Yes, a bit confusing. Another part that might cause this: we also call init-finalize multiple times in a row.

If that's the case, there are probably other issues. We should try to make the whole thing more robust instead of protecting pop on empty container.

ax3l:
This is what we try to do:
- MPI Init
- AMReX Init
- AMReX Finalize
- AMReX Init
- AMReX Finalize
- AMReX Init
- AMReX Finalize 
- MPI Finalize

Maybe AMReX finalize is called more often? (Printf debug)

ax3l:
Yes, I think the logic somehow called finalize twice...
Pretty weird, because I check `amrex::Initialized()` before calling it, but could not reproduce a valid state that shows it.

https://github.com/ECP-WarpX/impactx/pull/524

PR #3733, opened 2024-01-30
Add helper function for setting id and cpu simultaneously.

### LINKED ISSUE #3731: Assigning `idcpu` in one go
I think it would be efficient if we had a function that is `make64bit(cpu, id)` that we can assign straight to the `idcpu` array value:
```C++
pa_idcpu[ip] = newFunction(cpuid, pid+ip);
```

Do we have such a thing already?

I work around this now via
```C++
pa_idcpu[ip] = 0;
amrex::ParticleIDWrapper{pa_idcpu[ip]} = pid+ip;
amrex::ParticleCPUWrapper{pa_idcpu[ip]} = cpuid;
```
to avoid valgrind (uninit data read) warnings during the first assign.

x-ref: https://github.com/ECP-WarpX/WarpX/pull/4653

atmyers:
Also sneaking in a small ignore_unused fix.

Close #3731.

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_Particle.H
+++ b/Src/Particle/AMReX_Particle.H
@@ -195,8 +195,9 @@ struct ConstParticleCPUWrapper
  */
 std::uint64_t SetParticleIDandCPU (Long id, int cpu) {
     std:uint64_t idcpu = 0;
-    ParticleIDWrapper(idcpu) = id;
-    ParticleCPUWrapper(idcpu) = cpu;
+    ParticleIDWrapper{idcpu} = id;
+``
+    ParticleCPUWrapper{idcpu} = cpu;
     return idcpu;
 }
 
--- a/Src/Particle/AMReX_Particle.H
+++ b/Src/Particle/AMReX_Particle.H
@@ -194,7 +194,7 @@ struct ConstParticleCPUWrapper
  * to avoid writing twice into the same memory bank.
  */
 std::uint64_t SetParticleIDandCPU (Long id, int cpu) {
-    std:uint64_t idcpu = 0;
+    std::uint64_t idcpu = 0;
     ParticleIDWrapper{idcpu} = id;
     ParticleCPUWrapper{idcpu} = cpu;
     return idcpu;
--- a/Src/Particle/AMReX_Particle.H
+++ b/Src/Particle/AMReX_Particle.H
@@ -188,6 +188,11 @@ struct ConstParticleCPUWrapper
     operator int () const noexcept { return static_cast<int>(m_idata & 0x00FFFFFF); }
 };
 
+/** Set the idcpu value at once, based on a particle id and cpuid
+ *
+ * This can be used in initialization and assignments,
+ * to avoid writing twice into the same memory bank.
+ */
 std::uint64_t SetParticleIDandCPU (Long id, int cpu) {
     std:uint64_t idcpu;
     ParticleIDWrapper(idcpu) = id;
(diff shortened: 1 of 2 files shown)


Src/Particle/AMReX_Particle.H
```diff
@@ -188,6 +188,12 @@ struct ConstParticleCPUWrapper
     operator int () const noexcept { return static_cast<int>(m_idata & 0x00FFFFFF); }
 };
 
+std::uint64_t SetParticleIDandCPU (Long id, int cpu) {
+    std:uint64_t idcpu;
```
ax3l: to avoid read-before-write warning:
```suggestion
    std:uint64_t idcpu = 0;
```

This is in register (not in global memory like the original work-around), so this should be good.

Src/Particle/AMReX_Particle.H
```diff
@@ -188,6 +188,12 @@ struct ConstParticleCPUWrapper
     operator int () const noexcept { return static_cast<int>(m_idata & 0x00FFFFFF); }
 };
 
+std::uint64_t SetParticleIDandCPU (Long id, int cpu) {
```
ax3l:
```suggestion
/** Set the idcpu value at once, based on a particle id and cpuid
 *
 * This can be used in initialization and assignments,
 * to avoid writing twice into the same memory bank.
 */
std::uint64_t SetParticleIDandCPU (Long id, int cpu) {
```

ax3l (approved):
Thank you!

Src/Particle/AMReX_Particle.H
```diff
@@ -188,6 +188,17 @@ struct ConstParticleCPUWrapper
...
 };
 
+/** Set the idcpu value at once, based on a particle id and cpuid
+ *
+ * This can be used in initialization and assignments,
+ * to avoid writing twice into the same memory bank.
+ */
+std::uint64_t SetParticleIDandCPU (Long id, int cpu) {
+    std:uint64_t idcpu = 0;
+    ParticleIDWrapper(idcpu) = id;
```
atmyers:
```suggestion
    ParticleIDWrapper{idcpu} = id;
``

File: Src/Particle/AMReX_Particle.H [hunk 4]
```diff
@@ -188,6 +188,17 @@ struct ConstParticleCPUWrapper
...
+    ParticleCPUWrapper(idcpu) = cpu;
```
atmyers:
```suggestion
    ParticleCPUWrapper{idcpu} = cpu;
```

ax3l (approved):

Src/Particle/AMReX_Particle.H
```diff
@@ -188,6 +188,18 @@ struct ConstParticleCPUWrapper
...
 
+/** Set the idcpu value at once, based on a particle id and cpuid
+ *
+ * This can be used in initialization and assignments,
+ * to avoid writing twice into the same memory bank.
+ */
+std::uint64_t SetParticleIDandCPU (Long id, int cpu) {
+    std:uint64_t idcpu = 0;
+    ParticleIDWrapper{idcpu} = id;
+``
```
ax3l:
```suggestion
```

Src/Particle/AMReX_Particle.H
```diff
@@ -188,6 +188,17 @@ struct ConstParticleCPUWrapper
     operator int () const noexcept { return static_cast<int>(m_idata & 0x00FFFFFF); }
 };
 
+/** Set the idcpu value at once, based on a particle id and cpuid
+ *
+ * This can be used in initialization and assignments,
+ * to avoid writing twice into the same memory bank.
+ */
+std::uint64_t SetParticleIDandCPU (Long id, int cpu) {
+    std:uint64_t idcpu = 0;
```
atmyers:
```suggestion
    std::uint64_t idcpu = 0;
```

PR #3735, opened 2024-01-31
Add `ParticleIDWrapper::make_invalid()`

ax3l:
## Summary

A cheaper way to swap validity sign on particle ids, as needed to select and track particles from one kernel to another (e.g., boundary condition treatment, re-emission physics, scraping of particles, etc.).

With our current encoding, `ParticleIDWrapper::make_invalid()` is the same as `id = -id`, but cheaper.

Improvements:
- less code emitted, faster execution
- less code emitted, less instructions to store for occupancy limits on GPU
- faster, streamlined code: no jump calls
- explicit valid/invalid calls instead of swaps with input-dependent outcome

### Host Code

https://godbolt.org/z/KPjzExWz1

id_flips

### CUDA Device Code

PTX: https://godbolt.org/z/6En5rK14o
SASS for SM_80: https://godbolt.org/z/d6zYfxaKG

- validation: old/new are same register usage and cost; new code uses a few less logical ops
- mark_(in)valid vs. `id = -id`: now saves 4 registers :tada: 
  - this is the one we often use in already heavy physics kernels - a win for occupancy

Interesting: there are still no 64bit shifts / but shuffles in CUDA hardware...

```
ptxas info    : 0 bytes gmem
ptxas info    : Compiling entry function '_Z12new_is_validPmPd' for 'sm_80'
ptxas info    : Function properties for _Z12new_is_validPmPd
    0 bytes stack frame, 0 bytes spill stores, 0 bytes spill loads
ptxas info    : Used 8 registers, 368 bytes cmem[0]

ptxas info    : Compiling entry function '_Z12old_is_validPmPd' for 'sm_80'
ptxas info    : Function properties for _Z12old_is_validPmPd
    0 bytes stack frame, 0 bytes spill stores, 0 bytes spill loads
ptxas info    : Used 8 registers, 368 bytes cmem[0]

ptxas info    : Compiling entry function '_Z14new_make_validPm' for 'sm_80'
ptxas info    : Function properties for _Z14new_make_validPm
    0 bytes stack frame, 0 bytes spill stores, 0 bytes spill loads
ptxas info    : Used 8 registers, 360 bytes cmem[0]

ptxas info    : Compiling entry function '_Z16new_make_invalidPm' for 'sm_80'
ptxas info    : Function properties for _Z16new_make_invalidPm
    0 bytes stack frame, 0 bytes spill stores, 0 bytes spill loads
ptxas info    : Used 8 registers, 360 bytes cmem[0]

ptxas info    : Compiling entry function '_Z15old_invert_signPm' for 'sm_80'
ptxas info    : Function properties for _Z15old_invert_signPm
    0 bytes stack frame, 0 bytes spill stores, 0 bytes spill loads
ptxas info    : Used 12 registers, 360 bytes cmem[0]
```

## Checklist

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_Particle.H
+++ b/Src/Particle/AMReX_Particle.H
@@ -102,6 +102,41 @@ struct ParticleIDWrapper
         r = (sign) ? lval : -lval;
         return r;
     }
+
+    /** Mark the particle as invalid
+     *
+     * Swaps the is_valid (sign) bit to invalid.
+     * This is NOT identical to id = -id, but it is equally reversible via make_valid().
+     */
+    AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+    void make_invalid () noexcept
+    {
+        // RHS mask: 0111...
+        m_idata &= ~(uint64_t(1) << 63);
+    }
+
+    /** Mark the particle as valid
+     *
+     * Swaps the is_valid (sign) bit to valid.
+     * This is NOT identical to id = -id, but it is equally reversible via make_invalid().
+     */
+    AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+    void make_valid () noexcept
+    {
+        // RHS mask: 1000...
+        m_idata |= uint64_t(1) << 63;
+    }
+
+    /** Check the particle is valid, via the sign of the id.
+     *
+     * Returns true if the particle is valid (the id is positive), otherwise false (invalid particle).
+     */
+    [[nodiscard]] AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+    bool is_valid () const noexcept
+    {
+        // the leftmost bit is our id's valid sign
+        return m_idata >> 63;
+    }
 };
 
 struct ParticleCPUWrapper
@@ -172,6 +207,17 @@ struct ConstParticleIDWrapper
         Long lval = static_cast<Long>(val);  // bc we take -
         r = (sign) ? lval : -lval;
         return r;
+    }
+
+    /** Check the sign of the id.
+     *
+     * Returns true if the id is positive, otherwise false (invalid particle).
+     */
+    [[nodiscard]] AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+    bool is_valid () const noexcept
+    {
+        // the leftmost bit is our id's valid sign
+        return m_idata >> 63;
     }
 };
 


ax3l:
Also adding a more efficient `.negative()` test:
https://godbolt.org/z/K1Yjve6f4

image

Src/Particle/AMReX_Particle.H
```diff
@@ -102,6 +102,28 @@ struct ParticleIDWrapper
...
+
+    /** Check the sign of the id.
+     *
+     * Returns false if the id is positive, otherwise true (invalid particle).
+     */
+    AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+    bool negative () noexcept
+    {
+        // the leftmost bit is our id sign
+        return value >> 63;
```
ax3l, line 125: According to godbolt, I think I wrote the write test here (if the bit is 1 it is negative).

I am confused with the `Long ()` implementation above reading:
```C++
        Long lval = static_cast<Long>(val);  // bc we take -
        r = (sign) ? lval : -lval;
```

Shouldn´t that read
```C++
        r = (sign) ? lval : -lval;
```
?

Src/Particle/AMReX_Particle.H
```diff
@@ -102,6 +102,28 @@ struct ParticleIDWrapper
...
+
+    /** Check the sign of the id.
+     *
+     * Returns false if the id is positive, otherwise true (invalid particle).
+     */
+    AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+    bool negative () noexcept
+    {
+        // the leftmost bit is our id sign
+        return m_idata >> 63;
```
ax3l, line 138: According to [godbolt](https://godbolt.org/z/K1Yjve6f4), I think I wrote the correct test here ("if the bit is 1, this evaluates to true, so it is negative").

I am confused with the `Long ()` implementation above reading:
```C++
        Long lval = static_cast<Long>(val);  // bc we take -
        r = (sign) ? lval : -lval;
```

Shouldn't that read
```C++
        r = (sign) ? -lval : lval;
```
?

ax3l: For currently unknown reason, we swap the id sign bit before storing the other 39 bits of it.

Thus, this part needs to read
```suggestion
        return !(m_idata >> 63);
```
for our current convention... (updated)

atmyers: The leftmost bit is `0` if the `Long` id is negative and `1` if it's positive, i.e. we store the sign, not it's inverse.

ax3l: Wait, in a (signed) int, the leftmost bit is 1 if it is negative, no?

ax3l: I think there is a misunderstanding here in our nomenclature:
- in encoding of signed ints, the sign bit is only non-zero if a number is negative [[1]](https://en.wikipedia.org/wiki/Signed_number_representations) [[2]](https://www.simonv.fr/TypesConvert/?integers)
- here, we currently store if a number has positive sign (the exact opposite)

I think we could simplify this by changing the convention to the "usual" int sign bit convention (0: positive, 1: negative), but I would not do this in the current PR. Let's chat tomorrow :)

Src/Particle/AMReX_Particle.H
```diff
@@ -95,13 +95,47 @@ struct ParticleIDWrapper
...
+    /** Swap the sign of the id.
+     *
+     * Positive ids become negative and negative ids become positive.
+     */
+    AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+    void negate () noexcept
+    {
+        // The uppermost bit is our id inverse sign.
+        // We will XOR it with 1 to flip it.
+        m_idata ^= static_cast<uint64_t>(1) << 63;
```
ax3l: Need to double check that this is correct:
https://en.wikipedia.org/wiki/Signed_number_representations

The cool thing about this is, that if applied twice it will restore the id. But I think it does not yet set `id = -id`, if we value that.

Alternatively, we could call this function `flipValid` and use it for exactly the purpose we use it for right now - marking particles for move/remove but keeping their id for analysis after flipping back - at a very cheap price.

ax3l: Unless at the min/max range, we could use something based on https://stackoverflow.com/a/64325162/2719194

but we might need to shift in/out the cpu part again, making it more costly.

ax3l:
@WeiqunZhang @atmyers @AlexanderSinn ready for review now - let me know if this looks legit

WeiqunZhang:
We will wait till the next release tomorrow.

ax3l:
It's just adding and now changing stuff, so it should be pretty safe, but I will also just need it after the release tomorrow, so no rush.

ax3l:
As another optimization, I explored using 32bit registers via tricks like:
```C++
    bool is_valid () const noexcept
    {
        // the leftmost bit is our id's inverse sign
        auto const * const i32 = (uint32_t*)&m_idata;
        return *i32 >> 31;

    }
```
This does what one expects on CPU (DWORD over QWORD, 32bit register used over 64bit one) and on CUDA GPUs (SM_80) it demotes a 64bit load to a 32bit one & reduces one `ISETP.GT.*` op in SASS code... but since this buys in endianness handling (increment pointer +1 for big endian?), and does not reduce registers and cmem usage in our test according to `ptxas`, I would leave it be for now.

AlexanderSinn:
Replacing a 64 bit load with a 32 bit one? Don’t do this, the 64 bit load would be coalesced but the 32 bit load not because there would be a gap to the next thread. The 32 bit version might be slower.

WeiqunZhang (approved):

ax3l:
> Replacing a 64 bit load with a 32 bit one? Don’t do this, the 64 bit load would be coalesced but the 32 bit load not because there would be a gap to the next thread. The 32 bit version might be slower.

You are right. Yeah, I though to load coalesced and then copy into a 32bit register, do rest of ops there... but this micro-optimization seems not worth it.

PR #3737, opened 2024-01-31
[Draft] Simplify idcpu

ax3l:
## Summary

Simplify `idcpu` by moving the `id` including its sign bit in the
uppermost 40 of 64 bytes.

Use a signed type for `idcpu`

This:
- allows users to interpret `idcpu` as a signed integer that is positive for valid and negative for invalid particles
  - this is visible in output as well as in user-written code, e.g., from Python
  - contrary to `id`, a global id does not need conversion and can be interpreted as a valid type as is
- simplifies conversions (assignment, extraction) significantly
- enable us to judge the `id` sign from the `idcpu` integer
  representation
- make use of compiler optimizations when directly operating on
  it (e.g., with `*= -1` or `< 0` requests).

## Additional background

- X-ref #3735 (rebase this after merged)
- https://en.wikipedia.org/wiki/Signed_number_representations
- https://www.simonv.fr/TypesConvert/?integers

## Checklist

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_Particle.H
+++ b/Src/Particle/AMReX_Particle.H
@@ -67,40 +67,31 @@ struct ParticleIDWrapper
     AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
     ParticleIDWrapper& operator= (const Long id) noexcept
     {
-        // zero out the 40 leftmost bits, which store the sign and the abs of the id;
+        // zero out the 40 leftmost bits, store the id including its sign bit (0: +, 1: -)
         m_idata &= 0x00FFFFFF;
 
-        uint64_t val;
-        uint64_t sign = id >= 0;
-        if (sign)
-        {
-            // 2**39-1, the max value representable in this fashion
-            AMREX_ASSERT(id <= 549755813887L);
-            val = id;
-        }
-        else
-        {
-            // -2**39-1, the min value representable in this fashion
-            AMREX_ASSERT(id >= -549755813887L);
-            val = -id;
-        }
+        // note: although right-shifts of the sign-bit are implementation-
+        //       defined (arightmetic or logical), this operation will give the same result
+        int64_t const sign = (m_idata >> 63) << 63;  // leave only leftmost sign bit
 
-        m_idata |= (sign << 63);  // put the sign in the leftmost bit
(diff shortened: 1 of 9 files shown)


ax3l:
Why unfortunately the type change is not safe... The bir shift of the sign bit is implementation defined in C++...

Src/Particle/AMReX_Particle.H
```diff
@@ -67,51 +67,61 @@ struct ParticleIDWrapper
...
-            AMREX_ASSERT(id >= -549755813887L);
-            val = -id;
-        }
-
-        m_idata |= (sign << 63);  // put the sign in the leftmost bit
-        m_idata |= (val << 24);   // put the val in the next 39
+        // +/-2**39-1, the max value representable in this fashion
+        AMREX_ASSERT(std::abs(id) <= 549755813887L);
+
+        m_idata |= (id << 24);   // put the id in the 40 uppermost bits
```
AlexanderSinn: According to https://en.cppreference.com/w/cpp/language/operator_arithmetic#Bitwise_shift_operators
for negative id, `id << 24` is undefined behavior (until C++20).

ax3l: Ooops, I forgot to xor in and out the sign bit before C++20 (as done in `Long ()` below)

Src/Particle/AMReX_Particle.H
```diff
@@ -67,51 +67,61 @@ struct ParticleIDWrapper
...
         return *this;
     }
 
     AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
     operator Long () const noexcept
     {
-        Long r = 0;
+        // note: although right-shifts of the sign-bit are implementation-
+        //       defined (arightmetic or logical), this operation will give the same result
+        int64_t const sign = (m_idata >> 63) << 63;  // leave only leftmost sign bit
```
ax3l: Could also be an or with a 10000... mask

Src/Particle/AMReX_Particle.H
ax3l: Ah, and even with that, this only works for positive id assignments, since negative signed ids fill more upper bits (they are not just the same bits as positive ids with a flipped most significant bit).

PR #3742, opened 2024-02-01
Use long integer in GPU kernels

WeiqunZhang:
In the current implementation of ParallelFor, we use `int` for the linear cell index in the flatten one-dimension view. This limits the size of the box to less than 2^30 cells (half of INT_MAX). The factor of half comes from the grid-stride loop. This limitation has not been a serious issue until recently because GPUs did not have that much memory. However, the total memory on the latest GPUs has increased quite a lot. Issues have been reported by users.

In this PR, we have started using `std::uint64_t` as the linear cell index. An issue of using `std::uint64_t` is that 64-bit integer division is very expensive. Fortunately, we are able to "steal" the fast division code from https://github.com/NVIDIA/cutlass. Streaming tests have shown very good performance. On A100, the FArrayBox version of the triad test achieves 1.76 TB/s, which is the same as the rate of the much simpler 1D vector version. In fact, it's slightly faster than the rate of 1.72 TB/s from the current version in the development branch.

We have not made all kernel launches in AMReX safe for large sizes. Nevertheless, this PR is the first step, and it will be followed up by more PRs.

### DIFF
--- a/Src/Base/AMReX_INT.H
+++ b/Src/Base/AMReX_INT.H
@@ -31,4 +31,29 @@ namespace amrex {
 }
 #endif
 
+#if (defined(__x86_64) || defined (__aarch64__)) && !defined(_WIN32) && (defined(__GNUC__) || defined(__clang__))
+
+#define AMREX_INT128_SUPPORTED 1
+
+#if defined(__GNUC__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wpedantic"
+#endif
+
+typedef unsigned __int128 amrex_uint128_t; // NOLINT(modernize-use-using)
+typedef          __int128 amrex_int128_t;  // NOLINT(modernize-use-using)
+
+#ifdef __cplusplus
+namespace amrex {
+    using UInt128_t = amrex_uint128_t;
+    using  Int128_t = amrex_int128_t;
+}
+#endif
+
+#if defined(__GNUC__)
+#pragma GCC diagnostic pop
+#endif
+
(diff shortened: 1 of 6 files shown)


atmyers (approved):

Src/Base/AMReX_INT.H
```diff
@@ -31,4 +31,29 @@ namespace amrex {
...
+
+#define AMREX_INT128_SUPPORTED 1
+
+#if defined(__GNUC__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wpedantic"
+#endif
+
+typedef unsigned __int128 amrex_uint128_t; // NOLINT(modernize-use-using)
+typedef          __int128 amrex_int128_t;  // NOLINT(modernize-use-using)
```
ax3l, line 44: NVHPC seems unhappy about long ints:
https://github.com/ECP-WarpX/WarpX/pull/4678
```
-- The C compiler identification is NVHPC 21.11.0
-- The CXX compiler identification is NVHPC 21.11.0
-- Check for working CUDA compiler: /opt/nvidia/hpc_sdk/Linux_x86_64/21.11/compilers/bin/nvcc - skipped
```
```
"/home/runner/work/WarpX/WarpX/build/_deps/fetchedamrex-src/Src/Base/AMReX_INT.H", line 43: error: expected a ";"
  typedef unsigned __int128 amrex_uint128_t; 
                            ^

"/home/runner/work/WarpX/WarpX/build/_deps/fetchedamrex-src/Src/Base/AMReX_INT.H", line 48: error: identifier "amrex_uint128_t" is undefined
  using UInt128_t = amrex_uint128_t; 
                    ^
...
```

PR #3748, opened 2024-02-07
Minor new functions in AMReX_MPMD to provide flexibility for python binding 

siddanib:
## Summary
New functions (*void* **Initialize_without_split**, int **AppNum**) are created in Src/Base/AMReX_MPMD.* to provide additional flexibility of calling MPI_Comm_split equivalent in python instead of it being called inside of AMReX, which is the default behavior of *MPI_Comm* **Initialize**.  

The proposed changes:
- [ ] add minor new functions to AMReX_MPMD

### DIFF
--- a/Src/Base/AMReX_MPMD.H
+++ b/Src/Base/AMReX_MPMD.H
@@ -20,7 +20,7 @@ bool Initialized ();
 
 int MyProc ();   //! Process ID in MPI_COMM_WORLD
 int NProcs ();   //! Number of processes in MPI_COMM_WORLD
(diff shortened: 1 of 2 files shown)


WeiqunZhang:
I think `Initialize` now can call `Initialize_without_split` first and then call `MPI_Comm_split`. This cam reduce code duplication.

siddanib:
That's true! Thanks, @WeiqunZhang. Could you please let me know if these latest changes are consistent?

WeiqunZhang (approved):
LGTM

PR #3749, opened 2024-02-09
Add unsigned int as comm data type

WeiqunZhang:
Previously, we had char, unsigned long long and unsigned long long [8] as comm data type. The issue is we cannot use them for floats or ints. So we add unsigned int, which has a size of 4.

This should solve an issue seen in single precision WarpX runs.


WeiqunZhang:
There must be something else going. There should not be any issues with the current development because we add extra padding for the alignment. Unless that somehow did not as expected.

WeiqunZhang:
Currently we check alignment using
```
    return (reinterpret_cast<std::size_t>(p) % alignment) == 0;
```
Technically speaking, that is undefined behavior. We could change it to
```
inline bool is_aligned (const void* p, std::size_t alignment)
{
    auto* q = const_cast<void*>(p);
    auto space = alignment;
    return std::align(alignment, alignment, q, space) == p;
}
```
It's more expensive, but it should matter at all before the function is used in MPI communication functions.

PR #3756, opened 2024-02-13
Update SUNDIALS support to SUNDIALS v7.0.0

balos1:
## Summary

This PR updates the SUNDIALS interface to SUNDIALS v7.0.0, which will be released at the end of the month. 

## Additional background

This PR breaks support for SUNDIALS < v7. I don't think it would difficult to support both versions if that is desired.  A [release candidate](https://github.com/LLNL/sundials/releases/tag/v7.0.0-rc.1) for SUNDIALS v7 is 'released' and tagged, but there will be a few more updates in the final v7 that currently sit in sundials@develop. 

## Checklist

The proposed changes:


WeiqunZhang:
We do want to make sure the old versions work too. It should not be hard by using the `SUNDIALS_VERSION_MAJOR` macro. If you don't have time, I am happy to finish this PR. If we can get this merged before the end of this month, we can get this into the next release and then update spack.

balos1:
Sounds good. If you can finish it up that would be great. If you have any issues just let me know.

WeiqunZhang:
Great. Since I cannot push to your branch, I will close this PR. Then I will cherry pick your commits and start from there. Thanks!

PR #3758, opened 2024-02-13
Fix offset in send buffer of single precision particle communication

WeiqunZhang:
The bug manifests when a process needs to send more than 2GB single precision data. The offset must guarantee that the send buffer's data pointer incremented by the offset has the appropriate alignment for any MPI data type. Thus we use `Arena::align_size` as the requirement of the alignment.

Also remove an incorrect assertion. The confusion probably came from the misnamed function ParallelDescriptor::alignof_comm_data. Thus we rename it sizeof_selected_comm_data_type.


atmyers:
Let's wait until Eric confirms this fixes his issue.

atmyers (approved):

clarkse:
> Let's wait until Eric confirms this fixes his issue.

This appears to fix the problem.  Thanks!

PR #3767, opened 2024-02-15
add AMREX_LIKELY and AMREX_UNLIKELY

zingale:
## Summary

This adds support for C++20 `[[likely]]` and `[[unlikely]]`.

Compiler support can be found here:
https://en.cppreference.com/w/cpp/compiler_support

This works with GCC >= 9, nvcc >= 12, clang >= 12

The feature testing info is here:
https://en.cppreference.com/w/cpp/feature_test

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_Extension.H
+++ b/Src/Base/AMReX_Extension.H
@@ -200,8 +200,8 @@
 #    define AMREX_LIKELY [[likely]]
 #    define AMREX_UNLIKELY [[unlikely]]
 #else
-#    define AMREX_LIKELY ((void)0)
-#    define AMREX_UNLIKELY ((void)0)
+#    define AMREX_LIKELY
+#    define AMREX_UNLIKELY
 #endif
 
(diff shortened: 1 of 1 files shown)


Src/Base/AMReX_Extension.H
```diff
@@ -190,6 +190,20 @@
...
+//   - Clang >= 12.0
+//   - GCC >= 9.0
+//   - Intel >= 2021.7
+//   - MSVC >= 19.26
+//   - nvcc >= 12
+#if defined(__has_cpp_attribute) && __has_cpp_attribute(likely) >= 201803L
+#    define AMREX_LIKELY [[likely]]
+#    define AMREX_UNLIKELY [[unlikely]]
+#else
+#    define AMREX_LIKELY ((void)0)
```
WeiqunZhang: Should this be empty instead of `((void)0)`?

zingale: yes.  good point

WeiqunZhang (approved):

PR #3768, opened 2024-02-15
Make MFParallelFor safer from int overflow

WeiqunZhang:
This is continuation of the changes in #3742 making AMReX ready for big kernels.

We also store the number of points in BoxIndexer now because we always need that number in GPU kernels.


WeiqunZhang:
All regression tests have passed.

atmyers (approved):

PR #3778, opened 2024-02-24
Curl Curl solver: 4-color Gauss-Seidel smoother

WeiqunZhang:
This implements the 4-color Gauss-Seidel smoother of Li et. al. 2020.  "An Efficient Preconditioner for 3-D Finite Difference Modeling of the Electromagnetic Diffusion Process in the Frequency Domain", IEEE Transactions on Geoscience and Remote Sensing, 58, 500-509.

PR #3779, opened 2024-02-25
Update GMRES/MLMG interface

WeiqunZhang:
For the curl curl test, if beta = 1e-9 alpha/dx^2, the multigrid solver is able to reduce the residual by 10 orders of magnitude in 10 v-cycles. But for beta = 1e-14 alpha/dx^2, the multigrid solver's residual will stall at about 5e-6 of the original residual. However, it can be solved using GMRES with multigrid as preconditioner.


WeiqunZhang:
Depend on #3778

atmyers (approved):

PR #3787, opened 2024-03-03
Update GMRES/MLMG for nodal solver

WeiqunZhang:
We need to set RHS on Dirichlet nodes to zero for nodal solvers because the Dirichlet nodes are not unknowns and GMRES does not have the knowledge.


WeiqunZhang:
Depend on #3785

atmyers (approved):

PR #3788, opened 2024-03-05
Robin BC: Abort if solver is not safe for reuse

WeiqunZhang:
Because the coefficients stored inside LinOp objects are irreversibly modified for Robin BC, it's unsafe to reuse the solver when there is Robin BC, unless the scalars and the coefficients are reset.

### DIFF
--- a/Src/LinearSolvers/MLMG/AMReX_MLABecLaplacian.H
+++ b/Src/LinearSolvers/MLMG/AMReX_MLABecLaplacian.H
@@ -409,6 +409,8 @@ MLABecLaplacianT<MF>::update ()
     applyMetricTermsCoeffs();
 #endif
 
+    applyRobinBCTermsCoeffs();
+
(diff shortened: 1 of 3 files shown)


cgilet:
If setACoef and setScalars were re-called before every solve/apply, could it work to reuse the solver?

WeiqunZhang:
Yes that could work. Maybe we should try to be clever here.

cgilet:
I think we need to add `applyRobinBCTermsCoeffs();` in `update()` for both ABecLap and EBABecLap. Do you know if I edit the file on github, does that create changes for you to approve or not, or does it just push a commit?

WeiqunZhang:
If you push the change to your fork and let me know, I will cherry pick it.

cgilet:
okay. branch pr3788 on my fork

cgilet (approved):

Issue #3795, opened 2024-03-11
Windows (Shared): `omp_locks` failed Symbol


ax3l:
This is a Windows (shared/DLL) specific build regression to #3696.

Since this PR / the `24.02` release, we cannot build WarpX and ImpactX on Conda-Forge with a shared AMReX dependency anymore, due to this error:
```
[...]

PR #3796, opened 2024-03-11
`omp_locks`: C Array

ax3l:
## Summary

Use a plain C array over a `std::array` for `omp_locks`. Primarily because this causes linker issues on MSVC/Clang-Cl, secondarily because `omp_locks` might violate in some implementations the type requirements of `std::array` (MoveConstructible and MoveAssignable type `T`).
https://en.cppreference.com/w/cpp/container/array

## Additional background

Potentially a fix for #3795

Testing in https://github.com/conda-forge/impactx-feedstock/pull/28

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_OpenMP.H
+++ b/Src/Base/AMReX_OpenMP.H
@@ -5,7 +5,6 @@
 #ifdef AMREX_USE_OMP
 #include <AMReX_Extension.H>
 #include <omp.h>
-#include <array>
 
 namespace amrex::OpenMP {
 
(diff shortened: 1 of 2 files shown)


ax3l:
The error now is:
```
lld-link: error: undefined symbol: __declspec(dllimport) struct omp_lock_t *amrex::OpenMP::omp_locks
```

WeiqunZhang:
The types in OMP are likely simple C types. So it should not be an issue for std::array. Even if it is so, it should have failed before the link stage.

In llvm, omp_lock_t is a simple struct with a void*.

https://github.com/llvm/llvm-project/blob/ad8c8281363261929b53b0a519cd20e9e2445343/openmp/runtime/src/include/omp.h.var#L83

Anyway, we can merge this and try to fix it in another way.

WeiqunZhang (approved):

PR #3805, opened 2024-03-14
Add Tests/OpenMP/atomicAdd

WeiqunZhang:


Tests/CMakeLists.txt
```diff
@@ -143,6 +143,10 @@ else()
       list(APPEND AMREX_TESTS_SUBDIRS GPU)
    endif ()
 
+   if (NOT AMReX_GPU_BACKEND STREQUAL NONE)
```
ax3l:
```suggestion
   if (AMReX_OMP)
```

Tests/OpenMP/atomicAdd/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+if (NOT AMReX_GPU_BACKEND STREQUAL NONE)
```
ax3l, line 1:
```suggestion
if (NOT AMReX_OMP)
```

WeiqunZhang:
Oh, I did it the original way because the code does not work with GPU unless managed memory is used, but it does work even when OMP is off.

WeiqunZhang:
If we let it run for GPU builds, we have to fix the source code. Use RunOn::Device and add a streamSychronize, and possible disable tiling when GPU is on.  Also note that when both GPU and OMP are on, atomicAdd and lockAdd will do it on GPU. That effectively disable OMP.

ax3l:
Oh, ok. I checked the logs and I could not find the test being built in it o.0

WeiqunZhang:
Oh, I had an extra `NOT` in the logic of testing the backend

WeiqunZhang:
It should read `AMReX_GPU_BACKEND STREQUAL NONE`, but somehow I had an extra `NOT`.

Tests/CMakeLists.txt
```diff
@@ -143,6 +143,10 @@ else()
       list(APPEND AMREX_TESTS_SUBDIRS GPU)
    endif ()
 
+   if (AMReX_OMP)
```
WeiqunZhang:
```suggestion
   if (AMReX_GPU_BACKEND STREQUAL NONE)
```

Tests/OpenMP/atomicAdd/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+if (NOT AMReX_OMP)
```
WeiqunZhang:
```suggestion
if (AMReX_GPU_BACKEND STREQUAL NONE)
```

ax3l:
Looking through the build logs: should this show up as `Test_OpenMP...`?

Tests/OpenMP/atomicAdd/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+if (AMReX_GPU_BACKEND STREQUAL NONE)
```
ax3l:
```suggestion
if (NOT AMReX_GPU_BACKEND STREQUAL NONE)
```

WeiqunZhang: Thanks! That's right. Now I understand. I wrote the correct expression here first and then copy-and-pasted to the other place. Copy-and-paste is a major source of my bugs.

ax3l (approved):
Now I see the `Test_OpenMP_atomicAdd_*d` targets being built.

ax3l:
@WeiqunZhang should this test be included in `AMReX_TEST_TYPE=Small` so we can run it in CI of package managers (e.g., conda-forge #3795)?

WeiqunZhang:
> @WeiqunZhang should this test be included in `AMReX_TEST_TYPE=Small` so we can run it in CI of package managers (e.g., conda-forge #3795)?

Sure

PR #3806, opened 2024-03-15
New constructor for MPMD::Copier

siddanib:
## Summary
A new constructor, `Copier::Copier()`, is created in `MPMD::Copier`. This constructor is ONLY for a particular setup where the utilized `BoxArray` is exactly the same in both MPMD codes. The empty constructor builds its `BoxArray` based on the `BoxArray` supplied by the other code. Hence, the empty constructor MUST be used in conjunction with the other code's constructor being `Copier::Copier(ba, dm, true)`.


Src/Base/AMReX_MPMD.H
```diff
@@ -26,18 +26,27 @@ int MyProgId (); //! Program ID
 class Copier
 {
 public:
-    Copier (BoxArray const& ba, DistributionMapping const& dm);
+    Copier ();
```
WeiqunZhang: Default ctor is very special. So I think it's better to add a dummy argument (e.g., `bool`) to the ctor to make it non-default.

WeiqunZhang:
```suggestion
    Copier (bool);
```

Src/Base/AMReX_MPMD.H
```diff
@@ -26,18 +26,27 @@ int MyProgId (); //! Program ID
...
+    BoxArray get_my_ba();
+
+    DistributionMapping get_my_dm();
+
 private:
     std::map<int,FabArrayBase::CopyComTagsContainer> m_SndTags;
     std::map<int,FabArrayBase::CopyComTagsContainer> m_RcvTags;
     bool m_is_thread_safe;
+    Vector<Box> m_bv;
+    DistributionMapping m_dm;
```
WeiqunZhang:
```suggestion
    BoxArray m_ba;
    DistributionMapping m_dm;
```
BoxArray has a shared_ptr inside. If we store Vector<Box> instead of BoxArray, everytime `get_my_ba` is called a completely new BoxArray without sharing is created. We should try to avoid that.

Src/Base/AMReX_MPMD.H
```diff
@@ -26,18 +26,27 @@ int MyProgId (); //! Program ID
...
 
     template <typename FAB>
     void send (FabArray<FAB> const& mf, int icomp, int ncomp) const;
 
     template <typename FAB>
     void recv (FabArray<FAB>& mf, int icomp, int ncomp) const;
 
+    BoxArray get_my_ba();
+
+    DistributionMapping get_my_dm();
```
WeiqunZhang:
```suggestion
    BoxArray const& boxArray() () const;

    DistributionMapping const& DistributionMap () const;
```
We can change the function names for consistency. We use `boArray` and `DistributionMap` a lot in other places.

Src/Base/AMReX_MPMD.cpp
```diff
@@ -245,6 +264,124 @@ Copier::Copier (BoxArray const& ba, DistributionMapping const& dm)
...
+            }
+        }
+    }
+
+    for (auto& kv : m_SndTags) {
+        std::sort(kv.second.begin(), kv.second.end());
+    }
+    for (auto& kv : m_RcvTags) {
+        std::sort(kv.second.begin(), kv.second.end());
+    }
```
WeiqunZhang: No need to sort in this case. Make sure we do not sort either on the sender side for this special case.

Src/Base/AMReX_MPMD.cpp
```diff
@@ -245,6 +264,124 @@ Copier::Copier (BoxArray const& ba, DistributionMapping const& dm)
...
+    // MPI_COMM_WORLD.
+
+    // Build communication meta-data
+    m_is_thread_safe = ba.ixType().cellCentered();
+
+    std::vector<std::pair<int,Box> > isects;
+
+    for (int i = 0; i < this_nboxes; ++i) {
+        if (procs[i] == myproc) {
+            ba.intersections(m_bv[i], isects);
```
WeiqunZhang: No need to use BoxArray::intersections. because this is a special case that we copy from box i to box i, not to box j where j!=i.

WeiqunZhang (changes requested):

Src/Base/AMReX_MPMD.cpp
siddanib: Thanks for helping me with this, @WeiqunZhang ! For this special case if we ONLY do a copy of a (NODAL) box `i` to box `i` do you think we should also make `m_is_thread_safe = true` irrespective of the `BoxArray` being cell centered or not?

WeiqunZhang:
that's right

Src/Base/AMReX_MPMD.cpp
```diff
@@ -139,7 +139,8 @@ int MyProgId ()
     return (myproc == ParallelDescriptor::MyProc()) ? 0 : 1;
 }
 
-Copier::Copier (BoxArray const& ba, DistributionMapping const& dm)
+Copier::Copier (BoxArray const& ba, DistributionMapping const& dm,
+        bool send_ba)
```
WeiqunZhang, line 143:
```suggestion
Copier::Copier (BoxArray const& ba, DistributionMapping const& dm,
        bool send_ba)
        : m_ba(ba), m_dm(dm)
```

Src/Base/AMReX_MPMD.cpp
```diff
@@ -151,10 +152,11 @@ Copier::Copier (BoxArray const& ba, DistributionMapping const& dm)
         other_root = 0;
     }
 
+    m_ba.define(ba.boxList());
```
WeiqunZhang:
```suggestion
```

Src/Base/AMReX_MPMD.cpp
```diff
@@ -151,10 +152,11 @@ Copier::Copier (BoxArray const& ba, DistributionMapping const& dm)
...
     Vector<Box> bv = ba.boxList().data();
-
     int this_nboxes = static_cast<int>(ba.size());
     Vector<int> procs = dm.ProcessorMap();
+    m_dm.define(procs);
```
WeiqunZhang:
```suggestion
```

Src/Base/AMReX_MPMD.H
```diff
@@ -26,18 +26,27 @@ int MyProgId (); //! Program ID
...
+    Copier (BoxArray const& ba, DistributionMapping const& dm,
+            bool send_ba = false);
 
     template <typename FAB>
     void send (FabArray<FAB> const& mf, int icomp, int ncomp) const;
 
     template <typename FAB>
     void recv (FabArray<FAB>& mf, int icomp, int ncomp) const;
 
+    BoxArray const& boxArray () const;
```
WeiqunZhang:
```suggestion
    [[nodiscard]] BoxArray const& boxArray () const;
```

Src/Base/AMReX_MPMD.H
```diff
@@ -26,18 +26,27 @@ int MyProgId (); //! Program ID
...
+
+    DistributionMapping const& DistributionMap () const;
```
WeiqunZhang:
```suggestion
    [[nodiscard]] DistributionMapping const& DistributionMap () const;
```

WeiqunZhang (changes requested):

siddanib:
Thanks for helping me with this, @WeiqunZhang! I wanted to mention that I included a minor logic for `m_is_thread_safe` in the general `Copier(ba,dm,false)`. 

https://github.com/siddanib/amrex/blob/6d644be78f76d2aa930a45ffc256364deacf18b6/Src/Base/AMReX_MPMD.cpp#L251

My intention was to check if there is any overlap among cell-centered `Box`es in the `oba`.  `m_is_thread_safe = false` if such an overlap is detected. I hope it is consistent with the objective of `MPMD::Copier`.

Thanks!

WeiqunZhang:
I don't think the check is correct, because there are no race conditions for reading. I also don't think we need to check because under normal circumstances this is not any issue.

I just made some small changes. Could you please check if it works?

siddanib:
Thanks for looking into this, @WeiqunZhang! I was able to successfully perform my local MPMD tests based on the latest changes. I removed the unused `isects_o` variable that I ended up including in my previous commits.

WeiqunZhang (approved):

PR #3807, opened 2024-03-15
Fix a memory leak in TableData::operator=

WeiqunZhang:
X-Ref: https://github.com/ECP-WarpX/WarpX/pull/4779


ax3l (approved):
Thank you! I think the copy constructor might need the same logic usually, does it?

WeiqunZhang:
You mean copy assignment? It is deleted.

ax3l:
@WeiqunZhang I actually meant the move constructor, but I think that's not needed there after all (nothing to clear from a previous allocation).

WeiqunZhang:
Yes for constructor there is nothing to clear and no alias to check.

PR #3835, opened 2024-03-21
Update to SUNDIALS v7

WeiqunZhang:
This is the rebirth of #3756. But it's compatible with SUNDIALS 6.


WeiqunZhang:
@balos1 I am able to compile and run with both sundials 6 and 7 using gnu make. But there are still some issues with cmake. `SUNDIALS::core` is only available in v7. So I made it an optional component and check if it's found after find_package is called. But I got the following error,
```
CMake Error at Tools/CMake/AMReXThirdPartyLibraries.cmake:148 (find_package):
  Could not find a configuration file for package "SUNDIALS" that is
  compatible with requested version "6.0.0".

  The following configuration files were considered but not accepted:

    /home/wqzhang/mygitrepo/spack/opt/spack/linux-ubuntu20.04-skylake/gcc-9.4.0/sundials-develop-5wskgac5smdxvywok7yuojfsk6jfzr6a/lib/cmake/sundials/SUNDIALSConfig.cmake, version: 7.0.0
```
Is version 7.0.0 considered incompatible with 6.0.0 in the sundials' cmake config file? If we want to support both v7 and v6, how should we handle this?

@ax3l

WeiqunZhang:
Note that if I simply set the minimum version to 7.0.0 with core as optional, it works.

balos1:
@WeiqunZhang Unfortunately it looks like we setup the [CMake version file for SUNDIALS](https://github.com/LLNL/sundials/blob/main/CMakeLists.txt#L267) with the `SameMajorVersion` [compatibility mode](https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html#command:write_basic_package_version_file). Im going to change it to `AnyNewerVersion`, but that doesn't fix the problem for 7.0.0.  The workaround is to either do `find_package` twice (once with minimum 6.0.0 and once with minimum 7.0.0, check if `NOT SUNDIALS_FOUND` after the first one), or check the version manually like this:
```cmake
find_package(SUNDIALS CONFIG REQUIRED
                      COMPONENTS ${SUNDIALS_COMPONENTS}
                      OPTIONAL_COMPONENTS core)
set(SUNDIALS_MINIMUM_VERSION 6.0.0)
if (SUNDIALS_VERSION VERSION_LESS ${SUNDIALS_MINIMUM_VERSION})
   message(FATAL_ERROR "SUNDIALS ${SUNDIALS_MINIMUM_VERSION} or newer is required. Found version ${SUNDIALS_VERSION}.")
endif()

```

Src/Extern/SUNDIALS/AMReX_SUNMemory.cpp
```diff
@@ -1,13 +1,16 @@
+#include "AMReX_Sundials_Core.H"
```
ax3l:
```suggestion
#include <AMReX_Sundials_Core.H>
```
?

Tools/CMake/AMReXThirdPartyLibraries.cmake
```diff
@@ -143,9 +143,25 @@ if (AMReX_SUNDIALS)
...
+                    OPTIONAL_COMPONENTS core) # core only available for >= 7
+       if (SUNDIALS_VERSION VERSION_LESS ${SUNDIALS_MINIMUM_VERSION})
+          message(FATAL_ERROR "SUNDIALS_VERSION ${SUNDIALS_MINIMUM_VERSION} or newer is required. Found version ${SUNDIALS_VERSION}.")
+       endif ()
     endif ()
     foreach(D IN LISTS AMReX_SPACEDIM)
+        if (SUNDIALS_VERSION VERSION_GREATER_EQUAL 7)
+           if (NOT TARGET SUNDIALS::core)
+              message(FATAL_ERROR "\nSUNDIALS::core not found")
+           endif ()
```
ax3l: This should not be needed since the below link line will cause an issue if `SUNDIALS::core` is not found anyway.
```suggestion
```

ax3l (approved):
Thank you, looks good!

PR #3841, opened 2024-03-23
GCC: inline limit

WeiqunZhang:
By default, set GCC inline limit to a huge number (43210). GCC's default value is version dependent. It appears to be smaller than 2000 in recent versions. The motivation for setting this to a huge number is that GCC especially the recent versions (e.g., 12) no longer inline big lambda functions passed to ParallelFor, resulting in serious performance issues.


zingale:
I tested this with Castro and GCC 13.2 and I didn't see any measureable performance difference.

WeiqunZhang:
The hotspot functions in Castro are perhaps already inlined. Or they are so expensive that the cost of not inlining does not matter.

jmsexton03 (approved):

PR #3844, opened 2024-03-25
Clang-Tidy 17: modernize-type-traits

WeiqunZhang:


ax3l (approved):
C++17, njom njom njom.

PR #3848, opened 2024-03-25
`MultiFab::deepCopy()`

ax3l:
## Summary

Create a deep copy of this MultiFab, using the same Arena and factory.

## Additional background

See https://github.com/AMReX-Codes/pyamrex/pull/282

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_MultiFab.H
+++ b/Src/Base/AMReX_MultiFab.H
@@ -492,7 +492,7 @@ public:
      * This uses the same arena and factory in the copy and copies
      * all data.
      */
-    MultiFab deepCopy ();
+    [[nodiscard]] MultiFab deepCopy () const;
 
     /**
     * \brief Copy from src to dst including nghost ghost cells.
--- a/Src/Base/AMReX_MultiFab.H
+++ b/Src/Base/AMReX_MultiFab.H
@@ -488,6 +488,8 @@ public:
                      const IntVect&  nghost);
 
 
+    using FabArray<FArrayBox>::copy;
+
     /** Create a copy of this MultiFab, using the same Arena.
      */
     MultiFab copy ();
--- a/Src/Base/AMReX_MultiFab.H
+++ b/Src/Base/AMReX_MultiFab.H
@@ -486,6 +486,12 @@ public:
                      int             dstcomp,
                      int             numcomp,
                      const IntVect&  nghost);
+
+
+    /** Create a copy of this MultiFab, using the same Arena.
+     */
+    MultiFab copy ();
+
     /**
     * \brief Copy from src to dst including nghost ghost cells.
     * The two MultiFabs MUST have the same underlying BoxArray.
--- a/Src/Base/AMReX_MultiFab.H
+++ b/Src/Base/AMReX_MultiFab.H
@@ -487,12 +487,12 @@ public:
                      int             numcomp,
                      const IntVect&  nghost);
 
-
-    using FabArray<FArrayBox>::copy;
-
-    /** Create a copy of this MultiFab, using the same Arena.
+    /** Create a deep copy of this MultiFab and its data.
+     *
+     * This uses the same arena and factory in the copy and copies
+     * all data.
      */
-    MultiFab copy ();
+    MultiFab deepCopy ();
 
     /**
     * \brief Copy from src to dst including nghost ghost cells.
--- a/Src/Base/AMReX_MultiFab.cpp
+++ b/Src/Base/AMReX_MultiFab.cpp
@@ -168,7 +168,7 @@ MultiFab::Add (MultiFab& dst, const MultiFab& src,
 }
 
 MultiFab
-MultiFab::copy ()
+MultiFab::deepCopy ()
 {
     auto mf = MultiFab(
         this->boxArray(),
--- a/Src/Base/AMReX_MultiFab.cpp
+++ b/Src/Base/AMReX_MultiFab.cpp
@@ -167,6 +167,27 @@ MultiFab::Add (MultiFab& dst, const MultiFab& src,
     amrex::Add(dst, src, srccomp, dstcomp, numcomp, nghost);
 }
 
+MultiFab
+MultiFab::copy ()
+{
+    auto mf = MultiFab(
+        this->boxArray(),
+        this->DistributionMap(),
+        this->nComp(),
+        this->nGrowVect(),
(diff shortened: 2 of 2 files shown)


WeiqunZhang:
We should use a different name, because `MultiFab::copy` will hide `FabArray::copy` and therefore is a breaking change. Also we already have `MultiFab::Copy`. So it can be quite confusing. (We even deprecated `FabArray::copy` to avoid confusion.

Src/Base/AMReX_MultiFab.cpp
```diff
@@ -167,6 +167,27 @@ MultiFab::Add (MultiFab& dst, const MultiFab& src,
...
 
+MultiFab
+MultiFab::copy ()
+{
+    auto mf = MultiFab(
+        this->boxArray(),
+        this->DistributionMap(),
+        this->nComp(),
+        this->nGrowVect(),
+        MFInfo().SetArena(this->arena())
```
WeiqunZhang:
```suggestion
        MFInfo().SetArena(this->arena()),
        this->Factory()
```
Use the same factory too. For EB, the new MultiFab will contains EB information.

ax3l: Ah, that's good to know. Adding to https://github.com/AMReX-Codes/pyamrex/pull/282, too

ax3l:
> We should use a different name, because MultiFab::copy will hide FabArray::copy and therefore is a breaking change.

I am not sure it does - they have different arguments in their signatures...?

> Also we already have MultiFab::Copy. So it can be quite confusing. (We even deprecated FabArray::copy to avoid confusion.

I added the lowercase variant similar to the uppercase (`Add` - static) and lowercase (`add` - method) naming so far. Primarily, I added it because in Python containers like NumPy nd arrays, `.copy()` is the canonical way to do a by-value copy. https://github.com/AMReX-Codes/pyamrex/pull/282

Should we add this extra `copy()` method overload to `FabArray` instead?

WeiqunZhang:
That's how C++ works. https://isocpp.org/wiki/faq/strange-inheritance#hiding-rule (We could work around it by `using FabArray<FArrayBox>::copy`. But function overloading can be confusing to users.)

ax3l:
Haha, ah yes I sometimes forget that fun detail :D
We have do that in ImpactX elements, too.

Python has no overloading at all... unless we are in pybind11 :laughing: 

> We could work around it by using FabArray<FArrayBox>::copy. But function overloading can be confusing to users.

Sounds safe to me, since we do not have default parameters on the existing copies...?

WeiqunZhang:
What makes me uneasy is its implication.  We have so far repeatedly resisted the urge of  making it too easy to create a deep copied object. The copy ctor is deleted. We have a ctor for creating an alias MultiFab, which takes a parameter `amrex::MakeType`, but only `make_alias` is allowed. The new copy function is just like a copy ctor. One might be attempted to write the following correct but inefficient code.
```
MultiFab a = b.copy();
a = 0;
```

So let's have some time to think about this. Maybe we should make the function name more obvious like `createDeepCopy`. Maybe we could update the ctor to allow `amrex::MultiFab a(amrex::make_deep_copy, b);`.

ax3l:
Yes, I very much sympathize where this comes from. Here, I think there is a reasonable difference: in C++, the copy constructor and copy assignment are syntactically very reasonable to delete for large containers such as MultiFab, because they can be implicitly be called and lead to surprising resource costs.

This is essentially the same reason why Python does by-ref for everything by default - no implicit copies of containers ever.

Contrary to copy constructors and copy assignment operators, adding a `.copy()` method is explicit and a clear resource allocation when it is needed.
> One might be attempted to write the following correct but inefficient code.

Maybe I miss a point. Yes, we should definitely expose `make_alike` the same way so we create an exact shallow copy without the values being copied (i.e., for your example doing immediately an `a = 0;`)

```C++
MultiFab a = b.copy();
a *= -1;
```

```C++
MultiFab a = b.make_alike();  // or b.copy(/* deep= */ false)
a = 0;
```

Let's chat about this more. I think that is a reasonable syntax and explicit once a copy is actually needed. It also uses canonical naming for deep copies, so we definitely would want this downstream in pyAMReX.
- https://numpy.org/doc/stable/reference/generated/numpy.copy.html - deep
- https://docs.cupy.dev/en/stable/reference/generated/cupy.copy.html - deep
- https://docs.xarray.dev/en/latest/generated/xarray.DataArray.copy.html - `deep=True`  default
- https://pytorch.org/docs/stable/generated/torch.clone.html (close xD) - deep

My feeling is that making it equally explicit and not more verbose/complicated to create deep copies is desirable.

ax3l:
To avoid confusion with `FabArray<FAB>::copy` (which could be called `copyFrom()` and are deprecated for the use of `ParallelCopy` instead) we name this: `deepCopy()`.

In pyAMReX, because `FabArray<FAB>::copy` is deprecated, we will not expose it to Python and keep naming this `.copy()` as is canonical in Python.

WeiqunZhang (approved):

Src/Base/AMReX_MultiFab.H
```diff
@@ -486,6 +486,14 @@ public:
                      int             dstcomp,
                      int             numcomp,
                      const IntVect&  nghost);
+
+    /** Create a deep copy of this MultiFab and its data.
+     *
+     * This uses the same arena and factory in the copy and copies
+     * all data.
+     */
+    MultiFab deepCopy ();
```
WeiqunZhang:
```suggestion
    [[nodiscard]] MultiFab deepCopy () const;
```

Src/Base/AMReX_MultiFab.cpp
```diff
@@ -167,6 +167,28 @@ MultiFab::Add (MultiFab& dst, const MultiFab& src,
     amrex::Add(dst, src, srccomp, dstcomp, numcomp, nghost);
 }
 
+MultiFab
+MultiFab::deepCopy ()
```
WeiqunZhang:
```suggestion
MultiFab::deepCopy () const
```

Src/Base/AMReX_MultiFab.cpp
```diff
@@ -167,6 +167,28 @@ MultiFab::Add (MultiFab& dst, const MultiFab& src,
...
+{
+    auto mf = MultiFab(
```
WeiqunZhang:
```suggestion
    MultiFab mf(
```

PR #3861, opened 2024-03-29
Fix: Resize Component on Add

ax3l:
## Summary

Some object called "dummy mf" needs to be resized to avoid segfaults.

## Additional background

Follow-up to #3615

## Checklist

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_ParticleContainer.H
+++ b/Src/Particle/AMReX_ParticleContainer.H
@@ -1244,6 +1244,7 @@ public:
         m_num_runtime_real++;
         h_redistribute_real_comp.push_back(communicate);
         SetParticleSize();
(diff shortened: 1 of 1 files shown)


atmyers (approved):

Src/Particle/AMReX_ParticleContainer.H
```diff
@@ -1244,6 +1244,7 @@ public:
         m_num_runtime_real++;
         h_redistribute_real_comp.push_back(communicate);
         SetParticleSize();
+        this->resizeData();
```
ax3l, line 1247: @atmyers potentially, you could also move the logic in `AddRealComp` that reads `//resize runtime SoA` into `resizeData`.

PR #3866, opened 2024-04-01
Add `PODVector::assign(value)`

ax3l:
## Summary

```
vector.assign(vector.size(), 42);
```
is a bit verbose for a standard operation, even if it mirrors https://en.cppreference.com/w/cpp/container/vector/assign

Add another overload similar to `setVal(ue)` used in other AMReX containers.

## Additional background

https://github.com/AMReX-Codes/pyamrex/pull/222#discussion_r1384120148

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_PODVector.H
+++ b/Src/Base/AMReX_PODVector.H
@@ -512,6 +512,15 @@ namespace amrex
                                    (Allocator const&)(*this));
         }
 
+        /** Set the same value to every element of the vector
+         *
+         * @param a_value the value to assign
+         */
+        void assign (const T& a_value)
+        {
+            assign(m_size, a_value);
+        }
+
         [[nodiscard]] allocator_type get_allocator () const noexcept { return *this; }
 
         void push_back (const T& a_value)


Src/Base/AMReX_PODVector.H
```diff
@@ -512,6 +512,15 @@ namespace amrex
                                    (Allocator const&)(*this));
         }
 
+        /** Set the same value to every element of the vector
+         *
+         * @param a_value the value to assign
+         */
+        void assign (const & T a_value)
```
ax3l: We can also avoid an overload and call this `setVal(value)` in symmetry with MultiFab, what do you think?
```suggestion
        void setVal (const T& a_value)
```

ax3l: In MultiFabs, we further overload the assignment operator for `amrex::Real` and/or `int`, for convenience.

WeiqunZhang: I think `assign` is fine. It will still be asymmetric unless we add `assign` to MultiFab, which most versions would not make sense.

WeiqunZhang (approved):

Issue #3869, opened 2024-04-03
`MultiFab::sum_unique` w/ a region


ax3l:
Discussed today, raised by @RemiLehe 

For our applications, we would need a version of `MultiFab::sum_unique` that can be limited to a specific region (like `::max` and `::setVal` support it).

Suggested API:
```c++
MultiFab::sum_unique (amrex::Box const & region, int comp=0, bool local=false);
```

Thanks to @WeiqunZhang for volunteering to add this overload.

ax3l:
Implemented via #3871

PR #3871, opened 2024-04-03
Add MultiFab::sum(region) and sum_unique(region)

WeiqunZhang:
## Summary

Add two new functions to MultiFab that return the sum of the given region.

## Additional background

#3869 

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_MultiFab.H
+++ b/Src/Base/AMReX_MultiFab.H
@@ -247,7 +247,7 @@ public:
 
     /**
     * \brief Same as sum with \p local =false, but for non-cell-centered data, this
-    * skips non-unique points that are owned by multiple boxes.
+    * only adds non-unique points that are owned by multiple boxes once.
     */
     [[nodiscard]] Real sum_unique (int comp = 0,
                      bool local = false,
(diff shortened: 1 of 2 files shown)


asalmgren:
1) duplicatly is misspelled
2) I'm worried about people assuming this works correctly with periodicity even though the code has a comment that it doesn't account for that

WeiqunZhang:
Yes, there could be potential confusion even though logically speaking it should not because without being provided the periodicity information amrex cannot know how to handle it. So the caller has to assume periodicity is not part of the consideration. This is a feature request from WarpX, which does not need the periodic boundaries to be handled in this function. I didn't try to add periodicity support because it's actually quite awkward to implement. We would need to memcpy a vector of periodically shifted Boxes to device in order to do all these in one kernel. Maybe we can wait till someone wants the feature.

Src/Base/AMReX_MultiFab.H
```diff
@@ -249,6 +253,13 @@ public:
                      bool local = false,
                      const Periodicity& period = Periodicity::NonPeriodic()) const;
     /**
+    * \brief Returns the unique sum of component "comp" in the given
+    * region. Non-unique points owned by multiple boxes in the MultiFab are
+    * skipped. No ghost cells are included. This function does not take
```
ax3l:
```suggestion
    * region. Non-unique points owned by multiple boxes in the MultiFab are
    * only added once. No ghost cells are included. This function does not take
```

ax3l: (maybe same update to the doc string above of the original `sum_unique`)

ax3l (approved):
Thank you! :)

PR #3880, opened 2024-04-10
amrex::tupleToArray

WeiqunZhang:
Add a utility function that can convert an amrex::GpuTuple whose types are the same to an array.

### DIFF
--- a/Src/Base/AMReX_Tuple.H
+++ b/Src/Base/AMReX_Tuple.H
@@ -3,6 +3,7 @@
 #define AMREX_TUPLE_H_
 #include <AMReX_Config.H>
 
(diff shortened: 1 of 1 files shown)


AlexanderSinn:
Looks good! I think `constexpr` should be added to the functions. Possibly also `AMREX_GPU_HOST_DEVICE` but that would only make sense if it would return an `amrex::GpuArray` instead of `std::array`.

WeiqunZhang:
Yes, `constexpr` should be added. As for having it return GpuArray, I thought about it, but wasn't entirely sure if we should do it. Thinking about it again, there is probably not much downside.

AlexanderSinn (approved):

PR #3881, opened 2024-04-10
Set AMD wavefront size according to AMReX_AMD_ARCH

### LINKED ISSUE #3792: `__AMDGCN_WAVEFRONT_SIZE` only has expected value during device pass
This is in reference to a previous pull request #3777. The `__AMDGCN_WAVEFRONT_SIZE` only has the expected (correct) value when referenced in the device pass of compilation, and is left at a default value 64 when referenced in the host pass. Unfortunately, it is apparently nontrivial to expose the device-pass value during the host pass through this macro, and I don't have a suggestion from compiler folks about exposing this value at compile time another way. They aim to add a warning when the host-pass detects this macro is used, but as for compile-time value I'm not sure I have any good idea. Manual selection through`ifdef`ing on the architecture could be inconvenient to maintain.

WeiqunZhang:
At compile time set it to 64 except for gfx10*.

Close #3792


WeiqunZhang:
@mrowan137

ax3l (approved):
LGTM! :)

PR #3907, opened 2024-04-25
3d anisotropic eb

jmsexton03:
## Additional background
Creating draft PR to run tests
## Checklist

The proposed changes:


asalmgren:
I'm thinking it might be less confusing to just set axm (not axmt) to apx(i,j,k) so we don't have to redefine all of those at line 113 ... then just swap axm and axmt in lines 54-112.

asalmgren (approved):

Src/EB/AMReX_EB2_3D_C.cpp
```diff
@@ -76,10 +76,10 @@ void set_eb_data (const int i, const int j, const int k,
...
     }
 
-    Real dapx = axm - axp;
-    Real dapy = aym - ayp;
-    Real dapz = azm - azp;
-    Real apnorm = std::sqrt(dapx*dapx+dapy*dapy+dapz*dapz);
+    Real dapx = (axm - axp)*(dx[1]*dx[2]);
+    Real dapy = (aym - ayp)*(dx[0]*dx[2]);
+    Real dapz = (azm - azp)*(dx[0]*dx[1]);
+    const Real apnorm = std::sqrt(dapx*dapx+dapy*dapy+dapz*dapz) + 1.e-30_rt*std::sqrt(dx[0]*dx[1]*dx[2]);
```
WeiqunZhang, line 82: The second term has a different units than the first. The first term is L^2, whereas the second is L^(3/2).

Src/EB/AMReX_EB2_3D_C.cpp
```diff
@@ -121,7 +124,7 @@ void set_eb_data (const int i, const int j, const int k,
         return;
     }
 
-    Real bainv = 1.0_rt/barea(i,j,k);
+    Real bainv = bareascaling*bareascaling/apnorm;
```
WeiqunZhang, line 127: This is the only place bareascaling is used. barescaling is computed  with std::sqrt(..). So maybe sqrt is unnecessary.

PR #3910, opened 2024-04-29
Workaround for Windows omp_lock_t issue

WeiqunZhang:


WeiqunZhang:
Sorry there was a typo.

ax3l:
Currently testing this in the 24.05 release roll-outs on conda-forge.

ax3l:
@WeiqunZhang your work-around does the trick! :tada: 
https://github.com/conda-forge/warpx-feedstock/pull/83

WeiqunZhang:
I tested the code for Windows on Linux by changing `#if defined(_WIN32)` to `#if !defined(_WIN32)` and ran some tests. It worked.

ax3l (approved):
Thanks!

Ok, then let us use this type erasure work-around for now. I am still positive this is a conda-forge tools issue, because we cannot see it in CI here, but this is a good work-around :)

PR #3919, opened 2024-05-03
Fix assertions involving IntVect

WeiqunZhang:
A MultiFab used to have the same number of ghost cells in all directions. This has no longer been the case for many years. But many assertions have not been updated for this change. This commit fixes various assertions to make it more appropriate.

Also added are overloaded functions for element-wise comparison between an IntVect and an integer. This makes the code cleaner.


asalmgren (approved):

asalmgren:
BTW -- in cleaning up code, it would be good at some point to do an all-search-and-replace in amrex for BL_ASSERT --> AMREX_ASSERT.

PR #3941, opened 2024-05-13
Fix FillPatcher after enforcing PhysBCFunct operator() fills nghost

### LINKED ISSUE #3940: Segfault using FillPatcher with recent AMReX versions
Since commit [1735fd5](https://github.com/AMReX-Codes/amrex/commit/1735fd5c65d6c73d667233437357de91a629b9e0), I have been getting segfaults in my code when using the FillPatcher class in AmrCore.  It happens at [this line](https://github.com/AMReX-Codes/amrex/blob/af6e77226cdb1668a8936ee3b35cf0d441409869/Src/AmrCore/AMReX_FillPatcher.H#L393C14-L393C32) in FillPatcher.H, currently 393:

`        cbc(*m_cf_crse_data_tmp, 0, ncomp, nghost, time, cbccomp);`

From what I can gather, m_cf_crse_data_tmp doesn't have any ghost cells, so expanding m_cf_crse_data_tmp's validbox by the 'nghost' values of the parent MultiFabs (as per commit 1735fd5) can lead to it trying to fill in boundary conditions of non-existent cells.  Which might well be expected to cause segfaults.  However, I can't seem to reproduce the segfaults I'm getting with Tests/Amr/Advection_AmrCore, after turning off periodicity and uncommenting the 'walls (Neumann)' BC code lines.  So I'm not completely sure that it isn't something I'm doing wrong in my code.

For me, changing `nghost` in the code above to `m_cf_crse_data_tmp->nGrowVect()`, i.e. setting nghost to (0,0,0), gets rid of the segfaults.  But I'm not sure if this is the correct fix or not.

Would be good to hear what you think.  Thanks!

cgilet:
grow cells (PR 3914)

Also assert in operator that mf has nghost grow cells to fill.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_PhysBCFunct.H
+++ b/Src/Base/AMReX_PhysBCFunct.H
@@ -153,7 +153,7 @@ public:
 
         BL_PROFILE("PhysBCFunct::()");
 
-        AMREX_ASSERT(nghost <= mf.nGrowVect());
+        AMREX_ASSERT(nghost.allLE(mf.nGrowVect()));
 
         //! create a grown domain box containing valid + periodic cells
         const Box& domain = m_geom.Domain();
--- a/Src/Base/AMReX_PhysBCFunct.H
+++ b/Src/Base/AMReX_PhysBCFunct.H
@@ -153,6 +153,8 @@ public:
 
         BL_PROFILE("PhysBCFunct::()");
 
+        AMREX_ASSERT(nghost <= mf.nGrowVect());
+
         //! create a grown domain box containing valid + periodic cells
         const Box& domain = m_geom.Domain();
         Box gdomain = amrex::convert(domain, mf.boxArray().ixType());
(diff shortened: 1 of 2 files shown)


Src/Base/AMReX_PhysBCFunct.H
```diff
@@ -153,6 +153,8 @@ public:
 
         BL_PROFILE("PhysBCFunct::()");
 
+        AMREX_ASSERT(nghost <= mf.nGrowVect());
```
WeiqunZhang:
```suggestion
        AMREX_ASSERT(nghost.allLE(mf.nGrowVect()));
```
`IntVect::operator<=` does lexicographical comparison. (For example, (1,1,1) <= (0,0,2) is true.)

WeiqunZhang:
We also need to change Line 596.

WeiqunZhang (approved):

PR #3943, opened 2024-05-15
Delete copy constructor and assignment operator for ParticleTile

atmyers:
This prevents common programming errors involving accidently copying particle tiles.

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_ParticleTile.H
+++ b/Src/Particle/AMReX_ParticleTile.H
@@ -719,6 +719,8 @@ struct ParticleTile
     using ConstParticleTileDataType = ConstParticleTileData<StorageParticleType, NArrayReal, NArrayInt>;
 
     ParticleTile () = default;
+
+#ifndef _WIN32  // workaround windows compiler bug
     ~ParticleTile () = default;
 
     ParticleTile (ParticleTile const&) = delete;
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
The Windows CI failure looks like a compiler bug.

WeiqunZhang:
More specifically it complains about `std::pair<const Key, Value>(std::pair<const Key, Value&>` cannot be done because `Value` is not copyable, where range based for loop is used to access `std::map<Key, Value>`. But the `value_type` of std::map should be `std::pair<const Key, Value>`. Why does that `&` in `Value&` come from?

We could put the new changes inside `#ifndef _WIN32`.

atmyers:
Note - the WarpX test failure was a problem in WarpX that was uncovered by this change. It's already been fixed in development but is not part of the release yet.

WeiqunZhang (approved):

Src/Particle/AMReX_ParticleTile.H
```diff
@@ -720,6 +720,16 @@ struct ParticleTile
 
     ParticleTile () = default;
 
+#ifndef _WIN32  // workaround windows compiler bug
```
ax3l, line 723: @atmyers @WeiqunZhang This should probably read MSVC.

There are many compilers on Windows (LLVM/Clang, GCC aka MinGW, Intel, etc.) that should work file :)

ax3l: Ok, I see that LLVM/Clang failed, too.

PR #3944, opened 2024-05-15
Fix particle restart when finestLevel() < finest_level_in_file

dpgrote:
## Summary
In the restart for particles, there is a loop up to `lev = finest_level_in_file` but is accessing `ParticleDistributionMap(lev)`. In the case when `finestLevel() < finest_level_in_file`, this was causing an assertion error (with DEBUG) and a seg fault (without DEBUG). This fixes the loop to only go up to `finestLevel()`. This fixes the issue and still works properly (all of the particles are read in). This also removes the `if` block that would no longer ever be called.

## Additional background
Curiously, previously it would work with `finestLevel() < finest_level_in_file`, but none of this code had changed recently. I don't know how it worked before.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_ParticleIO.H
+++ b/Src/Particle/AMReX_ParticleIO.H
@@ -921,7 +921,7 @@ ParticleContainer_impl<ParticleType, NArrayReal, NArrayInt, Allocator, CellAssig
     }
 
     if (dual_grid) {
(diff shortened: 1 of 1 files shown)


atmyers:
Thank you! I think we also need to make the same change on line 924...

atmyers (approved):

PR #3945, opened 2024-05-16
add DEBUG_LIBSTDCXX option to GNU Make gcc for libstdc++ debug

zingale:
this catches bounds errors in the standard library containers

## Checklist

The proposed changes:

### DIFF
--- a/Tools/GNUMake/comps/gnu.mak
+++ b/Tools/GNUMake/comps/gnu.mak
@@ -104,7 +104,7 @@ else
 endif
 
 ifeq ($(DEBUG_LIBSTDCXX),TRUE)
-   CXXFLAGS += -D_GLIBCXX_DEBUG
+   CPPFLAGS += -D_GLIBCXX_DEBUG
 endif
 
 ifeq ($(WARN_ALL),TRUE)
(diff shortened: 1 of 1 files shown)


zingale:
see: https://gcc.gnu.org/onlinedocs/libstdc++/manual/debug_mode_using.html

zingale:
not sure if we need this as an option so I left it WIP.  It doesn't find issues in Castro.  I am not sure how much of the standard library containers we rely on.

WeiqunZhang:
Since the new option is opt-in, there is no harm I guess.

Tools/GNUMake/comps/gnu.mak
```diff
@@ -103,6 +103,10 @@ else
   endif
 endif
 
+ifeq ($(DEBUG_LIBSTDCXX),TRUE)
+   CXXFLAGS += -D_GLIBCXX_DEBUG
```
WeiqunZhang:
```suggestion
   CPPFLAGS += -D_GLIBCXX_DEBUG
```

WeiqunZhang (approved):

Issue #3955, opened 2024-05-23
Add more support for N dimensions


AlexanderSinn:
Some codes, such as HiPACE++, use a different number of dimensions in different parts of the code. Currently, AMREX_SPACEDIM has to be chosen as the maximum number of dimensions used, with the other places in the code not utilizing the extra dimensions by assigning a size of 1 to them. This can be very confusing, and in the case of ParallelFor and Array4 can lead to a slight performance loss. Additionally, it might be useful to have more than 3 dimensions sometimes. For example, a 3D loop with tiling on GPU could be implemented with a 6D ParallelFor.

This can be achieved by replacing the AMREX_SPACEDIM macro with a template parameter that specifies the number of dimensions.

I would volunteer to start working on this by converting `IntVect` to `IntVectND<>` and adding `using IntVect = IntVectND<AMREX_SPACEDIM>;`. Currently IntVect relies hevily on `AMREX_D_TERM`. This can be replaced by `sfinae`, `if constexpr` or for loops that will usually get unrolled since the bounds are known at compile time.

Later on, the same can be done for other types such as RealVect, Box, Geometry, Array4, as well as ParallelFor.

Even if not everything will be converted (BoxArray, MultiFab might require too much effort), just having N-dimensional indexing would be very useful.

WeiqunZhang:
Maybe we should have a meeting discussing this.

AlexanderSinn:
Good point, maybe at the next WarpX+AMReX development meeting?

AlexanderSinn:
Now that ParallelFor is N dimensional, I think a base level of functionality for N dimensions is provided. Below, I listed some additional classes that could be converted to N dimensions similarly to IntVect/Box/ParallelFor. Anybody can feel free to contribute.

- [x] IntVect #3969
- [x] IndexType #3988
- [x] Box #4016
- [x] ParallelFor #4052
- [x] ParallelFor CTO #4109
- [x] Reduce #4875
- [x] Array4 #4868
- [x] RealVect #4430
- [ ] RealBox
- [ ] Geometry
- [ ] BaseFab

BenWibking:
@AlexanderSinn @WeiqunZhang Is there any thinking on adding an `ArrayND` object to generalize `Array4`? This would be particularly useful for interop with external codes/libraries that use `std::mdspan`.

asalmgren:
Closing this issue due to lack of activity.  Please re-open it if you feel there is is still an issue!

[...]

PR #3969, opened 2024-06-01
Convert IntVect to IntVectND

AlexanderSinn:
## Summary

As described in #3955, this PR converts IntVect to the n dimensional IntVectND and adds
`using IntVect = IntVectND<AMREX_SPACEDIM>;`. Additionally, deduction guides, support for structured bindings and the helper functions `IntVectCat`, `IntVectSplit` and `IntVectResize` were added to IntVectND.

I didn't find a way to make `IntVect::Zero` and `IntVect::Unit` work with n dimensions, so I removed them (converting them to `static constexpr` or `inline static` doesn't work with MSVC). Instead, `IntVect::TheZeroVector()` and `IntVect::TheUnitVector()` can be used.

## Additional background

Using structured binding support of `amrex::GpuTuple`, the following code should work (on GPU):
```C++
int i, j, k, n;
...
// amrex::IntVectND<4>
amrex::IntVectND iv1 (i, j, k, n);
// amrex::IntVectND<8>
amrex::IntVectND iv2 = amrex::IntVectCat(iv1, iv1);
// ... = amrex::GpuTuple<amrex::IntVectND<3>,amrex::IntVectND<2>,amrex::IntVectND<2>,amrex::IntVectND<1>>
auto [iv3, iv4, iv5, iv6] = amrex::IntVectSplit<3,2,2,1>(iv2); 
// int, int, int = amrex::IntVectND<3>
auto [i2, j2, k2] = iv3;
// int = amrex::IntVectND<1>
auto [n2] = iv6;

assert(i==i2 && j==j2 && k==k2 && n==n2);
```

## Checklist

The proposed changes:


WeiqunZhang:
Should we have `extern template class IntVectND<AMREX_SPACEDIM>`, which might help reduce compile time?

AlexanderSinn:
Wouldn't that prevent the member functions from getting inlined, which could be bad when IntVect is used in a parallel for?

WeiqunZhang:
My understanding is it should not affect run time performance. But I might be wrong.

WeiqunZhang:
It is also not clear how GPU compilers treat it. So probably a bad idea and
let's not worry about it.

AlexanderSinn:
I tried to add `extern template class IntVectND<AMREX_SPACEDIM>;`however, without adding ` template class IntVectND<AMREX_SPACEDIM>; `to the cpp file. I expected to get a linker error however it compiled just fine.

WeiqunZhang:
Maybe it's because all those functions have already been inlined.

WeiqunZhang:
```
diff --git a/Src/Base/AMReX_IntVect.H b/Src/Base/AMReX_IntVect.H
index 70a12811a2..7c472742ae 100644
--- a/Src/Base/AMReX_IntVect.H
+++ b/Src/Base/AMReX_IntVect.H
@@ -710,6 +710,9 @@ public:
         return IntVectND<dim>(std::numeric_limits<int>::lowest());
     }
 
+    static const IntVectND<dim> Zero;
+    static const IntVectND<dim> Unit;
+
     static constexpr std::size_t size () noexcept {
         return static_cast<std::size_t>(dim);
@@ -767,6 +770,13 @@ private:
     int vect[dim] = {};
 };
 
+
+template <>
+inline constexpr IntVectND<AMREX_SPACEDIM> IntVectND<AMREX_SPACEDIM>::Zero{0};
+
+template <>
+inline constexpr IntVectND<AMREX_SPACEDIM> IntVectND<AMREX_SPACEDIM>::Unit{1};
+
 // Template deduction guide for IntVectND
 template<std::size_t dim>
 AMREX_GPU_HOST_DEVICE // __device__ for HIP
```

This seems to work. So we could get IntVect::Zero and IntVect::Unit back for backward compatibility.

Unfortunately `template <int dim> inline constexpr IntVectND<dim> IntVectND<dim>::Zero{0}` might only work for gcc. https://stackoverflow.com/questions/72409091/define-a-static-constexpr-member-of-same-type-of-a-template-class

WeiqunZhang:
Unfortunately, nvcc does not work.

AlexanderSinn:
Maybe instead of being IntVectND, Zero and Unit could have a different type that is implicitly convertible to IntVectND. This would allow them to be `static constexpr`.

WeiqunZhang:
That may not work in cases like
```
IntVect::Zero + sign(a_side) * 2 * BASISV(idir);
```
This is copied from a real code.

Maybe we can try to figure out how to make nvcc happy. The error message is
```
/home/runner/work/amrex/amrex/Src/Base/AMReX_IntVect.H:775:34: error: conflicting declaration ‘amrex::IntVectND<3> amrex::IntVectND<3>::Zero’
  775 | inline constexpr IntVectND<AMREX_SPACEDIM> IntVectND<AMREX_SPACEDIM>::Zero{0};
      |                                  ^~~~~~~~~~~~~
/home/runner/work/amrex/amrex/Src/Base/AMReX_IntVect.H:713:24: note: previous declaration as ‘const amrex::IntVectND<3> amrex::IntVectND<3>::Zero’
  713 |     static const IntVectND<dim> Zero;
      |                        ^~~~
```

WeiqunZhang:
Maybe it wants `constexpr const`?

AlexanderSinn:
It seems to work if both are just const with no constexpr.

WeiqunZhang:
Great. Maybe it will even work for all `<int dim>`.

AlexanderSinn:
constexpr const actually works somehow

WeiqunZhang:
@ax3l This PR will break pyamrex. How do you want to handle this? Should we wait till pyamrex has a PR ready?

ax3l:
Thanks for the ping. That is all we need to track.

Please go ahead here to merge and we update pyAMReX shortly after.

Src/Base/AMReX_IntVect.H
```diff
@@ -34,196 +35,239 @@ int coarsen (int i, int ratio) noexcept
...
     */
-    static const IntVect Zero;
-
-    ///
-    /**
-       This is an IntVect all of whose components are equal to one.
-    */
-    static const IntVect Unit;
+    constexpr IntVectND () noexcept {
+        for (int i=0; i<dim; ++i) {
```
WeiqunZhang: We could remove the loop because `int vect[dim] = {};` will initialize all the elements to zero.

Src/Base/AMReX_IntVect.H
```diff
@@ -34,196 +35,235 @@ int coarsen (int i, int ratio) noexcept
...
 
     //return coordinate with largest value
     [[nodiscard]] AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
     int maxDir(bool a_doAbsValue) const noexcept;
 
-    //! Returns a reference to the i'th coordinate of the IntVect.
-    [[nodiscard]] AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
-    int& operator[] (int i) noexcept { BL_ASSERT(i>=0 && i < AMREX_SPACEDIM); return vect[i]; }
+    //! Returns a reference to the i'th coordinate of the IntVectND.
+    [[nodiscard]] AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE constexpr
```
WeiqunZhang, line 224: Because of the use of `BL_ASSERT`, are we allowed to mark this as constexpr?

AlexanderSinn: Probably not since `amrex::Assert` is not constexpr. However regular `assert` apparently is since C++14 so maybe we can use that or make `amrex::Assert` constexpr.

WeiqunZhang: `static_assert(IntVect::TheNodeVector()[0] == 1);` compiles in debug mode.

`static_assert(IntVect::TheNodeVector()[3] == 1);` in debug mode fails with
```
main.cpp:10:49:   in ‘constexpr’ expansion of ‘amrex::IntVectND<3>::TheNodeVector().amrex::IntVectND<3>::operator[](3)’
/home/wqzhang/mygitrepo/amrex/Src/Base/AMReX_BLassert.H:45:54: error: call to non-‘constexpr’ function ‘void amrex::Assert(const char*, const char*, int, const char*)’
   45 | #define    BL_ASSERT(EX) (EX)?((void)0):amrex::Assert( # EX , __FILE__, __LINE__)
      |                                         ~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/wqzhang/mygitrepo/amrex/Src/Base/AMReX_IntVect.H:229:40: note: in expansion of macro ‘BL_ASSERT’
  229 |     int& operator[] (int i) noexcept { BL_ASSERT(i>=0 && i < dim); return vect[i]; }
      |                                        ^~~~~~~~~
```

Non-debug build fails with
```
   10 |         static_assert(IntVect::TheNodeVector()[3] == 1);
main.cpp:10:51: error: array subscript value ‘3’ is outside the bounds of array type ‘int [3]’
```

Maybe we can still leave it as constexpr.

AlexanderSinn: I see the same thing:
In a test it seems to kind of work as-is. When compiling in debug mode and using IntVect in a constexpr context, the index is known at compile time. If the index is within the bounds it compiles fine, however if the index is outside the bounds, it gives a compiler error
```
/home/asinn/hipace/src/particles/beam/BeamParticleContainer.cpp(508): error: expression must have a constant value
/home/asinn/amrex/Src/Base/AMReX_IntVect.H(229): note #2703-D: cannot call non-constexpr function "amrex::Assert"
/home/asinn/amrex/Src/Base/AMReX.H(167): here
```

WeiqunZhang: https://stackoverflow.com/questions/47726503/calling-non-constexpr-function-from-constexpr-allowed-in-some-conditions

AlexanderSinn: C++ assert works in the same way
```
<source>: In function 'int main()':
<source>:11:27:   in 'constexpr' expansion of 'func(15)'
<source>:5:5: error: call to non-'constexpr' function 'void __assert_fail(const char*, const char*, unsigned int, const char*)'
    5 |     assert(i < 10);
      |     ^~~~~~
/usr/include/assert.h:69:13: note: 'void __assert_fail(const char*, const char*, unsigned int, const char*)' declared here
   69 | extern void __assert_fail (const char *__assertion, const char *__file,
      |             ^~~~~~~~~~~~~
```

WeiqunZhang (approved):

Src/Base/AMReX_IntVect.H
```diff
@@ -34,196 +35,235 @@ int coarsen (int i, int ratio) noexcept
...
-    static const IntVect Unit;
-
-    ///
-    /**
-    * \brief Construct an IntVect whose components are all zero.
-    */
-    constexpr IntVect () noexcept {} // cannot use = default due to Clang bug // NOLINT
+    template <class...Args,
+        std::enable_if_t<
+            (sizeof...(Args)+2 == dim) &&
```
ax3l, line 97: I am currently having trouble to get this constructor to work with
```C++
if constexpr (dim > 1) {
    py_iv.def(py::init<AMREX_D_DECL(int, int, int)>());
}
```
in pyAMReX.

Why `+2`? Shouldn't this be `-2` the way the signature is written below?

WeiqunZhang: In 3D, `(int i, int j, Args...ks)` has `sizeof...(Args)` = 1. `1+2 == 3` is what we want, not `1-2`.

ax3l: Ah, cannot use the macro, obvs..

ax3l:
pyAMReX update coming via https://github.com/AMReX-Codes/pyamrex/pull/332 :)

PR #3977, opened 2024-06-08
Add structured binding support to GpuTuple

AlexanderSinn:
## Summary

This PR adds C++17 structured binding support to `amrex::GpuTuple`.

## Additional background

See case 2 in https://en.cppreference.com/w/cpp/language/structured_binding.
Note that structured bindings will also use the existing `amrex::get`.

## Checklist

The proposed changes:


WeiqunZhang:
Nice! It works as expected. Note that since specializing std::get (a function template) has been outlawed since C++20, but specializing std class templates are still legit.

WeiqunZhang (approved):

PR #3987, opened 2024-06-13
Multi-level composite solver using hypre

WeiqunZhang:
This adds a multi-level composite solver for cell-centered ABecLaplacian using hypre. Currently it uses the sstruct interface and BoomerAMG. What's different from our existing hypre solvers is that this one does not use amrex::MLMG at all and it can do multi-level composite solve entirely with hypre, whereas all previous solvers use hypre as a "bottom" solver of MLMG's v-cycle (even though the v-cycle could have only one level).


BenWibking:
Are there plans to add SSAMG support: https://github.com/hypre-space/hypre/blob/recmat/src/sstruct_ls/ssamg.c?

WeiqunZhang:
Yes, ssamg support will be added in the near future.

WeiqunZhang:
I just pushed a commit that adds SSAMG support. For a CPU test, it works on 1, 2, and 4 processes, but segfaults on 3 processes. Not sure if it's our bug or a bug in hypre. BoomerAMG works for all these cases on CPU.

We still have some issues on GPU.

asalmgren (approved):

PR #3994, opened 2024-06-20
New CMake option: AMReX_CCACHE

WeiqunZhang:
Add a new CMake option, AMReX_CCACHE[=OFF].

Enable ccache in a Windows CI that used to take about 1.5 hours. With ccache, it reduces to xxx.


WeiqunZhang:
Finally able to make cmake with visual studio generator use ccache. Unfortunately, some msvc flags are not supported.

PR #3996, opened 2024-06-20
Workaround std::exclusive_scan issue with old compilers

WeiqunZhang:
Some systems do not have std::exclusive_scan. Thus, we implement a workaround.


WeiqunZhang:
@marchdf

asalmgren (approved):

marchdf:
This built with clang-15 on Frontier.

PR #4006, opened 2024-06-25
HDF5: Correctly choose datatype of plotfile based on RD numBytes

arnav-singhal:
## Summary
If numBytes is 4, use H5T_NATIVE_FLOAT instead of H5T_NATIVE_DOUBLE. This fixes writing real values, as in ExaEpi.
## Additional background
As far as I can tell, whichRD (and therefore whichRDBytes) does not change between levels, so I perform this check outside of the level loop, but if it is indeed possible for these values to change, I can move it inside.
## Checklist

The proposed changes:


atmyers (approved):

atmyers:
Thanks!

PR #4009, opened 2024-06-26
Fix L2 norms in C_util/Convergence codes

ajnonaka:
## Summary

The multifab L2 norm routines simply return sum(a dot a). Thus, when you sum up norms from each grid you do NOT need to take them to the ^NORM power. The cell count (i.e. volume) scaling is performed lower

The formula is now implemented correctly, for p!=0:

L^p = ( 1/Ncell sum (phi_f - phi_c)^p ) ^(1/p)

(note that using cell volume instead of 1/Ncell accomplishes the proper scaling as well, which is what the code does)

## Checklist

The proposed changes:

### DIFF
--- a/Tools/C_util/Convergence/DiffSameDomainRefined.cpp
+++ b/Tools/C_util/Convergence/DiffSameDomainRefined.cpp
@@ -245,7 +245,7 @@ main (int   argc,
 
                     if (norm != 0)
                     {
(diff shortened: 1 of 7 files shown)


zingale:
note that it also worked for p =1

zingale (approved):

zingale:
this is a nice catch.  L2 never behaved well for me

PR #4013, opened 2024-07-02
Convexify AMR data

WeiqunZhang:
Add a function that "convexifies" AMR data by removing cells that are covered by fine levels from coarse level MultiFabs. This could be useful for visualization.

### DIFF
--- a/Src/Base/AMReX_MultiFabUtil.H
+++ b/Src/Base/AMReX_MultiFabUtil.H
@@ -410,7 +410,7 @@ namespace amrex
      * MultiFabs have no ghost cells. For nodal data, the nodes on the
      * coarse/fine interface exist on both levels.
      */
-    Vector<MultiFab> convexify (Vector<MultiFab const*> const& mf,
+    [[nodiscard]] Vector<MultiFab> convexify (Vector<MultiFab const*> const& mf,
                                 Vector<IntVect> const& refinement_ratio);
 }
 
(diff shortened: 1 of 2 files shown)


Src/Base/AMReX_MultiFabUtil.H
```diff
@@ -398,6 +398,20 @@ namespace amrex
...
+     *
+     * This function "convexifies" the AMR data by removing cells that are
+     * covered by fine levels from coarse level MultiFabs. This could be
+     * useful for visualization. The returned MultiFabs on coarse levels
+     * have different BoxArrays from the original BoxArrays. For the finest
+     * level, the data is simply copied to the returned object. The returned
+     * MultiFabs have no ghost cells. For nodal data, the nodes on the
+     * coarse/fine interface exist on both levels.
+     */
+    Vector<MultiFab> convexify (Vector<MultiFab const*> const& mf,
```
WeiqunZhang:
```suggestion
    [[nodiscard]] Vector<MultiFab> convexify (Vector<MultiFab const*> const& mf,
```

asalmgren:
Does "removes cells" really mean "removes boxes"?

I can't see how you would remove individual cells and still have a boxarray of what's left.

WeiqunZhang:
Right, they are not isolated cells. The cells belong to boxes. But those boxes might not be the original boxes in the boxarray.

asalmgren (approved):

WeiqunZhang:
Some AMReX users are using some visualization tools that need the shapes of "leaf" boxes to be convex. In AMReX, the coarse level Boxes might have a hole in the middle if we take out the region covered by fine. That was the motivation.

PR #3988, opened 2024-06-17
convert IndexType to IndexTypeND

AlexanderSinn:
## Summary

Similar to #3969 but for IndexType.

## Additional background

A maximum of 31 dimensions are supported so that (1u << dim) can fit into an unsigned int.

## Checklist

The proposed changes:
- [ ] fix a bug or incorrect behavior in AMReX
- [x] add new capabilities to AMReX
- [ ] changes answers in the test suite to more than roundoff level
- [ ] are likely to significantly affect the results of downstream AMReX users
- [ ] include documentation in the code and/or rst files, if appropriate

PR #4016, opened 2024-07-03
Convert Box to BoxND

AlexanderSinn:
## Summary

Similar to #3969 and #3988 but for Box.

## Additional background

It should be checked that the changes to BoxIndexer do not affect the compiled GPU code.
In my testing, it gives the same performance as development. 

Example usage:
```C++
amrex::BoxND b1{amrex::IntVectND{1,2,3}, amrex::IntVectND{4,5,6}, amrex::IntVectND{1,0,1}};
// ((1,2,3) (4,5,6) (1,0,1))
auto b2 = amrex::BoxCat(b1, b1, b1);
// ((1,2,3,1,2,3,1,2,3) (4,5,6,4,5,6,4,5,6) (1,0,1,1,0,1,1,0,1))
auto [b3, b4, b5, b6, b7] = amrex::BoxSplit<1, 4, 2, 1, 1>(b2);
// ((1) (4) (1))((2,3,1,2) (5,6,4,5) (0,1,1,0))((3,1) (6,4) (1,1))((2) (5) (0))((3) (6) (1))
auto b8 = amrex::BoxResize<2>(b4);
// ((2,3) (5,6) (0,1))
auto b9 = amrex::BoxResize<5>(b8);
// ((2,3,0,0,0) (5,6,0,0,0) (0,1,0,0,0))
```

## Checklist

The proposed changes:


AlexanderSinn:
I added `IntVectND::dim3(int fill_extra)` to make it easier to implement BoxND functions such as end and length. It looks like pyamrex needs an overload_cast for dim3() now.

PR: https://github.com/AMReX-Codes/pyamrex/pull/345

ax3l:
https://github.com/AMReX-Codes/pyamrex/pull/345 merged - thanks!

Src/Base/AMReX_Box.H
```diff
@@ -205,180 +191,215 @@ public:
...
-    * \brief Returns the number of points contained in the Box.
+    * \brief Returns the number of points contained in the BoxND.
     */
     [[nodiscard]] AMREX_GPU_HOST_DEVICE
     Long numPts () const noexcept {
-        return ok() ? AMREX_D_TERM( static_cast<Long>(length(0)),
-                                   *static_cast<Long>(length(1)),
-                                   *static_cast<Long>(length(2)))
-                    : Long(0);
+        AMREX_IF_ON_HOST((checkOverflow();))
```
WeiqunZhang, line 348: Could we put this inside `ifdef AMREX_DEBUG`? I don't think we need to check for overflow here for non-debug builds, given it's very unlikely overflow could happen here.

AlexanderSinn: I also added `defined(AMREX_USE_ASSERTION)` as for me debug mode is often too slow to use.

Src/Base/AMReX_Box.H
```diff
@@ -1565,225 +1653,323 @@ Box adjCell (const Box& b, Orientation face, int len=1) noexcept
...
-#elif (AMREX_SPACEDIM == 3)
-    return Box(IntVect(amrex::max(b1.smallend[0], b2.smallend[0]),
-                       amrex::max(b1.smallend[1], b2.smallend[1]),
-                       amrex::max(b1.smallend[2], b2.smallend[2])),
-               IntVect(amrex::min(b1.bigend  [0], b2.bigend  [0]),
-                       amrex::min(b1.bigend  [1], b2.bigend  [1]),
-                       amrex::min(b1.bigend  [2], b2.bigend  [2])),
-               typ);
-#endif
+BoxND<dim> minBox (BoxND<dim> const& b1, BoxND<dim> const& b2, IndexTypeND<dim> typ) noexcept
```
WeiqunZhang: Let's remove this function. I don't think it's being used anywhere now. The name of this function is very misleading. So to avoid confusion, let's just remove it.

WeiqunZhang:
LGTM. I will do one more pass tomorrow.

Src/Base/AMReX_Box.H
AlexanderSinn: All minBox functions or just that one?

WeiqunZhang: Just this one. This one has different behavior from others. It was added by me 5 years ago. I made a mistake in naming.

WeiqunZhang (approved):

PR #4032, opened 2024-07-16
ParmParse: Math expression support

WeiqunZhang:
ParmParse: Math expression support

Support math expression for integers and floating point numbers in
ParmParse. For example,

    n_cell = 128
    amrex.n_cell = {n_cell*2} 8 {16**2}

becomes

    n_cell = 128
    amrex.n_cell = 256 8 256

Note that the user must use `{}` to indicate that the expression needs to be
evaluate with amrex::Parser. If the expression contains another variable, it
will be looked up by ParmParse.

ParmParse's constructor now accepts an optional second argument,
`parser_prefix`. When a variable in a math expression is being looked up, it
will first try to find it by using the exact name of the variable. If this
attempt fails and the `ParmParse` object has a non-empty `parser_prefix`, it
will try again, this time looking up the variable by prefixing its name with
`parser_prefix` followed by a `.`. For example,

    physical_constants.c = 3.e8
    amrex.pi = 3.14
    amrex.foo = { sin(amrex.pi/2) * c**2 }

will give `foo = 8.999997146e+16` if we do

    amrex::ParmParse pp("amrex", "physical_constants");
    double foo;
    pp.get("foo", foo);

We also have to make a change to amrex::Parser and iParser. The variable
names can contain `.` now.


WeiqunZhang:
Depend on #4031

BenWibking:
What is the order of evaluation of variables? Will it complain if I do:
```
code.a = {code.b}
code.b = {code.a}
```

WeiqunZhang:
The recursion will not stop until it segfaults. @BenWibking thanks for pointing this out. We should give a warning.

As for evaluation orders, I have added an example to the documentation. Does that answer your question?

```
The variables in :cpp:`ParmParse` math expressions are not evaluated until
they are referenced. If a variable is defined multiple times, the last
occurrence will override previous ones even if it appears after the variable
has been referenced. This behavior is demonstrated in the following example.

.. highlight:: python

::

    foo.a = 1                                                                                                                                                                
    foo.b = {foo.a}                                                                                                                                                          
    foo.a = 2                                                                                                                                                                

will become

.. highlight:: python

::

    foo.a = 2                                                                                                                                                                
    foo.b = 2                 
```

WeiqunZhang:
We just discussed this in WarpX meeting.

* @AlexanderSinn prefers `"` `"` to `{}`.
* @AlexanderSinn suggested that a static `parser_prefix` in `ParmParse` (in addition to the non-static member)
* @AlexanderSinn and @dpgrote mentioned that ideally it would be easier for the users if neither `"` nor `{}` is needed. But I think this creates ambiguities and it's actually error prone for the users because the result depends on which `ParmParse` functions (`get`, `getarr` etc.) are used. For example
    ```
    amr.n_cell = 64 64 +16 128 # four numbers or three?
    foo = sin(x) -cos(y) # is foo a scalar or two numbers sin(x) and -cos(y)?
    ```

WeiqunZhang:
```
code.a = {code.b}
code.b = {code.a}
```
will result in
```
amrex::Error::0::ParmParse: recursive reference to {code.b} is not allowed !!!
SIGABRT
```

AlexanderSinn:
How about if the user does not add `{}` or `""`, the expression is not allowed to contain a space (and maybe also not equals), but otherwise it works without `{}` or `""`.

WeiqunZhang:
This is from https://github.com/Hi-PACE/hipace/blob/development/examples/laser/inputs_SI
```
my_constants.kp_inv = 10.e-6
geometry.prob_lo     = -6.*kp_inv   -6.*kp_inv   -8.*kp_inv
geometry.prob_hi     =  6.*kp_inv    6.*kp_inv    6.*kp_inv
```
They contain spaces. So for this to work, we would need to use `{}` or `""`. In Hipace++, this is handled with a special `fillWithParserArr` function. I haven't tried yet. But if we use `""` as follows, it seems that it should work with both Hipace++ and this PR (modified to use `""`).
```
my_constants.kp_inv = 10.e-6
geometry.prob_lo     = "-6.*kp_inv"   "-6.*kp_inv"   "-8.*kp_inv"
geometry.prob_hi     =  "6.*kp_inv"    "6.*kp_inv"    "6.*kp_inv"
```
On the other hand, if we use `{}`, it might be easier for you to detect in Hipace++ and try to maintain backward compatibility.

AlexanderSinn:
With "the expression is not allowed to contain a space" I mean the individual expressions that are given to the parser, so that ParmParse can simply split the line by spaces so that
```
geometry.prob_lo     = -6.*kp_inv   -6.*kp_inv   -8.*kp_inv
geometry.prob_hi     = "-6. * kp_inv"   "-6. * kp_inv"   "-8. * kp_inv"
```
would work, but not
```
geometry.prob_lo     = -6. * kp_inv   -6. * kp_inv   -8. * kp_inv
```
The second version would currently get an error in hipace along the lines of: wrong number of arguments, expected 3 got 9.
One special thing about the current implementation in hipace is that for input parameters that are known to be a scalar value, it is not split by spaces and everything including the spaces is parsed to the Parser (more precisely it is split by spaces by ParmParse querryarr but then all elements are combined again). However, I'm fine with changing this so that scalar parameters also don't allow spaces without escaping.

AlexanderSinn:
Note that `geometry.prob_lo` and `geometry.prob_hi` are amrex parameters, but they are parsed and replaced by the hipace functions.

WeiqunZhang:
I am worried about it being too sensitive to empty spaces. Once the users start using math expressions, I suspect some of them might be quite long. It would then suddenly fail, if it accidentally has an unwanted space.

dpgrote:
WarpX is doing the same thing as hipace. When a scalar is read in, it reads in the parameter as an array of strings and concatenates them and then parses the result. Then for arrays, there can be no spaces in each element of the array. This gives the very nice well defined and easy to read expressions like the example above.
```
geometry.prob_lo     = -6.*kp_inv   -6.*kp_inv   -8.*kp_inv
```
I've taken this same code and put it in MauiC as well, giving the same nice format. It would be a shame to uglify this and force the use of `{}` or `""`.

In many places in WarpX and MauiC, there are already checks on the number of elements when arrays are read in so this will catch any extraneous spaces that mighty appear anyway. To make every input parameter less nice visually just in case that rare long expression might have an extra space seems like a poor trade off.

WeiqunZhang:
If we have
```
a = 1 -2
```
and
```
ParmParse pp;
int a;
pp.get("a", a);
```
What should `a` be? @dpgrote @AlexanderSinn  I think @dpgrote 's answer is -1, whereas @AlexanderSinn 's is 1.

dpgrote:
I'm sure that hipace copied the code from WarpX so it would be doing the same thing and get -1. His description is the same "it is split by spaces by ParmParse querryarr but then all elements are combined again".

WeiqunZhang:
That's not backward compatible. Currently with the development branch we get 1.

WeiqunZhang:
If we don't use `{}` or `""`, all white spaces need to be significant for int and real unless they are inside `""`. We cannot just combine the whole line into a single scalar.

WeiqunZhang:
I suspect that that might create more issues for WarpX and Hipace++ if both codes migrate to the new ParmParse.

WeiqunZhang:
By the way, in all seriousness, my opinion is python would be a better language if it uses `{}`.

dpgrote:
> That's not backward compatible. Currently with the development branch we get 1.

Ok, so for scalars parameters now it only reads in the first element of a space separate array. I would be fine keeping it this way, specifying no spaces in scalar expressions. This would be consistent with no spaces in array elements. Alex also is fine with this. This is not backwards compatible with WarpX but is probably not a problem since uses of spaces in scalars is probably quite rare anyway.

dpgrote:
> I suspect that that might create more issues for WarpX and Hipace++ if both codes migrate to the new ParmParse.

Requiring `{}` or `""` would create far more issues. Almost every input file would need to be updated with extensive changes. With requiring no spaces in scalars, there would probably be very few updates required.

WeiqunZhang:
WarpX has things like `my_constants.n0 = beam_charge / (q_e * sigmax * sigmay * sigmaz * (2.*pi)**(3./2.))`. Maybe we can add functions like `ParmParse::queryWithParser` that will combine everything into one string before parsing.

AlexanderSinn:
> If we have
> 
> ```
> a = 1 -2
> ```
> 
> and
> 
> ```
> ParmParse pp;
> int a;
> pp.get("a", a);
> ```
> 
> What should `a` be? @dpgrote @AlexanderSinn I think @dpgrote 's answer is -1, whereas @AlexanderSinn 's is 1.

To me this should either be  `-1` or error: expected one element got two.

AlexanderSinn:
The behavior where not everything is read in (which would result in 1) is intentionally avoided and asserted in hipace for all parameters.

WeiqunZhang:
https://github.com/AMReX-Codes/amrex/blob/bc86eacae9e7204ba1689fd9f05af4e5e62529af/Docs/sphinx_documentation/source/Basics.rst#math-expressions

WeiqunZhang:
This PR also depends on #4038.

WeiqunZhang:
As for relationship with #4035, either this one or #4035 can be merged after #4031. The two are likely to have conflicts, but I will fix them after one is merged.

Docs/sphinx_documentation/source/Basics.rst
```diff
@@ -275,6 +275,147 @@ by default returns the last one. The difference between :cpp:`query` and
...
+    double g[] = 7.2 11.6
+    w = 1 with query
+    w = -1 with queryParser
+    foo = 5, bar = 5
+    double prob_lo[] = {1, 1, 1}
+    double prob_hi[] = {1, 1, 1}
+
+Note that the empty spaces are significant for math expressions unless they
+are inside a pair of ``"`` or explicitly parsed by
+:cpp:`ParmParse::queryWithParer` or :cpp:`ParmParse::getWithParser`. If the
```
dpgrote: Typo `queryWithParser`

WeiqunZhang:
```suggestion
:cpp:`ParmParse::queryWithParser` or :cpp:`ParmParse::getWithParser`. If the
```

Docs/sphinx_documentation/source/Basics.rst
```diff
@@ -275,6 +275,147 @@ by default returns the last one. The difference between :cpp:`query` and
...
+expression contains another variable, it will be looked up by
+:cpp:`ParmParse`. :cpp:`ParmParse`'s constructor accepts an optional second
+argument, ``parser_prefix``. When a variable in a math expression is being
+looked up, it will first try to find it by using the exact name of the
+variable. If this attempt fails and the :cpp:`ParmParse` object has a
+non-empty non-static member ``parser_prefix``, it will try again, this time
+looking up the variable by prefixing its name with the value of
+``parser_prefix`` followed by a ``.``. If this attempt also fails and the
+:cpp:`ParmParse` class has a non-empty static member ``ParserPrefix`` (which
```
dpgrote, line 392: Interesting. Is there a reason why you have two different parser prefix options?

WeiqunZhang: Initially there was only one that was stored as a non-static member. It's safer, but it would require passing the prefix string to the constructor. So a static member was added to give people more options.

WeiqunZhang:
Note that allowing spaces inside `""` comes for free because that's how we handle strings.

dpgrote (approved):
Looks good!

AlexanderSinn:
I tested this PR by using it in HiPACE++ and replacing the custom parser functions with amrex pp.get and pp.query. It did mostly work with existing input scripts. I had to add a RealVect overload. I didn't use queryWithParser and there actually were silent errors in the script where "" would be needed (as expected). To make parser functions that are evaluated on the GPU, it would be nice if some of the variable reading functionality from pp_parser could be reused by users. One issue I found was that for `beam1.num_particles = 5e9` which is queried as Long, the iParser couldn't parse `5e9` and after I changed it to `5*10^9` it overflowed to 705032704 without giving an error.

WeiqunZhang:
We could change iParser to use long int. As for `5e9`, we could support that if that does not cause any confusion. We can do that in a separate PR.

Could you elaborate on "To make parser functions that are evaluated on the GPU, it would be nice if some of the variable reading functionality from pp_parser could be reused by users."?

AlexanderSinn:
We use amrex::Parser and amrex::ParserExecutor to evaluate functions on the GPU that come from the input script. For this we currently get them as a string from ParmParse, give that to the Parser and then replace all the unknown symbols with constants from ParmParse, just like it's done in pp_parser.

WeiqunZhang:
Okay, I think I get what you mean. We can add a `Parser ParmParse::makeParser (const char* name)`, where `name` is a ParmParse name like `f(x,y)` in `f(x,y) = a*x+b*y`, or we we add a `Parser ParmParse::makeParser (std::string const& function_body)`, where `function_body` is `a*x+b*y`. In both cases, we will set known constants. Or should we implement both versions? How to avoid confusion?

Or we can take the hipace++ approach. https://github.com/Hi-PACE/hipace/blob/5d8dda3cad53b5fc051eaa98937ba3eb8d3c9bb0/src/utils/Parser.H#L510

WeiqunZhang:
As for iparser, maybe it's better to support `5'000'000'000` than `5e9`.

AlexanderSinn:
I prefer the `std::string const& function_body` version as its more generall and can be combined with get/query/getarr/queryarr and other sources as needed.

AlexanderSinn:
For iParser `5'000'000'000` would be nice, but I think also `5e9` makes sense so long as there is an assert that the number is an integer and fits inside int/long (`4.32e1`and `5e20` would not be allowed). `amrex.the_arena_init_size` should also benefit from this.

WeiqunZhang:
Added `makeParser` and `makeIParser`. I will open another PR for switching to long long int in IParser.

Src/Base/AMReX_ParmParse.H
```diff
@@ -997,9 +1012,70 @@ public:
...
+    }
+
     //! Remove given name from the table.
     int remove (const char* name);
 
+    //! Make Parser using given string `func` as function body and `vars` as
+    //! variable names. Constants known to ParmParse will be set. It's a
+    //! runtime error, if there are unknown symbols in `func`.
+    Parser makeParser (std::string const& func,
+                       Vector<std::string> const& vars) const;
```
WeiqunZhang:
```suggestion
    [[nodiscard]] Parser makeParser (std::string const& func,
                                     Vector<std::string> const& vars) const;
```

Src/Base/AMReX_ParmParse.H
```diff
@@ -997,9 +1012,70 @@ public:
...
+
+    //! Make IParser using given string `func` as function body and `vars`
+    //! as variable names. Constants known to ParmParse will be set. It's a
+    //! runtime error, if there are unknown symbols in `func`.
+    IParser makeIParser (std::string const& func,
+                         Vector<std::string> const& vars) const;
```
WeiqunZhang:
```suggestion
    [[nodiscard]] IParser makeIParser (std::string const& func,
                                       Vector<std::string> const& vars) const;
```

atmyers (approved):

PR #4035, opened 2024-07-17
ParmParse: Refactoring II

WeiqunZhang:
Use std::unordered_map instead of std::vector. This speeds up significantly when there are thousands of parameters.


WeiqunZhang:
Depends on #4031

asalmgren (approved):

PR #4039, opened 2024-07-18
FillPatch for Face Data: Fix Geometry used in creating mask

WeiqunZhang:


cgilet (approved):
This fixes the issue. Thank you.

PR #4043, opened 2024-07-21
Nodal Solver: Use multi-color Gauss-Seidel on GPU

WeiqunZhang:
Because the nodal solver's Gauss-Seidel smoother is not thread safe on GPU, we have been using Jacobi. But recently an amr-wind test shows that the Jacobi smoother is not stable for some hard problems. Using a multi-color (4 in 2d and 8 in 3d) Gauss-Seidel smoother appears to work well.


WeiqunZhang:
This has passed our regression tests on CPU and GPU. The amr-wind test with anisotropic resolution works.

asalmgren (approved):

PR #4048, opened 2024-07-25
Fix typos in FaceDivFree interior interpolation.

cgilet:
The problem was masked in test in amrex/Tests/DivFreePatch because all three components were the same.

## Checklist

The proposed changes:


BenWibking:
This will affect us once we turn on AMR for our MHD problems. Could the test also be updated to catch the bug?

kngott:
I don't have the math in front of me, so I can't fully confirm this is correct, but the change _looks_ correct, based on the pattern, so I trust it. 😄

kngott (approved):

cgilet:
@BenWibking , yes, I'm looking at updating the test too. That's part of why this is still a draft. 

Do you also use the fortran interface for divergence-preserving interpolation? I haven't looked at that, or anything specific to E&M, since my use cases don't have that.

BenWibking:
> @BenWibking , yes, I'm looking at updating the test too. That's part of why this is still a draft.
> 
> Do you also use the fortran interface for divergence-preserving interpolation? I haven't looked at that, or anything specific to E&M, since my use cases don't have that.

No, only the C++ interface.

PR #4049, opened 2024-07-25
AmrData: remove dependency on Fortran

WeiqunZhang:
This allows us to compile Amrvis without a Fortran compiler.

Also removed AmrData::Interp which is never being used anywhere.


drummerdoc (approved):
Looks good!  Thanks weiqun.

WeiqunZhang:
I also updated the makefile in Amrvis to remove Fortran.

PR #4055, opened 2024-07-31
Documentation: Runtime Parameters

WeiqunZhang:
* Update the chapter on runtime parameters.

* Add amrex prefix to a number of obscure parameters. This should not cause any backward compatibility issues.

* For uncommon parameters, use query instread of queryAdd. So the user will not see them when print out all parameters.

* Remove fab.init_snan because there is now amrex.init_snan.


WeiqunZhang:
Ready for review, but don't merge before the next release.

asalmgren (approved):

atmyers (approved):

PR #4060, opened 2024-08-01
Use squeryWithParser instead of squeryval in pp_make_parser

AlexanderSinn:
## Summary

When an input parameter refers to a user constant or another input parameter, this change will enable the user constant/second parameter to contain spaces without the need for `""` instead of only parsing it up to the first space.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_ParmParse.cpp
+++ b/Src/Base/AMReX_ParmParse.cpp
@@ -868,6 +868,24 @@ saddarr (const std::string& name, const std::vector<T>& ref)
     ++entry.m_count;
 }
 
+template <class T>
+bool squeryWithParser (const ParmParse::Table& table,
+                       const std::string&      parser_prefix,
+                       const std::string&      name,
+                       T&                      ref)
+{
+    std::vector<std::string> vals;
+    bool exist = squeryarr(table, parser_prefix, name, vals,
+                           ParmParse::FIRST, ParmParse::ALL, ParmParse::LAST);
+    if (!exist) { return false; }
+
+    std::string combined_string;
+    for (auto const& v : vals) {
+        combined_string.append(v);
+    }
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
I am not sure this is desired. The issue is we get different values depending on whether we query it directly or indirectly.

AlexanderSinn:
Right now, one would get different values compared to using queryWithParser directly. My thinking is that ignoring part of the input (after space) makes no sense for constants, as they are always scalar. At least there should be an assert that there is no space.

WeiqunZhang:
Maybe we should treat query and query parser differently during recursion.

AlexanderSinn:
That would still have the issue as it is possible for constants to be used by multiple parameters.
```bash
# works with pp.queryWithParser, silently incorrect with pp.query
my_constants.kp_inv = clight / wp
my_constants.kp = wp / clight
my_constants.wp = sqrt( ne * q_e^2/(m_e *epsilon0) )
my_constants.ne = 1.0505e23
# pp.query
geometry.prob_lo     = -18*kp_inv   -18*kp_inv   -1.5*kp_inv
# pp.query
geometry.prob_hi     =  18*kp_inv    18*kp_inv    1.5*kp_inv 
# pp.queryWithParser
laser.w0 = 3 * kp_inv 
# pp.query, would be kp_inv with amrex develepment, 2.083*kp_inv with this PR, ideally this would give an error
laser.L0 = -prob_lo/18 
```

WeiqunZhang:
Maybe add a new method, `queryarrWithParser`?

AlexanderSinn:
Yes it would be nice to have a parser function for an array that asserts that the length of the input is correct.

WeiqunZhang:
@AlexanderSinn Could you give #4073 a try?

PR #4063, opened 2024-08-06
Workaround for CUDA 12.6

WeiqunZhang:
It appears that in CUDA 12.6 cub explicitly include nvtx3. This causes a conflict with nvtx included by AMReX_TinyProfiler.cpp and AMReX_GpuDevice.cpp. This is a workaround for the issue.

I think this is a cub bug. It's incompatible with the default `#include <nvTools.h>` included by nvcc.


WeiqunZhang:
Fix it by using NVTX3 (#4064)

PR #4064, opened 2024-08-06
Use NVTX3 to avoid conflict with cub in CUDA 12.6

WeiqunZhang:


ax3l (approved):

Src/Base/AMReX_TinyProfiler.H
```diff
@@ -5,14 +5,6 @@
 #include <AMReX_INT.H>
 #include <AMReX_REAL.H>
 
-#ifdef AMREX_USE_CUDA
-#include <nvToolsExt.h>
-#endif
-
-#if defined(AMREX_USE_HIP) && defined(AMREX_USE_ROCTX)
-#include <roctracer/roctx.h>
```
ax3l, line 13: Where did those go?

PR #4073, opened 2024-08-08
ParmParse: queryarrWithParser

WeiqunZhang:
Add a new function queryarrWithParser that will query an array of values using Parser. The function takes the number of elements as an argument too. If the size of the input found with ParmParse does not match the size argument, it's a runtime error. For each individual array element, queryWithParser is used for the query.

The behavior of queryWithParser has changed. Previously, the query function is used for unresolved symbols. Now the queryWithParser function is used instead. The difference is that the latter will ignore the spaces in the input and combine the entire input into a single string.


AlexanderSinn:
I'm out of office, so I can't try it now, but this should fix the issue for me as I can now use pp_parser with use_querywithparser = true directly.

asalmgren (approved):

PR #4078, opened 2024-08-14
Make sure ChopGrids does not violate refinement ratio.

WeiqunZhang:
When using ChopGrids on fine level BoxArrays, we must make sure the box sizes are multiples of refinement ratios.


asalmgren (approved):

WeiqunZhang:
Note that the issue we are trying to fix only affected odd ref rations like 3.

PR #4079, opened 2024-08-14
BLBackTrace: Check for `addr2line` in path first

### LINKED ISSUE #4077: Hardcoded path to `addr2line` or `eu-addr2line` in `BLBackTrace`
The paths to `addr2line`and `eu-addr2line` are hardcoded to
```
/usr/bin/addr2line
```
and
```
/usr/bin/eu-addr2line
```
respectively in [AMReX_BLBackTrace.cpp](https://github.com/AMReX-Codes/amrex/blob/e792933f9561c9ab0e47238c041a6c09818f0908/Src/Base/AMReX_BLBackTrace.cpp#L210-L226). These might not be the ones we want to use if we are not using the system binutils (or elfutils) (e.g., we have built a newer binutils using Spack). This has caused issues for me and @draenog where the newer binutils we are using compresses the debug info in a way that the older `addr2line` from the system binutils is not able to cope with and we get the following errors
```
/usr/bin/addr2line: /path/to/my_amrex_application.ex: unable to initialize decompress status for section .debug_info
/usr/bin/addr2line: /path/to/my_amrex_application.ex: File format not recognized
```

This code should instead check what `addr2line` or `eu-addr2line` is available in the `PATH` first and can then resort to the above hardcoded paths if the command is not found.

mirenradia:
## Summary
This fixes #4077 by first checking if `addr2line` or `eu-addr2line` is in the `PATH` first and then falls back to `/usr/bin/addr2line` or `/usr/bin/eu-addr2line` if not.

## Additional background
See #4077 for why using the hardcoded path is a problem.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_BLBackTrace.cpp
+++ b/Src/Base/AMReX_BLBackTrace.cpp
@@ -182,12 +182,12 @@ namespace {
 #ifdef __linux__
     bool command_exists(std::string const &cmd)
     {
-      // command -v is part of POSIX so should be available
-      std::string check_command = "command -v " + cmd + " > /dev/null 2>&1";
-      int r = std::system(check_command.c_str());
-      // return value of std::system is implementation defined and can be
-      // decoded using WEXITSTATUS but it should be 0 on success
-      return r == 0;
+        // command -v is part of POSIX so should be available
+        std::string check_command = "command -v " + cmd + " > /dev/null 2>&1";
+        int r = std::system(check_command.c_str());
+        // return value of std::system is implementation defined and can be
+        // decoded using WEXITSTATUS but it should be 0 on success
+        return r == 0;
     }
 #endif
 }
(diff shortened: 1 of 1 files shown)


Src/Base/AMReX_BLBackTrace.cpp
```diff
@@ -209,19 +222,32 @@ BLBackTrace::print_backtrace_info (FILE* f)
...
         std::string eu_cmd;
         {
-            have_eu_addr2line = file_exists("/usr/bin/eu-addr2line");
+          if (command_exists("eu-addr2line")) {
+            have_eu_addr2line = 1;
+            eu_cmd = "eu-addr2line";
+          } else {
+            std::string eu_fallback_path = "/usr/bin/eu-addr2line";
+            have_eu_addr2line = file_exists(eu_fallback_path.c_str());
+            eu_cmd = eu_fallback_path;
```
WeiqunZhang: We could move the string here.

Src/Base/AMReX_BLBackTrace.cpp
```diff
@@ -209,19 +222,32 @@ BLBackTrace::print_backtrace_info (FILE* f)
...
         std::string cmd;
         {
-            have_addr2line = file_exists("/usr/bin/addr2line");
+          if (command_exists("addr2line")) {
+            have_addr2line = 1;
+            cmd = "addr2line";
+          } else {
+            std::string fallback_path = "/usr/bin/addr2line";
+            have_addr2line = file_exists(fallback_path.c_str());
+            cmd = fallback_path;
```
WeiqunZhang: We could move the string here.

WeiqunZhang (changes requested):
LGTM. Could we use 4 spaces for indentation?

WeiqunZhang (approved):

PR #4081, opened 2024-08-14
Add overload for `getParticleCell` that returns local cell index

roelof-groenewald:
## Summary
In WarpX we have a few routines that requires the local cell index of particles (i.e. the cell number in the current box). The AMReX function `getParticleCell` currently only has logic to return the global cell index (relative to the domain low end). This PR adds a new overload for that function which does not shift the cell index based on the global box index.

## Additional background
See https://github.com/ECP-WarpX/WarpX/pull/5118 for WarpX context.

## Checklist

The proposed changes:


atmyers:
Hi @roelof-groenewald - thanks for this! Could you add short docstrings to these functions that explain the difference between them?

atmyers (approved):

PR #4083, opened 2024-08-15
MLEBNodeFDLaplacian: Make it work with AMREX_USE_EB but with no EB

WeiqunZhang:
This will allow the use of the linear solver when AMReX is built with EB support at compile time but no EB is built at runtime.

X-Ref: https://github.com/ECP-WarpX/WarpX/pull/4865

### DIFF
--- a/Src/LinearSolvers/MLMG/AMReX_MLEBNodeFDLaplacian.cpp
+++ b/Src/LinearSolvers/MLMG/AMReX_MLEBNodeFDLaplacian.cpp
@@ -433,7 +433,7 @@ MLEBNodeFDLaplacian::Fapply (int amrlev, int mglev, MultiFab& out, const MultiFa
         Array4<int const> const& dmarr = dmask.const_array(mfi);
 #ifdef AMREX_USE_EB
         bool cutfab = edgecent[0] && edgecent[0]->ok(mfi);
-        if (cutfab) {
+        if (cutfab && factory) { // clang-tidy is not that smart
             AMREX_D_TERM(Array4<Real const> const& ecx = edgecent[0]->const_array(mfi);,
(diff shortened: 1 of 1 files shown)


ax3l (approved):
Cool, thanks!

ax3l:
Oopsi, I see some regressions with this PR in https://github.com/ECP-WarpX/WarpX/pull/5126#issuecomment-2292419396 affecting EB-compiled (and EB-runtime-used) tests that now fail to converge.

Src/LinearSolvers/MLMG/AMReX_MLEBNodeFDLaplacian.cpp
```diff
@@ -422,12 +432,12 @@ MLEBNodeFDLaplacian::Fapply (int amrlev, int mglev, MultiFab& out, const MultiFa
...
 #ifdef AMREX_USE_EB
-        bool cutfab = edgecent[0]->ok(mfi);
-        if (cutfab) {
+        bool cutfab = edgecent[0] && edgecent[0]->ok(mfi);
+        if (cutfab && factory) { // clang-tidy is not that smart
             AMREX_D_TERM(Array4<Real const> const& ecx = edgecent[0]->const_array(mfi);,
                          Array4<Real const> const& ecy = edgecent[1]->const_array(mfi);,
                          Array4<Real const> const& ecz = edgecent[2]->const_array(mfi));
-            auto const& levset = levset_mf.const_array(mfi);
+            auto const& levset = factory->getVolFrac().const_array(mfi);
```
ax3l, line 440: This looks like a bug
```suggestion
            auto const& levset = factory->getLevelSet().const_array(mfi);
```

PR #4088, opened 2024-08-17
Time integrator interface updates

gardner48:
## Summary

Remove `const` on input state to time integrator right-hand side (RHS) functions to allow filling ghost cells within the RHS call. Remove the `set_pre_rhs_action` and `set_post_update` functions in the time integrator interface. Update documentation to reflect changes from #3984.

## Additional background

Before the updates in #3984 a `set_post_update` was used attach a function for filling ghost cells after computing a new stage or solution in a time integrator. #3984 expanded the time integrator interface with SUNDIALS to support implicit and ImEx methods which require updating the ghost cells for RHS evaluations inside iterative solvers. To support these methods, a new `set_pre_rhs_action` function was added to attach a function for filling ghost cells. This created some confusion/inconsistencies in how to fill ghost cells between explicit and implicit methods. To simplify things, this PR removes `const` from the input state to a RHS function so its ghost cells can be filled as part of the RHS callback and removes the `pre_rhs_action`/`set_post_update` functions. 

## Checklist

The proposed changes:

cc: @ajnonaka

### DIFF
--- a/Docs/sphinx_documentation/source/TimeIntegration_Chapter.rst
+++ b/Docs/sphinx_documentation/source/TimeIntegration_Chapter.rst
@@ -16,13 +16,8 @@ A Simple Time Integrator Setup
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 This is best shown with some sample code that sets up a time integrator and
-asks it to step forwards by some interval ``dt``. The user needs to supply at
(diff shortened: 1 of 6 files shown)


ajnonaka (approved):
The amrex-tutorial SUNDIALS example checks out fine.  Thanks for the updates and the documentation!

Issue #4087, opened 2024-08-16
`amrex::ParmParse::dumpTable` shows duplicate values


ax3l:
@JustinRayAngus discovered that since recently, the function `amrex::ParmParse::dumpTable` starts to plot values multiple times, likely because they were changed. The old values should not be shown.

Used in WarpX and ImpactX to plot the ultimately used inputs, e.g. here:
https://github.com/ECP-WarpX/WarpX/blob/24.08/Source/ablastr/utils/UsedInputsFile.cpp

WeiqunZhang:
The "issue" here is mainly a conscious choice of WarpX. WarpX processes many ParmParse parameters using Parser first and then explicitly uses `add` or `addarr` to add the post-processed values to ParmParse.

I tested with 24.07, which does not have the recent ParmParse changes. It has the same "issue". The only difference is in the old way the entries were not sorted so it was harder to notice.

WeiqunZhang:
For example,
```
$ git grep -n "pp.*add" Source/
Source/Diagnostics/BTD_Plotfile_Header_Impl.H:263:    /** Append particle info of the newly added box, namely,
Source/Diagnostics/FlushFormats/FlushFormatOpenPMD.cpp:33:    pp_diag_name.add("openpmd_backend", openpmd_backend);
Source/Diagnostics/ReducedDiags/ParticleHistogram2D.cpp:68:    pp_rd_name.add("openpmd_backend", m_openpmd_backend);
Source/EmbeddedBoundary/WarpXInitEB.cpp:104:            pp_eb2.add("geom_type", geom_type); // use all_regular by default
Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp:423:    // modified if such support wants to be added
Source/Initialization/WarpXAMReXInit.cpp:49:            pp_amr.add("blocking_factor", 1);
Source/Parallelization/WarpXComm.cpp:1051:    // overlapping points among boxes. Using the mask, we can add the
Source/Particles/PhysicalParticleContainer.cpp:392:    pp_species_name.queryarr("addIntegerAttributes", m_user_int_attribs);
Source/Particles/PhysicalParticleContainer.cpp:407:    pp_species_name.queryarr("addRealAttributes", m_user_real_attribs);
Source/Utils/WarpXUtil.cpp:55:        a_pp.addarr(input_str, input_array);
Source/Utils/WarpXUtil.cpp:93:    pp_geometry.addarr("prob_lo", prob_lo);
[...]

PR #4093, opened 2024-08-19
Fix a bug in ParmParse::remove

WeiqunZhang:
Should have used prefixed name.

### DIFF
--- a/Src/Base/AMReX_ParmParse.cpp
+++ b/Src/Base/AMReX_ParmParse.cpp
@@ -1854,7 +1854,7 @@ int
 ParmParse::remove (const char* name)
 {
     auto const pname = prefixedName(name);
-    auto n = m_table->erase(name);
+    auto n = m_table->erase(pname);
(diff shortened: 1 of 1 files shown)


ax3l:
Thanks! Just to connect it well, this fixes #4087 ?

Src/Base/AMReX_ParmParse.cpp
```diff
@@ -1854,7 +1854,7 @@ int
 ParmParse::remove (const char* name)
 {
     auto const pname = prefixedName(name);
```
ax3l, line 1856: Interesting that this did not throw an unused variable warning. Should we double check CI covers this?

ax3l (approved):

ax3l:
Huh, why does the WarpX app build still fail? We merged https://github.com/ECP-WarpX/WarpX/pull/5126 the other day.

Update: aha, this checks out WarpX 24.08 not its development.

PR #4096, opened 2024-08-19
Fortran BoxArray: Add nboxes function returning the number of boxes.

### LINKED ISSUE #4094: Easiest way to access boxes in fortran
Hello,

I'm using the amrcore fortran interface, and I'm trying to gain access to the details of all the boxes at each level. I can access the array of geometry objects which gives me info about extent, periodicity, and cell size. I can also access the BoxArray object at each level, but I haven't been able to identify an easy way to loop through each box in the BoxArray and extract - say - its range of indices. Is there a simple way to do that?

I think I understand how it's done using MFiter starting from a multifab, but it's not exactly what I'm looking for for a few reasons:
- I'm interested in looking into what the grid looks like, independently of the data stored on that grid
- I'm interested in doing that on a single processor (doing I/O, for example). I understand that a boxarray is a global concept, so it seems all processors should have all the details of all the boxes. It seems however that mfiter would only loop through the box owned specifically by the processor doing the call.

Is there a simple way to do that from fortran? Maybe I'm fundamentally misunderstanding things. For example, I could imagine a scenario where the concept of a box is fundamentally tied to the multifab, i.e., each multifab might have a different distribution of boxes, and therefore my question is meaningless. But I feel that it's probably not the case, and that given an amrcore, all zero-ghost single-component multifabs would have exactly the same boxes at all levels.

Thanks for your help!

Olivier

WeiqunZhang:
Note that the function returns the total number of boxes including ones not owned by the MPI process calling the function.

### DIFF
--- a/Src/F_Interfaces/Base/AMReX_boxarray_fi.cpp
+++ b/Src/F_Interfaces/Base/AMReX_boxarray_fi.cpp
@@ -47,6 +47,11 @@ extern "C" {
         ba->maxSize(iv);
     }
 
+    Long amrex_fi_boxarray_nboxes (const BoxArray* ba)
+    {
+        return ba->size();
+    }
+
     void amrex_fi_boxarray_get_box (const BoxArray* ba, int i, int lo[3], int hi[3])
     {
         const Box& bx = (*ba)[i];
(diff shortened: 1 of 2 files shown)


asalmgren (approved):

desjardi:
Great, thanks for the blazing fast turn-around! Quick question though - I see that you have the number of boxes returned as a long/integer(amrex_long), but the get_box function takes in an integer. That seems inconsistent, and may need to be changed so that both procedures use long and integer(amrex_long)?

WeiqunZhang:
Yes, you are right that the two are not consistent. Unfortunately the inconsistency has already existed in the C++ code.

desjardi:
Fair enough - thanks again for your help!

PR #4098, opened 2024-08-20
Fix start_comp offset in tracer particle routines

atmyers:
This makes the routines work as expected given the variable names (currently, `ncomp` was being treated as `end_comp`) 

The proposed changes:


asalmgren (approved):

RevathiJambunathan (commented):
Thanks for fixing this @atmyers !

PR #4100, opened 2024-08-21
TinyProfiler:: Add more runtime parameters

### LINKED ISSUE #3895: Tiny Profiler: Runtime Options (Enable/Disable, Filename)
As discussed today with @WeiqunZhang @AlexanderSinn in the WarpX developer meeting:
Adding a few more runtime options in AMReX to optionally redirect the tiny profiler table summary into a file (and also to disable prints) would be awesome (e.g., for Python runs).

## Current option

A C++ output stream that would be user-implemented (in apps), e.g., to pass an fstream.

## New Options?
A runtime option that defaults to `1` if TP is compiled, but can be disabled to supress printing the table (or profiling altogether even? Maybe two options?)
```ini
tiny_profiler.enabled = 1
tiny_profiler.print = 1
```

A runtime option to set an fstream as a runtime option (so apps can default it and users can overwrite it)
```ini
tiny_profiler.outname = "diags/profile.txt"  # default: none, using stdout
```

WeiqunZhang:
Add the following ParmParse parameters to TinyProfiler:

  * tiny_profiler.enabled

    This parameter can be used to disable tiny profiling including CArena memory profiling at run time.

  * tiny_profiler.memprof_enabled

    This parameter can be used to disable tiny profiler's memory arena profiling at run time. If tiny_profiler.enabled is false, this parameter has no effects.

  * tiny_profiler.output_file

    If this parameter is empty, the output of tiny profiling is dumped on the default out stream of AMReX. If it's not empty, it specifies the file name for the output. Note that `/dev/null` is a special name that mean a null file.

Close #3895

### DIFF
--- a/Docs/sphinx_documentation/source/RuntimeParameters.rst
+++ b/Docs/sphinx_documentation/source/RuntimeParameters.rst
@@ -1204,3 +1204,35 @@ enabled.
   which are unnecessary for correctness, could potentially degrade the
   performance.
 
+.. py:data:: tiny_profiler.enabled
+   :type: bool
+   :value: true
+
+   .. versionadded:: 24.09
+      Runtime parameter `tiny_profiler.enabled``.
+
+   This parameter can be used to disable tiny profiling including
+   :cpp:`CArena` memory profiling at run time.
+
+.. py:data:: tiny_profiler.memprof_enabled
+   :type: bool
+   :value: true
+
+   .. versionadded:: 24.09
+      Runtime parameter ``tiny_profiler.memprof_enabled``.
+
+   This parameter can be used to disable :cpp:`CArena` memory profiling at
+   run time. If ``tiny_profiler.enabled`` is false, this parameter has no
+   effects.
+
+.. py:data:: tiny_profiler.output_file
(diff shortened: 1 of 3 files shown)


atmyers (approved):

Docs/sphinx_documentation/source/RuntimeParameters.rst
```diff
@@ -1204,3 +1204,35 @@ enabled.
...
+   :type: string
+   :value: [empty]
+
+   .. versionadded:: 24.09
+      Runtime parameter ``tiny_profiler.output_file``.
+
+   If this parameter is empty, the output of tiny profiling is dumped on the
+   default out stream of AMReX. If it's not empty, it specifies the file
+   name for the output. Note that ``/dev/null`` is a special name that mean
+   a null file.
```
ax3l, line 1238: Maybe `that means no output.`?

WeiqunZhang: I will add that to the TinyProfiler::flush PR I am preparing.

PR #4106, opened 2024-08-22
Support for multiple periods in FillBoundary and ParallelCopy

WeiqunZhang:
This is a feature requested by a user who wants to run simulations with the number of ghost cells more than the domain length in periodic directions.


asalmgren (approved):

asalmgren:
My only comment is about the PR description: "This is a feature requested by a user who wants to run simulations with the number of ghost cells less than the domain length in periodic directions."   

That should read "This is a feature requested by a user who wants to run simulations with the number of ghost cells MORE than the domain length in periodic directions."

PR #4110, opened 2024-08-26
CI: WarpX on `development`

ax3l:
## Summary

Avoid that breaking changes slip in between weekly updates and AMReX release tagging.

## Checklist

The proposed changes:

### DIFF
--- a/.github/workflows/apps.yml
+++ b/.github/workflows/apps.yml
@@ -62,15 +62,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
-    - name: Get Latest Release Tag
-      run: |
-        WARPX_TAG=$(wget https://github.com/ECP-WarpX/WarpX/releases/latest 2>&1 | grep Location: | awk '{print $2}' | awk -F/ '{print $NF}')
-        echo "WARPX_TAG=$WARPX_TAG" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


.github/workflows/apps.yml
```diff
@@ -62,15 +62,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
-    - name: Get Latest Release Tag
-      run: |
-        WARPX_TAG=$(wget https://github.com/ECP-WarpX/WarpX/releases/latest 2>&1 | grep Location: | awk '{print $2}' | awk -F/ '{print $NF}')
```
ax3l, line 67: Discussed: ideal would be an update to find the latest commit that has the words `AMReX`, `Update` (case insensitive and arbitrary order) in it
https://github.com/ECP-WarpX/WarpX/pull/5172
or `Release`
https://github.com/ECP-WarpX/WarpX/pull/5125
or is the latest tag, whatever is newer.

ax3l:
@WeiqunZhang do we want to merge this for now? We can idealize it as described above if WarpX should surprisingly break without an AMReX update on `development` (I am not sure how often that happens at all for the CPU compile test that is covered here - most breakage we see should be from breaking AMReX changes until we adopt for them weekly).

WeiqunZhang (approved):

PR #4116, opened 2024-08-28
`AMREX_DEVICE_PRINTF`: Host

ax3l:
## Summary

In AMReX, device means host if compiled for CPUs.
Add support for printf debugging on CPUs for `AMREX_DEVICE_PRINTF`.

Currently, the macro was undefined, unless for the special case of SYCL compilation for host code paths where it worked.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_GpuPrint.H
+++ b/Src/Base/AMReX_GpuPrint.H
@@ -23,6 +23,9 @@
 #  define AMREX_DEVICE_PRINTF(...) std::printf(__VA_ARGS__);
 #elif defined(AMREX_USE_HIP)
 #  define AMREX_DEVICE_PRINTF(...) ::printf(__VA_ARGS__);
+#else
+#  define AMREX_DEVICE_PRINTF(format,...) { \
+      std::printf(format, __VA_ARGS__); }
 #endif
 
 #endif  // !defined(__APPLE__)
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
It was intentional, because AMREX_DEVICE_PRINTF really should be for debugging only.

WeiqunZhang:
It was intended as a feature, not a bug.

ax3l:
Yes, but both on host as well as on device this is breaking compiler optimizations and bloating registers.

I want to use it exclusively inside a
```C++
#if defined(AMREX_DEBUG) || defined(DEBUG)
#endif
```
block.

ax3l:
A common construct where this macro is used that I saw is this:
```C++
#if defined(AMREX_DEBUG)
    AMREX_IF_ON_DEVICE((
        AMREX_DEVICE_PRINTF("Message %d\n", n);
    ))
    AMREX_IF_ON_HOST((
        std::stringstream ss;
        ss << "Message " << n << "\n";
        amrex::Abort(ss.str());
    ))
#endif
```

Maybe we could provide a macro for this?
The difference is:
- only defined in debug
- will abort on the host, too

```C++
#define AMREX_DEBUG_DEVICE_PRINTF(Format,...) ...
```

WeiqunZhang (approved):

PR #4118, opened 2024-08-29
Simplify cache cleanup actions

WeiqunZhang:
For the CleanUpCache workflow run triggered by PR workflows, there is a simple way of getting the PR number. The full event webhook payload is available as a json file and it contains the number of the triggering PR. So we don't need to upload the PR number as an artifact.

For the post-PR cleanup, we could also clean up the cache directly without uploading the PR number and triggering a workflow run for cleaning.


WeiqunZhang:
Unfortunately, the changes only work for PRs from the same repo.

PR #4119, opened 2024-08-29
Capability adds for ParmParse enum

baperry2:
## Summary

- The ParmParse functions for AMReX Enums added in #4069 did not carry the optional arguments found in other ParmParse functions. Those are added so that, for example, querying an enum has the same interface as querying other types.
- These functions shouldn't modify the ParmParse object, so they are made `const`
- The `prefixedName` function is made a public member instead of a protected member.

## Additional background

The use case for these changes is https://github.com/erf-model/ERF/pull/1772

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_ParmParse.H
+++ b/Src/Base/AMReX_ParmParse.H
@@ -1393,9 +1393,9 @@ public:
 
     static std::string ParserPrefix;
 
-protected:
-
     [[nodiscard]] std::string prefixedName (const std::string_view& str) const;
+
+protected:
 
     std::string m_prefix; // Prefix used in keyword search
     std::string m_parser_prefix; // Prefix used by Parser
--- a/Src/Base/AMReX_ParmParse.H
+++ b/Src/Base/AMReX_ParmParse.H
@@ -1200,6 +1200,8 @@ public:
             try {
                 ref = amrex::getEnum<T>(s);
             } catch (...) {
+                amrex::Print() << "amrex::ParmParse::query (input name: "
+                               << this->prefixedName(name) << "):" << std::endl;
                 throw;
             }
         }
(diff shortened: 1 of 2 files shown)


baperry2:
I got clang-tidy complaints about ParmParse `add` and `addarr` functions not being `const` - I don't think that should be related to what I changed. In any event, I tried making all those const, which seems to be fine other than breaking the pyamrex interface. @WeiqunZhang and @asalmgren, any thoughts on what to do with this issue? Otherwise, this PR should be ready for review.

By the way, I really like the added capability for parmparsing enums. I can already think of a ton of places where that can drastically simplify code in various apps.

WeiqunZhang:
Re: constness, one could argue that `add` and `addarr` are expected to change the *logical* state of ParmParse. So I think we should leave them as non-const functions. If clang-tidy complains, let's use `NOLINT(readability-make-member-function-const)`.

Yes, we should add `const` to query and get.

https://isocpp.org/wiki/faq/const-correctness#logical-vs-physical-const

baperry2:
Agreed that add/addarr should intuitively not be const despite what clang-tidy thinks. I added some NOLINTs for those functions  instead of making const.

Src/Base/AMReX_ParmParse.H
```diff
@@ -1192,14 +1192,16 @@ public:
...
+    int query (const char* name, T& ref, int ival = FIRST) const
     {
         std::string s;
-        int exist = this->query(name, s);
+        int exist = this->query(name, s, ival);
         if (exist) {
             try {
                 ref = amrex::getEnum<T>(s);
             } catch (...) {
+                amrex::Print() << "amrex::ParmParse::query (input name: "
```
WeiqunZhang: We need to allow the users to turn off warning messages. So we either remove this, or put this inside `if (amrex::Verbose() > xxx)`, or add the message to what is being thrown by amrex::getEnum.

baperry2: Done.

WeiqunZhang (approved):

PR #4126, opened 2024-09-03
`ParmParse`: Prefix to `FILE`

ax3l:
## Summary

For CI/CD workflows and out-of-source tests we often want to include dependent inputs files via `FILE = <filename>`. For development, we prefer to run in temporary run directories but want to avoid having to copy over the latest inputs file from a source directory (mostly to avoid confusion between source and copy and to enable rapid development cycles).

Now, the environment variable `AMREX_INPUTS_FILE_PREFIX` can be set to prefix every `FILE = <filename>` with a custom path. We will use this in the CTests integration of WarpX.

## Additional background

CC @EZoni
https://github.com/ECP-WarpX/WarpX/pull/5068

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_ParmParse.cpp
+++ b/Src/Base/AMReX_ParmParse.cpp
@@ -9,6 +9,7 @@
 
 #include <algorithm>
 #include <cctype>
+#include <cstdlib>
 #include <iostream>
 #include <limits>
 #include <numeric>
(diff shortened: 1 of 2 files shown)


Src/Base/AMReX_ParmParse.cpp
```diff
@@ -407,6 +408,14 @@ read_file (const char* fname, ParmParse::Table& tab)
     //
     if ( fname != nullptr && fname[0] != 0 )
     {
+        std::string filename = fname;
+
+        // optional prefix to search files in
+        char const *amrex_inputs_file_prefix = std::getenv("AMREX_INPUTS_FILE_PREFIX");
+        if (amrex_inputs_file_prefix != nullptr) {
+            filename = std::string(amrex_inputs_file_prefix) + filename;
```
ax3l: Make more robust to append a `/` (or `\`) if missing in the prefix.

ax3l: done (universal `/` added)

ax3l:
Checked by @EZoni today: this should be ready for merge now :)

WeiqunZhang (approved):

PR #4129, opened 2024-09-04
MLMG: Minimum domain width

WeiqunZhang:
By default, the minimum domain width at the coarsest multigrid level is 2 and 4, for non-EB and EB runs, respectively. Previously, it was set to 4 for runs compiled with EB support but without EB at run time. So the results of those runs would be different from runs not compiled with EB support. This is not a correctness issue. Nevertheless, for the sake of consistence, we set the minimum domain width to 4 only if there is non-trivial EB at run time.


ax3l:
Thanks a lot, testing in https://github.com/ECP-WarpX/WarpX/pull/4865 now.

ax3l (approved):
Fantastic, that solves the discrepancies on results and performance :tada:

PR #4131, opened 2024-09-05
EB Boundary Area: Fix issues for anisotropic cell size

WeiqunZhang:
* In 2D, the scaling in incorrect. For example, if dx >> dy and the boundary is parallel to the x-direction, it would produce inf for the scaled boundary area.

* In 3D, the scaling has been modified so that it's easy to convert from the dimensionless boundary area to physical units.


asalmgren (approved):

WeiqunZhang:
I will update the documentation and then merge.

WeiqunZhang:
@RemiLehe

Issue #3670, opened 2023-12-14
add support for 2D spherical geometry


zingale:
We'd like to model a thin surface layer on a star in 2d spherical geometry (r, theta) with phi axisymmetry.  Here theta is the angle from the +z axis.

The geometry should look like the attached:

[2d_spherical.pdf](https://github.com/AMReX-Codes/amrex/files/13676456/2d_spherical.pdf)

We'll use this issue to track the changes needed in AMReX

zingale:
We want to allow for `Coord() == 2` in 2-d

Here are the places where the geometry matters and what needs to be done (if anything):

- [ ] volume and area factors need to be added to `AMReX_COORDSYS_2D_C.H` and we need to make sure that `AMReX_Geometry.cpp` doesn't abort for our geometry
- [ ] we need to ensure that the interpolation works with this geometry.  In particular we want the conversed interpolation to work.  Note: we can debug this with piecewise constant to start
- [ ] the restriction should "just work" so long as the volume factors are in place.  But we should check the `avgDown` functions to see if there are aborts
- [ ] we rely on the muligrid apply_op to compute the diffusion term.  At the moment, in Castro, we seem to do this by applying metric factors on the application side, so we need to add those metrics.  Alternately, the linear solvers can now handle geometry for us, so we should consider switching to that and then we need to make sure that this new geometry is supported.
- [ ] the flux correction for hyperbolic should also be fine, since the area factors are in place

asalmgren:
Closing this issue due to lack of activity.  Please re-open it if you feel there is is still an issue!

PR #4141, opened 2024-09-08
add geometric terms for spherical 2D support.

zhichen3:
## Summary
This adds the appropriate geometric terms (area and volume) for spherical 2D geometry. We assume dx[0]=dr,  and dx[1]=d$`\theta`$.
## Additional background
This intended to deal with issue #3670 
## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_CoordSys.cpp
+++ b/Src/Base/AMReX_CoordSys.cpp
@@ -164,6 +164,7 @@ CoordSys::UpperIndex(const Real* point) const noexcept
     IntVect ix;
     for (int k = 0; k < AMREX_SPACEDIM; k++)
     {
+        // +1 for UpperIndex?
         ix[k] = (int) ((point[k]-offset[k])/dx[k]);
     }
     return ix;
@@ -330,6 +331,8 @@ CoordSys::GetEdgeVolCoord (Vector<Real>& vc,
     GetEdgeLoc(vc,region,dir);
     //
     // In R direction of RZ, vol coord = (r^2)/2
+    // In R direction of SPHERICAL, vol coord = (r^3)/3
+    // In theta direction of SPHERICAL, vol coord = -cos(theta)
     //
 #if (AMREX_SPACEDIM == 2)
     if (dir == 0 && c_sys == RZ)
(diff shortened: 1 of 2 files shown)


WeiqunZhang (approved):

WeiqunZhang:
LGTM

Src/Base/AMReX_CoordSys.cpp
```diff
@@ -461,7 +513,10 @@ CoordSys::Volume (const Real xlo[AMREX_SPACEDIM],
                             *(xhi[2]-xlo[2]));
 #if (AMREX_SPACEDIM==2)
     case RZ:
-        return static_cast<Real>(0.5*TWOPI)*(xhi[1]-xlo[1])*(xhi[0]*xhi[0]-xlo[0]*xlo[0]);
+        return static_cast<Real>(0.5*TWOPI)*(xhi[1]-xlo[1])*(xhi[0]+xlo[0])*(xhi[0]-xlo[0]);
```
zingale: this will cause roundoff diffs when merged

zingale:
this looks good to me too, but I don't really understand where getEdgeVolCoord is used.

Src/Base/AMReX_CoordSys.cpp
zhichen3: should I change it back?

Src/Base/AMReX_CoordSys.cpp
```diff
@@ -164,6 +164,7 @@ CoordSys::UpperIndex(const Real* point) const noexcept
     IntVect ix;
     for (int k = 0; k < AMREX_SPACEDIM; k++)
     {
+        // +1 for UpperIndex?
```
zhichen3: btw, I was wondering if `UpperIndex` should have a +1

zingale: it looks like it should, but it also appears that this isn't used anywhere

Src/Base/AMReX_CoordSys.cpp
```diff
@@ -164,7 +164,7 @@ CoordSys::UpperIndex(const Real* point) const noexcept
     IntVect ix;
     for (int k = 0; k < AMREX_SPACEDIM; k++)
     {
-        ix[k] = (int) ((point[k]-offset[k])/dx[k]);
```
WeiqunZhang, line 167: Maybe we should do +1 to the int, not the real number before the cast.

Src/Base/AMReX_CoordSys.cpp
WeiqunZhang: Yes, that would be better.

WeiqunZhang:
> this looks good to me too, but I don't really understand where getEdgeVolCoord is used.

`getEdgeVolCoord` was used in BoxLib's Intepolater. But at one point, we decided that amrex will only support 1d spherical, 2d rz, cartesian coordinates, and inlined the edge vol corrds into the interpolation code for simplicity.

For you 2D, spherical, you might need to modify interpolaters for better accuracy.

WeiqunZhang:
You want to make sure CellConservativeLinear does conserve mass in 2d spherical coordinates.

Src/Base/AMReX_CoordSys.cpp
zhichen3: okay, I changed it back.

Src/Base/AMReX_CoordSys.cpp
zhichen3: I moved the +1 outside int()

PR #4143, opened 2024-09-10
AMREX_ENUM: Add more capabilites

WeiqunZhang:
Add support for `=`. For example,

```c++
    AMREX_ENUM(MyColor,
               red,
               chi=red,
               green,
               blue,
               Default = green);
```
The underlying values are then
```c++
    { red = 0,
      chi = 0,
      green = 1,
      blue = 2,
      Default = 1}
```
We can then have
```c++
    assert(amrex::getEnum<MyColor2>("red") == MyColor2::red);
    assert(amrex::getEnum<MyColor2>("chi") == MyColor2::chi);
    assert(amrex::getEnum<MyColor2>("green") == MyColor2::green);
    assert(amrex::getEnum<MyColor2>("blue") == MyColor2::blue);
    assert(amrex::getEnum<MyColor2>("Default") == MyColor2::Default);
```
Because of there are duplicated values, using value to get name string will return the first matched string. Currently, I don't see a way to fix it. But, since `MyColor::red == MyColor::chi` is true, it's also not outrageously wrong.
```c++
    assert(amrex::getEnumNameString(MyColor2::red) == "red");
    assert(amrex::getEnumNameString(MyColor2::chi) == "red");
    assert(amrex::getEnumNameString(MyColor2::green) == "green");
    assert(amrex::getEnumNameString(MyColor2::blue) == "blue");
    assert(amrex::getEnumNameString(MyColor2::Default) == "green");
```
P.S. Chi means red in Chinese. https://en.wikipedia.org/wiki/Radical_155

### DIFF
--- a/Src/Base/AMReX_Enum.H
+++ b/Src/Base/AMReX_Enum.H
@@ -15,49 +15,91 @@
 template <typename T>
 using amrex_enum_traits = decltype(amrex_get_enum_traits(std::declval<T>()));
 
-namespace amrex {
+namespace amrex
+{
+    template <typename T, typename ET = amrex_enum_traits<T>,
+              std::enable_if_t<ET::value,int> = 0>
+    std::vector<std::pair<std::string,T>> getEnumNameValuePairs ()
+    {
+        auto tmp = amrex::split(std::string(ET::enum_names), ",");
+        std::vector<std::pair<std::string,T>> r;
+        r.reserve(tmp.size());
+        int next_value = 0;
+        for (auto const& item : tmp) {
+            auto const& kv = amrex::split(item, "=");
+            if (kv.size() == 1) {
+                r.emplace_back(amrex::trim(kv[0]), static_cast<T>(next_value));
+                ++next_value;
+            } else if (kv.size() == 2) {
+                auto const& value_string = amrex::trim(kv[1]);
+                auto it = std::find_if(r.begin(), r.end(),
+                                       [&] (auto const& x) -> bool
+                                           { return x.first == value_string; });
+                auto this_value = it->second;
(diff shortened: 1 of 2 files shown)


atmyers (approved):

baperry2:
This is great - just yesterday I was thinking capability like this would be very helpful!

One additional extension that would be nice would be to somehow let arbitrary parmparse strings be set for each quantity in the enum rather than exactly matching the enum names (which this almost, but doesn't quite do by setting `someEnum = desiredName`). The example use case I'm thinking of is in ERF with this enum: `AMREX_ENUM(PBLType, None, MYNN25, YSU);`. Here `MYNN25` is actually supposed to be `MYNN2.5`; the `.` can't be in the enum name but it would still be nice to have it be in the option that users set in their input file (`erf.pbl_type = MYNN2.5`). Obviously the current solution of just omitting the `.` isn't a huge deal, but being able to do something like `AMREX_ENUM(PBLType, None, MYNN25 = "MYNN2.5", YSU);` would be handy if not too difficult to implement, and I'm sure would be useful in other instances as well.

WeiqunZhang:
I don't see any obvious solutions, because `MYNN25 = "MYMN2.5"` is not allowed in standard C++'s enumerator.

baperry2:
Ok, that makes sense.

WeiqunZhang:
But we could add an optional argument to `query_enum_case_insensitive`. Something like
```
AMREX_ENUM(PBLType, None, MYNN25, YSU);

pp.query_enum_case_insensitive("erf.pbl_type", pbl_type, ".");
```
All characters in the last argument will be ignored.

baperry2:
I like that as a good way to achieve the same effect.

PR #4145, opened 2024-09-11
Add ParmParse::query_enum_sloppy that can ignore characters

WeiqunZhang:
The characters in the `ignores` argument will be ignored as if they don't exist in the value part of `name = value`.


ax3l (approved):
Nice, thanks for the test, too!

PR #4146, opened 2024-09-11
Vis multicuts

ejyoo921:
## Summary
Visualization of multi-cut locations.
1) In the input file, you can add eb2.plt_multiple_cuts = "true" to produce plot files to see the multi-cut cells.
-> The description was added to the RuntimeParameters.rst file.
* Note: If "eb2.plt_multiple_cuts" is true, we abort the code after creating the plot files. This is in the AMReX_EB2_Level.H.

2) For a 3D case, you would get up to three plot files: plt.multicut.x, plt.multicut.y, and plt.multicut.z.

This depends on which face has multi-cut issues.
-> See AMReX_EB2_3D_C.cpp

3) For a 2D case, the output name is plt.multicut.
-> See AMReX_EB2_2D_C.cpp

4) For any dimension, a cell value of 2 indicates regular, 0 means covered, and values greater than 2 indicate multi-cut cells.

## Checklist

The proposed changes:

### DIFF
--- a/Docs/sphinx_documentation/source/RuntimeParameters.rst
+++ b/Docs/sphinx_documentation/source/RuntimeParameters.rst
@@ -787,6 +787,15 @@ Embedded Boundary
    .. tip::  Because AMReX currently does not support multi-cut cells, it
              would be a runtime error if multi-cut cells are left unfixed.
 
+.. py:data:: eb2.plt_multiple_cuts
+   :type: bool
+   :value: false
+
+   If this parameter is set to true and multi-cut cells are present,
+   we produce plot files to see the location of the multi-cut issue.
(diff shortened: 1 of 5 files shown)


WeiqunZhang:
Plotfile is for cell-centered data only. But the data you want to save are face data.

ejyoo921:
> Plotfile is for cell-centered data only. But the data you want to save are face data.

I can change them into cell-centered data. Would that be okay?

ejyoo921:
I changed the face-centered cells to cell-centered cells for visualizations. It cannot precisely show which face has the multi-cut issue; however, the name of the plot files shows what direction of faces has a problem. For example, "plt. multi cut.x" would contain multi-cut cells having issues on x-direction faces.

PR #4148, opened 2024-09-12
Fix roundoff issue in SUNDIALS evolve()

ajnonaka:
## Summary
Fix a roundoff issue in the SUNDIALS evolve() function that caused the integrator to run a second time with a very tiny dt in a given time step.

## Checklist

The proposed changes:
- [X ] fix a bug or incorrect behavior in AMReX

### DIFF
--- a/Src/Base/AMReX_RKIntegrator.H
+++ b/Src/Base/AMReX_RKIntegrator.H
@@ -252,7 +252,7 @@ public:
         {
             // Adjust step size to reach output time
             // protect against roundoff
-            if (std::abs(time_out - time_current - dt) <= 1.e5 * std::numeric_limits<amrex::Real>::epsilon() * std::abs(dt)) {
+            if (almostEqual(std::abs(time_out-time_current), std::abs(dt)), 10.) {
                 dt = time_out - time_current;
                 stop = true;
             }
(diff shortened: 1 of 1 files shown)


Src/Base/AMReX_RKIntegrator.H
```diff
@@ -251,7 +251,8 @@ public:
         for (int step_number = 0; step_number < BaseT::max_steps && !stop; ++step_number)
         {
             // Adjust step size to reach output time
-            if (time_out - time_current < dt) {
+            // protect against roundoff
+            if (std::abs(time_out - time_current - dt) <= 1.e5 * std::numeric_limits<amrex::Real>::epsilon() * std::abs(dt)) {
```
WeiqunZhang: What if Real is float?

We could also do `amrex::almostEqual(time_current+dt, time_out)`. `almostEqual` has an optional argument that can be used to fine tune what is considered almost equal.

ajnonaka: That sounds good, I'll test it.

ajnonaka: @WeiqunZhang how is this?

WeiqunZhang:
Any reasons why `std::abs` is needed? Also the last argument should be an int (e.g., `10`, not double like `10.`.

ajnonaka:
@WeiqunZhang fixed

WeiqunZhang (approved):

PR #4151, opened 2024-09-16
Adding computation of complete elliptic integrals into amrex::Math so…

clarkse:
… that it can be executed on accelerator devices.

## Summary
Currently the parser recognizes comp_ellint_1 and comp_ellint_2, but only executes on CPU when compiled with gcc. An implementation of the calculations have been added to amrex::Math to compute these quantities using the method described here:
https://dlmf.nist.gov/19.8

It is a quadratically converging iterative method that is compatible with device execution.

Gauss's Arithmetic-Geometric Mean
 
## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_Math.H
+++ b/Src/Base/AMReX_Math.H
@@ -10,7 +10,6 @@
 #include <cstdlib>
 #include <type_traits>
 #include <utility>
-#include <iostream>
 
 #ifdef AMREX_USE_SYCL
 #  include <sycl/sycl.hpp>
--- a/Src/Base/AMReX_Math.H
+++ b/Src/Base/AMReX_Math.H
@@ -232,10 +232,7 @@ T comp_ellint_1 (T k)
 {
     // Computing K based on DLMF
     // https://dlmf.nist.gov/19.8
-    T tol = 1e-12;
-    if constexpr (std::is_same<T, float>::value) {
-        tol = 1e-6;
-    }
+    T tol = std::numeric_limits<T>::epsilon();
     
     T a0 = 1.0;
     T g0 = std::sqrt(1.0 - k*k);
@@ -261,10 +258,7 @@ T comp_ellint_2 (T k)
     // Computing E based on DLMF
     // https://dlmf.nist.gov/19.8
     T Kcomp = amrex::Math::comp_ellint_1<T>(k);
-    T tol = 1e-12;
-    if constexpr (std::is_same<T, float>::value) {
-        tol = 1e-6;
-    }
+    T tol = std::numeric_limits<T>::epsilon();
 
     // Step Zero
     T a0 = 1.0;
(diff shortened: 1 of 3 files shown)


Src/Base/AMReX_Math.H
```diff
@@ -225,6 +226,70 @@ std::uint64_t umulhi (std::uint64_t a, std::uint64_t b)
 }
 #endif
 
+template <typename T>
+AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+T comp_ellint_1 (T m)
+{
+    // Computing K based on DLMF
+    // https://dlmf.nist.gov/19.8
+    T tol = 1e-12;
```
WeiqunZhang: What if T is float?

Src/Base/AMReX_Math.H
```diff
@@ -225,6 +226,70 @@ std::uint64_t umulhi (std::uint64_t a, std::uint64_t b)
...
+    // Find Arithmetic Geometric mean
+    while(std::abs(a0 - g0) > tol) {
+        a = 0.5*(a0 + g0);
+        g = std::sqrt(a0 * g0);
+
+        a0 = a;
+        g0 = g;
+    }
+
+    return 0.5*pi<T>()/a;
```
WeiqunZhang, line 251: If m is 0, a is uninitialized.

WeiqunZhang: For `comp_ellint_1(0.5)`, I get 1.685750355 with std::comp_ellint_1, but 1.854074677 with amrex::Math::comp_ellint_1. The error seems to be too big.

Src/Base/AMReX_Math.H
clarkse: I will add a check on it to adjust the tolerance for a float to 1e-6.

Src/Base/AMReX_Math.H
clarkse: This is a definition issue. In most cases the integrals are defined with m as the argument. In this case with gcc it is k = sqrt(m). I will update it to be consistent and I will additionally add some test figures doing comparisons.

clarkse: @WeiqunZhang in general though do you think this is the right spot for this calculation?

WeiqunZhang: Yes, I think so.

WeiqunZhang: It seems that we could use `T tol = std::numeric_limits<T>::epsilon();`. If we want to be conservative, we could make it 10x bigger than epsilon. In my testing, it only increases the number of iteration slightly.

clarkse: I think the latest push should resolve these concerns. I have also added a comparison of the AGM algorithm to what is done in Python's ellipm and ellipk routines.

image

clarkse: The above figure is the difference between the two cases for the Z-components of the analytical solution to a current loop.

Src/Base/AMReX_Math.H
clarkse: This should be fixed now.

WeiqunZhang (approved):

PR #4163, opened 2024-09-23
Named SoA Support

### LINKED ISSUE #3614: Named Particle SoA Components
X-ref:
- https://github.com/AMReX-Codes/pyamrex/pull/220
- https://github.com/AMReX-Codes/pyamrex/pull/88

It would be nice if AMReX supported the functionality to name particle attributes (as strings). In WarpX, we added this as `NamedParticleContainer`, but as a mainstream implementation we could avoid deriving yet another class type (because if we do, we will again have additional compile-time to bind that custom type in pyAMReX), and instead could just extend the particle classes with these runtime details.

ax3l:
## Summary

Add support to optionally name SoA Real and Int components, both for compile-time and runtime components.

- [x] define and defaults
- [x] single tile `Get*Data(std::string)`
- [x] refactor out default name generation for reuse, finalize defaults in `AddReal/IntComp`
- [x] test

## Additional background

Close #3614 

## Checklist

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_ParticleContainer.H
+++ b/Src/Particle/AMReX_ParticleContainer.H
@@ -1274,8 +1274,7 @@ public:
 
     void AddRealComp (int communicate=1)
     {
-        std::string default_name = "runtime_real" + std::to_string(m_num_runtime_real);
-        AddRealComp(default_name, communicate);
+        AddRealComp(getDefaultCompNameReal<ParticleType>(NArrayReal+m_num_runtime_real), communicate);
     }
 
     void AddIntComp (std::string const & name, int communicate=1)
@@ -1303,8 +1302,7 @@ public:
 
     void AddIntComp (int communicate=1)
     {
-        std::string default_name = "runtime_int" + std::to_string(m_num_runtime_int);
-        AddIntComp(default_name, communicate);
+        AddIntComp(getDefaultCompNameInt<ParticleType>(NArrayInt+m_num_runtime_int), communicate);
     }
 
     int NumRuntimeRealComps () const { return m_num_runtime_real; }
--- a/Src/Particle/AMReX_ParticleContainer.H
+++ b/Src/Particle/AMReX_ParticleContainer.H
@@ -1249,9 +1249,7 @@ public:
 
     Long superParticleSize() const { return superparticle_size; }
 
-    template <typename T,
-              std::enable_if_t<std::is_same_v<T,bool>,int> = 0>
-    void AddRealComp (std::string const & name, T communicate=true)
+    void AddRealComp (std::string const & name, bool communicate=true)
     {
         m_soa_rdata_names.push_back(name);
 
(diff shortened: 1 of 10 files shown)


Src/Particle/AMReX_ParticleContainerI.H
```diff
@@ -60,10 +64,51 @@ ParticleContainer_impl<ParticleType, NArrayReal, NArrayInt, Allocator, CellAssig
...
+        }
+        for (int i=first_r_name; i<NArrayReal; ++i)
+        {
+            m_soa_rdata_names.push_back("real_comp" + std::to_string(i-first_r_name));
+        }
+        for (int i=0; i<NArrayInt; ++i)
+        {
+            m_soa_idata_names.push_back("int_comp" + std::to_string(i));
+        }
+
```
ax3l: We discussed that we want to make this two functions (one for real, one for int) that take the ParticleType and index to return a default name on the current index.

This way, we can reuse this in the:
- current I/O logic (existing naming scheme we took over here)
- here
- in `AddRealComp`/`AddIntComp` without `name` argument

atmyers:
@ax3l I think this is ready now - want to give it a review?

Src/Particle/AMReX_ParticleContainer.H
```diff
@@ -1270,10 +1272,15 @@ public:
         }
     }
 
-    template <typename T,
-              std::enable_if_t<std::is_same_v<T,bool>,int> = 0>
-    void AddIntComp (T communicate=true)
+    void AddRealComp (int communicate=1)
     {
+        AddRealComp(getDefaultCompNameReal<ParticleType>(NArrayReal+m_num_runtime_real), communicate);
```
ax3l, line 1277: zero indexed and used before `m_num_runtime_real++`, looks correct.

Src/Particle/AMReX_StructOfArrays.H
```diff
@@ -19,11 +23,18 @@ struct StructOfArrays {
     using RealVector = amrex::PODVector<ParticleReal, Allocator<ParticleReal> >;
     using IntVector = amrex::PODVector<int, Allocator<int> >;
 
-    void define (int a_num_runtime_real, int a_num_runtime_int)
+    void define (
+        int a_num_runtime_real,
+        int a_num_runtime_int,
+        std::vector<std::string>* soa_rdata_names=nullptr,
+        std::vector<std::string>* soa_idata_names=nullptr
```
ax3l, line 30: Yes, this is indeed not super save... Should we store a reference to a PC instead?

Or store the names in `StructOfArrays` and look them up when we need them from PC and ParticleTile?

Src/Particle/AMReX_StructOfArrays.H
```diff
@@ -226,13 +291,20 @@ struct StructOfArrays {
...
     std::array<RealVector, NReal> m_rdata;
     std::array< IntVector,  NInt> m_idata;
 
+    // runtime data
     std::vector<RealVector> m_runtime_rdata;
     std::vector<IntVector > m_runtime_idata;
 
+    // names of both compile-time and runtime Real and Int data
+    std::vector<std::string>* m_rdata_names = nullptr;
+    std::vector<std::string>* m_idata_names = nullptr;
```
ax3l, line 305: We could own the names here and let PC and ParticleTile reference them...

ax3l (commented):
Very nice!

One thought on maybe moving the ownership around for safer function calls. Currently, there is a small risk someone builds their own `StructOfArrays` outside of a PC and then passes non-owned names into it, which would easily create memory violations when the SoA outlives the passed names.

ax3l:
Discussed today: the case here is not very common, because we create SoAs from PCs and from other SoAs.

In tiling, we actually create an SoA per tile in the grid of tiles, which if we store the string in those would duplicate the names quite a bit (yet, not as much as the particle arrays also owned by those SoA's).

We conclude, it is unlikely to create issues right now in this design in the common usage :)

atmyers (approved):

PR #4165, opened 2024-09-23
Env Variable: `AMREX_DEFAULT_INIT`, etc.

ax3l:
## Summary

Introduce control of backtrace handling and exception throwing via the environment variable `AMREX_DEBUG`.

This is very handly when running CI/CD and IDE workflows to quickly debug a failing test without overwriting default init parameters.

## Checklist

The proposed changes:


Src/Base/AMReX.cpp
```diff
@@ -334,18 +335,30 @@ amrex::Initialize (int& argc, char**& argv, bool build_parm_parse,
...
+        if (str_amrex_debug == "1" || str_amrex_debug == "on" || str_amrex_debug == "true") {
+            amrex_debug = true;
+        }
+    }
+
     system::exename.clear();
 //    system::verbose = 0;
     system::regtest_reduction = false;
-    system::signal_handling = true;
+    system::signal_handling = !amrex_debug;
```
ax3l: Without this, I cannot track memory violations in gdb:
```
gdb -ex r -ex bt --args warpx ...
```

ax3l:
I implemented this downstream now via CLI options, controlled in tests via extra CMake test options.

ax3l:
Ah no, I did not solve this for Python modules/apps.

ax3l:
A possible alternative approach we could do to solve this is to introduce an
```bash
export AMREX_DEFAULT_INIT='amrex.the_arena_init_size=0 amrex.abort_on_unused_inputs=1 amrex.abc="a b c"'
```
env var, which can be used to inject options into `amrex_initialize()` -- possibly appended to overwrite defaults. (or make sure all defaults are `queryAdd` in codes)

Would solve workflows like: https://github.com/BLAST-WarpX/warpx/pull/5785 for Python tests (and also works for regular app/executable tests).

WeiqunZhang (approved):

PR #4172, opened 2024-09-29
Add issame functions for boxarray and distromap. Add "hdf5" to dep.py

dunhamsj:
## Summary

Adding two functions: one to check whether two boxarrays are the same or not; the other to check whether to distromaps are the same or not. Also, adding "hdf5" to dep.py to suppress warnings.

## Checklist

The proposed changes:


dunhamsj:
Sorry there are 146 commits, I didn't know how to squash them. There are only minor changes to the code though.

PR #4173, opened 2024-09-29
Add comparison operator for boxarray and distromap. Add hdf5 to dep.py

dunhamsj:
## Summary

Add `issame` function for boxarray and for distromap for the Fortran interface. Also add "hdf5" to dep.py to suppress warnings.

## Checklist

The proposed changes:

### DIFF
--- a/Src/F_Interfaces/Base/AMReX_boxarray_fi.cpp
+++ b/Src/F_Interfaces/Base/AMReX_boxarray_fi.cpp
@@ -87,7 +87,7 @@ extern "C" {
         return ba->intersects(bx);
     }
 
-    bool amrex_fi_boxarray_issame (const BoxArray& baa, const BoxArray& bab)
+    int amrex_fi_boxarray_issame (const BoxArray* baa, const BoxArray* bab)
     {
         return baa == bab;
     }
--- a/Src/F_Interfaces/Base/AMReX_boxarray_fi.cpp
+++ b/Src/F_Interfaces/Base/AMReX_boxarray_fi.cpp
@@ -86,4 +86,9 @@ extern "C" {
         Box bx(IntVect(lo), IntVect(hi), ba->ixType());
         return ba->intersects(bx);
     }
+
+    bool amrex_fi_boxarray_issame (const BoxArray& baa, const BoxArray& bab)
+    {
+        return baa == bab;
+    }
 }
--- a/Src/F_Interfaces/Base/AMReX_boxarray_fi.cpp
+++ b/Src/F_Interfaces/Base/AMReX_boxarray_fi.cpp
@@ -89,6 +89,6 @@ extern "C" {
 
     int amrex_fi_boxarray_issame (const BoxArray* baa, const BoxArray* bab)
     {
-        return baa == bab;
+        return *baa == *bab;
     }
 }
--- a/Src/F_Interfaces/Base/AMReX_boxarray_mod.F90
+++ b/Src/F_Interfaces/Base/AMReX_boxarray_mod.F90
@@ -9,7 +9,8 @@ module amrex_boxarray_module
 
   private
 
-  public :: amrex_boxarray_build, amrex_boxarray_destroy, amrex_print
+  public :: amrex_boxarray_build, amrex_boxarray_destroy, amrex_print, &
+            amrex_boxarray_issame
 
   type, public :: amrex_boxarray
      logical     :: owner = .false.
@@ -128,6 +129,12 @@ module amrex_boxarray_module
        type(c_ptr), value, intent(in) :: ba
        integer, intent(in) :: lo(*), hi(*)
      end function amrex_fi_boxarray_intersects_box
+
+     pure logical function amrex_fi_boxarray_issame (baa, bab) bind(c)
+       import
+       implicit none
+       type(c_ptr), value, intent(in) :: baa, bab
+     end function amrex_fi_boxarray_issame
   end interface
 
 contains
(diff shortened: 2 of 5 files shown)


Src/F_Interfaces/Base/AMReX_boxarray_fi.cpp
```diff
@@ -86,4 +86,9 @@ extern "C" {
         Box bx(IntVect(lo), IntVect(hi), ba->ixType());
         return ba->intersects(bx);
     }
+
+    bool amrex_fi_boxarray_issame (const BoxArray& baa, const BoxArray& bab)
```
WeiqunZhang: We should use `const BoxArray* ` instead of `const BoxArray&`, because `const &` is not part of C. We probably should also use `int` on the C/C++ side and `c_int` on the Fortran side, because Fortran `logical` may not be compatible.

WeiqunZhang:
Could we add `operator(==)`?

Src/F_Interfaces/Base/AMReX_boxarray_fi.cpp
dunhamsj: That makes sense, I'll make those changes.

dunhamsj:
Sorry, I don't know what you mean by adding `operator(==)`, but I'm happy to try it.

WeiqunZhang:
Something like
```
--- a/Src/F_Interfaces/Base/AMReX_boxarray_mod.F90
+++ b/Src/F_Interfaces/Base/AMReX_boxarray_mod.F90
@@ -9,8 +9,7 @@ module amrex_boxarray_module
 
   private
 
-  public :: amrex_boxarray_build, amrex_boxarray_destroy, amrex_print, &
-            amrex_boxarray_issame
+  public :: amrex_boxarray_build, amrex_boxarray_destroy, amrex_print, operator(==)
 
   type, public :: amrex_boxarray
      logical     :: owner = .false.
@@ -37,6 +36,10 @@ module amrex_boxarray_module
 #endif
   end type amrex_boxarray
 
+  interface operator(==)
+     module procedure amrex_boxarray_issame
+  end interface operator(==)
+
   interface amrex_boxarray_build
      module procedure amrex_boxarray_build_bx
      module procedure amrex_boxarray_build_bxs
@@ -266,8 +269,8 @@ contains
   end function amrex_boxarray_intersects_box
 
   pure function amrex_boxarray_issame(baa, bab) result(r)
-    class(amrex_boxarray), intent(in) :: baa
-    class(amrex_boxarray), intent(in) :: bab
+    type(amrex_boxarray), intent(in) :: baa
+    type(amrex_boxarray), intent(in) :: bab
     logical :: r
     r = amrex_fi_boxarray_issame(baa%p, bab%p)
   end function amrex_boxarray_issame
```

This will allow the users to write `baa == bab` or `baa .eq. bab` instead of `amrex_boxarray_issame(baa,bab)`. I also changed from `class` to `type`, because the types involved are not polymorphic.

dunhamsj:
I see, thanks. I made those changes, and now in my fortran code, `baa .eq. bab` returns 0 if the two boxarrays are equal. I just want to be sure, I should then do something like `if( ( baa .eq. bab ) .eq. 0 )` to check if they are equal?

WeiqunZhang:
I am not sure what you are talking about. The fortran function should still return `logical`, not 0 or 1. Something like `return c_function_that_returns_int(...) .ne. 0`. Why don't you push what you have so that I can see what changes you have made?

dunhamsj:
Ah, okay. I think I misunderstood what you were suggesting. I changed all the functions to return integers. I will push my changes though

WeiqunZhang:
My suggestion is the bind(c) functions use `c_int` and Fortran native code logical. It's logically true is a c_int is not 0.

WeiqunZhang:
The reason is I mentioned previously that Fortran logical may not be compatible with C/C++ bool.

With gfotran,
```
program main
  logical :: a = .true.
  print*, "sizeof(logical) = ", sizeof(a)
end program main
```
produces ` sizeof(a) = 4`.

But with g++, the sizeof `bool` is 1. They are not compatible. We could probably use `c_bool` in Fortran. But `_Bool` is a C type, not a C++ type. So the safest way is to use `int` (i.e., `c_int`) for C interface code and manually convert to `logical` in Fortran.

dunhamsj:
That makes sense, thanks. I'm trying it now using logical for the fortran code, integer for the bind(c) code, and int for the c code. I think it's working, I just want to be sure before I push it

dunhamsj:
Sorry, I couldn't get it to work right. Before the changes, the function would return true when it should have, but after the changes it returns false. I just pushed my changes

Src/F_Interfaces/Base/AMReX_boxarray_fi.cpp
```diff
@@ -86,4 +86,9 @@ extern "C" {
         Box bx(IntVect(lo), IntVect(hi), ba->ixType());
         return ba->intersects(bx);
     }
+
+    int amrex_fi_boxarray_issame (const BoxArray* baa, const BoxArray* bab)
+    {
+        return baa == bab;
```
WeiqunZhang:
```suggestion
        return *baa == *bab;
```

WeiqunZhang:
Okay, I will try to fix it and ask you to test.

dunhamsj:
That did it, thanks! I just pushed the changes

WeiqunZhang:
We should also use `.ne. 0` to convert from c_int to logical. Could you test?

dunhamsj:
Yes, that works

WeiqunZhang (approved):

PR #4175, opened 2024-09-30
Use The_Comms_Arena in particle communication.

atmyers:
The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_NeighborParticlesGPUImpl.H
+++ b/Src/Particle/AMReX_NeighborParticlesGPUImpl.H
@@ -256,8 +256,10 @@ updateNeighborsGPU (bool boundary_neighbors_only)
 
     clearNeighbors();
 
-    snd_buffer.setArena(The_Comms_Arena());
-    rcv_buffer.setArena(The_Comms_Arena());
+    if (ParallelDescriptor::UseGpuAwareMpi()) {
+        snd_buffer.setArena(The_Comms_Arena());
+        rcv_buffer.setArena(The_Comms_Arena());
+    }
 
     packBuffer(*this, neighbor_copy_op, neighbor_copy_plan, snd_buffer);
     if (ParallelDescriptor::UseGpuAwareMpi())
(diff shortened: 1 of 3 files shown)


WeiqunZhang:
Because `The_Comms_Arena` is an alias to `The_Pinned_Arena` when gpu aware mpi is not used, the code like below might be problematic.
```
    // assuming ParallelDescriptor::UseGpuAwareMpi() is false                                                                                                                

    amrex::PODVector<char, PolymorphicArenaAllocator<char> > rcv_buffer;
    rcv_buffer.setArena(The_Comms_Arena()); // pinned memory if not using gpu aware mpi                                                                                      

    if (ParallelDescriptor::UseGpuAwareMpi())
    {
        // ...                                                                                                                                                               
    }
    else
    {
        ...
        // rcv_buffer is not device memory                                                                                                                                   
        Gpu::htod_memcpy_async(rcv_buffer.dataPtr(), pinned_rcv_buffer.dataPtr(), pinned_rcv_buffer.size());
    }
```

atmyers:
Maybe we can use `The_Comms_Arena()` with gpu-aware MPI, and `The_Arena()` without it.

ax3l (approved):

WeiqunZhang:
> Maybe we can use `The_Comms_Arena()` with gpu-aware MPI, and `The_Arena()` without it.

Yes, that would require the least amount of changes.

Src/Particle/AMReX_NeighborParticlesGPUImpl.H
```diff
@@ -255,6 +255,10 @@ updateNeighborsGPU (bool boundary_neighbors_only)
     }
 
     clearNeighbors();
+
+    snd_buffer.setArena(The_Comms_Arena());
```
WeiqunZhang: Should these two lines inside `if (ParallelDescriptor::UseGpuAwareMpi())` like in AMReX_ParticleContainerI.H?

atmyers: Yes.

WeiqunZhang (approved):

ax3l:
This is causing a memory overhead and performance overhead in ImpactX: https://github.com/ECP-WarpX/impactx/issues/769

Let's discuss next week how we proceed (e.g., opt-in to this PR's features, changing defaults, etc.).

PR #4176, opened 2024-10-01
SmallMatrix: Matrix class with compile time size

WeiqunZhang:
Add amrex::SmallMatrix class with compile time size.

Useful for, e.g.:
- https://github.com/ECP-WarpX/impactx/pull/714
- https://github.com/ECP-WarpX/impactx/pull/702

### DIFF
--- a/Tests/SmallMatrix/main.cpp
+++ b/Tests/SmallMatrix/main.cpp
@@ -1,0 +1,144 @@
+#include <AMReX.H>
+#include <AMReX_Math.H>
+#include <AMReX_Print.H>
+#include <AMReX_REAL.H>
+#include <AMReX_SmallMatrix.H>
+
+using namespace amrex;
+
+int main (int argc, char* argv[])
+{
+    static_assert(Order::C == Order::RowMajor &&
+                  Order::F == Order::ColumnMajor);
+
+    amrex::Initialize(argc, argv);
+    {
+        SmallMatrix<Real,3,4> m34{};
+        for (int j = 0; j < 4; ++j) {
+            for (int i = 0; i < 3; ++i) {
+                AMREX_ALWAYS_ASSERT(m34(i,j) == 0.0_rt);
+            }
+        }
+    }
+    {
+        SmallVector<Real,3> cv{};
+        SmallRowVector<Real,3> rv{};
+        SmallVector<int,3> cv2{1,2,3};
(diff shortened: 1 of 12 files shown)


zingale:
it would be good to have a `.setVal()` method to set all elements to a value.

WeiqunZhang:
Notes on why SmallMatrix matrix{} is zero initialized.

SmallMatrix is not an aggregate, because it has a user declared default
constructor. The rule is that, for `SmallMatrix matrix{}` with an empty
brace-enclosed initializer list, value-initialization is performed. The
effects of value-initialization of SmallMatrix (which has a user-declared
but not user-provided default constructor) are that the matrix object is
first zero-initialized and then the object's default constructor is
applied. Since the default constructor does nothing, the final result is
the object is zero-initialized.

Why is SmallMatrix's default constructor user-declared not user-provided?
It's because we first declare it with `SmallMatrix () = default`.

Reference:
  https://en.cppreference.com/w/cpp/language/list_initialization
  https://en.cppreference.com/w/cpp/language/value_initialization
  https://en.cppreference.com/w/cpp/language/zero_initialization

ax3l:
Is `SmallMatrix<Real, N, M> m;` also zero-defined then or is it allocated but values are undefined?

ax3l:
FYI @cemitch99

Tests/SmallMatrix/main.cpp
```diff
@@ -0,0 +1,144 @@
...
+                AMREX_ALWAYS_ASSERT(m34(i,j) == v++);
+            }
+        }
+    }
+    {
+        auto v3 = SmallVector<double,3>::Zero();
+        v3[0] = 1.;
+        v3(1) = 2.;
+        v3[2] = 3.;
+        auto m33 = SmallMatrix<double,3,3>::Identity();
```
ax3l, line 66: Nice, very useful!

WeiqunZhang:
It's uninitialized by default.
```
/**
 * \brief Default constructor
 *
 * The data are uninitialized by default. If you want to initialize
 * to zero, you can do `SmallMatrix<T,NRows,NCols> M{};`.
 */
constexpr SmallMatrix () = default;
```

ax3l (approved):
That is super useful, I think that is all we currently need :rocket: :sparkles:

PR #4181, opened 2024-10-03
Add more build options for linear solvers

WeiqunZhang:
Add two new CMake build options, AMReX_LINEAR_SOLVERS_INCFLO and AMReX_LINEAR_SOLVERS_EM. Both options depend on AMReX_LINEAR_SOLVERS, and are enabled by default when AMReX_LINEAR_SOLVERS is enabled. By default, the cell-centered MLABecLaplacian and MLPoisson solvers are always included. However, these new build options provide finer control over which solvers to build.

* AMReX_LINEAR_SOLVERS_INCFLO can be used to control the inclusion of solvers commonly used by incompressible flow codes, such as nodal projection and Navier-Stokes tensor solvers.

* AMReX_LINEAR_SOLVERS_EM can be used to control the inclusion of solvers commonly used by electromagnetic applications, such as the curl curl and tensor Laplacian solvers used by WarpX.

These options are also available in GNU Make as USE_LINEAR_SOLVERS_INCFLO and USE_LINEAR_SOLVER_EM.


ax3l (approved):
Thank you for this! :rocket:

PR #4186, opened 2024-10-08
Fix queryktharr() return value/behaviour.

philip-blakely:
## Summary

Up to AMReX 24.07 (at least), ParmParse::queryktharr() used to return 0 if the given k (2nd parameter) was out of range.
As of AMReX 24.10, this is not the case, and ppindex() attempts to access found->second.m_vals out-of-bounds, resulting in seg-fault/undefined behaviour.

This PR reverts to the original behaviour, which is now consistent with the documentation for querykth() in AMReX_ParmParse.H. It also makes minor corrections to the associated comments to align with this behaviour.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_ParmParse.cpp
+++ b/Src/Base/AMReX_ParmParse.cpp
@@ -365,7 +365,7 @@ ppindex (const ParmParse::Table& table, int n, const std::string& name)
     if (n == ParmParse::LAST) {
         return &(found->second.m_vals.back());
     } else {
-      if(found->second.m_vals.size() < (std::size_t)(n+1)) {
+      if(found->second.m_vals.size() < (std::size_t)n + 1) {
         return nullptr;
       }
       return &(found->second.m_vals[n]);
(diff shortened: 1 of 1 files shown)


Src/Base/AMReX_ParmParse.cpp
```diff
@@ -365,7 +365,10 @@ ppindex (const ParmParse::Table& table, int n, const std::string& name)
     if (n == ParmParse::LAST) {
         return &(found->second.m_vals.back());
     } else {
-        return &(found->second.m_vals[n]);
+      if(found->second.m_vals.size() < (std::size_t)n + 1) {
+        return nullptr;
+      }
+      return &(found->second.m_vals[n]);
```
WeiqunZhang:
```suggestion
        if(found->second.m_vals.size() < (std::size_t)n + 1) {
            return nullptr;
        }
        return &(found->second.m_vals[n]);
```

WeiqunZhang (approved):

PR #4187, opened 2024-10-09
SoA: Public Getter for Names

ax3l:
## Summary

Expose the names to public member functions.

## Additional background

Needed for Python bindings. Follow-up to #4163 

## Checklist

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_StructOfArrays.H
+++ b/Src/Particle/AMReX_StructOfArrays.H
@@ -60,7 +60,7 @@ struct StructOfArrays {
             return *m_rdata_names;
         }
         else {
-            throw std::runtime_error("SoA: no Real names defined!");
+            return std::vector<std::string>();
         }
     }
 
@@ -71,7 +71,7 @@ struct StructOfArrays {
             return *m_idata_names;
         }
         else {
-            throw std::runtime_error("SoA: no int names defined!");
+            return std::vector<std::string>();
         }
     }
 
--- a/Src/Particle/AMReX_StructOfArrays.H
+++ b/Src/Particle/AMReX_StructOfArrays.H
@@ -8,6 +8,7 @@
 
 #include <algorithm>
 #include <array>
+#include <stdexcept>
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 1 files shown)


Src/Particle/AMReX_StructOfArrays.H
```diff
@@ -52,6 +53,28 @@ struct StructOfArrays {
...
     [[nodiscard]] const std::array< IntVector,  NInt>& GetIntData  () const { return m_idata; }
 
+    /** Get the names for the real SoA components **/
+    [[nodiscard]] std::vector<std::string> GetRealNames () const
+    {
+        if (m_rdata_names) {
+            return *m_rdata_names;
+        }
+        else {
+            throw std::runtime_error("SoA: no Real names defined!");
```
ax3l: We could also return:
```suggestion
            return std::vector<std::string>();
```

Src/Particle/AMReX_StructOfArrays.H
```diff
@@ -52,6 +53,28 @@ struct StructOfArrays {
...
+    }
+
+    /** Get the names for the int SoA components **/
+    [[nodiscard]] std::vector<std::string> GetIntNames () const
+    {
+        if (m_idata_names) {
+            return *m_idata_names;
+        }
+        else {
+            throw std::runtime_error("SoA: no int names defined!");
```
ax3l: We could also return:
```suggestion
            return std::vector<std::string>();
```

ax3l:
Let me know what you prefer.

In fact in Python bindings I prefer the alternative of returning an empty list over having to handle an exception.

Src/Particle/AMReX_StructOfArrays.H
WeiqunZhang: I think return is better.

Src/Particle/AMReX_StructOfArrays.H
```diff
@@ -8,6 +8,7 @@
 
 #include <algorithm>
 #include <array>
+#include <stdexcept>
```
ax3l:
```suggestion
```

atmyers (approved):

WeiqunZhang (approved):

PR #4188, opened 2024-10-10
SmallMatrix: Support 1-based indexing

WeiqunZhang:
To avoid confusion, operations involving both 0 and 1-base indexing are not allowed.


WeiqunZhang:
Note that if the C++ standard supports matrix in the future, it's very unlikely 1-based indexing will be supported.

ax3l (approved):
Thank you so much!

PR #4189, opened 2024-10-11
SmallMatrix: Structured binding support

WeiqunZhang:


WeiqunZhang:
Note that because structured binding to const reference extends the lifetime of the temporary object, the following code is correct.

    auto const& [a, b] = M * v

where M is a matrix and v is a vector.

ax3l (approved):
:tada: Cool, thanks!!

PR #4193, opened 2024-10-17
amrex::FFT

WeiqunZhang:
Add parallel FFT capability to AMReX. It relies on FFTW3, cuFFT, rocFFT and oneMKL, for CPU, CUDA, HIP and SYCL builds, respectively.


WeiqunZhang:
This is still work in progress. But the cmake part is ready for review. @sayerhs, when you have time, could you review the cmake part?

asalmgren (approved):

WeiqunZhang:
The CMake stuff has been tested in GitHub CI, my workstation, perlmutter, frontier, and aurora. Both CPU and GPU builds work on these systems. I suppose it's good enough for now. We can fix any issues later.

PR #4195, opened 2024-10-21
Bounded sampling

AMLattanzi:
## Summary
The `get_line_data` and `get_slice_data` routines do not currently allow for a natural way to bound the returned MultiFab. In the former case, the line length is dictated by the bottom and top of the boxarray at that level. In the latter case, the plane is dictated by the problem domain at that level.

This PR adds an optional `Box` argument to the `get_line_data` call and an optional `RealBox` argument to the `get_slice_data` call. These optional arguments bound the returned MuliFab. This feature has application in the ERF code in [PR 1897](https://github.com/erf-model/ERF/pull/1897) where it is of great value to sample velocities in planes/lines just upstream of a wind turbine. Since the wind turbine is small in relation to the problem domain, the bounding feature ensures the returned data is actually what the user wants.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_MultiFabUtil.H
+++ b/Src/Base/AMReX_MultiFabUtil.H
@@ -1025,7 +1025,7 @@ MF get_line_data (MF const& mf, int dir, IntVect const& cell, Box const& bnd_bx)
     } else {
         for (int i = 0; i < nboxes; ++i) {
             Box const& b   = ba[i];
-            Box const& b1d = bnd_box & b;
+            Box const& b1d = bnd_bx & b;
             if (b1d.ok()) {
                 bl.push_back(b1d);
                 procmap.push_back(dm[i]);
--- a/Src/Base/AMReX_MultiFabUtil.H
+++ b/Src/Base/AMReX_MultiFabUtil.H
@@ -170,7 +170,7 @@ namespace amrex
                                              const MultiFab& cc,
                                              const Geometry& geom, int start_comp, int ncomp,
                                              bool interpolate=false,
-                                             RealBox bnd_rbx = RealBox());
+                                             RealBox const& bnd_rbx = RealBox());
 
     /**
      * \brief Get data in a cell of MultiFab/FabArray
@@ -189,7 +189,7 @@ namespace amrex
      * specified by a direction and a cell.
      */
     template <typename MF, std::enable_if_t<IsFabArray<MF>::value,int> FOO = 0>
-    MF get_line_data (MF const& mf, int dir, IntVect const& cell, Box bnd_bx = Box());
+    MF get_line_data (MF const& mf, int dir, IntVect const& cell, Box const& bnd_bx = Box());
 
     //! Return an iMultiFab that has the same BoxArray and DistributionMapping
     //! as the coarse MultiFab cmf. Cells covered by the coarsened fine grids
@@ -997,7 +997,7 @@ Vector<typename MF::value_type> get_cell_data (MF const& mf, IntVect const& cell
 }
 
 template <typename MF, std::enable_if_t<IsFabArray<MF>::value,int> FOO>
-MF get_line_data (MF const& mf, int dir, IntVect const& cell, Box bnd_bx)
+MF get_line_data (MF const& mf, int dir, IntVect const& cell, Box const& bnd_bx)
 {
     bool do_bnd = (!bnd_bx.isEmpty());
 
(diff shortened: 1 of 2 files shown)


Src/Base/AMReX_MultiFabUtil.H
```diff
@@ -996,26 +997,45 @@ Vector<typename MF::value_type> get_cell_data (MF const& mf, IntVect const& cell
...
+                procmap.push_back(dm[i]);
+                index_map.push_back(i);
+            }
+        }
+    } else {
+        for (int i = 0; i < nboxes; ++i) {
+            Box const& b = ba[i];
+            IntVect lo = bnd_bx.smallEnd();
+            IntVect hi = bnd_bx.bigEnd();
+            if (!b.contains(lo)) { lo[dir] = b.smallEnd(dir); }
```
WeiqunZhang: The logic here is a little complicated. So it's hard for me to follow it through. But is `Box const& b1d = b & bnd_bx` what you are trying to do here? If that is not the case, could you explain a little?

Src/Base/AMReX_MultiFabUtil.H
```diff
@@ -996,26 +997,45 @@ Vector<typename MF::value_type> get_cell_data (MF const& mf, IntVect const& cell
 }
 
 template <typename MF, std::enable_if_t<IsFabArray<MF>::value,int> FOO>
-MF get_line_data (MF const& mf, int dir, IntVect const& cell)
+MF get_line_data (MF const& mf, int dir, IntVect const& cell, Box bnd_bx)
```
WeiqunZhang: We could use `Box const&` instead of `Box`?

Src/Base/AMReX_MultiFabUtil.cpp
```diff
@@ -24,12 +24,11 @@ namespace {
     }
 
 
-    std::unique_ptr<MultiFab> allocateSlice(int dir, const MultiFab& cell_centered_data,
-                                            int ncomp, const Geometry& geom, Real dir_coord,
-                                            Vector<int>& slice_to_full_ba_map) {
+    std::unique_ptr<MultiFab> allocateSlice (int dir, const MultiFab& cell_centered_data,
+                                             int ncomp, const Geometry& geom, Real dir_coord,
+                                             Vector<int>& slice_to_full_ba_map, RealBox real_slice) {
```
WeiqunZhang, line 29: `RealBox const&` instead of `RealBox` can be used.

Src/Base/AMReX_MultiFabUtil.H
AMLattanzi: `lo` is initially set to the smallEnd of the bounding box. This point should be the lowest point in the output line. So we first check if this point lies within the given box we are looping over. If it does not, we follow the logic like the unbounded case and we set the lo to the `smallEnd` of the given box we are looping over. This means that when the `lo` point is contained in the current box, we ensure that the lowest point in the line is in fact the `smallEnd` of our bounding box. If the box does not contain the bounding point, we still need to build the line upward through the boxarray until we get to our upper bounding point.

Src/Base/AMReX_MultiFabUtil.H
```diff
@@ -996,26 +997,45 @@ Vector<typename MF::value_type> get_cell_data (MF const& mf, IntVect const& cell
...
+            IntVect hi = bnd_bx.bigEnd();
+            if (!b.contains(lo)) { lo[dir] = b.smallEnd(dir); }
+            if (b.contains(lo)) {
+                if (!b.contains(hi)) { hi[dir] = b.bigEnd(dir); }
+                if (!bnd_bx.contains(hi)) { continue; }
+                Box b1d(lo,hi,b.ixType());
+                bl.push_back(b1d);
+                procmap.push_back(dm[i]);
+                index_map.push_back(i);
+            }
```
WeiqunZhang: Could you explain what you try to achieve, not what your code does? If I try harder, I might be able to understand the code. It seems that I need to check 3x3=9 cases depending the comparison results of various lo and hi to understand it does. But I suspect it's equivalent to the following 
```suggestion
            Box const& b1d = bnd_box & b;
            if (b1d.ok()) {
                bl.push_back(b1d);
                procmap.push_back(dm[i]);
                index_map.push_back(i);
            }
```

AMLattanzi: I would like a multifab returned that contains data on the bounding box only. What you have commented above is a cleaner and more efficient way to achieve that. I will add it to the PR.

WeiqunZhang (approved):

PR #4197, opened 2024-10-23
Skip some CIs if only Docs/ has changed

WeiqunZhang:


WeiqunZhang:
@EZoni This might work for WarpX.

EZoni:
Thanks, @WeiqunZhang.

I guess we can try this on WarpX too, although, if really it isn't feasible to wait until the `paths-ignore` filter is fixed on the GitHub end, I was still hoping to come up with an "in-house" workaround, which we can control 100%, rather than relying on yet another external solution.

Reference for everyone, WarpX PR where we started this discussion: https://github.com/ECP-WarpX/WarpX/pull/5387.

WeiqunZhang:
> I was still hoping to come up with an "in-house" workaround, which we can control 100%, rather than relying on yet another external solution.

I don't think it's hard to do. But that's exactly why it's fine to use an external solution, because we can replace it.

asalmgren (approved):

PR #4200, opened 2024-10-24
Add option to do stable redistribute with GPUs

atmyers:
Currently the order of particles in memory will not be consistent if you run the same simulation twice on the GPU. This usually doesn't matter, but for cases where it does, this adds a slower, stable option. 

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_ParticleCommunication.H
+++ b/Src/Particle/AMReX_ParticleCommunication.H
@@ -154,7 +154,9 @@ struct ParticleCopyPlan
         m_box_counts_d.resize(num_buckets+1, 0);
         m_box_offsets.resize(num_buckets+1);
         auto* p_dst_box_counts = m_box_counts_d.dataPtr();
-        auto getBucket = pc.BufferMap().getBucketFunctor();
+        auto getBucket = pc.stableRedistribute() ? pc.BufferMap().getHostBucketFunctor() : pc.BufferMap().getBucketFunctor();
+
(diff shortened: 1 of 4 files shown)


Src/Particle/AMReX_ParticleContainerBase.H
```diff
@@ -231,6 +231,10 @@ public:
 
     void SetVerbose (int verbose) { m_verbose = verbose; }
 
+    int stableRedistribute () const {return m_stable_redistribute; }
```
WeiqunZhang: One day clang-tidy will want a `[[nodiscard]]` here.

Src/Particle/AMReX_ParticleCommunication.H
```diff
@@ -154,7 +154,9 @@ struct ParticleCopyPlan
         m_box_counts_d.resize(num_buckets+1, 0);
         m_box_offsets.resize(num_buckets+1);
         auto* p_dst_box_counts = m_box_counts_d.dataPtr();
-        auto getBucket = pc.BufferMap().getBucketFunctor();
+        auto getBucket = pc.stableRedistribute() ? pc.BufferMap().getHostBucketFunctor() : pc.BufferMap().getBucketFunctor();
+
+        Gpu::HostVector<unsigned int> h_box_counts(m_box_counts_d.size(), 0);
```
WeiqunZhang: Should we make the initial size of this vector zero if it's not stable redistribution?

WeiqunZhang (approved):

PR #4238, opened 2024-11-20
EBData: New class containing Array4's for all EB data

WeiqunZhang:
This can make the code cleaner because we can get access to all available EB data in one object. This also reduces many GPU kernels' register pressure because the GPU device lambdas are much smaller without the need to capture multiple Array4's. For example, the 3D EBABecLap's gsrb kernel's number of registers reduces from 198 to 130.

Add a new function that returns a random point on EB for WarpX.


WeiqunZhang:
Needed by https://github.com/ECP-WarpX/WarpX/pull/5493

atmyers (approved):

PR #4257, opened 2024-12-07
Use async sends in particle communication

atmyers:
This changes a few instances of `ParallelDescriptor::Send` to `ParallelDescriptor::Asend`. This improves the overall runtime of the redistribute benchmark by a few percent for multi-node runs:

```
Gpus      Send        Asend
4	  4.125       4.123
8	  4.29 	      4.198
16	  4.472       4.297
32	  4.62 	      4.456
```

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_ParticleCommunication.H
+++ b/Src/Particle/AMReX_ParticleCommunication.H
@@ -571,8 +571,8 @@ void communicateParticlesStart (const PC& pc, ParticleCopyPlan& plan, const SndB
         AMREX_ASSERT(plan.m_snd_counts[i] % ParallelDescriptor::sizeof_selected_comm_data_type(plan.m_snd_num_particles[i]*psize) == 0);
         AMREX_ASSERT(Who >= 0 && Who < NProcs);
 
-        ParallelDescriptor::Send((char const*)(snd_buffer.dataPtr()+snd_offset), Cnt, Who, SeqNum,
-                                 ParallelContext::CommunicatorSub());
+        ParallelDescriptor::Asend((char const*)(snd_buffer.dataPtr()+snd_offset), Cnt, Who, SeqNum,
+                                  ParallelContext::CommunicatorSub());
     }
 
     amrex::ignore_unused(pc);
--- a/Src/Particle/AMReX_ParticleCommunication.H
+++ b/Src/Particle/AMReX_ParticleCommunication.H
@@ -543,10 +543,7 @@ void communicateParticlesStart (const PC& pc, ParticleCopyPlan& plan, const SndB
     plan.m_particle_rreqs.resize(plan.m_nrcvs);
 
     plan.m_particle_sstats.resize(0);
-    plan.m_particle_sstats.resize(plan.m_NumSnds);
-
     plan.m_particle_sreqs.resize(0);
-    plan.m_particle_sreqs.resize(plan.m_NumSnds);
 
     const int SeqNum = ParallelDescriptor::SeqNum();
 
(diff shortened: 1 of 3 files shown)


Src/Particle/AMReX_ParticleCommunication.H
```diff
@@ -571,8 +571,8 @@ void communicateParticlesStart (const PC& pc, ParticleCopyPlan& plan, const SndB
         AMREX_ASSERT(plan.m_snd_counts[i] % ParallelDescriptor::sizeof_selected_comm_data_type(plan.m_snd_num_particles[i]*psize) == 0);
         AMREX_ASSERT(Who >= 0 && Who < NProcs);
 
-        ParallelDescriptor::Send((char const*)(snd_buffer.dataPtr()+snd_offset), Cnt, Who, SeqNum,
-                                 ParallelContext::CommunicatorSub());
+        ParallelDescriptor::Asend((char const*)(snd_buffer.dataPtr()+snd_offset), Cnt, Who, SeqNum,
```
WeiqunZhang: This will lead to a memory leak, because MPI_Request is not "freed" with MPI_Wait. We had this kind of leaks before, and it caused issues for long runs. So we know this does leak a small amount of memory in practice, not just in theory. See also https://stackoverflow.com/questions/43004420/is-a-mpi-wait-necessary-after-a-mpi-isend.

Since we use C++17 now. I suggest we make ParallelDescriptor::Asend `[[nodiscard]]`.

Src/Particle/AMReX_ParticleCommunication.cpp
```diff
@@ -169,8 +170,9 @@ void ParticleCopyPlan::buildMPIStart (const ParticleBufferMap& map, Long psize)
         AMREX_ASSERT(Who >= 0 && Who < NProcs);
         AMREX_ASSERT(Cnt < std::numeric_limits<int>::max());
 
-        ParallelDescriptor::Send((char*) snd_data[i].data(), Cnt, Who, SeqNum,
-                                 ParallelContext::CommunicatorSub());
+        auto msg = ParallelDescriptor::Asend((char*) snd_data[i].data(), Cnt, Who, SeqNum,
```
WeiqunZhang: We only need to save `ParallelDescriptor::Asend(...).req()`, and later we can call Waitall on the vector.

WeiqunZhang (changes requested):

Src/Particle/AMReX_ParticleCommunication.H
```diff
@@ -571,9 +581,10 @@ void communicateParticlesStart (const PC& pc, ParticleCopyPlan& plan, const SndB
         AMREX_ASSERT(plan.m_snd_counts[i] % ParallelDescriptor::sizeof_selected_comm_data_type(plan.m_snd_num_particles[i]*psize) == 0);
         AMREX_ASSERT(Who >= 0 && Who < NProcs);
 
-        ParallelDescriptor::Asend((char const*)(snd_buffer.dataPtr()+snd_offset), Cnt, Who, SeqNum,
-                                  ParallelContext::CommunicatorSub());
+        plan.m_particle_sreqs[isnd++] = ParallelDescriptor::Asend((char const*)(snd_buffer.dataPtr()+snd_offset), Cnt, Who, SeqNum,
+                                                                  ParallelContext::CommunicatorSub()).req();
     }
+    AMREX_ASSERT(isnd == plan.m_NumSnds - 1);
```
WeiqunZhang: Is that right?

WeiqunZhang (approved):

PR #4258, opened 2024-12-07
[WIP] MLABecLaplacian spherical 2d support.

zhichen3:
## Summary
add support spherical 2d support for MLABecLaplacian.

## Additional background
I'm trying to get add spherical 2d geometry support for MLABecLaplacian for MLMG since Castro uses the `apply()` function in MLMG to compute thermal diffusion source term. I think I mainly need to change `applyMetricTerms()` and `unapplyMetricTerms()`.

I think the logic of the current code is to use `applyMetricTerms()` within `applyMetricTermsCoeffs` to modify the coefficients by the extra geometric term from divergence, i.e. $r^2$ for spherical in r-dir, and r for cylindrical in r-dir. Then `unapplyMetricTerms` is applied to the entire $L(\phi)$ after everything is calculated, which corresponds to the 1/r^2 and 1/r outside of the derivative in divergence.

My current logic is to modify the coefficient by $r^2 \sin(\theta)$ when dir=0 and $\sin(\theta)$ when dir=1, and when we unapply we divide by $\frac{1}{r^2\sin(\theta)}$ altogether, so that it is consistent with the divergence calculation.

I think I should also make sure `compResidual()`, `getFluxes()`, and `getGradSolution()` to work although it doesn't affect the `apply()` function.

`getGradSolution()` probably works since I modified `compGrad()`.

## Checklist

The proposed changes:


atmyers:
Hi @zhichen3, are you still actively working on this? I see that it was mentioned in a recent PR to Castro. Is this still something that you'd like to see merged?

zhichen3:
> Hi @zhichen3, are you still actively working on this? I see that it was mentioned in a recent PR to Castro. Is this still something that you'd like to see merged?

I'll close this one for now. We decided to evaluate the diffusion term in castro for convenience. I'll reopen the pr if we decided to switch back. Thanks.

PR #4264, opened 2024-12-13
Add version of PC::Checkpoint that allows toggling off certain components.

atmyers:
This is useful in the case that some components can be re-initialized rather than read from the checkpoint file. 

## Checklist

The proposed changes:


dpgrote (approved):
Thanks!

WeiqunZhang (approved):

WeiqunZhang:
The WarpX CI failure was fixed in #4267 by removing a line.

PR #4268, opened 2024-12-14
Make FFT safe for slabs

WeiqunZhang:
Support FFT on domains that have one cell in some dimensions.

It also supports Poisson solves on slab domains. However, for FFT::PoissonHybrid that treats the z-direction in a special way, the z-direction must have more than one cell.


WeiqunZhang:
The CI failure was fixed in #4267

atmyers (approved):

PR #4272, opened 2024-12-18
Fix CMake Compiler Flag Summary: ID Regex Matching

### LINKED ISSUE #2978: CMake build system
Hi,

I'm a bit confused so I'll try my best to not make a mess of this...

1) When compiling with the Intel classic compilers, CMake uses -O3 whereas the Make uses -O2, both for non-debug builds. Can you please clarify why this is the case?

2) When compiling with the Intel classic compilers, for a debug build, CMake uses the flags in `AMReXFlagsTargets.cmake` for both the classic and llvm Intel compilers. Is this on purpose? I'm a bit of a noob in CMake so I don't really know how these flags are picked up in the first place.

3) When compiling an application that uses AMReX, I can't seem to be able to make `AMReX::Flags_CXX` work in my `CMakeLists.txt`:
    ```
    # CMake minimum version
    cmake_minimum_required (VERSION 3.15)
    
    # Project name
    project(foobar)
    
    # All cpp source files
    file(GLOB_RECURSE src_files *.cpp)
    
    # Import AMReX settings and targets
    find_package(AMReX REQUIRED)
    
    # Generate obj files
    add_executable(${PROJECT_NAME} ${src_files})
    target_link_libraries(${PROJECT_NAME} AMReX::amrex AMReX::Flags_CXX)
    ```
    No compilation flags are added. However, using `-DCMAKE_BUILD_TYPE=<build_type>` does add the compilation flags used to compile AMReX no matter if `AMReX::Flags_CXX` is present or not... Curiously, if `<build_type>` is set to `Debug` and I use the Intel classic compilers, then only the flags for the classic compiler in `AMReXFlagsTargets.cmake` are picked up (unlike what happens above when compiling AMReX itself). Is this the intended behaviour? It'd be nice if there was an option to compile an application with the same compilation flags that were used to compile AMReX, is this possible with CMake?

Cheers,
-Nuno

nmnobre:
Fixes #2978.

## Summary
To understand what I'm trying to fix here, first note:

https://github.com/AMReX-Codes/amrex/blob/b3f67385e62f387b548389222840486c0fffca57/Tools/CMake/AMReXFlagsTargets.cmake#L71-L95

Then, with Intel's classic compilers, configure AMReX with `-DCMAKE_BUILD_TYPE=Debug` and note the C++ flags listed under "AMReX configuration summary", before and after the changes suggested here. The gist is `Intel` as a compiler id was matching `IntelLLVM` (and `Clang` was also matching `AppleClang` though the flags in that case are the same, so it's harder to tell).
EDIT: this actually only affects what's printed in the summary, the flags that are _actually_ used are already correct!

The proposed changes:

### DIFF
--- a/Tools/CMake/AMReXGenexHelpers.cmake
+++ b/Tools/CMake/AMReXGenexHelpers.cmake
@@ -279,7 +279,7 @@ function ( eval_genex _list _lang _comp )
       string(REGEX REPLACE "\\$<PLATFORM_ID:[A-Za-z]*>" "0"  _in "${_in}")
 
       # Genex in the form $<*_COMPILER_ID:compiler_ids>
-      string(REGEX REPLACE "\\$<${_lang}_COMPILER_ID:[^>]*${_comp}[^>]*>" "1"  _in "${_in}")
+      string(REGEX REPLACE "\\$<${_lang}_COMPILER_ID[^>]*[:,]${_comp}[>,]" "1"  _in "${_in}")
       string(REGEX REPLACE "\\$<[A-Za-z]*_COMPILER_ID:[A-Za-z]*>" "0"  _in "${_in}")
 
       # Genex in the form $<*_COMPILER_VERSION:version>
@@ -290,7 +290,7 @@ function ( eval_genex _list _lang _comp )
       string(REGEX REPLACE "\\$<[A-Za-z]*_COMPILER_VERSION:[^\$>]*>" "0"  _in "${_in}")
 
       # Genex in the form $<COMPILE_LANG_AND_ID:language,compiler_ids>
-      string(REGEX REPLACE "\\$<COMPILE_LANG_AND_ID:${_lang},[^\$>]*${_comp}[^\$>]*>" "1"  _in "${_in}")
+      string(REGEX REPLACE "\\$<COMPILE_LANG_AND_ID:${_lang}[^\$>]*,${_comp}[>,]" "1"  _in "${_in}")
       string(REGEX REPLACE "\\$<COMPILE_LANG_AND_ID:[A-Za-z,]*>" "0"  _in "${_in}")
 
       # Genex in the form $<COMPILE_LANGUAGE:languages>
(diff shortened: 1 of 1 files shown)


Tools/CMake/AMReXGenexHelpers.cmake
```diff
@@ -279,7 +279,7 @@ function ( eval_genex _list _lang _comp )
       string(REGEX REPLACE "\\$<PLATFORM_ID:[A-Za-z]*>" "0"  _in "${_in}")
 
       # Genex in the form $<*_COMPILER_ID:compiler_ids>
-      string(REGEX REPLACE "\\$<${_lang}_COMPILER_ID:[^>]*${_comp}[^>]*>" "1"  _in "${_in}")
+      string(REGEX REPLACE "\\$<${_lang}_COMPILER_ID[^>]*[:,]${_comp}[>,]" "1"  _in "${_in}")
```
ax3l, line 282: Can you help me understand what the change in regex does for each part?
- Does `[:,]` now match a literal `:` or `,`?
- Does `[>,]` now match a literal `>` or `,`?

nmnobre: > * Does `[:,]` now match a literal `:` or `,`?
> * Does `[>,]` now match a literal `>` or `,`?

Yes to both, the idea is that we will now exactly (without prefixes or suffixes) match `${_comp}` because it must be surrounded by commas (if it's in the middle of the list) or immediately follow the colon at the beginning of the list or immediately precede the chevron at the end of the list.

WeiqunZhang: It would also match `$<CXX_COMPILER_ID,someid>`, right?

nmnobre: Right, but that's illegal CMake syntax and will cause CMake to complain with "Expression did not evaluate to a known generator expression" or "Expression syntax not recognized."

ax3l: Thanks! So why do we need to match the `,`?

ax3l (approved):
LGTM, more robust :+1:

PR #4281, opened 2025-01-02
Silence Thread Sanitizer in `InitRandom`

### LINKED ISSUE #4279: ThreadSanitizer: data race `amrex::InitRandom` w/ OMP
With OpenMP, we see from the Clang's ThreadSanitizer that there is a data race flagged in `amrex::InitRandom`:
https://github.com/ECP-WarpX/WarpX/pull/5492#issuecomment-2518839376

ax3l:
## Summary

Either making the shared variables constant (not sensible in the function signature) or using `firstprivate` should silence the warnings we see.

## Additional background

Fix #4279
cc @EZoni

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_Random.cpp
+++ b/Src/Base/AMReX_Random.cpp
@@ -98,10 +98,10 @@ InitRandom (ULong cpu_seed, int nprocs, ULong gpu_seed)
 #endif
 
 #ifdef AMREX_USE_OMP
-#pragma omp parallel firstprivate(cpu_seed, nprocs)
+#pragma omp parallel for
 #endif
+    for (int tid = 0; tid < nthreads; tid++) {
     {
-        int tid = OpenMP::get_thread_num();
         ULong init_seed = cpu_seed + tid*nprocs;
         generators[tid].seed(init_seed);
     }
--- a/Src/Base/AMReX_Random.cpp
+++ b/Src/Base/AMReX_Random.cpp
@@ -98,7 +98,7 @@ InitRandom (ULong cpu_seed, int nprocs, ULong gpu_seed)
 #endif
 
 #ifdef AMREX_USE_OMP
-#pragma omp parallel
+#pragma omp parallel firstprivate(cpu_seed, nprocs)
 #endif
     {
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
It does not seem to make thread sanitizer happy.

Another thing you can try is to turn `omp parallel` into `omp parallel for` and add a for-loop with nthreads iterations so that we don't need call `OpenMP::get_thread_num()`. Maybe its issue is thread sanitizer does not understand the omp thread id is unique.

Src/Base/AMReX_Random.cpp
```diff
@@ -98,10 +98,10 @@ InitRandom (ULong cpu_seed, int nprocs, ULong gpu_seed)
 #endif
 
 #ifdef AMREX_USE_OMP
-#pragma omp parallel
+#pragma omp parallel for
 #endif
+    for (int tid = 0; tid < nthreads; tid++) {
```
AlexanderSinn:
```suggestion
    for (int tid = 0; tid < nthreads; tid++)
```
With the for loop, the parallel region isn't even required anymore.

ax3l: I can type...
```suggestion
    for (int tid = 0; tid < nthreads; tid++)
```

ax3l: Yes, I was thinking the same. Not sure it matters...

WeiqunZhang:
I think it's a false positive. In the past on our own machines, we had to compile gcc from source with certain flags to make thread sanitizer work for OpenMP. For the WarpX CI, maybe similar things need to be done for clang. Not sure if this old stackoverflow post is still relevant. https://stackoverflow.com/questions/33004809/can-i-use-thread-sanitizer-for-openmp-programs

For this particular case, we could simply not using openmp. However, this is probably the first OMP parallel region in the code. I believe the later ones would fail too even if we remove OMP here.

ax3l:
Agreed, I would not change it and rather make sure an OpenMP-enabled sanitizer is used.

PR #4286, opened 2025-01-08
Make using The_Comms_Arena() for particle communication opt-in

atmyers:
For mesh data it is better to use a small, separate Arena for communication. But, we found that for particle communication this approach uses too much memory (see the ImpactX Issue here: https://github.com/ECP-WarpX/impactx/issues/769) 

This restores the old default behavior (prior to PR #4175), but allows users to opt-in to using The_Comms_Arena if desired.

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_NeighborParticlesGPUImpl.H
+++ b/Src/Particle/AMReX_NeighborParticlesGPUImpl.H
@@ -256,7 +256,7 @@ updateNeighborsGPU (bool boundary_neighbors_only)
 
     clearNeighbors();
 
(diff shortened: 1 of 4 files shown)


WeiqunZhang (approved):

ax3l:
Thank you, @atmyers !

Quick question, is it required to use the comms arena for GPU-aware MPI/GPU-direct comms or is it also opt-in independent of that?

atmyers:
It's independent of whether `amrex.use_gpu_aware_mpi` is on or not.

ax3l:
Thanks! I thought so, the prior control vars changed in this diff just confused me :)

PR #4290, opened 2025-01-09
Fix a memory leak in Open BC solver

WeiqunZhang:
We forgot a test to make sure We only build the plans once.

X-Ref:
- https://github.com/ECP-WarpX/impactx/pull/790
- https://github.com/ECP-WarpX/WarpX/issues/5547

### DIFF
--- a/Src/FFT/AMReX_FFT_R2C.H
+++ b/Src/FFT/AMReX_FFT_R2C.H
@@ -473,7 +473,7 @@ void R2C<T,D,S>::prepare_openbc ()
 #if (AMREX_SPACEDIM == 3)
     if (m_do_alld_fft) { return; }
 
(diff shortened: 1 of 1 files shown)


ax3l:
I saw there is also a `defined2` variable for another plan, does related code need to be updated/guarded as well?

ax3l (approved):
Tested, fixes the leaks we saw in ImpactX MPI runs :tada:

WeiqunZhang:
> I saw there is also a `defined2` variable for another plan, does related code need to be updated/guarded as well?

No. The bug is limited to the open bc solver. The other plan is used for other types of solvers with mixed boundaries.

ax3l:
Ok perfect. Thank you again! :)

PR #4291, opened 2025-01-11
Make ParmParse Parser more robust

WeiqunZhang:
It's now allowed to call ParmParse Parser functions (e.g., queryAddWithParser) outside OpenMP parallel regions before ParmParse is initialized.


ax3l (approved):
Thank you! :tada:

PR #4297, opened 2025-01-17
Use bigint in HypreSolver test

WeiqunZhang:
Also fix a bug when HYPRE_Int is long long. The issue was in the previous code the discard branch of `if constexpr` was still checked by the compiler, because it's not in a function template, even though it's in a member function of a class template. We now turn the function into a template.

### DIFF
--- a/Src/Extern/HYPRE/AMReX_HypreSolver.H
+++ b/Src/Extern/HYPRE/AMReX_HypreSolver.H
@@ -108,6 +108,7 @@ public:
 #endif
     fill_local_id (Marker const& marker);
 
+    template <typename AI>
     void fill_global_id ();
 
     template <class Filler,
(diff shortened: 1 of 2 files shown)


Src/Extern/HYPRE/AMReX_HypreSolver.H
```diff
@@ -254,7 +255,13 @@ HypreSolver<MSS>::HypreSolver (Vector<IndexType>   const& a_index_type,
     }
     AMREX_ASSERT(proc_end == proc_begin + m_nrows_proc);
 
-    fill_global_id();
+    // To generate global ids for Hypre, we need to remove duplicates on
+    // nodes shared by multiple Boxes with OverrideSync.  So we need to use
+    // a type that supports atomicAdd.  HYPRE_Int is either int or long
+    // long.  The latter does not have native atomicAdd support in CUDA/HIP.
```
ax3l: Maybe a last sentence, that `unsigned long long` is supported on CUDA/HIP, would describe the new logic fully.

ax3l (approved):
LGTM, small comment how to add the final thought on the inline comment.

ax3l (approved):

PR #4299, opened 2025-01-17
PC: Ensure Uniqueness of SoA Names

ax3l:
## Summary

Avoid user errors. Happened immediately to me: I added a runtime component with the same name as a compile-time SoA component and no error was thrown (now it will).

## Additional background

https://github.com/AMReX-Codes/pyamrex/pull/382

## Checklist

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_ParticleContainer.H
+++ b/Src/Particle/AMReX_ParticleContainer.H
@@ -60,6 +60,7 @@
 #include <memory>
 #include <numeric>
 #include <random>
+#include <stdexcept>
 #include <string>
 #include <tuple>
 #include <type_traits>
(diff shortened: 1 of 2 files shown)


ax3l:
@WeiqunZhang @atmyers I think the CI failures here are unrelated.

WeiqunZhang:
For the PETSc test, I have been seeing it fails in my fork too. The issue is in the installation script. Maybe we should switch to a newer version.

For ICC test, I don't know. We can try to rerun it. We might also consider removing it because it is deprecated.

ax3l:
I restarted ICC already a few times. But admittedly, the original ICC is dead and EOL since over a year or so now. We could remove it.
> icpc: remark #10441: The Intel(R) C++ Compiler Classic (ICC) is deprecated and will be removed from product release in the second half of 2023. The Intel(R) oneAPI DPC++/C++ Compiler (ICX) is the recommended compiler moving forward. Please transition to use this compiler. Use '-diag-disable=10441' to disable this message.

ax3l:
This is ready. I opened another PR to remove ICC from CI.

atmyers (approved):

PR #4308, opened 2025-01-22
Fix includes

WeiqunZhang:
We should not include std headers inside amrex namespace directly or indirectly.


ax3l (approved):

ax3l:
Ah, small updates still needed in call sites like `Src/Particle/AMReX_ParticleIO.H`

AlexanderSinn:
This could maybe fix it, although we will only truly know after merging this and then trying to build hipace and impactx documentation.

WeiqunZhang:
In any case, we need to fix this.

PR #4310, opened 2025-01-26
Speed up CUDA Toolkit installation in Windows CI

WeiqunZhang:
Only install the cuda sub-packages we need. This reduces cuda-toolkit time from 15-20 minutes to a few minutes. Apparently it's very slow to install the visual_studio_integration package. We also use ninja as CMake's generator now.


asalmgren (approved):

ax3l:
Nice!!

PR #4314, opened 2025-01-28
Fix Doxygen with default void std::enable_if_t

AlexanderSinn:
## Summary

Since the Doxygen version changed from 1.9.1 to 1.9.8 in the job runner, that generates
https://amrex-codes.github.io/amrex/docs_xml/doxygen/amrex-doxygen-web.tag.xml, there is a problem where if a function is overloaded where the only difference is the std::enable_if in the return value and the second argument of std::enable_if is not specified, so it defaults to void, then errors are caused in downstream doxygen builds causing CI failures.

## Additional background

In the example below, it can be seen how if there is a second argument in the std::enable_if expression, both overloads get combined into one in the documentation where the return value is replaced by the second argument of std::enable_if. But if there is no second argument both versions end up in the xml file with the full std::enable_if expressions.
```C++
template <typename T>
std::enable_if_t<std::is_same<std::decay_t<decltype(std::declval<T>().box())>, Box>::value,
                 Long>
get_tag_size (T const& tag) noexcept
{...}

template <typename T>
std::enable_if_t<std::is_integral<std::decay_t<decltype(std::declval<T>().size())> >::value,
                 Long>
get_tag_size (T const& tag) noexcept
{...}

template <typename T, typename F>
std::enable_if_t<std::is_same<std::decay_t<decltype(std::declval<T>().box())>, Box>::value>
tagparfor_call_f (
#ifdef AMREX_USE_SYCL
    sycl::nd_item<1> const& item,
#endif
    int icell, T const& tag, F&& f) noexcept
{...}

template <typename T, typename F>
std::enable_if_t<std::is_integral<std::decay_t<decltype(std::declval<T>().size())> >::value>
tagparfor_call_f (
#ifdef AMREX_USE_SYCL
    sycl::nd_item<1> const& item,
#endif
    int i, T const& tag, F&& f) noexcept
{...}
```

```xml
    <member kind="function">
      <type>Long</type>
      <name>get_tag_size</name>
      <anchorfile>namespaceamrex_1_1detail.html</anchorfile>
      <anchor>a5dabf23fc1d6b0482df819bdd5d465c9</anchor>
      <arglist>(T const &amp;tag) noexcept</arglist>
    </member>
    <member kind="function">
      <type>AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE std::enable_if_t&lt; std::is_same&lt; std::decay_t&lt; decltype(std::declval&lt; T &gt;().box())&gt;, Box &gt;::value</type>
      <name>tagparfor_call_f</name>
      <anchorfile>namespaceamrex_1_1detail.html</anchorfile>
      <anchor>a3d1c62850c82bc09be970a6c864d52aa</anchor>
      <arglist>(int icell, T const &amp;tag, F &amp;&amp;f) noexcept</arglist>
    </member>
    <member kind="function">
      <type>AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE std::enable_if_t&lt; std::is_integral&lt; std::decay_t&lt; decltype(std::declval&lt; T &gt;().size())&gt; &gt;::value</type>
      <name>tagparfor_call_f</name>
      <anchorfile>namespaceamrex_1_1detail.html</anchorfile>
      <anchor>ad534e2d6a6e6d52339476c692cbccd78</anchor>
      <arglist>(int i, T const &amp;tag, F &amp;&amp;f) noexcept</arglist>
    </member>
```

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_TagParallelFor.H
+++ b/Src/Base/AMReX_TagParallelFor.H
@@ -125,7 +125,7 @@ get_tag_size (T const& tag) noexcept
 
 template <typename T, typename F>
 AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
-std::enable_if_t<std::is_same<std::decay_t<decltype(std::declval<T>().box())>, Box>::value>
+std::enable_if_t<std::is_same<std::decay_t<decltype(std::declval<T>().box())>, Box>::value, void>
 tagparfor_call_f (
 #ifdef AMREX_USE_SYCL
     sycl::nd_item<1> const& item,
@@ -150,7 +150,7 @@ tagparfor_call_f (
 
 template <typename T, typename F>
 AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
-std::enable_if_t<std::is_integral<std::decay_t<decltype(std::declval<T>().size())> >::value>
+std::enable_if_t<std::is_integral<std::decay_t<decltype(std::declval<T>().size())> >::value, void>
 tagparfor_call_f (
 #ifdef AMREX_USE_SYCL
     sycl::nd_item<1> const& item,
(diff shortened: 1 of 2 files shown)


WeiqunZhang (approved):

ax3l:
Thank you, @AlexanderSinn !! :rocket: :sparkles:

PR #4318, opened 2025-01-29
Enable GPU-aware MPI by default

### LINKED ISSUE #2967: enable GPU-aware MPI when performance conditions are met
GPU-aware MPI significantly improves performance over the default host-pinned buffers in AMReX if two conditions are satisfied:

1. `CUDA_VISIBLE_DEVICES` is _not_ set. (e.g., when using SLURM, `--gpu-bind=none`)
2. Managed memory is _not_ used (setting `amrex.the_arena_is_managed=0`)

If both of these are satisfied, then OpenMPI (at least) uses CUDA IPC (using the UCX `cuda_ipc` transport) to perform device-to-device copies over NVLink. In my tests, this is _significantly_ faster than using AMReX's host-pinned buffers on A100 + NVLink systems, leading to on-node scaling that is essentially perfect (95-99% scaling efficiency).

Scaling tests: https://github.com/BenWibking/quokka/pull/121
OpenMPI issue: https://github.com/open-mpi/ompi/issues/10871

BenWibking:
## Summary
This turns on GPU-aware MPI by default.

## Additional background
On all current machines, simulations run faster with GPU-aware MPI enabled. Two technical issues that prevented this are now resolved: AMReX now has the communication arena, which does not use managed memory, and SLURM no longer uses cgroup isolation for GPU bindings by default.

Closes https://github.com/AMReX-Codes/amrex/issues/2967.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_ParallelDescriptor.cpp
+++ b/Src/Base/AMReX_ParallelDescriptor.cpp
@@ -10,6 +10,9 @@
 
 #ifdef BL_USE_MPI
 #include <AMReX_ccse-mpi.H>
+#if __has_include(<mpi-ext.h>) && defined(OPEN_MPI)
+#         include <mpi-ext.h>
+#endif
 #endif
 
 #ifdef AMREX_PMI
@@ -57,7 +60,7 @@ namespace amrex::ParallelDescriptor {
 #endif
 
 #ifdef AMREX_USE_GPU
-    bool use_gpu_aware_mpi = true;
+    bool use_gpu_aware_mpi = false;
 #else
     bool use_gpu_aware_mpi = false;
 #endif
@@ -1510,6 +1513,28 @@ ReadAndBcastFile (const std::string& filename, Vector<char>& charBuf,
 void
 Initialize ()
 {
+#if defined(AMREX_USE_CUDA)
+
+#if (defined(OMPI_HAVE_MPI_EXT_CUDA) && OMPI_HAVE_MPI_EXT_CUDA) || (defined(MPICH) && defined(MPIX_GPU_SUPPORT_CUDA))
+    use_gpu_aware_mpi = (bool) MPIX_Query_cuda_support();
+#endif
+
+#elif defined(AMREX_USE_HIP)
+
+#if defined(OMPI_HAVE_MPI_EXT_ROCM) && OMPI_HAVE_MPI_EXT_ROCM
+    use_gpu_aware_mpi = (bool) MPIX_Query_rocm_support();
+#elif defined(MPICH) && defined(MPIX_GPU_SUPPORT_HIP)
+    use_gpu_aware_mpi = (bool) MPIX_Query_hip_support();
+#endif
+
+#elif defined(AMREX_USE_SYCL)
+
+#if defined(MPICH) && defined(MPIX_GPU_SUPPORT_ZE)
+    use_gpu_aware_mpi = (bool) MPIX_Query_ze_support();
+#endif
+
+#endif
+
 #ifndef BL_AMRPROF
     ParmParse pp("amrex");
     pp.queryAdd("use_gpu_aware_mpi", use_gpu_aware_mpi);
--- a/Src/Base/AMReX_ParallelDescriptor.cpp
+++ b/Src/Base/AMReX_ParallelDescriptor.cpp
@@ -57,7 +57,7 @@ namespace amrex::ParallelDescriptor {
 #endif
 
 #ifdef AMREX_USE_GPU
-    bool use_gpu_aware_mpi = false;
+    bool use_gpu_aware_mpi = true;
 #else
     bool use_gpu_aware_mpi = false;
 #endif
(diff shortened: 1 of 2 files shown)


BenWibking:
It looks like the HIP EB failure is due to flaky CI.

WeiqunZhang:
I think maybe we want to do something like
```
diff --git a/Src/Base/AMReX_ParallelDescriptor.cpp b/Src/Base/AMReX_ParallelDescriptor.cpp
index 67848505f2..83d55bb273 100644
--- a/Src/Base/AMReX_ParallelDescriptor.cpp
+++ b/Src/Base/AMReX_ParallelDescriptor.cpp
@@ -10,6 +10,9 @@
 
 #ifdef BL_USE_MPI
 #include <AMReX_ccse-mpi.H>
+#if __has_include(<mpi-ext.h>) && defined(OPEN_MPI)
+#         include <mpi-ext.h>
+#endif
 #endif
 
 #ifdef AMREX_PMI
@@ -57,7 +60,7 @@ namespace amrex::ParallelDescriptor {
 #endif
 
 #ifdef AMREX_USE_GPU
-    bool use_gpu_aware_mpi = true;
+    bool use_gpu_aware_mpi = false;
 #else
     bool use_gpu_aware_mpi = false;
 #endif
@@ -1510,6 +1513,28 @@ ReadAndBcastFile (const std::string& filename, Vector<char>& charBuf,
 void
 Initialize ()
 {
+#if defined(AMREX_USE_CUDA)
+
+#if (defined(OMPI_HAVE_MPI_EXT_CUDA) && OMPI_HAVE_MPI_EXT_CUDA) || (defined(MPICH) && defined(MPIX_GPU_SUPPORT_CUDA))
+    use_gpu_aware_mpi = (bool) MPIX_Query_cuda_support();
+#endif
+
+#elif defined(AMREX_USE_HIP)
+
+#if defined(OMPI_HAVE_MPI_EXT_ROCM) && OMPI_HAVE_MPI_EXT_ROCM
+    use_gpu_aware_mpi = (bool) MPIX_Query_rocm_support();
+#elif defined(MPICH) && defined(MPIX_GPU_SUPPORT_HIP)
+    use_gpu_aware_mpi = (bool) MPIX_Query_hip_support();
+#endif
+
+#elif defined(AMREX_USE_SYCL)
+
+#if defined(MPICH) && defined(MPIX_GPU_SUPPORT_ZE)
+    use_gpu_aware_mpi = (bool) MPIX_Query_ze_support();
+#endif
+
+#endif
+
 #ifndef BL_AMRPROF
     ParmParse pp("amrex");
     pp.queryAdd("use_gpu_aware_mpi", use_gpu_aware_mpi);
```

Src/Base/AMReX_ParallelDescriptor.cpp
```diff
@@ -1510,6 +1513,28 @@ ReadAndBcastFile (const std::string& filename, Vector<char>& charBuf,
...
+#if (defined(OMPI_HAVE_MPI_EXT_CUDA) && OMPI_HAVE_MPI_EXT_CUDA) || (defined(MPICH) && defined(MPIX_GPU_SUPPORT_CUDA))
+    use_gpu_aware_mpi = (bool) MPIX_Query_cuda_support();
+#endif
+
+#elif defined(AMREX_USE_HIP)
+
+#if defined(OMPI_HAVE_MPI_EXT_ROCM) && OMPI_HAVE_MPI_EXT_ROCM
+    use_gpu_aware_mpi = (bool) MPIX_Query_rocm_support();
+#elif defined(MPICH) && defined(MPIX_GPU_SUPPORT_HIP)
+    use_gpu_aware_mpi = (bool) MPIX_Query_hip_support();
```
WeiqunZhang:
```suggestion
    int is_supported = 0;
    if (MPIX_GPU_query_support(MPIX_GPU_SUPPORT_HIP, &is_supported) == MPI_SUCCESS) {
        use_gpu_aware_mpi = (bool) is_supported;
    }
```

At least for the cray-mpich version I tested on Frontier, `MPIX_Query_hip_support()` does not exist. But it does have `MPIX_GPU_query_support()` and `MPIX_GPU_SUPPORT_HIP` is defined.

Src/Base/AMReX_ParallelDescriptor.cpp
```diff
@@ -1510,6 +1513,28 @@ ReadAndBcastFile (const std::string& filename, Vector<char>& charBuf,
...
+#endif
+
+#elif defined(AMREX_USE_SYCL)
+
+#if defined(MPICH) && defined(MPIX_GPU_SUPPORT_ZE)
+    use_gpu_aware_mpi = (bool) MPIX_Query_ze_support();
```
WeiqunZhang:
```suggestion
    int is_supported = 0;
    if (MPIX_GPU_query_support(MPIX_GPU_SUPPORT_ZE, &is_supported) == MPI_SUCCESS) {
        use_gpu_aware_mpi = (bool) is_supported;
    }
```

WeiqunZhang (changes requested):

WeiqunZhang:
I tested this on Frontier. With `export MPICH_GPU_SUPPORT_ENABLED=1`, `MPIX_GPU_query_support(MPIX_GPU_SUPPORT_HIP, &is_supported)` sets `is_supported` to 1. With `unset MPICH_GPU_SUPPORT_ENABLED`, it becomes 0.

WeiqunZhang:
Tests on Perlmutter using both cray-mpich and openmpi also passed.

WeiqunZhang (approved):

PR #4319, opened 2025-01-31
Remove HYPRE version requirement

nischild:
## Summary
If Hypre dependency is added through CMake `find_package` AMReX enforces version 2.20.0.
This is version seems to be outdated (at least 5 years old). Therefore, I suggest to remove the version requirement for HYPRE from AMReX (or update it if a specific version is promoted.)

## Additional background
I added HYPRE using CMake `find_package` as a dependency to AMReX and found that the required version was v2.20.0.
Using HYPRE v2.20.0 lead to a SegFault within the AMReX HYPRE interface.
This SegFault could be removed by using the latest HYPRE version. To prevent such confusions in future removing the version requirement would be a solution.

### DIFF
--- a/Tools/CMake/AMReXThirdPartyLibraries.cmake
+++ b/Tools/CMake/AMReXThirdPartyLibraries.cmake
@@ -143,7 +143,7 @@ endif ()
 # HYPRE
 #
 if (AMReX_HYPRE)
-    find_package(HYPRE 2.20.0 REQUIRED)
+    find_package(HYPRE REQUIRED)
     if(AMReX_CUDA)
         find_package(CUDAToolkit REQUIRED)
 


WeiqunZhang:
Are the new versions considered incompatible with 2.20.0 by cmake?

Could you tell us more details about the segfault? Is it a GPU run? Are you using managed memory? Early versions of Hypre might depend on managed memory in some setups.

If we change this, we probably should also update Tools/CMake/AMReXConfig.cmake.in.

nischild:
It seems that `find_package` does not enforces the version number if the `EXACT` keyword is not provided.

I compiled only for CPU using
gcc - 14.2.1
openmpi - 4.1.7

and configured AMReX with
```
cmake -DAMReX_HYPRE=ON -DAMREX_LINEAR_SOLVERS=ON -DAMReX_SPACEDIM=3
```

Since I can currently not reproduce the SegFault I guess it was related to my toolchain before and not actually related to v2.20.0 of HYPRE.

Therefore, I close the PR.

Tools/CMake/AMReXThirdPartyLibraries.cmake
```diff
@@ -143,7 +143,7 @@ endif ()
 # HYPRE
 #
 if (AMReX_HYPRE)
-    find_package(HYPRE 2.20.0 REQUIRED)
```
ax3l, line 146: This is the minimally supported version (not the exact one).
https://cmake.org/cmake/help/latest/command/find_package.html#typical-usage

In dependencies that support `Config` mode, one can specify if major versions are compatible or not (not often used yet in the wild), which could limit the upper version that is supported.

There is a newer feature where one can specify version ranges, or one can just run `find_package(...)` again (w/o `REQUIRED`) for the major version step, if both are supported.

PR #4322, opened 2025-02-04
Fix: Names in `PC::make_alike`

ax3l:
## Summary

The ParticleContainer's `make_alike` function must not forget the names of the SoA components. Currently, it creates a new container with the same components, but default names.

## Additional background

First seen in https://github.com/ECP-WarpX/WarpX/pull/5481

## Checklist

The proposed changes:


WeiqunZhang (approved):

ax3l:
Wait, there is more to it. Have to also set the compile-time names.

ax3l:
Now it's complete.

PR #4326, opened 2025-02-07
GPU Mapping

WeiqunZhang:
For perlmutter and frontier, if there are multiple devices available, we will try to map GPUs to the closest core.

For an FFT test on perlmutter using 256 nodes, the correct mapping reduced the run time from 0.172 to 0.127. Note that you can achieve the similar effect with `srun ... bash -c "export
CUDA_VISIBLE_DEVICES=\$((3-SLURM_LOCALID)); ..."` by manually limiting the number of visible devices. But in this commit, we are trying to do this automatically for the user. Also note that MPI appears to crash with gpu-bind=closest on perlmutter. So we need to use gpu-bind=none.

For frontier, you could use gpu-bind=closest. But if your use gpu-bind=none, this commit will try to do the correct mapping for you.

In this commit, we also removed the old machine stuff and added new code for machine detection.


Src/Base/AMReX_GpuDevice.cpp
```diff
@@ -208,7 +209,11 @@ Device::Initialize ()
     }
     else {
         if (amrex::Verbose() && ParallelDescriptor::IOProcessor()) {
-            amrex::Warning("Multiple GPUs are visible to each MPI rank. This is usually not an issue. But this may lead to incorrect or suboptimal rank-to-GPU mapping.");
+            if ((Machine::name() != "nersc.perlmutter") &&
+                (Machine::name() != "olcf.frontier"))
+            {
+                amrex::Warning("Multiple GPUs are visible to each MPI rank. This is usually not an issue. But this may lead to incorrect or suboptimal rank-to-GPU mapping.");
```
ax3l: I think with the implementation below, we need to be more precise now: for the machines we implement logic for, we should post something like: "Fixing GPU assignment for Frontier according to heuristics..." or so?

ax3l (approved):
Thank you! LGTM :+1:

PR #4328, opened 2025-02-10
Adding templating to the MFInterp functions

bsrunnels:
## Summary

Adding templating to the MFInterp functions.

## Additional background

This enables codes that use BaseFab<T> (instead of MultiFabs) to use the MFInterp methods. 

## Checklist

The proposed changes:

### DIFF
--- a/Src/AmrCore/AMReX_MFInterp_1D_C.H
+++ b/Src/AmrCore/AMReX_MFInterp_1D_C.H
@@ -3,9 +3,10 @@
 
 namespace amrex {
 
+template<typename T>
 AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
-void mf_cell_cons_lin_interp_limit_minmax_llslope (int i, int, int, Array4<Real> const& slope,
-                                      Array4<Real const> const& u, int scomp, int ncomp,
+void mf_cell_cons_lin_interp_limit_minmax_llslope (int i, int, int, Array4<T> const& slope,
+                                      Array4<T const> const& u, int scomp, int ncomp,
                                       Box const& domain, IntVect const& ratio, BCRec const* bc) noexcept
 {
     Real sfx = Real(1.0);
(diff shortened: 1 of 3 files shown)


Src/AmrCore/AMReX_MFInterp_1D_C.H
```diff
@@ -3,9 +3,10 @@
...
 
+template<typename T>
 AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
-void mf_cell_cons_lin_interp_limit_minmax_llslope (int i, int, int, Array4<Real> const& slope,
-                                      Array4<Real const> const& u, int scomp, int ncomp,
+void mf_cell_cons_lin_interp_limit_minmax_llslope (int i, int, int, Array4<T> const& slope,
+                                      Array4<T const> const& u, int scomp, int ncomp,
                                       Box const& domain, IntVect const& ratio, BCRec const* bc) noexcept
 {
     Real sfx = Real(1.0);
```
WeiqunZhang, line 12: Should we use T here and other places in this function too?

atmyers:
@bsrunnels - is this still something you're working on?

bsrunnels:
@atmyers Yes though not recently. I ended up duplicating some of the code in alamo but would prefer to merge this capability into amrex if possible.

WeiqunZhang:
> @atmyers Yes though not recently. I ended up duplicating some of the code in alamo but would prefer to merge this capability into amrex if possible.

Could you answer the question I raised here? https://github.com/AMReX-Codes/amrex/pull/4328#discussion_r1950243749

What is `T` in your case for the following code where type of `u()` is `T`?
```
Real df = Real(2.0) * (u(i+1,0,0,nu) - u(i  ,0,0,nu));
```
We could merge your code as it is. But it's unclear what it tries to do. Does it rely on T being implicitly cast to Real? Or it needs an overloaded `operator*`? If we refactor this code later, how could we know it won't break your code?

PR #4333, opened 2025-02-13
Fix optional component communication for pure SoA particles

atmyers:
If using the legacy layout, the `AMREX_SPACEDIM` position components are counted separately. But if using pure SoA, the position components are assumed to be part of `NArrayReal` - hence the `h_redistribute_real_comp` vector was the wrong size.

The proposed changes:


ax3l (approved):
Thank you!!

PR #4338, opened 2025-02-19
Fix FillRandomNormal for odd number of length

WeiqunZhang:
In both CUDA and HIP, the length must be even for the normal distribution generator we use. This PR fixes it by generating the last one on the host if the total length is odd.


ajnonaka (approved):
Thanks Weiqun!

PR #4341, opened 2025-02-21
Copy comm flags in make_alike

atmyers:
Need to make sure these are preserved (we use them as IO flags in WarpX)

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_ParticleContainer.H
+++ b/Src/Particle/AMReX_ParticleContainer.H
@@ -1376,13 +1376,16 @@ public:
 
         // add runtime real comps to tmp
         for (int ic = 0; ic < this->NumRuntimeRealComps(); ++ic) {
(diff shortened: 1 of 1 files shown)


Src/Particle/AMReX_ParticleContainer.H
```diff
@@ -1376,14 +1376,17 @@ public:
 
         // add runtime real comps to tmp
         for (int ic = 0; ic < this->NumRuntimeRealComps(); ++ic) {
-            tmp.AddRealComp(real_names.at(ic + NArrayReal), false);
+            tmp.AddRealComp(real_names.at(ic + NArrayReal));
```
ax3l, line 1379: Instead of `true` by default, can we copy the current property?

ax3l: Ah wait, you copy it below...

ax3l (approved):
Nice, good catch!

PR #4348, opened 2025-02-26
Add a way to reuse TagParallelFor Tags

AlexanderSinn:
## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_TagParallelFor.H
+++ b/Src/Base/AMReX_TagParallelFor.H
@@ -159,13 +159,13 @@ struct TagVector {
     TagVector (const TagVector& other) = delete;
     TagVector& operator= (const TagVector& other) = delete;
     TagVector (TagVector&& other) noexcept
-        : h_buffer{other.h_buffer}
-          d_buffer{other.d_buffer}
-          d_tags{other.d_tags}
-          d_nwarps{other.d_nwarps}
-          ntags{other.ntags}
-          ntotwarps{other.ntotwarps}
-          nblocks{other.nblocks}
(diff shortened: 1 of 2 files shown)


Src/Base/AMReX_TagParallelFor.H
```diff
@@ -101,31 +101,153 @@ struct VectorTag {
...
+    }
+
+    ~TagVector () {
+        if (defined) {
+            undefine();
+        }
+    }
+
+    TagVector (const TagVector& other) = delete;
+    TagVector (TagVector&& other) = default;
```
WeiqunZhang: Would this cause double free?

Src/Base/AMReX_TagParallelFor.H
```diff
@@ -241,20 +334,60 @@ ParallelFor_doit (Vector<TagType> const& tags, F && f)
...
+{
+    AMREX_ALWAYS_ASSERT(tv.is_defined());
+
+    if (tv.ntags == 0) { return; }
+
+    const auto d_tags = tv.d_tags;
+    const auto ntags = tv.ntags;
+
+#ifdef AMREX_USE_OMP
+#pragma omp parallel
```
WeiqunZhang:
```suggestion
#pragma omp parallel for
```

Src/Base/AMReX_TagParallelFor.H
AlexanderSinn: Good catch, looks like it would

WeiqunZhang (approved):

PR #4349, opened 2025-02-26
Remove stream sync in build_arrays

AlexanderSinn:
## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_FabArray.H
+++ b/Src/Base/AMReX_FabArray.H
@@ -1623,7 +1623,10 @@ FabArray<FAB>::build_arrays () const
 #ifdef AMREX_USE_GPU
         m_arrays.dp = (A*)m_dp_arrays;
         m_const_arrays.dp = (AC*)m_dp_arrays + n;
-        Gpu::htod_memcpy(m_dp_arrays, m_hp_arrays, n*2*sizeof(A));
+        Gpu::htod_memcpy_async(m_dp_arrays, m_hp_arrays, n*2*sizeof(A));
+        if (!Gpu::inNoSyncRegion()) {
+            Gpu::streamSynchronize();
+        }
 #endif
     }
 }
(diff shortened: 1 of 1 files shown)


Src/Base/AMReX_FabArray.H
```diff
@@ -1633,9 +1636,12 @@ void
 FabArray<FAB>::clear_arrays ()
 {
 #ifdef AMREX_USE_GPU
-    The_Pinned_Arena()->free(m_hp_arrays);
-    The_Arena()->free(m_dp_arrays);
-    m_dp_arrays = nullptr;
+    if (m_dp_arrays) {
+        Gpu::streamSynchronize();
```
WeiqunZhang: I assume this is necessary for the correctness of the test you have been running. But is it possible to avoid streamSynchronize here and do it somewhere else?

AlexanderSinn: Indeed my test case crashed before I added that stream sync. Since we need to deallocate pinned memory, there needs to be a stream sync as that bit of memory could immediately be reallocated by something else and written to by the CPU before the next stream operation finishes. The only way to avoid this specific stream sync is to check if some other place has already synchronized the stream since the memcpy was done. This would require a per-stream counter to be updated every time stream sync is called, which I think could be a good addition as part of a bigger update to amrex streamsync.

WeiqunZhang: What I mean is that it should be the user's responsibility to know if an async kernel is still using the data in a FabArray, when it has reached its end of life. If that's case, the user needs to explicitly sync it. Note that before the metadata are cleared, the data stored in Fabs have already been cleared. So even if we want to insert a sync, it should be at the beginning of ~FabArray. But I still think it's the user's responsibility in this failure mode.

WeiqunZhang (approved):

PR #4352, opened 2025-02-26
Put loop over iter in MLMG smooth

AlexanderSinn:
## Checklist

The proposed changes:

### DIFF
--- a/Src/LinearSolvers/MLMG/AMReX_MLLinOp.H
+++ b/Src/LinearSolvers/MLMG/AMReX_MLLinOp.H
@@ -358,7 +358,7 @@ public:
      * \param niter number of (multicolor) smoothing iterations
      */
     virtual void smooth (int amrlev, int mglev, MF& sol, const MF& rhs,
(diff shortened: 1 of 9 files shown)


Src/LinearSolvers/MLMG/AMReX_MLLinOp.H
```diff
@@ -355,9 +355,10 @@ public:
      * \param sol               unknowns
      * \param rhs               RHS
      * \param skip_fillboundary flag controlling whether ghost cell filling can be skipped.
+     * \param niter number of (multicolor) smoothing iterations
      */
     virtual void smooth (int amrlev, int mglev, MF& sol, const MF& rhs,
-                         bool skip_fillboundary=false) const = 0;
+                         bool skip_fillboundary=false, int niter=1) const = 0;
```
WeiqunZhang: Since we are changing the interface here, let's make both parameters non-optional. Having optional parameters in virtual functions is said to be an anti-pattern.

WeiqunZhang (approved):

PR #4355, opened 2025-02-26
Use SingleStreamRegion and NoSyncRegion in MLMG solve

AlexanderSinn:
## Checklist

The proposed changes:

### DIFF
--- a/Src/LinearSolvers/MLMG/AMReX_MLMG.H
+++ b/Src/LinearSolvers/MLMG/AMReX_MLMG.H
@@ -155,6 +155,8 @@ public:
     void setNSolve (int flag) noexcept { do_nsolve = flag; }
     void setNSolveGridSize (int s) noexcept { nsolve_grid_size = s; }
 
+    void setNoGpuSync (bool do_not_sync) noexcept { do_no_sync_gpu = do_not_sync; }
+
 #if defined(AMREX_USE_HYPRE) && (AMREX_SPACEDIM > 1)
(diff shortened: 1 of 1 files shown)


Src/LinearSolvers/MLMG/AMReX_MLMG.H
```diff
@@ -374,6 +374,9 @@ MLMGT<MF>::solve (const Vector<AMF*>& a_sol, const Vector<AMF const*>& a_rhs,
 {
     BL_PROFILE("MLMG::solve()");
 
+    [[maybe_unused]] Gpu::SingleStreamRegion single_stream_reg;
```
WeiqunZhang: Maybe we should make this a runtime option. For example, maybe we can add a member function MLMG::setNoGpuSync to control whether we want to avoid gpu sync if possible (and use single stream for safety).

AlexanderSinn: hmm NoSyncRegion is a bit difficult to use with a runtime parameter

WeiqunZhang (approved):

marchdf:
@AlexanderSinn thanks for pushing this through! So do you have a call somewhere where you do: `setNoGpuSync(true)`? In your code? Can you point me to that?

WeiqunZhang:
Amr-wind does not build MLMG directly. But, there is a function in amr-wind that gives you a reference to an mlmg object. You can call it at any time before `solve`.

PR #4357, opened 2025-02-27
Fixed EB data for anisotropic grids: boundary centroid, normal vector (backward compatible)

skang67:
## Summary

1. This PR corrects two EB data quantities: the boundary centroid (`bcent`) in `set_eb_data` called within `build_cells`, and the normal vector (`normal`) in `EBToPVD::EBToPolygon`.
2. This correction only affects anisotropic grids (dx != dy != dz) and does not affect computations for isotropic grids (dx==dy==dz).

## Additional background

1. The current `bcent`, normalized using a single parameter (`(nx*dx[0])*(nx*dx[0]) + (ny*dx[1])*(ny*dx[1]) + (nz*dx[2])*(nz*dx[2])`) for all x-, y-, and z-components, results in inappropriate non-dimensional values. The proper physical coordinates can be calculated in user-end code (e.g. ERF) by re-normalizing the `bcent` parameters. 
2. However, when constructing multilevel EB data by coarsening the finest EB data in `Level::coarsenFromFine`, the `bcent` from the finest grid is used to construct coarse EB data. This process produces incorrect `bcent` on coarse grids, which cannot be recovered in user-end code. Therefore, I propose this PR.

@nataraj2 I am submitting a PR to fix normalization of `bcent` for anisotropic grids. So far, I have used the `bcent` as it is, by renormalizing it in ERF. However, I encountered an issue that cannot be fixed from ERF when coarsening the EB data from the fine grid to coarse grids. I discussed this issue with @WeiqunZhang.

## Checklist

The proposed changes:

### DIFF
--- a/Src/EB/AMReX_EBToPVD.cpp
+++ b/Src/EB/AMReX_EBToPVD.cpp
@@ -54,15 +54,19 @@ void EBToPVD::EBToPolygon(const Real* problo, const Real* dx,
                Real azm = apz(i  ,j  ,k  );
                Real azp = apz(i  ,j  ,k+1);
 
-               Real apnorm = std::sqrt((axm-axp)*(axm-axp) + (aym-ayp)*(aym-ayp) + (azm-azp)*(azm-azp));
+               Real adx = (axm-axp) * dx[1] * dx[2];
+               Real ady = (aym-ayp) * dx[0] * dx[2];
+               Real adz = (azm-azp) * dx[0] * dx[1];
+
+               Real apnorm = std::sqrt(adx*adx + ady*ady + adz*adz);
                Real apnorminv = Real(1.0)/apnorm;
 
                std::array<Real,3> normal, centroid;
                std::array<std::array<Real,3>,8> vertex;
 
-               normal[0] = (axp-axm) * apnorminv;
-               normal[1] = (ayp-aym) * apnorminv;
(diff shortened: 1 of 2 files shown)


nataraj2:
Sounds good.

Src/EB/AMReX_EBToPVD.cpp
```diff
@@ -54,15 +54,19 @@ void EBToPVD::EBToPolygon(const Real* problo, const Real* dx,
                Real azm = apz(i  ,j  ,k  );
                Real azp = apz(i  ,j  ,k+1);
 
-               Real apnorm = std::sqrt((axm-axp)*(axm-axp) + (aym-ayp)*(aym-ayp) + (azm-azp)*(azm-azp));
+               Real adx = (axm-axp) * dx[1] * dx[2];
+               Real ady = (aym-ayp) * dx[0] * dx[2];
+               Real adz = (azm-azp) * dx[0] * dx[1];
+
+               Real apnorm = std::sqrt(adx*adx + ady*ady + adz*adz);
```
WeiqunZhang, line 61: @drangara You added this file. Could you please review the change in this file?

drangara: The changes in this file look good to me

WeiqunZhang (approved):

PR #4364, opened 2025-03-05
ParmParse: Type Hints

ax3l:
## Summary

Today, we use the `ParmParse` object not only anymore to read text-based key-value pairs from input files and CLI options, but additionally and sometimes exclusively populate it heavily from the C++ and Python APIs directly.

This proposes a concept to store the type of the latest `add` with each entry, which can help to expose `ParmParse` options, e.g., when exporting them to Python dictionaries and other file formats (YAML, TOML, JSON, XML, etc.). When a typehint is found, we can cast to the correct type automatically for the value, instead of returning a string.

## Additional background

https://github.com/AMReX-Codes/pyamrex/pull/417#issuecomment-2699220142

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_ParmParse.H
+++ b/Src/Base/AMReX_ParmParse.H
@@ -3,8 +3,10 @@
 #include <AMReX_Config.H>
 
 #include <AMReX_BLassert.H>
+#include <AMReX_Box.H>
 #include <AMReX_Enum.H>
 #include <AMReX_INT.H>
+#include <AMReX_IntVect.H>
 #include <AMReX_IParser.H>
 #include <AMReX_Parser.H>
 #include <AMReX_TypeTraits.H>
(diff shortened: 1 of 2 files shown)


Src/Base/AMReX_ParmParse.H
```diff
@@ -1612,6 +1616,7 @@ public:
         // vector<vector<std::string>>.
         std::vector<std::vector<std::string>> m_vals;
         mutable Long                          m_count = 0;
+        std::variant<std::string, bool, int, long, long long, amrex::IntVect, amrex::Box, float, double> m_typehint = std::string{};
```
WeiqunZhang: Maybe we could make it cheaper by using `T*` instead of T.

ax3l: Good point(er), hehe

ax3l: Updated :)

ax3l:
I was wondering if I should also just store the vector type as a hint or if it comes naturally from m_vals being size >1...

ax3l:
Tested in https://github.com/AMReX-Codes/pyamrex/pull/429, works like a charm :)

WeiqunZhang (approved):

PR #4377, opened 2025-03-15
Update hypreSynchronize

WeiqunZhang:
hypre_SyncCudaDevice has been removed since v2.32.0-31.

### DIFF
--- a/Src/Base/AMReX_GpuDevice.cpp
+++ b/Src/Base/AMReX_GpuDevice.cpp
@@ -1081,7 +1081,7 @@ Device::profilerStop ()
 void hypreSynchronize ()
 {
 #ifdef AMREX_USE_GPU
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
Pretty confident that this is ready for merge.

Src/Base/AMReX_GpuDevice.cpp
```diff
@@ -1081,8 +1081,12 @@ Device::profilerStop ()
 void hypreSynchronize ()
 {
 #ifdef AMREX_USE_GPU
+#if (HYPRE_RELEASE_NUMBER > 23200) || (HYPRE_RELEASE_NUMBER == 23200 && HYPRE_DEVELOP_NUMBER >= 4)
```
WeiqunZhang:
```suggestion
#if (HYPRE_RELEASE_NUMBER > 23200)
```
Since this PR was submitted, there is a new hypre release, 2.33.0. So the logic here that deals with the hypre development version between 2.32 and 2.33 can be simplified now.

WeiqunZhang:
Merged into #4165

PR #4379, opened 2025-03-17
Update CMake for Cray compiler

### LINKED ISSUE #4378: CMake 3.28+ Cray compiler ID is not recognized by AMReX
Beginning in CMake/3.28, Cray's CCE compiler is identified by `CMAKE_CXX_COMPILER_ID` as `CrayClang`, instead of `Cray`, as it is in cmake <3.28. See the CMake release notes here: https://cmake.org/cmake/help/latest/release/3.28.html . This causes AMReX to fail to build with:
```
CMake Error at /.../amrex/Tools/CMake/AMReXGenerateConfigHeader.cmake:54 (message):
  Compiler 'CrayClang' not supported by AMReX developers! Try to configure
  with -DAMReX_DIFFERENT_COMPILER=ON
```

While we can and have temporarily added `-DAMReX_DIFFERENT_COMPILER=ON` to our builds, is that the correct solution? This CMake file appears to add some compiler definitions, and I wonder if the `__CRAYC` definition would still be important here.

Link to relevant portion of AMReX code: https://github.com/AMReX-Codes/amrex/blob/c35770e39300d308d604114bba0927317e67bfd2/Tools/CMake/AMReXGenerateConfigHeader.cmake#L40

Thanks,
Nick

WeiqunZhang:
Since CMake/3.28, Cray compiler has been identified as `CrayClang`. Update compiler macro too because Cray compiler no longer defines `__CRAYC`.


ax3l (approved):
Good to know, thanks for the update!

Issue #4382, opened 2025-03-20
Loss of significance error in mf_cell_cons_interp routines


sw561:
The mf_cell_cons_interp interpolater object limits the output to not create new minima or maxima. However there is a corner case where a numerical loss of significance error can lead to a negative number occuring in the fine mesh even while every cell on the coarse mesh is >= 0. A simple program to reproduce the error is attached below.

When investigating this I found that the slope limiting is based on ensuring that the reconstructed field on the cell does not exceed the maximum or drop below the minimum in any of the fine cells. When working with volume fractions, this means you can have a fine cell with exactly zero volume fraction, when the overlapping coarse cell is non-zero. (Although in fact this is the subtraction of two almost exactly equal numbers, leaving junk which may turn out to be negative)

This is different to what I'd expect: I would expect the slope limiting to be based on the extreme corner of the cell (and thus independent of the refinement ratio) rather than based on the value of the cell nearest the extreme corner.

[main.txt](https://github.com/user-attachments/files/19374338/main.txt)

WeiqunZhang:
You mentioned volume fractions. Are you using mf_cell_cons_interp for cut cells? That probably will not work well. You might want to use eb_mf_cell_const_interp.

As for the roundoff error, maybe you can add a protection against negative values somewhere.

asalmgren:
Closing this issue due to lack of activity.  Please re-open it if you feel there is is still an issue!

PR #4383, opened 2025-03-20
Slope limiting to corner of cell, not fine cell next to corner of cell

sw561:
## Summary

Proposed change to illustrate the issue being described in #4382
Slope limiting is based on value in the fine cell (and thus dependent on refinement ratio) instead of the corner of the cell.

## Additional background

See issue #4382

## Checklist

The proposed changes:
- [X] changes answers in the test suite to more than roundoff level
- [X] are likely to significantly affect the results of downstream AMReX users


WeiqunZhang:
I am reluctant to accept the changes. It might affect a lot of codes with unknown consequences. Maybe we could add a new option for this instead of replacing the current behavior.

asalmgren:
Also -- imagine a bilinear function -- the slope should be limited in that case by the fine corner ghost cell, not the corner of the cell

PR #4391, opened 2025-03-25
TagParallelFor: Fix launch bounds

WeiqunZhang:
We must pass block size as a template parameter to amrex::launch, otherwise the incorrect launch bounds will be used if AMREX_GPU_MAX_THREADS is set to be less than 256 at compile time.


WeiqunZhang:
Pretty confident that this is ready for merge.

AlexanderSinn (approved):

WeiqunZhang:
Merged into #4165

PR #4392, opened 2025-03-25
Avoid calling bisection in MFParallelFor

WeiqunZhang:
Also change the default number of threads per block in MFParallelFor to 128 for CUDA.


WeiqunZhang:
This contains #4414. If this is merged first, I will close #4414.

asalmgren (approved):

PR #4395, opened 2025-03-27
Added bounds checking to GPUable arrays

ankithadas:
## Summary
Implemented bounds checking for GPU arrays. 

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_Array.H
+++ b/Src/Base/AMReX_Array.H
@@ -148,8 +148,8 @@ namespace amrex {
         }
 
 #if defined(AMREX_DEBUG) || defined(AMREX_BOUND_CHECK)
-        AMREX_GPU_HOST_DEVICE inline
-        void index_assert (int i) const
+        AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+        static void index_assert (int i)
         {
             if (i < 0 || static_cast<unsigned int>(i) >= N) {
                 AMREX_IF_ON_DEVICE((
@@ -290,8 +290,8 @@ namespace amrex {
             return p;
         }
 #if defined(AMREX_DEBUG) || defined(AMREX_BOUND_CHECK)
-        AMREX_GPU_HOST_DEVICE inline
-        void index_assert (int i) const
+        AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+        static void index_assert (int i)
         {
             if (i<XLO || i>XHI) {
                 AMREX_IF_ON_DEVICE((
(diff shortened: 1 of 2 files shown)


WeiqunZhang:
One of the tests can be fixed with
```
diff --git a/Tests/LinearSolvers/NodeTensorLap/MyTest.cpp b/Tests/LinearSolvers/NodeTensorLap/MyTest.cpp
index 07eb465cf7..46d52a4c10 100644
--- a/Tests/LinearSolvers/NodeTensorLap/MyTest.cpp
+++ b/Tests/LinearSolvers/NodeTensorLap/MyTest.cpp
@@ -168,7 +168,11 @@ MyTest::initData ()
 
                 Real x = i*dx[0];
                 Real y = j*dx[1];
+#if (AMREX_SPACEDIM == 2)
+                Real z = 0;
+#else
                 Real z = k*dx[2];
+#endif
 
                 phi(i,j,k) = (std::cos(tpi*x) * std::cos(tpi*y) * std::cos(tpi*z))
                     + 0.25 * (std::cos(fpi*x) * std::cos(fpi*y) * std::cos(fpi*z));
```

As for the other one, I think we should make index_assert a static function.

ankithadas:
Not sure why the Smoke test failed.

WeiqunZhang:
Don't worry. That `../../../installdir/lib/cmake/AMReX/AMReXConfig.cmake, version: unknown` error is a cmake 4.0.0 "feature". There is a workaround in #4399.

Src/Base/AMReX_Array.H
```diff
@@ -137,6 +147,24 @@ namespace amrex {
...
         }
 
+#if defined(AMREX_DEBUG) || defined(AMREX_BOUND_CHECK)
+        AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+        static void index_assert (int i)
+        {
+            if (i < 0 || static_cast<unsigned int>(i) >= N) {
+                AMREX_IF_ON_DEVICE((
+                    AMREX_DEVICE_PRINTF(" (%d) is out of bound (0:%d)\n",
+                                        i, N);
```
WeiqunZhang:
```suggestion
                                        i, N-1);
```

Src/Base/AMReX_Array.H
```diff
@@ -137,6 +147,24 @@ namespace amrex {
...
+                    amrex::Abort();
+                ))
+                AMREX_IF_ON_HOST((
+                    std::stringstream ss;
+                    ss << " (" << i << ") is out of bound (0:" << N <<")";
```
WeiqunZhang:
```suggestion
                    ss << " (" << i << ") is out of bound (0:" << N-1 <<")";
```

Src/Base/AMReX_Array.H
ankithadas: Good catch.

WeiqunZhang (approved):

PR #4396, opened 2025-03-30
fix multi-level ParticleToMesh when zero_out_input is false

BenWibking:
## Summary
When more than one level is passed to `amrex::ParticleToMesh`, the input MultiFabs are always zeroed out before they are overwritten with the particle deposition, even if `zero_out_input` is set to false. This fixes the behavior for the multi-level case when `zero_out_input` is false.

## Checklist

The proposed changes:

### DIFF
--- a/Src/AmrCore/AMReX_AmrParticles.H
+++ b/Src/AmrCore/AMReX_AmrParticles.H
@@ -246,9 +246,9 @@ ParticleToMesh (PC const& pc, const Vector<MultiFab*>& mf,
     for (int lev = lev_min; lev <= lev_max; lev++)
     {
       if (zero_out_input) {
-        mf[lev]->ParallelCopy(mf_part[lev],0,0,mf_part[lev].nComp(),0,0,FabArrayBase::COPY);
-      } else {
-        mf[lev]->ParallelCopy(mf_part[lev],0,0,mf_part[lev].nComp(),0,0,FabArrayBase::ADD);	
+        mf[lev]->ParallelCopy(mf_part[lev],0,0,mf_part[lev].nComp(),0,0);
+      } else { // do not zero out input
+        mf[lev]->ParallelAdd(mf_part[lev],0,0,mf_part[lev].nComp(),0,0);
     }
 }
 
--- a/Src/AmrCore/AMReX_AmrParticles.H
+++ b/Src/AmrCore/AMReX_AmrParticles.H
@@ -249,6 +249,7 @@ ParticleToMesh (PC const& pc, const Vector<MultiFab*>& mf,
         mf[lev]->ParallelCopy(mf_part[lev],0,0,mf_part[lev].nComp(),0,0);
       } else { // do not zero out input
         mf[lev]->ParallelAdd(mf_part[lev],0,0,mf_part[lev].nComp(),0,0);
+      }
     }
 }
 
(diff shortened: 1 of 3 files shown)


BenWibking:
I am still working to clean up the test for this fix.

chongchonghe:
@BenWibking An alternative solution is #4397 . Instead of doing a ParallelAdd in the end, I made `mf_part` a copy of `mf`. Then, the `sum_fine_to_coarse` and `average_down` process is more robust (I think? I don't quite understand how `average_down` works).

BenWibking:
Your solution involves two copies, whereas this PR only requires one copy.

Why do you need to average down the pre existing MultiFab contents? If it is the hydro state (or a copy of it), it should already be averaged down. If it is another particle species deposition, it was also already averaged down.

BenWibking:
Averaging down is a linear operation. The average down applied to mf_A + mf_B is the same (up to round off error) as averaging down them separately, then adding them.

chongchonghe:
> Averaging down is a linear operation. The average down applied to mf_A + mf_B is the same (up to round off error) as averaging down them separately, then adding them.

OK, I agree. Your solution is more efficient.

BenWibking:
There is something odd I don't understand. When I change this from `ParallelCopy` to `ParallelAdd`, I get a different answer, even when the original multifabs are identically zero:
```
➜  ParticleMeshMultiLevel git:(fix-particletomesh-amr-zero_out_input-false) ✗ ./main3d.gnu.MPI.ex inputs
Initializing AMReX (25.03-25-g2fd8109e381f-dirty)...
MPI initialized with 1 MPI processes
AMReX (25.03-25-g2fd8109e381f-dirty) initialized

Number of particles per cell : 10
Size of domain               : 128 128 128
Total number of particles    : 20971520

AMReX (25.03-25-g2fd8109e381f-dirty) finalized
➜  ParticleMeshMultiLevel git:(fix-particletomesh-amr-zero_out_input-false) ✗ ~/amrex/Tools/Plotfile/fcompare.gnu.ex plt00000_v1 plt00000_v0

            variable name            absolute error            relative error
                                        (||A - B||)         (||A - B||/||A||)
 ----------------------------------------------------------------------------
 level = 0
 density                                          0                         0
 level = 1
 density                                          0                         0
 PLOTFILE AGREE
➜  ParticleMeshMultiLevel git:(fix-particletomesh-amr-zero_out_input-false) ✗ ./main3d.gnu.MPI.ex inputs.dont_zero_out_input
Initializing AMReX (25.03-25-g2fd8109e381f-dirty)...
MPI initialized with 1 MPI processes
AMReX (25.03-25-g2fd8109e381f-dirty) initialized

Number of particles per cell : 10
Size of domain               : 128 128 128
Total number of particles    : 20971520

AMReX (25.03-25-g2fd8109e381f-dirty) finalized
➜  ParticleMeshMultiLevel git:(fix-particletomesh-amr-zero_out_input-false) ✗ ~/amrex/Tools/Plotfile/fcompare.gnu.ex plt00000_v1 plt00000_v0

            variable name            absolute error            relative error
                                        (||A - B||)         (||A - B||/||A||)
 ----------------------------------------------------------------------------
 level = 0
 density                                65307500.31              0.1372404896
 level = 1
 density                                          0                         0
```

atmyers (approved):

Issue #4380, opened 2025-03-18
Test using PolymorphicArenaAllocator for ParticleContainer


ax3l:
As discussed: we will need a unit test in AMReX that uses the pure SoA particle container with the `PolymorphicArenaAllocator` 🙏 

Needed for:
- https://github.com/AMReX-Codes/pyamrex/issues/425
- https://github.com/BLAST-WarpX/warpx/issues/5779
- https://github.com/BLAST-ImpactX/impactx/issues/901
- https://github.com/Hi-PACE/hipace/issues/1217

ax3l:
Ping @atmyers can you help with this example? :)

ax3l:
@atmyers @ax3l we should add coverage for the polymorphic particle container in AMReX tests directly

ax3l:
Added via #4859

Issue #3570, opened 2023-09-28
Pure SoA Particles: ParticleTileData Size Optimization


ax3l:
Follow-up marked in #3290 by @AlexanderSinn:

The `ParticleTileData` struct that is used to access SoA data can be very large, over 300 bytes in some cases. I recommend changing the array-of-pointers to a full 2D array. This way `ParticleTileData` could be as small as 24 bytes (real pointer, int pointer, stride).

@atmyers what do you think?

ax3l:
@AlexanderSinn do you like to post a draft PR for this change?

asalmgren:
Closing this issue due to lack of activity.  Please re-open it if you feel there is is still an issue!

PR #4404, opened 2025-04-07
Add version of ParticleTile using runtime-only 2D array

AlexanderSinn:
## Summary

This PR adds a simpler version of ParticleTile #3570 

The new tile only has runtime components and always has a polymorphic arena allocator as described in #4380. To avoid the array of pointer style of the current particle tile, here a 2D array over particle ids and component ids is used. As a simplification both AoS and SoA members have been removed as well as all the push_back functions.

Performance test with HiPACE++:

dev, ParticleContainerPureSoA<11, 1>
```
TinyProfiler total time across processes [min...avg...max]: 21.44 ... 21.44 ... 21.44
TinyProfiler total time across processes [min...avg...max]: 21.42 ... 21.42 ... 21.42
TinyProfiler total time across processes [min...avg...max]: 21.4 ... 21.4 ... 21.4
TinyProfiler total time across processes [min...avg...max]: 21.43 ... 21.43 ... 21.43
TinyProfiler total time across processes [min...avg...max]: 21.39 ... 21.39 ... 21.39
```
ParticleContainerPureSoA<0, 0> using only runtime components
```
TinyProfiler total time across processes [min...avg...max]: 21.93 ... 21.93 ... 21.93
TinyProfiler total time across processes [min...avg...max]: 22.05 ... 22.05 ... 22.05
TinyProfiler total time across processes [min...avg...max]: 21.95 ... 21.95 ... 21.95
TinyProfiler total time across processes [min...avg...max]: 21.96 ... 21.96 ... 21.96
TinyProfiler total time across processes [min...avg...max]: 21.94 ... 21.94 ... 21.94

```
this PR, ParticleContainerPureSoA2<amrex::Real, int>
```
TinyProfiler total time across processes [min...avg...max]: 21.4 ... 21.4 ... 21.4
TinyProfiler total time across processes [min...avg...max]: 21.38 ... 21.38 ... 21.38
TinyProfiler total time across processes [min...avg...max]: 21.4 ... 21.4 ... 21.4
TinyProfiler total time across processes [min...avg...max]: 21.44 ... 21.44 ... 21.44
TinyProfiler total time across processes [min...avg...max]: 21.38 ... 21.38 ... 21.38
```

## Additional background

ParticleReal and int are replaced with template types to make implementing the const versions easier and to allow for more flexibility in the future (use double and float in the same application, use double2 or float4, use 64 bit int etc).

## Checklist

The proposed changes:


atmyers:
Hi @AlexanderSinn - is this still something you're working on? Do you think it would be worthwhile to keep pushing on this?

AlexanderSinn:
Yes. There is still a lot left to do. I am waiting for Axel to return and give feedback first

Src/Particle/AMReX_NeighborParticles.H
```diff
@@ -9,6 +9,7 @@
 #include <AMReX_NeighborList.H>
 #include <AMReX_OpenMP.H>
 #include <AMReX_ParticleTile.H>
+#include <AMReX_ParticleTile2.H>
```
ax3l: How about we call this thing `RuntimeParticleTile` or so, describing that it only holds runtime components?

Src/Particle/AMReX_ParIter.H
```diff
@@ -18,6 +18,9 @@ struct Particle;
 template <int NArrayReal, int NArrayInt>
 struct SoAParticle;
 
+template <class RType, class IType>
+struct SoAParticle2;
```
ax3l: `SOARuntimeParticle`?

Src/Particle/AMReX_ParIter.H
```diff
@@ -87,11 +90,11 @@ public:
...
     [[nodiscard]] SoARef GetStructOfArrays () const { return GetParticleTile().GetStructOfArrays(); }
 
-    [[nodiscard]] int numParticles () const { return GetParticleTile().numParticles(); }
+    [[nodiscard]] auto numParticles () const { return GetParticleTile().numParticles(); }
 
-    [[nodiscard]] int numRealParticles () const { return GetParticleTile().numRealParticles(); }
+    [[nodiscard]] auto numRealParticles () const { return GetParticleTile().numRealParticles(); }
 
-    [[nodiscard]] int numNeighborParticles () const { return GetParticleTile().numNeighborParticles(); }
+    [[nodiscard]] auto numNeighborParticles () const { return GetParticleTile().numNeighborParticles(); }
```
ax3l, line 97: Are there other return values than `int`/`long` that are useful here?

Src/Particle/AMReX_Particle.H
```diff
@@ -366,6 +368,12 @@ struct alignas(sizeof(double)) Particle
 #endif
     }
 
+    [[nodiscard]] AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+    uint64_t& idcpu () & { return this->m_idcpu; }
+
+    [[nodiscard]] AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
```
ax3l, line 452: Good general addition. This part can go in even without this PR (e.g., separate PR while this PR still ripens).

ax3l (commented):
This looks like a great start!

Let's finalize the naming to be more descriptive and implementation :)

Src/Particle/AMReX_ParIter.H
AlexanderSinn: The new ParticleTile uses `std::size_t` as index type for `numParticles()`

ax3l:
@atmyers can you pls prioritize this PR in the coming week(s)? :) :pray:

AlexanderSinn:
This PR depends on #4529 btw

AlexanderSinn:
I think now this is ready for review. It does lack a bit in documentation and is not compatible with everything yet, but the same is true for regular PureSoA.

Src/Particle/AMReX_ParticleTransformation.H
```diff
@@ -146,6 +147,60 @@ void swapParticle (const ParticleTileData<T_ParticleType, NAR, NAI>& dst,
...
+ */
+template <class DRType, class DIType, class SRType, class SIType>
+AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+void copyParticle (const ParticleTileDataRT<DRType, DIType>& dst,
+                   const ParticleTileDataRT<SRType, SIType>& src,
+                   typename ParticleTileDataRT<SRType, SIType>::size_type src_i,
+                   typename ParticleTileDataRT<DRType, DIType>::size_type dst_i) noexcept
+{
+    dst.idcpu(dst_i) = src.idcpu(src_i);
+
```
atmyers, line 170: Right now this routine and the one below silently truncate components if they don't match. Should this assert that `dst.m_n_real` and 'src.m_n_real` are the same, and the same for the ints?

Src/Particle/AMReX_ParIter.H
atmyers: I'm concerned that changing this to an unsigned type will lead to a bunch of mismatched comparisons in downstream code if we adopt this in e.g. WarpX.

Src/Particle/AMReX_ParticleTileRT.H
```diff
@@ -0,0 +1,828 @@
...
+    }
+
+    [[nodiscard]] AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+    explicit operator T* () const {
+        return m_data;
+    }
+};
+
+template <class T>
+ArrayView(T* data, std::size_t capacity) -> ArrayView<T>;
```
atmyers, line 67: Could this reuse `ArrayND`?

Src/Particle/AMReX_ParticleTransformation.H
AlexanderSinn: I addend an assert as the copy and swap functions for AoS/PureSoA also have the same assert

Src/Particle/AMReX_ParticleTileRT.H
AlexanderSinn: I switched the index type to amrex::Long now, lets see if that works.

ArrayView is needed for code that expects a PODVector to still work if operator[], data(), dataPtr(), begin() and end() are used. ArrayND<1> does not have all of those functions and could have overflow issues as it uses int as an index type.

atmyers (approved):

Src/Particle/AMReX_WriteBinaryParticleData.H
```diff
@@ -238,80 +238,98 @@ void packParticleIDs (I* idata, const std::uint64_t idcpu, bool is_checkpoint) n
...
+        for (int j = 0; j < ptd.m_num_runtime_int; ++j) {
+            if (write_int_comp[PTD::ParticleType::NInt + PTD::NAI + j]) {
+                idata_ptr[iout_index] = ptd.m_runtime_idata[j][idx];
+                iout_index++;
+            }
+        }
+    } else {
+        for (int j = 0; j < ptd.m_n_int; ++j) {
+            if (write_int_comp[j]) {
+                idata_ptr[iout_index] = ptd.rdata(j)[idx];
```
atmyers:
```suggestion
                idata_ptr[iout_index] = ptd.idata(j)[idx];
```

atmyers (changes requested):

atmyers:
Maybe you should also add an assertion to Redistribute that at least `AMReX_SPACEDIM` runtime real components have been added, since that function does assume they are there.

Src/Particle/AMReX_WriteBinaryParticleData.H
AlexanderSinn: Good catch!

Src/Particle/AMReX_ParIter.H
ax3l: With the other changes below: should this just be `Long`?

atmyers (approved):

PR #4405, opened 2025-04-07
Periodicity boundary support for multi-ghost communication

### LINKED ISSUE #4045: Periodicity with multiple layers of ghost nodes
Right now periodicity is not possible if there are two or more ghost nodes (i.e. if `FabArrayBase::m_multi_ghost` is true) - per

https://github.com/AMReX-Codes/amrex/blob/930e75ea9d02a46278a802ea886d950f4c8d1a90/Src/Base/AMReX_FabArrayBase.cpp#L879

Is it possible to update `FabArrayBase` to allow for periodicity with multiple layers of ghost nodes?

WeiqunZhang:
Close #4045

### DIFF
--- a/Src/Base/AMReX_BoxArray.H
+++ b/Src/Base/AMReX_BoxArray.H
@@ -810,6 +810,9 @@ public:
     [[nodiscard]] BoxList complementIn (const Box& b) const;
     void complementIn (BoxList& bl, const Box& b) const;
 
+    //! Return box - periodic boxarray
+    [[nodiscard]] BoxList complementIn (const Box& b, const Periodicity& period) const;
+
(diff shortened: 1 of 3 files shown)


WeiqunZhang:
@bsrunnels Could you give this PR a try?

bsrunnels (approved):
@WeiqunZhang checked it out and am now able to work with periodic boundaries with multiple ghost nodes.

(Note: I cannot verify yet that everything is working properly. I think there may still be some lingering issues. Specifically, I've identified some seemingly incorrect behavior with ParallelCopy and FillBoundary. However those are separate issues; I'll submit separate issues with MWEs for those.)

Thanks!

asalmgren (approved):

PR #4407, opened 2025-04-08
Fix a bug in 1d Curl Curl Solver

WeiqunZhang:
Ex is a cell centered variable. So we need to skip it at the higher end of the nodal domain. Note that it's okay for interior ghost cells, because we do have valid data there.


atmyers (approved):

RevathiJambunathan:
Thanks for the bug fix! @WeiqunZhang

PR #4408, opened 2025-04-08
HIP Debug Build

### LINKED ISSUE #4387: Build Option for Flags_FPE
@guj reports that she sees ROCm/HIP build issues on Frontier (OLCF) building WarpX in Debug mode, which is caused by `Flags_FPE`, in particular `-ftrapv` that we add by default in debug builds.

We might need to add an `-DAMReX_` option to control this, so we can set non-defaults and debug on HIP.

WeiqunZhang:
Disable fpe setting and make sure -O1 is use.

No inline for index_assert functions.

### DIFF
--- a/Src/Base/AMReX_Array.H
+++ b/Src/Base/AMReX_Array.H
@@ -148,7 +148,11 @@ namespace amrex {
         }
 
 #if defined(AMREX_DEBUG) || defined(AMREX_BOUND_CHECK)
+#if defined(AMREX_USE_HIP)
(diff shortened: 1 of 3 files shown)


BenWibking:
I tried this, but it's been linking Quokka for > 1 hour now... maybe we have too many templates.

WeiqunZhang:
I tested Castro. The link took with 4 seconds this PR. With the development branch, it has not finished linking after 10 minutes.

asalmgren (approved):

PR #4417, opened 2025-04-13
AMREX_GPU_MAX_THREADS: 256 -> 128 for CUDA

WeiqunZhang:
Except for reduction, change the default number of threads per block from 256 to 128 for CUDA.


ax3l:
Just a quick check: why? :)

WeiqunZhang:
It appears that 128 is better than 256 for some cases. So I opened this for testing.

PR #4420, opened 2025-04-15
Refactor Array4: Compute strides on the fly

WeiqunZhang:
This reduces the size of Array4 from 64 to 40 bytes.

### DIFF
--- a/Src/Base/AMReX_Array4.H
+++ b/Src/Base/AMReX_Array4.H
@@ -126,7 +126,7 @@ namespace amrex {
 #if defined(AMREX_DEBUG) || defined(AMREX_BOUND_CHECK)
             index_assert(i,j,k,0);
 #endif
-            return p[(i-begin.x)+(j-begin.y)*jstride()+(k-begin.z)*kstride()];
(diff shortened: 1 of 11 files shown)


Src/Base/AMReX_Array4.H
```diff
@@ -141,7 +135,7 @@ namespace amrex {
 #if defined(AMREX_DEBUG) || defined(AMREX_BOUND_CHECK)
             index_assert(i,j,k,n);
 #endif
-            return p[(i-begin.x)+(j-begin.y)*jstride+(k-begin.z)*kstride+n*nstride];
+            return p[(i-begin.x)+(j-begin.y)*jstride()+(k-begin.z)*kstride()+n*nstride()];
```
AlexanderSinn: maybe try
```suggestion
            return p[(i-begin.x)+Long(len.x)*((j-begin.y)+Long(len.y)*((k-begin.z)+Long(len.z)*n))];
```

zingale:
did this make any performance difference?

WeiqunZhang:
Not much difference, except for Castro. :) The Castro test crashed on perlmutter and I never figured out why.

PR #4425, opened 2025-04-20
Parser: Support User defined functions

WeiqunZhang:
The user defined functions can have up to 4 arguments. If more is needed, we can support it in the future. Getting device function pointers requires GPU relocatable device code (RDC) support. Moreover, these user defined device functions should be defined in cpp files not headers to ensure the function pointers are unique.

For now, we only support CPU, CUDA, and HIP. In the future, we can try to support SYCL.


WeiqunZhang:
Depend on #4442.

atmyers (approved):

PR #4427, opened 2025-04-22
Parser Optimization: Add ternary operators,  square, cubic and hypot

WeiqunZhang:


WeiqunZhang:
This helps performance on CPU but it's bad for GPU.

PR #4430, opened 2025-04-26
Convert RealVect to RealVectND

AlexanderSinn:
## Summary

Similar to #3969, #3988 and #4016 but for RealVect.

## Checklist

The proposed changes:


WeiqunZhang:
This would require changes in pyamrex. Could you also prepare a PR for pyamrex so that we can merge them at the same time.

WeiqunZhang (approved):

PR #4441, opened 2025-05-02
add conservative linear interpolation for spherical 2d geometry

zhichen3:
## Summary
this adds conservative linear interpolation for spherical 2d geometry.
It calculates `xoff` and `yoff` in "volume" space. For xoff, its the same as spherical 1d. For yoff, it uses cosine as weighting factor, since cell volume in spherical geometry $\propto (\cos(\theta_r) - \cos(\theta_l)) (r_r^3 - r_l^3)$
## Checklist

The proposed changes:

### DIFF
--- a/Src/AmrCore/AMReX_Interpolater.cpp
+++ b/Src/AmrCore/AMReX_Interpolater.cpp
@@ -868,7 +868,33 @@ CellConservativeLinear::interp (const FArrayBox& crse,
         });
     } else
 #elif (AMREX_SPACEDIM == 2)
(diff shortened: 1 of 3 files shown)


zhichen3:
[sph_interpolation.pdf](https://github.com/user-attachments/files/20007833/sph_interpolation.pdf)

I attached my notes on deriving `xoff` and `yoff`. Also wrote out the case where we have 2x2 refinement and it shows that this is conservative.

WeiqunZhang (approved):

PR #4449, opened 2025-05-06
CMake: NVTX is header-only for CUDA >= 11.2

BenWibking:
## Summary
The old NVTX library is removed in CUDA 12.9, so we have to remove it as a link-time dependecy. AMReX does not build with CUDA 12.9 otherwise.

Note that the new NVTXv3 is header-only and is already supported by AMReX.

## Checklist

The proposed changes:


WeiqunZhang:
But we still need that for CUDA < 11.2. Should we make it conditional?

WeiqunZhang:
> But we still need that for CUDA < 11.2. Should we make it conditional?

Do you want me to help?

BenWibking:
> But we still need that for CUDA < 11.2. Should we make it conditional?

@WeiqunZhang Ok, I've fixed it.

WeiqunZhang (approved):

PR #4451, opened 2025-05-07
CArena free unused memory before allocate

AlexanderSinn:
## Summary

In specific special cases, an application may ask for chunks of memory with slightly increasing size. If these do not fit into the existing space of the arena, a new chunk of system memory has to be allocated every time. The old chunks that are slightly too small to be reused could accumulate and consume a lot of memory and would never get freed with the default release threshold of max long. This PR adds a check to free unused system memory before allocating new memory, while still keeping the total amount of memory the same to avoid excessive system allocations and frees.

## Additional background

The fix from #4044 is reimplemented differently in the PR so that freeUnused_protected can be called.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_CArena.cpp
+++ b/Src/Base/AMReX_CArena.cpp
@@ -376,9 +376,9 @@ CArena::freeUnused_protected ()
 
     // deallocate_system can call cudafree which may perform implicit synchronization
     // of all cuda streams. In case amrex::Gpu::Elixir is used, a cudaLaunchHostFunc can be
-    // in the steam which calls CArena::free that aquires carena_mutex.
+    // in the steam which calls CArena::free that acquires carena_mutex.
     // So here carena_mutex needs to be unlocked first to avoid a deadlock.
-    // Note that other threads to allocate/free memory from the CArena in the mean time.
+    // Note that other threads to allocate/free memory from the CArena in the meantime.
     carena_mutex.unlock();
     for (auto& a : to_free) {
         deallocate_system(a.first, a.second);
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
A minor cleanup. We don't need to include MFIter.H anymore.

WeiqunZhang (approved):

WeiqunZhang:
@AlexanderSinn This seems to cause performance regression for Castro on Frontier. A test's run time went up from 30 to 36.

AlexanderSinn:
Interesting that it affects performance at all. Does Castro use OMP and HIP at the same time?

WeiqunZhang:
No OMP.

WeiqunZhang:
But it uses a lot of Elxirs.

WeiqunZhang:
I think need to avoid calling freeUnused protected inside mfiter.

WeiqunZhang:
If we do it, will hipace++ have the memory issue again?

AlexanderSinn:
So I guess it's periodically above the release threshold but it is actually better if it dosen't release the memory; otherwise, it keeps allocating and freeing hip memory. Does the performance fix itself if the release threshold is set to 64 GiB? Im not sure if hipace only allocates within mfiter or also outside.

Src/Base/AMReX_CArena.cpp
```diff
@@ -44,11 +43,7 @@ CArena::alloc_protected (std::size_t nbytes)
     }
 #endif
 
-    if (static_cast<Long>(m_used+nbytes) >= arena_info.release_threshold
-#ifdef AMREX_USE_GPU
-        && (MFIter::currentDepth() == 0)
-#endif
-        ) {
```
AlexanderSinn, line 51: maybe just adding this back would fix it

PR #4462, opened 2025-05-14
Communication Optimization: Cache GPU kernel metadata

WeiqunZhang:


WeiqunZhang:
If this works well, we will extend this to MPI buffer packing and unpacking.

asalmgren (approved):

WeiqunZhang:
Will merge after #4486.

PR #4468, opened 2025-05-15
Print max CArena usage

AlexanderSinn:
## Summary

The CArena::PrintUsage() called in amrex finalize prints the currently used and allocated memory. This isn't very useful, as all memory should already be freed by that time. This PR prints the per-CArena maximum memory usage and allocation instead (but without the initialize allocation).

## Checklist

The proposed changes:


Src/Base/AMReX_Arena.H
```diff
@@ -188,7 +188,7 @@ public:
     static std::size_t align (std::size_t sz);
 
     static void Initialize (bool minimal);
-    static void PrintUsage ();
+    static void PrintUsage (bool print_max_usage);
```
WeiqunZhang: We might want to make this argument optional. Some codes like Nyx are using it.

WeiqunZhang (approved):

WeiqunZhang (changes requested):
LGTM other than PrintUsage

WeiqunZhang (approved):

PR #4478, opened 2025-05-21
Avoid freeing GPU memory inside MFIter loop

WeiqunZhang:
This reverts part of the changes in #4451, which caused performance regression for some applications on Frontier. For example, a Castro test had a 20% performance hit.

### DIFF
--- a/Src/Base/AMReX_CArena.cpp
+++ b/Src/Base/AMReX_CArena.cpp
@@ -2,6 +2,7 @@
 #include <AMReX_CArena.H>
 #include <AMReX_BLassert.H>
 #include <AMReX_Gpu.H>
+#include <AMReX_MFIter.H>
 #include <AMReX_ParallelReduce.H>
 
 #include <utility>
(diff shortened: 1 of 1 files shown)


Src/Base/AMReX_CArena.cpp
```diff
@@ -369,25 +382,13 @@ CArena::freeUnused_protected ()
                                      {
                                          it = m_freelist.erase(it);
                                          nbytes += a.second;
-                                         to_free.emplace_back(a.first, a.second);
+                                         deallocate_system(a.first,a.second);
```
AlexanderSinn, line 385: I think this could deadlock if an Elixir were to be used outside an MFIter

WeiqunZhang: In principle, yes. But we do not use multiple streams outside MFIter.

AlexanderSinn: Even with a single stream I suspect it is possible
```C++
{
    Elixir eli(...);
    ParallelFor(...);
}
... = carena->alloc(...); // triggers freeUnused_protected -> deallocate_system ->
// cudaFree -> deviceSync -> elixir free -> carena free -> needs carena mutex, deadlock
```

WeiqunZhang:
Unfortunately this does not fix the performance issue. Looks like the real issue is the defragmentation.

WeiqunZhang:
Maybe we need to make defragmentation an opt-in feature. #4479

PR #4480, opened 2025-05-22
Reuse MLEBNodeFDLaplacian and MLCurlCurl

WeiqunZhang:
Allow these two linear operators to be reused.


ajnonaka (approved):
Thanks Weiqun!

PR #4490, opened 2025-06-03
Add an --output_dir option to makebuildinfo_C.py

yut23:
## Summary

This allows specifying the directory where `AMReX_buildInfo.cpp` is written (instead of the current directory), which I needed to avoid a race condition with simultaneous builds from a single source directory.

## Checklist

The proposed changes:


zingale (approved):

WeiqunZhang:
Not sure if we should merge this right now or wait for #4486.

zingale:
It can wait

PR #4493, opened 2025-06-04
Delay gpu stream sync in MFIter

WeiqunZhang:
Previously we performed gpu stream sync during MFIter::Initialize in case that kernels on non-default streams inside MFIter loop might depend on the result of kernels on the default amrex stream launched before the MFIter loop. Now we delay the sync if needed until the end of the first iteration.

### DIFF
--- a/Src/Base/AMReX_MFIter.cpp
+++ b/Src/Base/AMReX_MFIter.cpp
@@ -282,11 +282,12 @@ MFIter::Initialize ()
             "Nested or multiple active MFIters is not supported by default.  This can be changed by calling MFIter::allowMultipleMFIters(true)".);
     }
 
-#ifdef AMREX_USE_GPU
-    if (device_sync) {
-#ifdef AMREX_USE_OMP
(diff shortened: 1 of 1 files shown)


Src/Base/AMReX_MFIter.cpp
```diff
@@ -534,6 +535,17 @@ MFIter::operator++ () noexcept
 
 #ifdef AMREX_USE_GPU
         if (Gpu::inLaunchRegion()) {
+            if (device_sync && (streams > 1) && (OpenMP::get_num_threads() == 1)
+                && (currentIndex == 1) && isValid())
+            {
+                // Because omp num threads is 1, gpu stream sync has not
+                // been called in Initialize. We need to sync stream 0
+                // before launching kerenles on stream 1, because the user
```
atmyers:
```suggestion
                // before launching kernels on stream 1, because the user
```

atmyers (approved):

PR #4495, opened 2025-06-05
Added option to use either always use BNorm or ResNorm for calculatin…

ankithadas:
…g target convergence criteria in MLMG. The default behaviour is using Greater of the two Norms.

## Checklist

The proposed changes:

### DIFF
--- a/Docs/sphinx_documentation/source/LinearSolvers.rst
+++ b/Docs/sphinx_documentation/source/LinearSolvers.rst
@@ -159,13 +159,20 @@ to solve the problem given an initial guess and a right-hand side.
 Zero is a perfectly fine initial guess.  The two :cpp:`Reals` in the argument
 list are the targeted relative and absolute error tolerances. The relative
 error tolerance is hard-coded to be at least :math:`10^{-16}`.
-Given the linear system :math:`Ax=b`, the solver will terminate when the
-max-norm of the residual (:math:`b-Ax`) is less than
-:cpp:`std::max(a_tol_abs, a_tol_rel*max_norm)` where :cpp:`max_norm`
-is the max-norm of the rhs, :math:`b`, if the flag :cpp:`always_use_bnorm` is
-set to True or if the rhs max-norm is greater than or equal to the max-norm error
-of the initial guess, otherwise :cpp:`max_norm` is equal to the max-norm error
-of the initial guess.  Set the absolute tolerance to zero if one does not have a
+Given the linear system :math:`Ax = b`, the solver will terminate when the
+max-norm of the residual (:math:`b - Ax`) is less than
+:cpp:`std::max(a_tol_abs, a_tol_rel * max_norm)` where :cpp:`max_norm` is
+equal to the greater of rhs max-norm and residual max-norm. This behavior can
+be modified using the ``MLMG`` member function :cpp:`setConvergenceNormType (NormType norm)`
+where the available options are
+
+- :cpp:`MLMG::NormType::Greater`: The default.
+
+- :cpp:`MLMG::NormType::BNorm`: :cpp:`max_norm` is set to rhs max-norm.
+
+- :cpp:`MLMG::NormType::ResNorm`: :cpp:`max_norm` is set to residual max-norm.
+
+Set the absolute tolerance to zero if one does not have a
 good value for it.  The return value of :cpp:`solve` is the max-norm error.
 
 After the solver returns successfully, if needed, we can call
--- a/Docs/sphinx_documentation/source/LinearSolvers.rst
+++ b/Docs/sphinx_documentation/source/LinearSolvers.rst
@@ -161,16 +161,16 @@ list are the targeted relative and absolute error tolerances. The relative
 error tolerance is hard-coded to be at least :math:`10^{-16}`.
 Given the linear system :math:`Ax = b`, the solver will terminate when the
 max-norm of the residual (:math:`b - Ax`) is less than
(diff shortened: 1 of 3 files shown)


Src/LinearSolvers/MLMG/AMReX_MLMG.H
```diff
@@ -32,6 +32,9 @@ public:
     using BottomSolver = amrex::BottomSolver;
     enum class CFStrategy : int {none,ghostnodes};
 
+    // Norm used to evaluate the target convergence criteria.
+    enum class NormType: int {Greater, BNorm, ResNorm};
```
WeiqunZhang: We could use AMREX_ENUM ( https://amrex-codes.github.io/amrex/docs_html/Basics.html?highlight=amrex_enum#enum-class) Then we can get the string name easily.

Src/LinearSolvers/MLMG/AMReX_MLMG.H
```diff
@@ -358,6 +362,28 @@ MLMGT<MF>::MLMGT (MLLinOpT<MF>& a_lp)
...
 template <typename MF> MLMGT<MF>::~MLMGT () = default;
 
+template <typename MF>
+void
+MLMGT<MF>::setAlwaysUseBNorm (int flag) noexcept
+{
+    if (flag) {
+        norm_type = NormType::BNorm;
+    } else {
+        norm_type = NormType::Greater;
```
WeiqunZhang: If I call `setAlwaysUseBNorm(true)` followed by `setAlwaysUseResNorm(false);`, the norm type will be set to Greater. That seems counterintuitive.

Src/LinearSolvers/MLMG/AMReX_MLMG.H
```diff
@@ -2006,6 +2045,19 @@ MLMGT<MF>::checkPoint (const Vector<MultiFab*>& a_sol,
 
         HeaderFile.precision(17);
 
+        std::string norm_name;
```
WeiqunZhang: If we use AMREX_ENUM, this can be `auto norm_name = amrex::getEnumNameString(norm_type);`.

WeiqunZhang (changes requested):
Could you update https://amrex-codes.github.io/amrex/docs_html/LinearSolvers.html#?

Src/LinearSolvers/MLMG/AMReX_MLMG.H
ankithadas: I believe AMREX_ENUM functions won't work as intended inside a class as the `CLASS##_EnumTraits` would be inside `MLMG` scope.

Src/LinearSolvers/MLMG/AMReX_MLMG.H
ankithadas: Changed function to `setConvergenceNormType`. I have kept `setAlwaysUseBNorm(flag)` for backwards compatibility

WeiqunZhang: My main point is the else block should be removed. Consider this.
```
mlmg.setConvergenceNormType(MLMG::NormType::ResNorm);
mlmg.setAlwaysUseBNorm(false);
```
What do people expect?

Src/LinearSolvers/MLMG/AMReX_MLMG.H
WeiqunZhang: That's right. On the other hand, it might not be a bad idea to move the enum class out of MLMG and call it MLMGNormType. The issue is MLMG is a specialization of MLMGT<MultiFab>. So `mlmg.setConvergenceNormType(MLMG::NormType)` does not always work.

WeiqunZhang: Using AMREX_NUM will also make it easier if the user want to make the norm type a runtime parameter (e.g., mlmg_norm_type = ResNorm in inputs).

Src/LinearSolvers/MLMG/AMReX_MLMG.H
ankithadas: I would say that's an improper use of the two functions. Users should only use one of the functions. The else block in `setAlwaysUseBNorm()` was kept to match the previous behaviour of the function. However, if breaking changes are okay, I can remove `setAlwaysUseBNorm()`.

Docs/sphinx_documentation/source/LinearSolvers.rst
```diff
@@ -159,13 +159,20 @@ to solve the problem given an initial guess and a right-hand side.
...
-of the initial guess, otherwise :cpp:`max_norm` is equal to the max-norm error
-of the initial guess.  Set the absolute tolerance to zero if one does not have a
+Given the linear system :math:`Ax = b`, the solver will terminate when the
+max-norm of the residual (:math:`b - Ax`) is less than
+:cpp:`std::max(a_tol_abs, a_tol_rel * max_norm)` where :cpp:`max_norm` is
+equal to the greater of rhs max-norm and residual max-norm. This behavior can
+be modified using the ``MLMG`` member function :cpp:`setConvergenceNormType (NormType norm)`
+where the available options are
+
+- :cpp:`MLMG::NormType::Greater`: The default.
```
WeiqunZhang: We need to update these three lines.

WeiqunZhang (changes requested):

Docs/sphinx_documentation/source/LinearSolvers.rst
ankithadas: Oh sorry missed that. Updated.

WeiqunZhang (approved):

PR #4500, opened 2025-06-11
Replaced depreciated cub::Max, cub::Min cub::Sum with cuda::maximum<>…

ankithadas:
… cuda::minimum<>, cuda::std::plus<>

X-Ref: https://nvidia.github.io/cccl/cccl/3.0_migration_guide.html


WeiqunZhang:
I think we need to use `#if defined(CCCL_MAJOR_VERSION) && (CCCL_MAJOR_VERSION >= 3)` for the new functions.

WeiqunZhang (approved):

PR #4504, opened 2025-06-13
Incorrect kernel used in amrex::sum_fine_to_coarse()

ankithadas:
## Summary
The previous implementation resulted `amrex::average_down()` and `amrex::sum_fine_to_coarse()` to yield the same result. 

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_MultiFabUtil.cpp
+++ b/Src/Base/AMReX_MultiFabUtil.cpp
@@ -413,7 +413,7 @@ namespace amrex
             ParallelFor(crse_S_fine, nGrow, ncomp,
             [=] AMREX_GPU_DEVICE (int box_no, int i, int j, int k, int n) noexcept
             {
-                amrex_avgdown(i,j,k,n,crsema[box_no],finema[box_no],0,scomp,ratio);
+                amrex_sum_fine_to_crse(i,j,k,n,crsema[box_no],finema[box_no],0,scomp,ratio);
             });
             if (!Gpu::inNoSyncRegion()) {
                 Gpu::streamSynchronize();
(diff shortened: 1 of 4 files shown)


WeiqunZhang:
The function was originally added for particle codes. Note that it's not exactly the same as average_down, because how it treats ghost cells and the communication uses add instead of copy. I tend to agree with you that the behavior seems incorrect. But it might also be possible that it is the expected behavior to the original code that used it.

atmyers:
The original intention for these routines was inside particle-mesh interpolation near coarse-fine boundaries. The volume fraction was included to conserve mass since it changes depending on what level you're talking about. It may be the case that the function name is confusing or that this limits the general usefulness of these routines. But as is, I think will break NyX, PeleC, etc...

WeiqunZhang:
In some sense, the function is sum_fine_ghost_cells_to_coarse_valid_cells_and_dont_worry_about_coarse_valid_cell_covered_by_fine.

WeiqunZhang:
Maybe we should add some comments to the function to avoid potential confusion. The data are already density like. So we need to use average down to obtain the coarse value. But it also involves sum to add the average-downed density across MPI processes.

ankithadas:
Some documentation should definitely help. Potentially a function rename would be great.  Could we potentially move this function to AMR_Particle files ?

WeiqunZhang:
I would avoid breaking users' code. It's also not wrong to call it sum because it does involve sum of ghost cell values from different fine boxes.

ankithadas:
That's fair. In that case, I will close this PR and open a new one with comments added to the function to indicate its use-case.

PR #4509, opened 2025-06-17
Adding a new overload for MultiFab::norm2 (int comp, int numcomp)

siddanib:
## Summary
Adding an optional argument numcomp=1 to MultiFab::norm2 (int comp=0, int numcomp=1) enables getting norm2 over multiple components of a MultiFab.


WeiqunZhang:
If we want to avoid breaking pyamrex, we can add a new overload with both arguments being required instead of modifying the existing one, which can then call the new version.

WeiqunZhang (approved):

PR #4514, opened 2025-06-18
hypre with hip GNUmake improvements

baperry2:
## Summary

When compiling with HIP and HYPRE, use the `rocblas` and `rocsolver` libraries (similar to additional libraries for CUDA + HYPRE) so these don't have to be turned off when building HYPRE. Also, check in `lib64` for the HYPRE library, which is where it gets put when building HYPRE with cmake.

These modifications helped with building PeleLMeX with HYPRE support on Frontier.

## Checklist

The proposed changes:

### DIFF
--- a/Tools/GNUMake/packages/Make.hypre
+++ b/Tools/GNUMake/packages/Make.hypre
@@ -15,7 +15,7 @@ ifdef AMREX_HYPRE_HOME
   HYPRE_ABSPATH = $(abspath $(AMREX_HYPRE_HOME))
   SYSTEM_INCLUDE_LOCATIONS += $(HYPRE_ABSPATH)/include
   LIBRARY_LOCATIONS += $(HYPRE_ABSPATH)/lib $(HYPRE_ABSPATH)/lib64
-  LIBRARIES += -Wl,-rpath,$(HYPRE_ABSPATH)/lib -lHYPRE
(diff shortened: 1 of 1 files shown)


Tools/GNUMake/packages/Make.hypre
```diff
@@ -14,10 +14,15 @@ endif
 ifdef AMREX_HYPRE_HOME
   HYPRE_ABSPATH = $(abspath $(AMREX_HYPRE_HOME))
   SYSTEM_INCLUDE_LOCATIONS += $(HYPRE_ABSPATH)/include
-  LIBRARY_LOCATIONS += $(HYPRE_ABSPATH)/lib
+  LIBRARY_LOCATIONS += $(HYPRE_ABSPATH)/lib $(HYPRE_ABSPATH)/lib64
   LIBRARIES += -Wl,-rpath,$(HYPRE_ABSPATH)/lib -lHYPRE
```
WeiqunZhang: Should we also modify line 18 as well?

baperry2: Maybe? This worked for me without modifying line 18, but I don't really know much about how this stuff works so I'll add that as well if you think we should.

WeiqunZhang: With rpath, the linker will embed the library path into the executable so that we don't need to rely on `LD_LIBRARY_PATH` at run time.

baperry2: Makes sense, I added this modification.

WeiqunZhang (approved):

PR #4520, opened 2025-06-23
AMReX SIMD Helpers

ax3l:
## Summary

AMReX does not have a concept yet to help users write effective SIMD code on CPU, besides relying on auto-vectorization and pragmas, which are unreliable for any complex enough code. [1]

Lucky enough, C++ `std::datapar` was just accepted into C++26, which gives an easy in to write portable SIMD/scalar code. Yet, I did not find a compiler/stdlib yet with support for it, so I finally had play with the C++17 `<experimental/simd>` headers, which are not as complete as C++26 but a good in, especially if complemented with the https://github.com/mattkretz/vir-simd library.

This PR adds initial support for portable user-code by providing:
- build system support: `AMReX_SIMD` (default is OFF), relying on [vir-simd](https://github.com/mattkretz/vir-simd)
- an `AMReX_SIMD.H` header that handles includes & helper types
- `ParallelForSIMD<SIMD_WIDTH>(...)`

## Additional background

[1] Fun fact one: As written in the [story behind Intel's iscp compiler](https://pharr.org/matt/blog/2018/04/18/ispc-origins) and credited to [Tim Foley](http://graphics.stanford.edu/~tfoley/), *auto-vectorization is not a programming model.*

Fun fact two: This is as ad-hoc as the implementation for [data parallel types / SIMD in Kokkos](https://kokkos.org/kokkos-core-wiki/API/simd/simd.html), it seems.

## User-Code Examples & Benchmark

[Please see this ImpactX PR for details.](https://github.com/BLAST-ImpactX/impactx/pull/1002)

## Checklist

- [x] clean up commits (separate commits)
- [x] finalize fallbacks & CI checks
- [ ] add a `vir::stdx::simd` test in CI
- [x] CMake
- [ ] GnuMake
- [x] `AMReX_SIMD.H` 
- [x] `ParallelForSIMD` 
- [x] `ParticleIdWrapper::make_(in)valid(mask)`
- [x] clean up `sincos` support
- [x] `SmallMatrix`
- [x] Support for `GpuComplex` (minimal)
- [x] Support [passing WIDTH as compile-time meta-data](https://godbolt.org/z/7455hqrEc) to callee in `ParallelForSIMD`
- [x] add `vir::stdx::simd` in package managers:
  - [x] Spack [vir-simd](https://github.com/spack/spack-packages/pull/332)
  - [x] Conda [vir-simd](https://github.com/conda-forge/staged-recipes/pull/30377)

## Future Ideas / PRs

- allocate particle arrays aligned so we can use [stdx::vector_aligned](https://en.cppreference.com/w/cpp/experimental/simd/vector_aligned.html) (for [copies](https://en.cppreference.com/w/cpp/experimental/simd/simd/copy_from) into/out of vector registers - note: makes no difference anymore on modern CPUs)
- Support more/all functions in `ParticleIdWrapper`/`ParticleCpuWrapper`
- Support for [vir::simdize<std::complex<T>>](https://github.com/mattkretz/vir-simd/issues/42) instead of `GpuComplex<SIMD>`
- `ParallelFor` ND support
- `ParallelFor`/`ParallelForSIMD`: one could, maybe, with enable-if magic, etc fuse them into a single name again
- CMake superbuild: `vir-simd` auto-download for convenience (opt-out)
- Build system: "SIMD provider" selection, once we can opt-in to a C++26 compiler+stdlib instead of C++17 TS2 + vir-simd
- Update AMReX package in package management:
  - Spack [vir-simd](https://github.com/spack/spack-packages/pull/332)
  - Conda [vir-simd](https://github.com/conda-forge/staged-recipes/pull/30377)

### DIFF
--- a/Src/Base/AMReX_GpuComplex.H
+++ b/Src/Base/AMReX_GpuComplex.H
@@ -53,8 +53,9 @@ struct alignas(2*sizeof(T)) GpuComplex
     /**
      * \brief Add a real number to this complex number.
      */
+    template <typename U>
     AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
-    GpuComplex<T>& operator+= (const T& a_t) noexcept
+    GpuComplex<T>& operator+= (const U& a_t) noexcept
     {
         m_real += a_t;
         return *this;
@@ -63,8 +64,9 @@ struct alignas(2*sizeof(T)) GpuComplex
    /**
      * \brief Subtract a real number from this complex number.
      */
+    template <typename U>
     AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
-    GpuComplex<T>& operator-= (const T& a_t) noexcept
+    GpuComplex<T>& operator-= (const U& a_t) noexcept
     {
         m_real -= a_t;
         return *this;
@@ -73,8 +75,9 @@ struct alignas(2*sizeof(T)) GpuComplex
     /**
      * \brief Multiply this complex number by a real.
      */
+    template <typename U>
     AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
-    GpuComplex<T>& operator*= (const T& a_t) noexcept
+    GpuComplex<T>& operator*= (const U& a_t) noexcept
     {
         m_real *= a_t;
         m_imag *= a_t;
@@ -84,8 +87,9 @@ struct alignas(2*sizeof(T)) GpuComplex
     /**
      * \brief Divide this complex number by a real.
      */
+    template <typename U>
     AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
-    GpuComplex<T>& operator/= (const T& a_t) noexcept
+    GpuComplex<T>& operator/= (const U& a_t) noexcept
     {
         m_real /= a_t;
         m_imag /= a_t;
@@ -247,9 +251,9 @@ GpuComplex<T> operator+ (const T& a_x, const GpuComplex<T>& a_y) noexcept
 /**
  * \brief Multiply two complex numbers.
  */
-template <typename T>
+template <typename T, typename U>
 AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
-GpuComplex<T> operator* (const GpuComplex<T>& a_x, const GpuComplex<T>& a_y) noexcept
+GpuComplex<T> operator* (const GpuComplex<T>& a_x, const GpuComplex<U>& a_y) noexcept
 {
     GpuComplex<T> r = a_x;
     r *= a_y;
@@ -259,9 +263,9 @@ GpuComplex<T> operator* (const GpuComplex<T>& a_x, const GpuComplex<T>& a_y) noe
 /**
  * \brief Multiply a complex number by a real one.
  */
-template <typename T>
+template <typename T, typename U>
 AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
-GpuComplex<T> operator* (const GpuComplex<T>& a_x, const T& a_y) noexcept
+GpuComplex<T> operator* (const GpuComplex<T>& a_x, const U& a_y) noexcept
 {
     GpuComplex<T> r = a_x;
     r *= a_y;
--- a/Src/Base/AMReX_GpuLaunchFunctsC.H
+++ b/Src/Base/AMReX_GpuLaunchFunctsC.H
@@ -4,6 +4,24 @@
 
 namespace amrex {
 
+/** Helper type to store/access the SIMD width in ParallelForSIMD lambdas
+ *
+ * Use instead of int as the running index i. Used to pass the
+ * SIMD WIDTH as compile-time meta-data into a called function/method.
+ *
+ * @tparam WIDTH SIMD width in elements
+ * @tparam N index type (integer)
+ */
+template<int WIDTH, class N=int>
+struct SIMDindex
+{
+    /** SIMD width in elements */
+    static constexpr int width = WIDTH;
+
+    /** The linear loop index of ParallelFor(SIMD) */
+    N index = 0;
+};
+
 namespace detail {
 
     // call_f_scalar_handler
@@ -22,16 +40,6 @@ namespace detail {
         noexcept -> decltype(f(0,Gpu::Handler{}))
     {
         f(i, Gpu::Handler{});
-    }
-
-    // call_f_vector_handler
-
-    template <int SIMD_WIDTH, typename F, typename N>
-    AMREX_FORCE_INLINE
-    auto call_f_vector_handler (F const& f, N i)
-        noexcept -> decltype(f.template operator()<SIMD_WIDTH>(0))
-    {
-        f.template operator()<SIMD_WIDTH>(i);
     }
 
     // call_f_intvect_inner
@@ -190,23 +198,23 @@ void ParallelFor (T n, L&& f) noexcept
  * SIMD load/Write-back operations need to be performed before/after calling this.
  *
  * @tparam WIDTH SIMD width in elements
- * @tparam T index type (integer)
+ * @tparam N index type (integer)
  * @tparam L function/functor to call per SIMD set of elements
  */
-template <int WIDTH, typename T, typename L, typename M=std::enable_if_t<std::is_integral_v<T>> >
+template <int WIDTH, typename N, typename L, typename M=std::enable_if_t<std::is_integral_v<N>> >
 AMREX_ATTRIBUTE_FLATTEN_FOR
-void ParallelForSIMD (T n, L const& f) noexcept
+void ParallelForSIMD (N n, L const& f) noexcept
 {
-    T i = 0;
+    N i = 0;
     // vectorize full lanes
     for (; i + WIDTH <= n; i+=WIDTH) {
-        detail::call_f_vector_handler<WIDTH>(f, i);
+        f(SIMDindex<WIDTH, N>{i});
     }
     // scalar handling of the remainder
     // note: we could make the remainder calls faster, by repeatedly
-    //       calling call_f_vector_handler with WIDTH / 2 until we reach 1
+    //       decreasing the SIMD width by 2 until we reach 1
     for (; i < n; ++i) {
-        detail::call_f_scalar_handler(f, i);
+        f(SIMDindex<1, N>{i});
     }
 }
 
--- a/Src/Base/AMReX_GpuLaunchFunctsC.H
+++ b/Src/Base/AMReX_GpuLaunchFunctsC.H
@@ -24,6 +24,16 @@ namespace detail {
         f(i, Gpu::Handler{});
     }
 
+    // call_f_vector_handler
+
+    template <int SIMD_WIDTH, typename F, typename N>
+    AMREX_FORCE_INLINE
+    auto call_f_vector_handler (F const& f, N i)
+        noexcept -> decltype(f.template operator()<SIMD_WIDTH>(0))
+    {
+        f.template operator()<SIMD_WIDTH>(i);
+    }
+
     // call_f_intvect_inner
 
     template <typename F, std::size_t...Ns, class...Args>
@@ -173,6 +183,31 @@ void ParallelFor (T n, L&& f) noexcept
 {
     amrex::ignore_unused(MT);
     ParallelFor(n, std::forward<L>(f));
+}
+
+/** ParallelFor with a SIMD Width (in elements)
+ *
+ * SIMD load/Write-back operations need to be performed before/after calling this.
+ *
+ * @tparam WIDTH SIMD width in elements
+ * @tparam T index type (integer)
+ * @tparam L function/functor to call per SIMD set of elements
+ */
+template <int WIDTH, typename T, typename L, typename M=std::enable_if_t<std::is_integral_v<T>> >
+AMREX_ATTRIBUTE_FLATTEN_FOR
+void ParallelForSIMD (T n, L const& f) noexcept
+{
+    T i = 0;
+    // vectorize full lanes
+    for (; i + WIDTH <= n; i+=WIDTH) {
+        detail::call_f_vector_handler<WIDTH>(f, i);
+    }
+    // scalar handling of the remainder
+    // note: we could make the remainder calls faster, by repeatedly
+    //       calling call_f_vector_handler with WIDTH / 2 until we reach 1
+    for (; i < n; ++i) {
+        detail::call_f_scalar_handler(f, i);
+    }
 }
 
 template <typename T, typename L, typename M=std::enable_if_t<std::is_integral_v<T>> >
--- a/Src/Base/AMReX_SIMD.H
+++ b/Src/Base/AMReX_SIMD.H
@@ -10,6 +10,8 @@
 #   include <vir/simd.h>  // includes SIMD TS2 header <experimental/simd>
 #endif
 
+#include <type_traits>
+
 
 namespace amrex::simd
 {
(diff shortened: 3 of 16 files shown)


Src/Particle/AMReX_Particle.H
```diff
@@ -2,14 +2,14 @@
...
-#include <AMReX_REAL.H>
-#include <AMReX_FArrayBox.H>
 #include <AMReX_IntVect.H>
-#include <AMReX_RealVect.H>
 #include <AMReX_ParmParse.H>
-#include <AMReX_Geometry.H>
+#include <AMReX_REAL.H>
+#include <AMReX_RealVect.H>
+#include <AMReX_SIMD.H>
+
```
ax3l: Looks like these include on the left are needed.

Src/Base/AMReX_GpuLaunchFunctsC.H
```diff
@@ -24,6 +24,16 @@ namespace detail {
         f(i, Gpu::Handler{});
     }
 
+    // call_f_vector_handler
+
+    template <int SIMD_WIDTH, typename F, typename N>
+    AMREX_FORCE_INLINE
+    auto call_f_vector_handler (F const& f, N i)
+        noexcept -> decltype(f(0))
```
AlexanderSinn:
```suggestion
        noexcept -> decltype(f.template operator()<SIMD_WIDTH>(0))
```
Also there should be a second version of this with Gpu::Handler{} to match the scalar version.

Src/Base/AMReX_GpuLaunchFunctsC.H
```diff
@@ -175,6 +185,31 @@ void ParallelFor (T n, L&& f) noexcept
...
+ *
+ * SIMD load/Write-back operations need to be performed before/after calling this.
+ *
+ * @tparam WIDTH SIMD width in elements
+ * @tparam T index type (integer)
+ * @tparam L function/functor to call per SIMD set of elements
+ */
+template <int WIDTH, typename T, typename L, typename M=std::enable_if_t<std::is_integral_v<T>> >
+AMREX_ATTRIBUTE_FLATTEN_FOR
+void ParallelForSIMD (T n, L const& f) noexcept
```
AlexanderSinn: The design of this seems to be missing a lot to me. How does this work if SIMD is not enabled? What does this do on GPU? I think currently this cannot take a lambda as input and needs a struct with both simd and non-simd operator().

ax3l: This one is fully standalone. The SIMD header brings the the types for it.

One could in general use this pattern on GPU as well, but not the target of this PR.

There is a solid helper for copy in/out needed that I currently have in user-code, which will probably need documentation eventually. I do not yet have a good strategy how to make this a helper inside AMReX, but maybe later we know a way:
- https://github.com/BLAST-ImpactX/impactx/blob/629b3d884c3ca974d75ee99d3a0a494c3dc0e6f8/src/elements/mixin/beamoptic.H#L105-L151
- https://github.com/BLAST-ImpactX/impactx/blob/629b3d884c3ca974d75ee99d3a0a494c3dc0e6f8/src/elements/mixin/beamoptic.H#L172-L198

ax3l:
@AlexanderSinn tested and it also works called as a Lambda, and we could consider adding a few helpers for the aligned reads & writes to AMReX (later on?): https://godbolt.org/z/Pbzs6dnbP

ax3l:
I had a godbolt test link somewhere, where I ensured that the writes of non-modified reads are are optimized out by the compiler in `-O3`, but have a hard time reproducing it in that last link: https://godbolt.org/z/P3K87rsea
(Simpler example: https://godbolt.org/z/41Teqz4bE )

That would be quite important, otherwise we have to manually track our modifications.

ax3l:
One can force it with metaprogramming, and then the user just needs to `const` their declarations properly, which we usually do:
https://godbolt.org/z/4czrsEP1n

Shorter: (thx @AlexanderSinn)
```C++
template <typename R, typename... Args>
constexpr bool is_nth_arg_non_const (R(*)(Args...), int n) {

    bool val_arr[sizeof...(Args)] {!std::is_const_v<std::remove_reference_t<Args>>...};

    return val_arr[n];
}

...

if constexpr (is_nth_arg_non_const(modify<double>, 0))
    ad.copy_to(ap + i, stdx::vector_aligned);
```

My update to also work with functors:
https://godbolt.org/z/hv43EGEKb

ax3l:
Ah nice, and @AlexanderSinn just showed me that the issue is because of the `std::vector` use in the example (and only for gcc, [ok for clang](https://godbolt.org/z/bjbGrEs86)), so I remembered my test correctly that the compiler can optimize this on its own for the usual pattern we use in AMReX: https://godbolt.org/z/8xYxTY5Mo and that I implemented in ImpactX.

Src/Base/AMReX_SIMD.H
```diff
@@ -0,0 +1,141 @@
...
+     *
+     * compute(part_x, part_y);
+     *
+     * if constexpr (is_nth_arg_non_const(compute<double>, 0))
+     *     part_x.copy_to(&m_part_x[i], stdx::element_aligned);
+     * if constexpr (is_nth_arg_non_const(compute<double>, 1))
+     *     part_y.copy_to(&m_part_y[i], stdx::element_aligned);
+     */
+    template<typename R, typename... Args>
+    constexpr bool is_nth_arg_non_const (R(*)(Args...), int n)
```
ax3l, line 134: Need to add support for functors as well (used as such in ImpactX).

Note: Functor `operator()` signature types always come as two types, const and non-const member functions.

Src/Base/AMReX_GpuLaunchFunctsC.H
ax3l: Thx for the fix!

> Also there should be a second version of this with Gpu::Handler{} to match the scalar version.

Yes, def worth for some GPU kernels, but needs extra work. So far I only use this in CPU code.

AlexanderSinn: I that case, maybe rename it to `call_f_vector` or `call_f_simd`

Src/Base/AMReX_GpuLaunchFunctsC.H
```diff
@@ -175,6 +185,31 @@ void ParallelFor (T n, L&& f) noexcept
     ParallelFor(n, std::forward<L>(f));
 }
 
+/** ParallelFor with a SIMD Width (in elements)
+ *
+ * SIMD load/Write-back operations need to be performed before/after calling this.
```
ax3l, line 198: Probably good to add a C++ example snippet here in the doc string, too.

Src/Base/AMReX_GpuLaunchFunctsC.H
ax3l: > I think currently this cannot take a lambda as input

Alex has a draft that would change the impl to support lambda: https://godbolt.org/z/Pbzs6dnbP

I added a little improvement here: https://godbolt.org/z/7455hqrEc

> [I think this] needs a struct with both simd and non-simd operator().

Correct, because one needs to be able to resolve the remainder. We could alternatively resolve the remainder using the SIMD operator() and `WIDTH=1`, but in practice one will template the implementation anyway as a `T_Real` for CPU/GPU portability.

Src/Base/AMReX_GpuLaunchFunctsC.H
```diff
@@ -4,6 +4,30 @@
...
+{
+    /** SIMD width in elements */
+    static constexpr int width = WIDTH;
+
+    /** The linear loop index of ParallelFor(SIMD) */
+    N index = 0;
+
+    /** behave like an integer */
+    AMREX_FORCE_INLINE
+    operator int() const {
```
AlexanderSinn:
```suggestion
    operator N() const {
```

Src/Base/AMReX_GpuLaunchFunctsC.H
```diff
@@ -13,15 +37,25 @@ namespace detail {
     auto call_f_scalar_handler (F const& f, N i)
         noexcept -> decltype(f(0))
     {
-        f(i);
+        f(SIMDindex<1, N>{i});
```
AlexanderSinn: This would add the simd index type to code using regular ParallelFor. Since there is only one way to call the user function in ParallelForSIMD (not with and without Gpu::Handler like the other ParallelFor), it would be best to not use the call_f funcitons and just do it directly in ParallelForSIMD.

ax3l: Agreed.

Src/Base/AMReX_GpuLaunchFunctsC.H
```diff
@@ -175,6 +209,31 @@ void ParallelFor (T n, L&& f) noexcept
...
+ * @tparam L function/functor to call per SIMD set of elements
+ */
+template <int WIDTH, typename T, typename L, typename M=std::enable_if_t<std::is_integral_v<T>> >
+AMREX_ATTRIBUTE_FLATTEN_FOR
+void ParallelForSIMD (T n, L const& f) noexcept
+{
+    T i = 0;
+    // vectorize full lanes
+    for (; i + WIDTH <= n; i+=WIDTH) {
+        detail::call_f_vector_handler<WIDTH>(f, i);
```
ax3l: We could probably skip that indirection to `call_f_vector_handler`, because we only support CPU for now.

Src/Base/AMReX_GpuLaunchFunctsC.H
```diff
@@ -175,6 +199,31 @@ void ParallelFor (T n, L&& f) noexcept
...
+ * @tparam N index type (integer)
+ * @tparam L function/functor to call per SIMD set of elements
+ */
+template <int WIDTH, typename N, typename L, typename M=std::enable_if_t<std::is_integral_v<N>> >
+AMREX_ATTRIBUTE_FLATTEN_FOR
+void ParallelForSIMD (N n, L const& f) noexcept
+{
+    N i = 0;
+    // vectorize full lanes
+    for (; i + WIDTH <= n; i+=WIDTH) {
```
ax3l, line 210: Oh, is that correct? Should probably read
```suggestion
    for (; i + WIDTH < n; i+=WIDTH) {
```

ax3l:
@WeiqunZhang @AlexanderSinn I think this is ready for a first merge as a preview, for further iterative improvement. The chance for negative side effects for other users should be very low, given the way this is implemented and non-default.

If we can merge it now still in this release cycle, I could use it in an LDRD and a paper that we are about to submit early August :pray:

Src/Particle/AMReX_Particle.H
```diff
@@ -93,27 +94,75 @@ namespace amrex {
...
+        if constexpr (std::is_same_v<T_Mask, bool>) {
+            // scalar
+            static_assert(std::is_same_v<T_SIMD, uint64_t>, "make_invalid: forgot template on wrapper?");
+            if (mask) { make_invalid(idcpu); }
+        } else {
+#ifdef AMREX_USE_SIMD
+            // SIMD
+            static_assert(!std::is_same_v<T_SIMD, uint64_t>, "make_invalid: forgot template on wrapper?");
+            // .__cvt() because of: https://github.com/VcDevel/std-simd/issues/41
+            simd::stdx::where(mask.__cvt(), idcpu) &= ~(T_SIMD{1} << 63);
```
ax3l, line 110: Memo to myself: in ImpactX tests, this part seems to cause issues (i.e., does not mark any particles as invalid). Not clear if an issue in AMReX or in the ImpactX usage if vectorized idcpu.

Code here looks correct...

Only used in apertures. Not needed for the paper we work on.

ax3l: lol, all good. I commented out the diagnostics that produced the test data in ImpactX.

Src/Base/AMReX_GpuLaunchFunctsC.H
```diff
@@ -175,6 +199,31 @@ void ParallelFor (T n, L&& f) noexcept
...
+ * @tparam L function/functor to call per SIMD set of elements
+ */
+template <int WIDTH, typename N, typename L, typename M=std::enable_if_t<std::is_integral_v<N>> >
+AMREX_ATTRIBUTE_FLATTEN_FOR
+void ParallelForSIMD (N n, L const& f) noexcept
+{
+    N i = 0;
+    // vectorize full lanes
+    for (; i + WIDTH < n; i+=WIDTH) {
+        f.template operator()<WIDTH>(SIMDindex<WIDTH, N>{i});
```
AlexanderSinn:
```suggestion
       f(SIMDindex<WIDTH, N>{i});
```
Needs to be like this if lambdas should work. Note that this requires changing the simd operator() in impactx.

ax3l: Ok, easy change. Testing and pushing in a second.

ax3l: fixed here and in ImpactX :)

Src/Base/AMReX_GpuLaunchFunctsC.H
```diff
@@ -175,6 +199,31 @@ void ParallelFor (T n, L&& f) noexcept
...
+ * @tparam N index type (integer)
+ * @tparam L function/functor to call per SIMD set of elements
+ */
+template <int WIDTH, typename N, typename L, typename M=std::enable_if_t<std::is_integral_v<N>> >
+AMREX_ATTRIBUTE_FLATTEN_FOR
+void ParallelForSIMD (N n, L const& f) noexcept
+{
+    N i = 0;
+    // vectorize full lanes
+    for (; i + WIDTH < n; i+=WIDTH) {
```
AlexanderSinn:
```suggestion
    for (; i + WIDTH <= n; i+=WIDTH) {
```
It was correct with `<=`. Think of the case where WIDTH is 1, then `i+1<=n` is the same as `i<n`

ax3l: oh true! The result was still correct and caught by the remainder now, but is less efficient. WIll revert!

Src/Base/AMReX_GpuComplex.H
```diff
@@ -53,8 +53,9 @@ struct alignas(2*sizeof(T)) GpuComplex
     /**
      * \brief Add a real number to this complex number.
      */
+    template <typename U>
     AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
-    GpuComplex<T>& operator+= (const T& a_t) noexcept
+    decltype(auto) operator+= (const U& a_t) noexcept
```
WeiqunZhang: Do we have to use `decltype(auto)` here? It's less readable.

ax3l: Good point. We could do trailing return type, but I think the math here is so simple that explicit `GpuComplex<T>&` is always the result, at least for the case where we use it right now: `T` being scalar or vector type and `U` being a scalar type.

Will change the return type back to the original signature.

ax3l: Fixed!

Src/Base/AMReX_SIMD.H
```diff
@@ -0,0 +1,154 @@
...
+    // TODO Check if a default with * 2 or similar is sensible.
+    template<int SIMD_WIDTH = native_simd_size_real>
+    using SIMDReal = std::experimental::fixed_size_simd<amrex::Real, SIMD_WIDTH>;
+
+    template<int SIMD_WIDTH = native_simd_size_particlereal>
+    using SIMDParticleReal = std::experimental::fixed_size_simd<amrex::ParticleReal, SIMD_WIDTH>;
+
+    // Type that has the same amount of IdCpu SIMD elements as the SIMDParticleReal type
+    template<typename T_ParticleReal = SIMDParticleReal<>>
+    using SIMDIdCpu = std::experimental::rebind_simd_t<uint64_t, T_ParticleReal>;
```
WeiqunZhang: We probably should include `<cstdint>` and in principle we should use `std::uint64_t` instead of `uint64_t`.

WeiqunZhang (commented):
LGTM. Just another minor comment.

Src/Base/AMReX_SIMD.H
ax3l: Thank you - fixed!

WeiqunZhang (approved):

zingale:
This broke Castro/Microphysics compilation.  Is there a way to opt out of this?

ax3l:
@zingale oh no, how did it break?

Fix in #4570, but maybe we can understand what breaks and update downstream?

WeiqunZhang:
It's because of `using amrex::Math::sincos` and they have their own `sincos`.

ax3l:
Fix merged in #4570

ax3l:
Update: This functionality made ImpactX astronomically faster on CPU for compute-heavy/bound kernels:
https://github.com/BLAST-ImpactX/impactx/pull/1002#issuecomment-3138910414

Issue #4498, opened 2025-06-10
[OpenMP] `AddParticles(pc & other)` Broken in ImpactX


ax3l:
When using more than one OpenMP thread in ImpactX, the `addParticles(pc & other)` does not work. This might be due to the particular init logic we use in ImpactX for particles when not using collective effects (i.e., `redistribute` is never called)?

Here is a reproducer:
- https://github.com/BLAST-ImpactX/impactx/pull/987
- uses https://github.com/AMReX-Codes/pyamrex/pull/447

In detail, we draw a particle distribution once for a series of tests (which is time consuming) and then copy in the exact same particle container over and over for following tests (to be reproducible in exact particles that we had drawn for other tests & be faster than drawing again).
```py
# once we do
sim.add_particles(bunch_charge_C, distr, npart)

sim.backup_beam = self.sim.particle_container().make_alike()  # note: not sure if the issue could also be here?
self.sim.backup_beam.add_particles(self.sim.particle_container())
```
```py
# then repeated for each test we do

# instead of drawing from the distribution again, create a 2nd
# particle container and copy the same initial particles again.
pc = sim.particle_container()
pc.clear_particles()
pc.add_particles(sim.backup_beam)  # note: uses default of local=False (no difference if I use local=True)

assert pc.total_number_of_particles() == npart  # this fails for more than one OMP thread
```
results in `total_number_of_particles()` (with default arguemtns) to return Npart / OpenMP threads.
https://github.com/AMReX-Codes/amrex/blob/cef9e0c5d73ff888292d1ac1eae098aa7600ad6d/Src/Particle/AMReX_ParticleContainer.H#L574-L582

CC @atmyers @WeiqunZhang

ax3l:
I think the fix will be similar to the fix in #2695.

Calling `Redistribute()` before `addParticles(other)` is a work-around for now. But in ImpactX, we have a central mode where we do never call `Redistribute()` for performance & perfect load balancing reasons.

ax3l:
Ok, but the downside of calling `Redistribute()` as a work-around for this benchmark is that the OpenMP parallelism does not work anymore... So we need to really check what is wrong in `addParticles()` (and potentially `make_alike`, too).

ax3l:
[...]

PR #4522, opened 2025-06-25
make_alike should also copy the tile_size

atmyers:
This should fix Issue #4498 observed in ImpactX.

The proposed changes:


atmyers:
Never mind - `tile_size` is static, but the fix was something else...

PR #4526, opened 2025-06-26
Reduce use of SuperParticle

AlexanderSinn:
## Summary

In preparation of #4404, this PR reduces the reliance of SuperParticle. The issue with getSuperParticle is that the compiler is instructed to read in all the data of the particle into registers even if most of it is not needed. Additionally, runtime data is not accessible through SuperParticle.

## Additional background

The previous version of packIOData with RunOnGpu had a bug when some particles were invalid.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_ParticleReduce.H
+++ b/Src/Particle/AMReX_ParticleReduce.H
@@ -66,18 +66,18 @@ auto call_f (F const& f,
  *                      return p.idata(0);
  *                  });
  *
- *        using PTDType = typename PC::ParticleTileType::ConstParticleTileDataType;
+ *        using PTDType = typename PC::PTDType;
  *        auto sm = amrex::ReduceSum(pc,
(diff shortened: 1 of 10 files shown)


Src/Particle/AMReX_ParticleReduce.H
```diff
@@ -756,18 +756,18 @@ ReduceLogicalAnd (PC const& pc, int lev, F&& f)
  *                      return p.id() > 0;
  *                  });
  *
- *        using SPType  = typename PC::SuperParticleType;
+ *        using SPType  = typename PC::;
```
atmyers:
```suggestion
 *        using SPType  = typename PC::SuperParticleType;
```

AlexanderSinn: Good spot, thanks!

atmyers (approved):
Thanks for this!

PR #4527, opened 2025-06-26
[DO NOT MERGE] GPU TX: Annotate Unpack

ax3l:
## Summary

- [x] `unpackRemotes`
- [x] `unpackBuffers`

## Additional background

Debugging memory leaks on AMD GPUs.

## Checklist

The proposed changes:


ax3l:
Ok, `resizeTiles` causes huge memory allocations with `hipMalloc` that are never going down again.
Screenshot from 2025-06-26 14-35-30

Running with `amrex.the_arena_release_threshold=0` reduces the memory overhead by 2x, but it is still too large (i.e., still 25%-50% comm overhead for a 8-box-per-gpu (large boxes), 4-gpu run with a single particle species and 8 ppc, all homogeneous & warm plasma)

PR #4528, opened 2025-06-27
Add getInvalidRandomEngine() for CPU

AlexanderSinn:
## Summary

#4443 deleted the default constructor of RandomEngine for GPU. However for CPU still the default constructor must be used, forcing \#ifdef to be used in code that uses the constructor. This PR adds a `getInvalidRandomEngine()` function to get a dummy RandomEngine. Note the only use for RandomEngine on CPU is to make GPU code work on CPU.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_RandomEngine.H
+++ b/Src/Base/AMReX_RandomEngine.H
@@ -68,6 +68,7 @@ namespace amrex
 
 #endif
 
+    AMREX_FORCE_INLINE
     RandomEngine getInvalidRandomEngine () {
 #ifdef AMREX_USE_GPU
         return RandomEngine{nullptr};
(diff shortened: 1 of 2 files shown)


WeiqunZhang:
The original intention was to make it really hard for people to write this type of code.

WeiqunZhang:
If one wants to avoid ifdef, one should use ParallelForRNG and then use the RandomEngine provided by the function. Of course, I understand this makes some existing Particle functions hard to use.

WeiqunZhang:
One might argue that RandomEngine{nullptr} does not look right on GPU. But I also did not expect that people would have used RandomEnginre{} on GPU and expected it to work.

AlexanderSinn:
I changed it so now it is `getInvalidRandomEngine()` to make it more clear.

WeiqunZhang (approved):

PR #4531, opened 2025-06-27
Tweak CArena Defragmentation Strategy

WeiqunZhang:
The previous strategy added in #4451 has a flaw. Suppose a CArena's initial size is small and we have n vectors each with a size of x. Now we are resizing these vectors one by one to size x+y, where y << x. Then we would end up with n new allocations each with a size of 2*x+y. We have doubled the memory usage in the end, because the unused spaces can not be combined.

In the new strategy, we only attempt to combine allocations when the combined amount is not less than the requested amount of allocation.

We also check the malloc error code now. If it fails, we will try to free more memory and call malloc again.

PR #4529, opened 2025-06-27
Clean up RedistributeCPU

AlexanderSinn:
## Summary

This PR simplifies RedistributeCPU to be independent of particle layout in preparation for #4404.
For this, push_back is replaced by a resize with a geometric growth strategy. push_back is error-prone due to having the possibility to desynchronize the sizes of the individual component vectors if used incorrectly.

## Checklist

The proposed changes:
- [ ] fix a bug or incorrect behavior in AMReX
- [ ] add new capabilities to AMReX
- [ ] changes answers in the test suite to more than roundoff level
- [ ] are likely to significantly affect the results of downstream AMReX users
- [ ] include documentation in the code and/or rst files, if appropriate

PR #4532, opened 2025-06-28
Use geometric grow for resize in RedistributeGPU

AlexanderSinn:
## Summary

In an attempt to help with #4531 I used the `resize_geometric_grow` function from #4529 to resize the particle tiles in RedistributeGPU. Additionally, this PR changes the default `amrex.vector_growth_factor` from 1.5 to 1.25.

## Checklist

The proposed changes:

### DIFF
--- a/Docs/sphinx_documentation/source/RuntimeParameters.rst
+++ b/Docs/sphinx_documentation/source/RuntimeParameters.rst
@@ -1142,7 +1142,7 @@ Memory
 
 .. py:data:: amrex.vector_growth_factor
    :type: amrex::Real
-   :value: 1.5
+   :value: 1.25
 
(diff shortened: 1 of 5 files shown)


AlexanderSinn:
I tested this on LUMI and it does not fix the problem

AlexanderSinn:
This does not fix the problem on its own; however, when combining this with a properly sized arane init size (like the default) and disabling periodic particle sorting (which reallocates some of the particle vectors), the issue is fixed.

Note: the particle sorting could be changed in amrex to be able to work with this without having the memory problem

WeiqunZhang:
This may have unknown consequences, so I don't think we should do it.

PR #4534, opened 2025-06-28
AMRErrorTag: Add PARSER

WeiqunZhang:


asalmgren (approved):

WeiqunZhang:
Note that amrex::ParmParse has a `makeParser` function that could be useful if the parser expression contains constants set by ParmParse.

PR #4537, opened 2025-06-30
Retry a few times if directory creation fails

WeiqunZhang:
This tries to fix the following error reported by @BenWibking

    Writing checkpoint chk02500
    amrex::UtilCreateDirectory:: path errno:  chk02500 :: File exists
    amrex::UtilCreateDirectory:: path errno:  chk02500/Level_2 :: Input/output error
    amrex::Error::0::Couldn't create directory: chk02500/Level_2 !!!

### DIFF
--- a/Src/Base/AMReX_Utility.cpp
+++ b/Src/Base/AMReX_Utility.cpp
@@ -116,7 +116,13 @@ bool
 amrex::UtilCreateDirectory (const std::string& path,
                             mode_t mode, bool verbose)
 {
-    return FileSystem::CreateDirectories(path, mode, verbose);
+    double sleep = 0.1;
+    while (sleep < 2.0) {
+        if (FileSystem::CreateDirectories(path, mode, verbose)) { return true; }
+        amrex::Sleep(sleep);
+        sleep *= 2;
+    }
+    return false;
 }
 
 void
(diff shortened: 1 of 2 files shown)


Src/Base/AMReX_Utility.cpp
```diff
@@ -116,7 +116,13 @@ bool
 amrex::UtilCreateDirectory (const std::string& path,
                             mode_t mode, bool verbose)
 {
-    return FileSystem::CreateDirectories(path, mode, verbose);
+    double sleep = 0.1;
+    while (sleep < 2.0) {
+        if (FileSystem::CreateDirectories(path, mode, verbose)) { return true; }
```
BenWibking, line 121: Maybe if this fails, it could do, e.g.:
```
amrex::Print() << "FileSystem::CreateDirectories() failed with errno: " << std::strerror(errno) << ". Retrying.\n";
```

WeiqunZhang: If verbose is true, `FileSystem::CreateDirectories` will print message. In most cases, we are supposed to have a way to not print.

WeiqunZhang: We are trying to follow this policy. https://github.com/xsdk-project/xsdk-community-policies/blob/master/package_policies/M11.md

atmyers (approved):

PR #4545, opened 2025-07-03
CMake: `AMReX_FASTMATH`

### LINKED ISSUE #1769: Fast Math: Unified Build Option
Checking our current build scripts, we don't yet expose control over compiler fast-math options for all platforms. Let's add a flag in GNUmake & CMake for user-facing control.

This should also level the playing field for benchmark comparisons between various platforms.

Current GNUmake options I found:
- CUDA:        w/ fastmath
- HIP/SYCL:    no fastmath
- serial/OpenMP via GCC:    no fastmath
- serial/OpenMP via Intel:    no fastmath
- serial/OpenMP via CCE:    w/ fastmath

CMake:
- CUDA: w/ fastmath (default: `AMReX_CUDA_FASTMATH=ON`)

ax3l:
## Summary

Add new general compiler flag for fast-math optimizations. https://en.wikipedia.org/wiki/Floating-point_arithmetic#%22Fast_math%22_optimization

**This PR does not yet change any defaults (see details below).**

Close #1769

## Proposed Breaking Change

By default, AMReX turned fast-math OFF for all backends (serial, CPU/OpenMP, SYCL GPU, HIP GPU) besides:
- CUDA GPUs (both GNUmake & CMake)
- a single, legacy [cray compiler in GNUmake](https://github.com/AMReX-Codes/amrex/blob/25.07/Tools/GNUMake/comps/cray.mak)

I propose to amend this PR to:
- remove the `AMReX_CUDA_FASTMATH` flag (default: ON)
  - alternatively: fade it out and set its default value to `AMReX_FASTMATH`
- turn `AMReX_FASTMATH` to `OFF` by default for all backends.

Contrary to turning it `ON` by default, users of most backends will not be surprised by a sudden change in numerical results.

### Breaking:

- Users that support fast-math, which might be many will see a sudden drop of performance on CUDA GPUs.

### Recommended migration:

- If you already benchmarked correctness on CUDA GPUs, I recommend to intentionally set in your projects `-DAMReX_FASTMATH=ON` for all backends, to get the performance benefits on CPUs, Intel and AMD GPUs as well.

## Additional background

As HPC team, we should strive to make our numerics as robust as possible under fast-math optimizations. Without those, your compiler will barely auto-vectorize, will not replace simple optimizations like `x / 2.` with `x * 0.5` and many other optimizations that rely on agressive re-ordering.

Fast-math floating point math is an important aspect to consider and support for significant performance improvements on all modern architectures.

## Checklist

The proposed changes:

### DIFF
--- a/Tools/CMake/AMReXOptions.cmake
+++ b/Tools/CMake/AMReXOptions.cmake
@@ -254,6 +254,18 @@ unset(_GPU_RDC_default)
 print_option(AMReX_GPU_RDC)
 
 #
+# Fast Math    ================================================================
+#
+set(_FASTMATH_default OFF)
+if(AMReX_GPU_BACKEND STREQUAL CUDA)  # note: historic settings
+# if(NOT AMReX_GPU_BACKEND STREQUAL NONE)  # note: this would be more consistent for GPUs
+    set(_FASTMATH_default ON)
+endif()
+option(AMReX_FASTMATH  "Enable fast-math optimizations" ${_FASTMATH_default})
+print_option(AMReX_FASTMATH)
+unset(_FASTMATH_default)
+
+#
 # Parallel backends    ========================================================
 #
 option( AMReX_MPI  "Enable MPI"  ON )
--- a/Tools/CMake/AMReX_Config.cmake
+++ b/Tools/CMake/AMReX_Config.cmake
@@ -72,7 +72,6 @@ function (configure_amrex AMREX_TARGET)
    )
 
    unset(_condition)
-   unset(_cxx_msvc)
 
    #
    # Setup OpenMP
@@ -146,10 +145,32 @@ function (configure_amrex AMREX_TARGET)
       endif()
    endif()
 
+   # fast math
+   if (AMReX_FASTMATH)
+       # GPU specific backends set in AMReXParallelBackends.cmake
+       if (AMReX_GPU_BACKEND STREQUAL NONE)
+           # See https://cmake.org/cmake/help/v4.1/variable/CMAKE_LANG_COMPILER_ID.html#variable:CMAKE_%3CLANG%3E_COMPILER_ID
+           target_compile_options(${AMREX_TARGET} PUBLIC
+               $<$<CXX_COMPILER_ID:AppleClang,Clang,CrayClang,GNU,IBMClang,IntelLLVM,XLClang>:-ffast-math>
+               $<${_cxx_msvc}:"/fp:fast">  # MSVC
+           )
+           if (CMAKE_Fortran_COMPILER_LOADED)
+               target_compile_options(${AMREX_TARGET} PUBLIC
+                   $<$<Fortran_COMPILER_ID:AppleClang,Clang,CrayClang,GNU,IBMClang,IntelLLVM,XLClang>:-ffast-math>
+               )
+           endif ()
+       endif()
+   endif()
+
    #
    # Setup third-party profilers
    #
    set_amrex_profilers(${AMREX_TARGET})
+
+   #
+   # clean up helpers
+   #
+   unset(_cxx_msvc)
 
 endfunction ()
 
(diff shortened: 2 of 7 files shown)


WeiqunZhang:
We could consider `-funsafe-math-optimizations -fno-finite-math-only`.

WeiqunZhang:
https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html#index-ffast-math

WeiqunZhang:
At least for some versions of Intel compiler, it turns on some fast math features with `-O3`. We had this in GNU Makefile.
```
# disable warning: comparison with infinity always evaluates to false in fast floating point modes [-Wtautological-constant-compare]                                         
#                  return std::isinf(m);                                                                                                                                     
# appeared since 2021.4.0                                                                                                                                                    
CXXFLAGS += -Wno-tautological-constant-compare
```

WeiqunZhang:
@ax3l Maybe we can set the default to `-ffast-math -fno-finite-math-only` and add AMReX_FINITE_MATH_ONLY (off by default). But nvcc does not appear to have a finite-math flag. For Intel, we might need to use `-fp-mode`. Also not sure about the behavior of hip and nvhpc.

WeiqunZhang:
The fast-math flag used in llvm based cray compiler in gnu make system was actually added for consistency, because the previous cray compilers did fast-math without any special fast-math flags.

ax3l:
Another flag I think would make sense to expose a control is the "honor parenthesis" flag. Quite often when building a stable algorithm, we kind of expect that parenthesis boundaries are not crossed in optimization, and depending on the compiler they do ignore them in fast-math.

ax3l:
https://github.com/AMReX-Codes/amrex/pull/4545#issuecomment-3042495008

Yes, Intel is known to do that [and other aggressive fast-math controls](https://en.wikipedia.org/wiki/Floating-point_arithmetic#%22Fast_math%22_optimization) in `-O3` (especially in Fortran).

ax3l:
I am not sure how much we safe from finite math, thus controlling it with an extra flag for the sake of easier adoption of fast-math is good I think, because it eliminates a large set of common adoption-issues.

Will be interesting to see what effect finite math / flush-to-zero / no-subnormals on its own has on our usual apps.

WeiqunZhang (approved):

Tools/CMake/AMReX_Config.cmake
```diff
@@ -146,11 +145,33 @@ function (configure_amrex AMREX_TARGET)
       endif()
    endif()
 
+   # fast math
+   if (AMReX_FASTMATH)
+       # GPU specific backends set in AMReXParallelBackends.cmake
+       if (AMReX_GPU_BACKEND STREQUAL NONE)
+           # See https://cmake.org/cmake/help/v4.1/variable/CMAKE_LANG_COMPILER_ID.html#variable:CMAKE_%3CLANG%3E_COMPILER_ID
+           target_compile_options(${AMREX_TARGET} PUBLIC
+               $<$<CXX_COMPILER_ID:AppleClang,Clang,CrayClang,GNU,IBMClang,IntelLLVM,XLClang>:-ffast-math>
```
ax3l, line 154: Hm, maybe we should provide a CMake interface target for the flags, similar as we do for AMReX_FPE -> `Flags_FPE` target. That way, downstream projects can use a pre-compiled AMReX (w/o FM) and turn FM on in the downstream project without mirroring the logic / lookup of flags as done here.

ax3l: #4591

Tools/CMake/AMReXOptions.cmake
```diff
@@ -253,6 +253,18 @@ cmake_dependent_option( AMReX_GPU_RDC "Enable Relocatable Device Code" ${_GPU_RD
...
 
+#
+# Fast Math    ================================================================
+#
+set(_FASTMATH_default OFF)
+if(AMReX_GPU_BACKEND STREQUAL CUDA)  # note: historic settings
+# if(NOT AMReX_GPU_BACKEND STREQUAL NONE)  # note: this would be more consistent for GPUs
+    set(_FASTMATH_default ON)
+endif()
+option(AMReX_FASTMATH  "Enable fast-math optimizations" ${_FASTMATH_default})
```
ax3l, line 264: This should have just been `OFF` and was leftover from development, because in the end we have `AMReX_CUDA_FASTMATH` already. Fix in #4591

PR #4547, opened 2025-07-03
Replace `std::numeric_limits<>::infinity()` with `max()`

AlexanderSinn:
## Summary

If fast-math is enabled from #4545, calculations with `std::numeric_limits<>::infinity()` won't work due to the implied `-ffinite-math-only` flag.

Unfortunately for these two places, using max instead of infinity doesn't fit very well, so I'm unsure if this should be merged.

## Checklist

The proposed changes:


WeiqunZhang:
I think we should keep infinity in ParmParse. If a user says `a = inf` in inputs. We should honor it.  As for CoordSys's inv_dx, we could use max instead of inf.

WeiqunZhang (approved):

PR #4550, opened 2025-07-04
amrex::FileStream

WeiqunZhang:
Implement our own file stream class that is somewhat similar to std::fstream. It can be used by amrex::NFilesIter in mesh and particle data I/O. By default amrex::NFilesIter still uses std::fstream. To use amrex::FileStream, one needs to use `USE_OWN_FILE_STREAM=TRUE` for GNU Make and `AMReX_OWN_FILE_STREAM=ON` for CMake.

amrex::FileStream::Read has not been tested and currently it is not being used for reading MultiFab and particles.


Src/Base/AMReX_FileStream.cpp
```diff
@@ -0,0 +1,425 @@
...
+
+FileStream& FileStream::write (char const* s, std::streamsize count)
+{
+    if (count <= 0) { return *this; }
+
+    if (m_fd == -1 || !m_good) {
+        throw std::runtime_error("FileStream::write: bad file descriptor or bad state");
+    }
+
+    if (m_buffer_mode == BufferMode::Read) {
```
BenWibking, line 208: From a quick skim of the code, I don't really understand the motivation for `BufferMode` or how it works. Is there a reason to have separate modes implemented by the same function, or can different modes have separate implementations in separate functions? I can go back and read it more carefully with more context.

WeiqunZhang: Currently we only need to open a stream for either write or read, but not both. If we want to support both, we will need to know the current mode of the buffer.

PR #4554, opened 2025-07-07
EB anisotropic meshes - changes for cut_face_2d routine

nataraj2:
This PR makes the necessary changes in the variables - `areafrac`, `nx`, `ny`, `centx`, `centy`, `Sx2`, `Sy2`, `Sxy`, in the `cut_face_2d` routine, for anisotropic meshes (ie. when mesh spacings are not equal). The derivation is in the attached pdf in the link below. The changes are in Eqns. 35-37, 43-44, 46-47, 49-50, 55-56.

[EB_anisotropic.pdf](https://github.com/user-attachments/files/21201504/EB_anisotropic.pdf)

## Checklist

The proposed changes:

### DIFF
--- a/Src/EB/AMReX_EB2_3D_C.cpp
+++ b/Src/EB/AMReX_EB2_3D_C.cpp
@@ -267,7 +267,7 @@ void cut_face_2d (Real& areafrac, Real& centx, Real& centy,
         Real dx4 = dx * (x_ym + x_yp) * (x_ym*x_ym + x_yp*x_yp);
         Real af1 = 0.5_rt*(axm+axp) + aa*0.5_rt*dx2*dxval/dyval;
         centx = 0.125_rt*(axp-axm) + aa*(1.0_rt/6._rt)*dx3*dxval/dyval;
-        Sx2 = (1.0_rt/24._rt)*(axm+axp) + aa*(1.0_rt/12._rt)*dx4;
+        Sx2 = (1.0_rt/24._rt)*(axm+axp) + aa*(1.0_rt/12._rt)*dx4*dxval/dyval;
 
         Real signy = (ny > 0.0_rt) ? 1.0_rt : -1.0_rt;
         Real y_xm = (-0.5_rt + axm)*signy;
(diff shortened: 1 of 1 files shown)


asalmgren:
@nataraj2 @WeiqunZhang -- take a look at the calling sequences -- note in cut_faces_2d we have 

Real axm, Real axp, Real aym, Real ayp, Real bcx, Real bcy, Real dxval, Real dyval) 

The point being that for the areas, the bcs and the d*val, the "x" version always comes before the "y" version.

Look now at the calling sequences, e.g.

lzm,lzp,lym,lyp,bcy,bcz,dx[2],dx[1])

Note that Mahesh's changes match -- dx[2] comes before dx[1] and lzm comes before lym ... but bcy comes before bcz.

This seems wrong??

WeiqunZhang:
Yes, I agree with @asalmgren that the orders seem wrong. I think they should always be dx[i] following by dx[j], where i < j. That's how we order face centroids. For y-face centroid, component 0 is for x and 1 for z.

WeiqunZhang:
Other than the order of dx, LGTM.

We probably also want to fix Sx2, Sy2 and Sxy in the future. They are used in build_cells that does the least squares fit.

WeiqunZhang:
I don't think this is correct. Are you saying there is a bug even in the uniform size case?

nataraj2:
See image below for the terminology used in the `cut_face_2d` calls - from @WeiqunZhang
```
 cut_face_2d(apz(i,j,k),fcz(i,j,k,0),fcz(i,j,k,1), // NOLINT(readability-suspicious-call-argument)
                            m2z(i,j,k,0),m2z(i,j,k,1),m2z(i,j,k,2),
                            lym,lyp,lxm,lxp,bcx,bcy);
```
`lym` denotes `axm` and not `aym`. Hence, the ordering of the `bcs` are correct as they are and the `dx` ordering has been corrected as @WeiqunZhang commented above.
The second order moments `Sx2`, `Sy2` and `Sxy` have been corrected for anisotropic meshes and the pdf attached in the very first comment of this PR has the derivations as well.

PXL_20250713_171220571

asalmgren (approved):

WeiqunZhang (approved):

PR #4562, opened 2025-07-11
SUNDIALS interface updates

gardner48:
## Summary

* Check SUNDIALS return flags
* Abort if a required callback function is called but not set
* Fix selecting a specific IMEX method with SUNDIALS
* Document time integrator runtime inputs
* Read base class parameters at construction
* Add options for selecting the algebraic solvers with implicit methods
* Add option to set a stop time
* Only output settings and final statistics with verbose output is enabled

## Additional background

Various fixes and enhancements from testing methods in FerroX

## Checklist

The proposed changes:

### DIFF
--- a/Docs/sphinx_documentation/source/RuntimeParameters.rst
+++ b/Docs/sphinx_documentation/source/RuntimeParameters.rst
@@ -17,8 +17,8 @@ to :cpp:`amrex::Initialize` and the function adds parameters to AMReX's
 .. important:: AMReX reserves the following prefixes in :cpp:`ParmParse`
                parameters: ``amr``, ``amrex``, ``blprofiler``, ``device``,
                ``DistributionMapping``, ``eb2``, ``fab``, ``fabarray``,
-               ``geometry``, ``particles``, ``tiny_profiler``, and
-               ``vismf``.
+               ``geometry``, ``integration``, ``particles``, ``tiny_profiler``,
+               and ``vismf``.
 
 AMR
 ---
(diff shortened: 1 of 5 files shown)


gardner48:
FYI @ajnonaka, various updates I made when testing with FerroX. I'll open a PR with some corresponding updates on the FerroX side as well.

Docs/sphinx_documentation/source/RuntimeParameters.rst
```diff
@@ -17,8 +17,8 @@ to :cpp:`amrex::Initialize` and the function adds parameters to AMReX's
 .. important:: AMReX reserves the following prefixes in :cpp:`ParmParse`
                parameters: ``amr``, ``amrex``, ``blprofiler``, ``device``,
                ``DistributionMapping``, ``eb2``, ``fab``, ``fabarray``,
-               ``geometry``, ``particles``, ``tiny_profiler``, and
```
jmsexton03, line 20: Could the prefix used be `time_integration` rather than `integration`? this would be more consistent with the object and folder naming. If this would require invasive changes elsewhere it's not required

ajnonaka: The issue here is on external codes.  It looks like the changes to amrex are minimal but all inputs files in applications would need to know to use time_integration.*

This could be addressed in a future PR if we want to change it.

ajnonaka (approved):

Issue #3948, opened 2024-05-21
CMake: Update CUDA Logic to 3.24+


ax3l:
In CMake 3.24, the `native` flag was introduced for `CUDA_ARCHITECTURES`, which is what we currently do in our old logic.
https://cmake.org/cmake/help/latest/prop_tgt/CUDA_ARCHITECTURES.html

Transitioning is now possible (CMake is at 3.29) and we can then drop `cmake_policy(SET CMP0104 OLD)` that we used for the transition period.
https://github.com/AMReX-Codes/amrex/blob/24.05/CMakeLists.txt#L19-L27

PR #4572, opened 2025-07-17
Update CMake to support newer GPU architectures

samuelpmish:
## Summary

I was trying out [amr-wind](https://github.com/Exawind/amr-wind) earlier and found its CMake build was unable to configure for Blackwell architecture GPUs: 

```
cmake . -Bbuild [...] -DCMAKE_CUDA_ARCHITECTURES=100
...
...
CMake Error at /usr/share/cmake-4.0/Modules/FindCUDA/select_compute_arch.cmake:245 (message):
  Unknown CUDA Architecture Name 10.0 in CUDA_SELECT_NVCC_ARCH_FLAGS
Call Stack (most recent call first):
  Tools/CMake/AMReXUtils.cmake:265 (cuda_select_nvcc_arch_flags)
  Tools/CMake/AMReXParallelBackends.cmake:99 (set_cuda_architectures)
  Src/CMakeLists.txt:40 (include)
```

It seems that the underlying cause was not in amr-wind itself, but in AMReX's use of some deprecated CMake CUDA features.

This PR makes a small change to the CMake build system to avoid those deprecated features, so that AMReX can compile with Hopper and Blackwell architecture GPUs. The configuration behavior is as follows:

`cmake . -DAMReX_GPU_BACKEND=CUDA` 
will default to the "native" option (which selects architecture based on the hardware present in the machine)

`cmake . -DAMReX_GPU_BACKEND=CUDA -DCMAKE_CUDA_ARCHITECTURES=100` 
builds for the explicitly-specified architecture(s)

## Tasks

- [ ] ensure "native" compilation still picks he local GPU, if present, with the same precedence of user hints as before
- [ ] ensure user interface does not break, e.g., `AMREX_CUDA_ARCH` env hint still works and has the same precendence
- [ ] Update docs/logic on Device LTO and avoid to break users.
- [ ] Ensure that for HPC machines, we pre-compile (i.e., at least when AMREX_CUDA_ARCH is set / `CMAKE_CUDA_ARCHITECTURES` is selected to a narrow set) down to SASS code, otherwise we will in an MPI context compile every process on startup from PTX to SASS, potentially 10's of thousands of times.
- [ ] CUDA 12.9 raises: `Support for offline compilation for architectures prior to '<compute/sm/lto>_75' will be removed in a future release` 

## Checklist

The proposed changes:


ax3l:
Thank you for this!

Yes #3948 is over-due and it looks like we also have some breakage with the latest CUDA 12.9 and CMake now with the old logic.

I am OoO for the rest of the week, but we should try to get this in for the next release of AMReX and I will try to help next week.

Bumping CMake to 3.24+ globally is fine now, please go ahead.

Please check that we can use the build mode of building for the local "native" GPU when one is discovered, to simplify development. Otherwise, let us keep the AMREX_ARCH env hint (if set) that we use so far to set a default for CMAKE_CUDA_ARCHITECTURES.

Tools/CMake/AMReXCUDAOptions.cmake
```diff
@@ -17,16 +17,7 @@ endmacro ()
 #
 message(STATUS "Enabled CUDA options:")
 
-set(AMReX_CUDA_ARCH_DEFAULT "Auto")
-if(DEFINED CMAKE_CUDA_ARCHITECTURES)
-   set(AMReX_CUDA_ARCH_DEFAULT "${CMAKE_CUDA_ARCHITECTURES}")
-endif ()
-if(DEFINED ENV{AMREX_CUDA_ARCH})
-   set(AMReX_CUDA_ARCH_DEFAULT "$ENV{AMREX_CUDA_ARCH}")
```
ax3l, line 25: Let us keep that `$ENV{AMREX_CUDA_ARCH}` hint to avoid breakage for users.

If found, we simply set the default for `CMAKE_CUDA_ARCHITECTURES` to it.

Tools/CMake/AMReXParallelBackends.cmake
```diff
@@ -109,8 +109,14 @@ if (  AMReX_GPU_BACKEND STREQUAL "CUDA"
           )
   endforeach()
 
-   # Take care of cuda archs
-   set_cuda_architectures(AMReX_CUDA_ARCH)
+   # default to building for the architecture of the available GPU 
+   set(AMREX_CUDA_ARCHS native)
```
ax3l, line 113: Great!

ax3l:
@samuelpmish There is a lot of legacy logic in `Tools/CMake/AMReXUtils.cmake` and others where we wrange CUDA archs and CUDA device LTO flags.

Can you remove/clean those out if you have a chance?

I can help next week, too.

ax3l:
@cyrush can you potentially update the Catalyst image we use in AMReX/WarpX to include CMake 3.24 or newer? :pray:

ax3l:
@c-wetterer-nelson can you potentially update the SENSEI image we use in AMReX/WarpX to include CMake 3.24 or newer? :pray:

c-wetterer-nelson:
> 

Hey Axel, are there still SENSEI users across AMReX/WarpX?

cyrush:
@ax3l our current ascent containers (0.9.4) are using CMake 3.28. The internal layout changed a bit, but I can address that.
It looks like you are using Ascent 0.9.2.

Ascent 0.9.3 (also available) is using CMake 3.26.3.

So a quick update to ascent 0.9.3 will get you beyond CMake 3.24, down the road I can help get 0.9.4 working your CI.

ax3l:
@c-wetterer-nelson 

> Hey Axel, are there still SENSEI users across AMReX/WarpX?

That is a good question, I think in WarpX not anymore.

@WeiqunZhang should we drop SENSEI throughout AMReX & BLAST codes?

ax3l:
I have to move this PR and testing it downstream into the 25.09 release cycle, due to other deadlines.

There is a hotfix for CUDA 12.9 for now in #4589

ax3l:
I added a task list to the PR description on things we will need to carefully check with downstream codes to avoid breakage.

samuelpmish:
Hey, sorry to disappear for a bit after posting this PR. I'm not sure I understand AMReX's build system enough to address some of the tasks on my own. Can someone help clarify the requirements for the listed tasks:

```
- ensure "native" compilation still picks he local GPU, if present, with the same precedence of user hints as before
- ensure user interface does not break, e.g., AMREX_CUDA_ARCH env hint still works and has the same precendence
- Update docs/logic on Device LTO and avoid to break users.
- Ensure that for HPC machines, we pre-compile (i.e., at least when AMREX_CUDA_ARCH is set / CMAKE_CUDA_ARCHITECTURES is selected to a narrow set) down to SASS code, otherwise we will in an MPI context compile every process on startup from PTX to SASS, potentially 10's of thousands of times.
- CUDA 12.9 raises: Support for offline compilation for architectures prior to '<compute/sm/lto>_75' will be removed in a future release
```

It seems like some of them are already satisfied (e.g. native compilation picking the local GPU). I believe the `-arch=native` will generate SASS for the available GPUs already.

Other ones like the warning about sm_75 can be suppressed with a flag, but I'm not sure that's always a good thing (as it hides important info from users with those cards).

clarkse:
Thanks for digging into this @samuelpmish — I hit the same Blackwell wall and wanted to share a narrower variant that might unblock the task-list concerns @ax3l raised, since the main worry here is preserving the existing user/downstream interface.

**Root cause, for reference.** The silent failure (not just the configure error you saw) comes from `convert_cuda_archs()` in `Tools/CMake/AMReXUtils.cmake`, whose `LESS 10` guards — added in #4589 as a CUDA 12.9 / CMake 3.30 workaround — *drop* any compute capability ≥ 10.0. The arch list then goes empty, `set_cuda_architectures()` falls through to the deprecated `cuda_select_nvcc_arch_flags`, and that helper returns the newest arch it knows (sm_86). So on a Blackwell box `-DAMReX_CUDA_ARCH=120` configures "successfully" but silently builds **sm_86**, which then runs only via PTX-JIT. (Verified on both CMake 3.28 and 4.3.4 — upgrading CMake doesn't help because the strip happens in AMReX before CMake's selector runs.)

**A minimal, interface-preserving alternative.** Rather than removing `AMReX_CUDA_ARCH` / the env hint and defaulting to `native` (which is what stalls on the downstream-break checklist, and which also requires a GPU at configure time for HPC login-node cross-compiles), this keeps `set_cuda_architectures()` and the whole existing interface, and *only* routes compute capability ≥ 10.0 to CMake-native integer `CUDA_ARCHITECTURES` (e.g. `120`), which nvcc ≥ 12.8 emits directly. Everything < 10.0, plus names and `Auto`, still goes through the legacy helper exactly as before:

```diff
 function (set_cuda_architectures _cuda_archs)

    set(_archs ${${_cuda_archs}})
-   convert_cuda_archs(_archs)

-   include(FindCUDA/select_compute_arch)
-   cuda_select_nvcc_arch_flags(_nvcc_arch_flags ${_archs})
-
-   # Extract architecture number: anything less than 6.0 must go
-   string(REPLACE "-gencode;" "-gencode=" _nvcc_arch_flags "${_nvcc_arch_flags}")
-
-   foreach (_item IN LISTS _nvcc_arch_flags)
-      # Match one time the regex [0-9]+.
-      # [0-9]+ means any number between 0 and 9 will be matched one or more times (option +)
-      string(REGEX MATCH "[0-9]+" _cuda_compute_capability "${_item}")
-
-      if (_cuda_compute_capability LESS 60)
-         message(STATUS "Ignoring unsupported CUDA architecture ${_cuda_compute_capability}")
+   # Compute capability >= 10.0 (Hopper-next / Blackwell: sm_100, sm_103, sm_120, sm_121) cannot be
+   # represented by the deprecated FindCUDA cuda_select_nvcc_arch_flags bundled with CMake, and
+   # convert_cuda_archs() drops it via its `LESS 10` guards -> the list goes empty and the helper
+   # falls back to the newest arch it knows (sm_86), silently mis-building for Blackwell. Route such
+   # archs straight to CMake-native integer CUDA_ARCHITECTURES (e.g. 120), which nvcc >= 12.8
+   # understands, and keep the legacy helper only for older archs / names / "Auto".
+   set(_native_archs)
+   set(_legacy_archs)
+   foreach (_a IN LISTS _archs)
+      string(REGEX REPLACE "\\-real$" "" _a "${_a}")
+      set(_norm "${_a}")
+      if (_norm MATCHES "\\.")
+         string(REPLACE "." "" _norm "${_norm}")  # 12.0 -> 120, 8.6 -> 86
+      endif ()
+      if (_norm MATCHES "^[0-9]+$" AND NOT (_norm LESS 100))
+         list(APPEND _native_archs ${_norm})
       else ()
-         list(APPEND _tmp ${_cuda_compute_capability})
+         list(APPEND _legacy_archs ${_a})
       endif ()
    endforeach ()

+   set(_tmp ${_native_archs})
+
+   if (_legacy_archs)
+      convert_cuda_archs(_legacy_archs)
+
+      include(FindCUDA/select_compute_arch)
+      cuda_select_nvcc_arch_flags(_nvcc_arch_flags ${_legacy_archs})
+
+      # Extract architecture number: anything less than 6.0 must go
+      string(REPLACE "-gencode;" "-gencode=" _nvcc_arch_flags "${_nvcc_arch_flags}")
+
+      foreach (_item IN LISTS _nvcc_arch_flags)
+         # Match one time the regex [0-9]+.
+         # [0-9]+ means any number between 0 and 9 will be matched one or more times (option +)
+         string(REGEX MATCH "[0-9]+" _cuda_compute_capability "${_item}")
+
+         if (_cuda_compute_capability LESS 60)
+            message(STATUS "Ignoring unsupported CUDA architecture ${_cuda_compute_capability}")
+         else ()
+            list(APPEND _tmp ${_cuda_compute_capability})
+         endif ()
+      endforeach ()
+   endif ()
+
    set(AMREX_CUDA_ARCHS ${_tmp} CACHE INTERNAL "CUDA archs AMReX is built for")

 endfunction()
```

**Properties / how this maps onto the checklist:**
- `AMReX_CUDA_ARCH` cache var and the `AMREX_CUDA_ARCH` env hint keep working with the same precedence — no interface change.
- No GPU required at configure time when an arch is given explicitly (preserves HPC login-node cross-compile).
- Explicit arch → emits both `sm_NN` SASS and `compute_NN` PTX, so the MPI-startup JIT-storm concern in the checklist is avoided for the pre-compiled case.
- Pre-Blackwell archs and `Auto`/names are completely unchanged (still the legacy path), so the downstream-break surface is small.
- Tradeoff: this keeps leaning on the deprecated FindCUDA helper for < 10.0, so it's more of a surgical fix than the full migration this PR does. The two aren't mutually exclusive — happy to fold this into the existing approach, or move the whole thing to native once the `Auto`/precedence behavior is nailed down.

**Tested** (WarpX RZ hybrid-PIC, 2× RTX 5090 / sm_120, CMake 4.3.4, nvcc 13.2): with the patch, `-DAMReX_CUDA_ARCH=120` resolves `AMREX_CUDA_ARCHS=120`, every target emits `compute_120 sm_120`, `cuobjdump` on the built `.so` shows `arch = sm_120`, and there's no PTX-JIT at load. Same path should cover B200 (sm_100).

I've implemented this on a branch — [`clarkse/amrex@cmake-cuda-arch-blackwell`](https://github.com/clarkse/amrex/tree/cmake-cuda-arch-blackwell), based on current `development` — and am happy to open a PR with it if that's useful, or fold it into this one, whatever's least duplicative. cc @WeiqunZhang @ax3l — you'll know best which direction you'd prefer here. Thanks again for kicking this off.

ax3l:
Sorry for the delay here, the last year was ... special ;D

I think updating to the new CUDA CMake logic as this PR does (i.e., going to #3948 3.24+) and dropping all the hold mangling of CUDA archs is the right way forward. In fact, this is the code path used by most downstream projects that bump CMake high enough already.

I will start from this PR to ensure the env var stays in place and we get this merged.

PR #4575, opened 2025-07-18
Update Math::sincos and sincospi

WeiqunZhang:
- Revert the change made to scalar types. They are now non-templates.

  - Use std::enable_if to restrict function templates to simd types.

  - Fix sincospi for simd types.


ax3l:
Thanks! Testing w/ ImpactX now.

ax3l (approved):
Works perfectly, thanks a lot!

PR #4576, opened 2025-07-19
ParmParse: Remove two Aborts

WeiqunZhang:
ParmParse now supports names containing special characters such as `/`, `(`, `)`, and `,` except for white space and `=` (e.g., `dAx_x/dx(x,y,t,zeval) = 12`). In fact, even before this PR, such names were already supported as long as they are not the first entry, due to the way input parsing was implemented. With this PR, these names are now allowed as the first entry as well.


ax3l (approved):
Awesome :D

ax3l:
x-ref https://github.com/BLAST-ImpactX/impactx/pull/964

PR #4595, opened 2025-08-03
Add ParallelForOMP

WeiqunZhang:
ParallelFor does not do any OMP threading, because we do coarse-grained OMP threading at the MFIter level with tiling. But there are cases we use ParallelFor outside MFIter and they could benefit from OMP threading. Thus, we add new ParallelForOMP functions in this PR. Note that OMP threading is used only for CPU builds with OMP, otherwise the new ParallelForOMP functions are equivalent to ParallelFor functions.

### DIFF
--- a/Src/Base/AMReX_GpuLaunch.H
+++ b/Src/Base/AMReX_GpuLaunch.H
@@ -245,7 +245,7 @@ namespace amrex {
 
 #if defined(AMREX_USE_GPU) || !defined(AMREX_USE_OMP)
 
-template <typename T, typename L, typename M=std::enable_if_t<std::is_integral<T>::value> >
+template <typename T, typename L, typename M=std::enable_if_t<std::is_integral_v<T>> >
 void ParallelForOMP (T n, L const& f) noexcept
 {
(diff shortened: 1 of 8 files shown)


WeiqunZhang:
If needed, we can extend this to support BoxND<dim>.

Src/Base/AMReX_GpuLaunch.H
```diff
@@ -241,4 +241,113 @@ namespace Gpu {
...
+}
+
+template <typename T, typename L, typename M=std::enable_if_t<std::is_integral_v<T>> >
+AMREX_ATTRIBUTE_FLATTEN_FOR
+void ParallelForOMP (Box const& box, T ncomp, L const& f) noexcept
+{
+    auto lo = amrex::lbound(box);
+    auto hi = amrex::ubound(box);
+#if (AMREX_SPACEDIM == 1)
+#pragma omp parallel for collspase(2)
```
atmyers:
```suggestion
#pragma omp parallel for collapse(2)
```

atmyers (approved):

PR #4603, opened 2025-08-10
Improve support for Particles with PolymorphicArenaAllocator

AlexanderSinn:
## Summary

This PR adds a `SetArena(Arena*)` function to ParticleContainerBase that allows setting a memory arena that is used for all the particle vectors if the allocator is PolymorphicArenaAllocator. The function has to be called before particle tiles are defined.

This functionality is used to fix a bunch of places where a polymorphic vector would previously not have its arena set properly.

Additionally the `RunOnGpu` logic in `AMReX_WriteBinaryParticleData.H` is extended to work with a polymorphic allocator.

Uses changes extracted from https://github.com/AMReX-Codes/amrex/pull/4404.

## Additional background

Previously all components of all particle tiles needed their arena set individually by the user if PolymorphicArenaAllocator was used. In case this was done this PR is a braking change due to the `AMREX_ALWAYS_ASSERT_WITH_MESSAGE(a_arena != nullptr` assert in `ParticleTile::define()`.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_ParticleContainerBase.H
+++ b/Src/Particle/AMReX_ParticleContainerBase.H
@@ -245,6 +245,14 @@ public:
     template <class MF>
     bool OnSameGrids (int level, const MF& mf) const { return m_gdb->OnSameGrids(level, mf); }
 
+    [[nodiscard]] Arena* arena () const {
+        return m_arena;
+    }
+
+    void SetArena (Arena* a) {
+        m_arena = a;
+    }
+
     static const std::string& CheckpointVersion ();
     static const std::string& PlotfileVersion ();
     static const std::string& DataPrefix ();
@@ -269,6 +277,7 @@ protected:
     std::unique_ptr<ParGDB> m_gdb_object = std::make_unique<ParGDB>();
     ParGDBBase* m_gdb{nullptr};
     Vector<std::unique_ptr<MultiFab> > m_dummy_mf;
+    Arena* m_arena = nullptr;
 
     mutable std::unique_ptr<iMultiFab> redistribute_mask_ptr;
     mutable int redistribute_mask_nghost = std::numeric_limits<int>::min();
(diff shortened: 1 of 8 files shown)


atmyers (approved):

Src/Particle/AMReX_ParticleContainerBase.H
```diff
@@ -245,6 +245,14 @@ public:
     template <class MF>
     bool OnSameGrids (int level, const MF& mf) const { return m_gdb->OnSameGrids(level, mf); }
 
+    [[nodiscard]] Arena* arena () const {
+        return m_arena;
+    }
+
+    void SetArena (Arena* a) {
+        m_arena = a;
+    }
```
ax3l, line 254: @AlexanderSinn @atmyers I am wondering: should we add an assert here that setting the arena can only be done once?

There are already good asserts / exceptions now if a PC with an unset allocator is trying to add particles. Thus, the logic here could be as easy as: assert if `m_arena` is not `nullptr` when callint `SetArena`. What do you think?

I want to avoid that someone calls `SetArena` for an already defined PC with live particles and gets into an unexpected state by that.

```suggestion
    void SetArena (Arena* a) {
        AMREX_ALWAYS_ASSERT_WITH_MESSAGE(m_arena == nullptr, "SetArena can only be called once per PC!");
        m_arena = a;
    }
```

(Technically, we can make it more complicated, like allowing to switch the arena until any tiles/particles are added. That could be useful in some usage patterns when specializing a user container that might set a default in its constructor.)

ax3l: x-ref: https://github.com/AMReX-Codes/pyamrex/issues/460#issuecomment-3145495944

AlexanderSinn: There is the `a_arena == GetStructOfArrays().GetIdCPUData().arena()` check in ParticleTile that should catch this issue, but I guess we could add another check to ParticleContainerBase

PR #4606, opened 2025-08-13
amrex::callNoinline: Call given function without inline

WeiqunZhang:
This works on lambdas, functors, normal functions. But it does not work on overloaded functions like std::sin. If needed, one could however wrap functions like std::sin inside a lambda function. It also does not work with normal
functions for SYCL and one would have to wrap it inside a lambda.     

Here is the motivation behind this PR. In this impactx PR (https://github.com/BLAST-ImpactX/impactx/pull/964), a GPU kernel uses 8 amrex::Parser's. The CUDA CI fails if more than one job is used in build. Apparently the kernel is too big because all those parser functions are inlined. This PR provides a way to reduce the size by forcing noinline.


ax3l (approved):
Thank you!

PR #4611, opened 2025-08-15
Curl Curl Solver: Support for 1D spherical and 1D cylindrical

WeiqunZhang:
This PR adds support for 1D spherical and 1D cylindrical. Note that the curl of curl of E is zero in the radial direction. For 1D spherical coordinates, E_theta and E_phi have reflect-odd symmetry, and are thus zero at R = 0. For 1D cylindrical coordinates, E_theta has reflect-odd symmetry, and is thus zero at r = 0. E_z has reflect-even symmetry at r = 0, where the curl-curl is -4(Ez(1)-Ez(0))/dr^2 in the z-direction.


WeiqunZhang:
Could we get this approved? It's a new feature that is used by WarpX in 1D. It will not break any other codes.

asalmgren (approved):

PR #4612, opened 2025-08-15
Add ParamParse::getPrefix

WeiqunZhang:
It was removed in #4031. But it is still used by ImpactX in an assert, and it is still useful.


ax3l (approved):
Thank you :)

PR #4616, opened 2025-08-18
AMReX Enum: Underlying Value

ax3l:
## Summary

Sometimes one needs the underlying (u)int value of an `enum class`, e.g., when building bitmasks. This adds a helper function to `AMREX_ENUM`.

## Additional background

https://github.com/BLAST-ImpactX/impactx/pull/1104

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_Enum.H
+++ b/Src/Base/AMReX_Enum.H
@@ -135,7 +135,7 @@ namespace amrex
      */
     template <typename T, typename ET = amrex_enum_traits<T>,
               std::enable_if_t<ET::value,int> = 0>
-    constexpr auto getEnumUnderlyingValue (T v) noexcept
+    constexpr auto toUnderlying (T v) noexcept
     {
         return static_cast<std::underlying_type_t<T>>(v);
     }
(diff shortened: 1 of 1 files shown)


Src/Base/AMReX_Enum.H
```diff
@@ -128,6 +128,17 @@ namespace amrex
...
+    /** Return the underlying (u)int of an enum value
+     *
+     * Useful when building bitmasks.
+     */
+    template <typename T, typename ET = amrex_enum_traits<T>,
+              std::enable_if_t<ET::value,int> = 0>
+    constexpr auto getEnumUnderlyingValue (T v) noexcept
+    {
+        return static_cast<std::underlying_type_t<T>>(v);
+    }
```
ax3l: I can make `T` also an `int`, because we hard-code `int` in `AMREX_ENUM` for now. Let me know what you prefer.

WeiqunZhang: I think this is fine.

WeiqunZhang: https://en.cppreference.com/w/cpp/utility/to_underlying.html We could also use a name that is similar to the future C++ standard, like `toUnderlying`.

ax3l:
@WeiqunZhang feel free to update. I am mostly family sick and rest in proposals this week -.-

WeiqunZhang (approved):

PR #4617, opened 2025-08-18
Enable zero-sized Type in TypeMultiplier

AlexanderSinn:
## Summary

In C++ it is not allowed to make a zero-sized array type like `int[0]` so a more verbose way using `amrex::TypeArray<int, 0>` needs to be used. Additionally, `std::declval` is now used instead of `Types{}`.

## Additional background

Zero-variants in https://github.com/BLAST-ImpactX/impactx/pull/1104

## Checklist

The proposed changes:


ax3l (approved):
Oh yay!

WeiqunZhang (approved):

PR #4618, opened 2025-08-18
Fix potential false sharing in ReduceOps.eval with OMP

AlexanderSinn:
## Summary

Looking at the code I noticed that eval() with OMP could be prone to false sharing.
TBD if this is actually the case and if this PR improves it

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_Reduce.H
+++ b/Src/Base/AMReX_Reduce.H
@@ -1059,19 +1059,22 @@ public:
 #ifdef AMREX_USE_OMP
 #pragma omp parallel
 #endif
-        for (MFIter mfi(mf,true); mfi.isValid(); ++mfi) {
-            Box const& b = mfi.growntilebox(nghost);
-            const int li = mfi.LocalIndex();
-            auto& rr_ref = reduce_data.reference(OpenMP::get_thread_num());
-            auto rr = rr_ref;
-            const auto lo = amrex::lbound(b);
-            const auto hi = amrex::ubound(b);
-            for (int k = lo.z; k <= hi.z; ++k) {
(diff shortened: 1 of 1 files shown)


Src/Base/AMReX_Reduce.H
```diff
@@ -1062,14 +1062,16 @@ public:
         for (MFIter mfi(mf,true); mfi.isValid(); ++mfi) {
             Box const& b = mfi.growntilebox(nghost);
             const int li = mfi.LocalIndex();
-            auto& rr = reduce_data.reference(OpenMP::get_thread_num());
+            auto& rr_ref = reduce_data.reference(OpenMP::get_thread_num());
+            auto rr = rr_ref;
```
WeiqunZhang: We could try to reduce the false sharing further.
```
#ifdef AMREX_USE_OMP
#pragma parallel
#endif
{
    auto& rr_ref = reduce_data.reference(OpenMP::get_thread_num());
    ReduceTuple rr;
    Reduce::detail::for_each_init<0, ReduceTuple, Ps...>(rr);
    for (MFIter ...) {
    }
    Reduce::detail::for_each_local<0,ReduceTuple,Ps...>(rr_ref, rr);
}
```

AlexanderSinn:
Initial benchmarks show only a very small difference, if any

WeiqunZhang:
In any way, it will not hurt.

WeiqunZhang (approved):

zingale:
this changes answers in Castro's CI -- is that to be expected?

AlexanderSinn:
A change in the order of machine precision is expected due to floating-point addition not being associative to that precision.

PR #4619, opened 2025-08-20
Updated applyBC to use ParallelFor on Vector of tags for GPU builds

ankithadas:
## Checklist

The proposed changes:


WeiqunZhang:
Sorry for taking so long. We have not got a chance to test it yet. But we have not forgot about it.

ankithadas:
Looks like I sprinkled some rather unfortunate bugs. Thanks for catching them!

WeiqunZhang:
Thank you for doing the bulk of the work!

WeiqunZhang (approved):

PR #4626, opened 2025-08-22
Incorrect cross term sum for 2D EB terms in MLEBTensor

ankithadas:
## Checklist

The proposed changes:


asalmgren:
It looks like this is "just" a sign change -- is that correct?  If so, could you modify this PR to change "+=" to +-=" and leave the rest the same?

asalmgren (approved):

ankithadas:
Thank you for approving the PR. I aimed to maintain the same structure as its 3D counterpart.

PR #4629, opened 2025-08-27
Adding support to memory collection of Particles and MultiFabs within a grid

jessdagostini:
## Summary
This PR adds functions to return memory usage from `Particles` and `FabArray`/`MultiFabs` within a grid/box. The main goal is to use this information for load balancing. We aim to consider the memory allocated to a worker as a factor in the load-balancing algorithms for AMReX, so there is a need to capture specific grid/box memory information.

## Checklist

The proposed changes:


Src/Base/AMReX_MultiFabUtil.H
```diff
@@ -1070,6 +1070,12 @@ void average_down_faces (const FabArray<FAB>& fine, FabArray<FAB>& crse,
     crse.ParallelCopy(ctmp,0,0,crse.nComp(),0,0,crse_geom.periodicity());
 }
 
+// Compute the memory capacity (in bytes) used by a Box
+// We have tried to work with more templated version, but it didn't work. What is the most appropriate way to do it?
+LayoutData<std::size_t> CapacityOfFabs (BoxArray const& ba,
+                 DistributionMapping const& dm,
+                 std::vector<MultiFab*> const& mfs);
```
WeiqunZhang: Could we remove BoxArray and DistributionMapping from the function arguments? The assumption here is all multifabs must have the same boxarray (ignoring the index type) and distributionmapping.

Could make argument be `Vector<MultiFab const*> const& mfs`. Vector provides bound check if assertion is on. We are not changing the data in mulitfabs. Thus we should pass them in as `const*`.

If we want it be consistent with how it's done in particle container, we could make this a member function of FabArray. That might be better. That will also solve the template "issue".

WeiqunZhang (changes requested):

Src/Base/AMReX_MultiFabUtil.H
jessdagostini: I believe it's possible to remove the `BoxArray` and the `DistributionMapping` if we have a way to access this information from `MultiFabUtil`. In the implementation at the `.cpp` file, we need those to create our `LayoutData` object that will hold the memory information for each box (https://github.com/AMReX-Codes/amrex/pull/4629/commits/7a29e92ef33cf1c404bd883e6bd546ede8fe54b8#diff-e104e0f29e7a0e37ad2ee839c42e2c317091f995556dcc35c18b47d66506761fR67)

Sounds good about the Vector! We weren't sure if our approach was the best :)

We have tried to implement this function over `FabArrayUtil`, and we had a hard time calling it from an "user-end" code. If you have a suggestion on how to make it work in the FabArray I would appreciate!
For context, I have tested both implementations using examples from the [AMReX Tutorials](https://github.com/AMReX-Codes/amrex-tutorials).

More specifically, I have used this example for the Particles https://github.com/AMReX-Codes/amrex-tutorials/blob/main/ExampleCodes/Particles/NeighborList/main.cpp and this for the MultiFabs https://github.com/AMReX-Codes/amrex-tutorials/blob/main/ExampleCodes/Basic/HeatEquation_EX0_C/Source/main.cpp
On the MultiFabs, I used the following code to get the memory information
```C++
std::vector<amrex::MultiFab*> multifabs_to_check;
multifabs_to_check.push_back(&phi_old);
multifabs_to_check.push_back(&phi_new);

amrex::LayoutData<std::size_t> mem_usage = amrex::CapacityOfFabs(ba, dm, multifabs_to_check);

long total_bytes = 0;
for (amrex::MFIter mfi(mem_usage); mfi.isValid(); ++mfi) {
       amrex::Print() << "Box " << mfi.index() << " uses "
                   << mem_usage[mfi] << " bytes\n";
 }
  ```

Please, let me know if I am missing something or if there's another approach for this!

Comment (WeiqunZhang) [reply Src/Base/AMReX_MultiFabUtil.H]: MultiFab is derived from FabArray, which has member functions that return BoxArray and DistributionMapping used to build the MultiFab. Something like
```C++
auto const& ba = mf_ptr->boxArray();
auto const& dm = mf_ptr->DistributionMap();
```

jessdagostini: Thank you! I have pushed an updated version with adjustments. I have tested with the same code example as shared above and got the same results, so I believe it is producing correct data.
About the point on 
> Vector provides bound check if assertion is on. 

I am not doing any assertions at this point. Should I?

Src/Particle/AMReX_ParticleContainer.H
```diff
@@ -576,6 +576,12 @@ public:
 
     Vector<Long> NumberOfParticlesInGrid  (int level, bool only_valid = true, bool only_local = false) const;
 
+    /**
+     * \brief Return capacity of memory for particles at specific grid
+     */
+
+    LayoutData<std::size_t> CapacityOfParticlesInGrid  (int lev) const;
```
atmyers: Could you rename this to "CapacityofParticleTiles", to make it more consistent with "CapacityofFabs"?

Src/Base/AMReX_MultiFabUtil.H
atmyers: I agree with @WeiqunZhang's suggestion to make `CapacityofFabs` a member function of `FabArray`. Other than that, looks good to me.

Src/Particle/AMReX_ParticleContainer.H
jessdagostini: I am not entirely familiar with AMReX nomenclatures, but are 'tiles' also a corresponding term for grid/box? In this function, we focus on determining the memory usage of particles within a Box.

atmyers: Ah, okay. Never mind then, your name is better.

jessdagostini: Cool, thanks!

Src/Base/AMReX_MultiFabUtil.H
jessdagostini: Ok! Do you have suggestions on how could I implement it over `FabArray` and make it callable on "user-end" functions? I have tried something similar to this on `FabArray`
```C++
template <class FAB, class foo = std::enable_if_t<IsBaseFab<FAB>::value> >
LayoutData<std::size_t> CapacityOfFabs (const Vector<const MultiFab*>& mfs)
{
    LayoutData<std::size_t> mem(mfs[0]->boxArray(), mfs[0]->DistributionMap());
    for (auto const& mf : mfs) {
        for (MFIter mfi(*mf, MFItInfo{}.DisableDeviceSync()); mfi.isValid(); ++mfi) {
            mem[mfi] += (*mf)[mfi].nBytesOwned();
        }
    }
    return mem;
}
```

and had trouble calling this with the following code:
```C++
amrex::Vector<const amrex::MultiFab*> multifabs_to_check;
multifabs_to_check.push_back(&phi_old);
multifabs_to_check.push_back(&phi_new);
amrex::LayoutData<std::size_t> mem_usage = amrex::CapacityOfFabs(multifabs_to_check);
```
(the above code is based on the amrex-tutorials https://github.com/AMReX-Codes/amrex-tutorials/blob/main/ExampleCodes/Basic/HeatEquation_EX0_C/Source/main.cpp)

The error I always got in this setting was something about `FabArray` not recognising the `MultiFab` vector

atmyers: Maybe the interface should be to have `CapacityOfFabs` take a `LayoutData<std::size_t>` into which to increment the result. Then you'd use it like:

```C++
LayoutData<std::size_t> mem(phi_old.boxArray(), phi_old.DistributionMap());
phi_old.CapacityOfFabs(mem);
phi_new.CapacityOfFabs(mem);
```

and similarly with `CapacityOfParticlesInGrid`.

jessdagostini: Got it! I have made the code changes to go on this layout, tested, and it worked fine! I have pushed the changes to this PR  in the https://github.com/AMReX-Codes/amrex/pull/4629/commits/db786f37f19eb0a4c32779319e7ef27cb407354a commit.

Src/Base/AMReX_FabArray.H
```diff
@@ -2506,6 +2510,16 @@ FabArray<FAB>::setDomainBndry (value_type val,
     }
 }
 
+template <class FAB>
+template <class F, std::enable_if_t<IsBaseFab<F>::value,int> FOO>
+void
+FabArray<FAB>::capacityOfFabs(LayoutData<std::size_t>& mem) {
+    // Should we have assertions here? What should be asserted?
```
atmyers: It would be good to assert that `mem` and `*this` use the same BoxArray and DistributionMap. You can use the `CellEqual` method for BoxArray and the `==` operator for the DistributionMap.

jessdagostini: Got it! Just pushed the assertion on https://github.com/AMReX-Codes/amrex/pull/4629/commits/7a29e92ef33cf1c404bd883e6bd546ede8fe54b8

Src/Base/AMReX_MultiFabUtil.H
atmyers: Nice, do you also want to change the particle version to use the same interface? I think this will be helpful when there is more than on `ParticleContainer` in an application (as there is in WarpX).

Src/Base/AMReX_MultiFabUtil.H
```diff
@@ -1070,6 +1070,10 @@ void average_down_faces (const FabArray<FAB>& fine, FabArray<FAB>& crse,
     crse.ParallelCopy(ctmp,0,0,crse.nComp(),0,0,crse_geom.periodicity());
 }
 
+// Compute the memory capacity (in bytes) used by a Box
+// We have tried to work with more templated version, but it didn't work. What is the most appropriate way to do it?
+LayoutData<std::size_t> CapacityOfFabs (const Vector<const MultiFab*>& mfs);
```
atmyers: Is this needed any more?

jessdagostini: Oh, no! Forgot to remove it! Pushing it now

Src/Base/AMReX_MultiFabUtil.H
jessdagostini: Yes, I think it makes sense to change `ParticleContainer` to this format! The only concern is that doing so will result in this function having a slightly different signature than other `ParticleContainer` functions that perform similar tasks, but are not directly memory-related (like https://github.com/AMReX-Codes/amrex/blob/development/Src/Particle/AMReX_ParticleContainerI.H#L496). Do you think this is ok?

atmyers: I'm okay with that. That function has a slightly different use case. I believe it was added for IO, where we write separate files for each species anyway, so we don't need to aggregate the totals over multiple species, as you do here.

jessdagostini: Ok! Working on it!

jessdagostini: Done! https://github.com/AMReX-Codes/amrex/pull/4629/commits/7fb20d436dc3b963608267a8b6e71b2b3ddf8bee

Src/Base/AMReX_FabArray.H
```diff
@@ -2506,6 +2510,17 @@ FabArray<FAB>::setDomainBndry (value_type val,
     }
 }
 
+template <class FAB>
+template <class F, std::enable_if_t<IsBaseFab<F>::value,int> FOO>
+void
+FabArray<FAB>::capacityOfFabs(LayoutData<std::size_t>& mem) {
+    AMREX_ASSERT(*this.CellEqual(mem.boxArray()) &&
```
atmyers:
```suggestion
    AMREX_ASSERT(this->CellEqual(mem.boxArray()) &&
```

Src/Base/AMReX_FabArray.H
```diff
@@ -2506,6 +2510,17 @@ FabArray<FAB>::setDomainBndry (value_type val,
     }
 }
 
+template <class FAB>
+template <class F, std::enable_if_t<IsBaseFab<F>::value,int> FOO>
+void
+FabArray<FAB>::capacityOfFabs(LayoutData<std::size_t>& mem) {
+    AMREX_ASSERT(this->CellEqual(mem.boxArray()) &&
```
atmyers:
```suggestion
    AMREX_ASSERT(this->boxArray().CellEqual(mem.boxArray()) &&
```

jessdagostini:
I have committed fixes for errors pointed out by two pipelines. There were two other pipelines "oneAPI" that failed because of a GPG key error. Not sure if I can fix this on my side

Src/Base/AMReX_FabArray.H
```diff
@@ -817,6 +817,10 @@ public:
     template <class F=FAB, std::enable_if_t<IsBaseFab<F>::value,int> = 0>
     void setDomainBndry (value_type val, int strt_comp, int ncomp, const Geometry& geom);
 
+    // Get capacity of the FabArray
+    template <class F=FAB, std::enable_if_t<IsBaseFab<F>::value,int> = 0>
+    void capacityOfFabs (LayoutData<std::size_t>& mem);
```
WeiqunZhang:
```suggestion
    void capacityOfFabs (LayoutData<std::size_t>& mem) const;
```

Src/Base/AMReX_FabArray.H
```diff
@@ -2506,6 +2510,17 @@ FabArray<FAB>::setDomainBndry (value_type val,
     }
 }
 
+template <class FAB>
+template <class F, std::enable_if_t<IsBaseFab<F>::value,int> FOO>
+void
+FabArray<FAB>::capacityOfFabs(LayoutData<std::size_t>& mem) {
```
WeiqunZhang:
```suggestion
FabArray<FAB>::capacityOfFabs (LayoutData<std::size_t>& mem) const {
```

Src/Base/AMReX_FabArray.H
```diff
@@ -2506,6 +2510,17 @@ FabArray<FAB>::setDomainBndry (value_type val,
...
+    AMREX_ASSERT(this->boxArray().CellEqual(mem.boxArray()) &&
+                 this->DistributionMap() == mem.DistributionMap());
```
WeiqunZhang:
```suggestion
    AMREX_ASSERT(amrex::isMFIterSafe(*this, mem));
```
`CellEqual` is more expensive and it returns false for cases that are safe (e.g., coarsened MultiFabs).

Src/Particle/AMReX_ParticleContainerI.H
```diff
@@ -586,6 +586,23 @@ Long ParticleContainer_impl<ParticleType, NArrayReal, NArrayInt, Allocator, Cell
...
 
+template <typename ParticleType, int NArrayReal, int NArrayInt,
+          template<class> class Allocator, class CellAssignor>
+void
+ParticleContainer_impl<ParticleType, NArrayReal, NArrayInt, Allocator, CellAssignor>::CapacityOfParticlesInGrid (LayoutData<std::size_t>& mem, int lev) const
+{
+    AMREX_ASSERT(lev >= 0 && lev < int(m_particles.size()));
+
+    AMREX_ASSERT(mem.boxArray().CellEqual(ParticleBoxArray(lev)) &&
+                 mem.DistributionMap() == ParticleDistributionMap(lev));
```
WeiqunZhang:
```suggestion
    AMREX_ASSERT(BoxArray::SameRefs(mem.boxArray(), ParticleBoxArray(lev)) &&
                 mem.DistributionMap() == ParticleDistributionMap(lev));
```

Src/Particle/AMReX_ParticleContainerI.H
```diff
@@ -586,6 +586,23 @@ Long ParticleContainer_impl<ParticleType, NArrayReal, NArrayInt, Allocator, Cell
...
+
```
WeiqunZhang, line 599:
```suggestion

    [[maybe_unused]] Gpu::NoSyncRegion no_sync{};
```

Src/Particle/AMReX_ParticleContainer.H
```diff
@@ -576,6 +576,13 @@ public:
 
     Vector<Long> NumberOfParticlesInGrid  (int level, bool only_valid = true, bool only_local = false) const;
 
+    /**
+     * \brief Return capacity of memory for particles at specific grid
+     *  Level is defaulted to 0.
+     */
+
+    void CapacityOfParticlesInGrid  (LayoutData<std::size_t>& mem, int lev = 0) const;
```
WeiqunZhang: It's probably a good idea to make `int lev` required.
```suggestion
    void CapacityOfParticlesInGrid  (LayoutData<std::size_t>& mem, int lev) const;
```

WeiqunZhang (changes requested):
The function is straightforward enough that I don't think we need to add tests for it.

Src/Base/AMReX_FabArray.H
```diff
@@ -817,6 +817,10 @@ public:
     template <class F=FAB, std::enable_if_t<IsBaseFab<F>::value,int> = 0>
     void setDomainBndry (value_type val, int strt_comp, int ncomp, const Geometry& geom);
 
+    // Get capacity of the FabArray
+    template <class F=FAB, std::enable_if_t<IsBaseFab<F>::value,int> = 0>
```
WeiqunZhang: For flexibility, we could also make the integer type a template typename here and for in particle containers. For example,
```C++
template <typename I, class F=FAB, std::enable_if_t<IsBaseFab<F>::value && std::is_integral_v<I> && sizeof(I) >= sizeof(Long), int> = 0>
void capacityOfFabs (LayoutData<I>& mem) const;
```
And in the implementation, we also want to do explicit static_cast before adding the values.

jessdagostini: Ok! Working on the update for this!

atmyers:
> The function is straightforward enough that I don't think we need to add tests for it.

I asked Jessica to add those. Maybe these new functions could just be called somewhere in an existing test, instead creating new ones?

Src/Base/AMReX_FabArray.H
jessdagostini: Actually, do you think it's worth doing it? Because I know that it is a best practice in C++ to use `std::size_t` for memory-related math (`size_t` was created for holding `sizeof` returns https://en.cppreference.com/w/cpp/types/size_t.html)

WeiqunZhang: Actually the best practice is proably to avoid unsigned integers in almost all cases.

Even Stroustrup does not like unsigned integers. https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1428r0.pdf Stroustrup said,
> Unfortunately, sizeof yields an unsigned (and it would be hard to change that), but we don’t have to
follow that for all types with something to do with sizes

Java does not even have build-in unsigned integer types. https://www.artima.com/articles/james-gosling-on-java-may-2001 Goshling said,
> In programming language design, one of the standard problems is that the language grows so complex that nobody can understand it. One of the little experiments I tried was asking people about the rules for unsigned arithmetic in C. It turns out nobody understands how unsigned arithmetic in C works. There are a few obvious things that people understand, but many people don't understand it.

jessdagostini: Hmm, got it! Interesting that the unsigned is such unknown territory. Will start working on including the template you suggested in both functions

jessdagostini:
> > The function is straightforward enough that I don't think we need to add tests for it.
> 
> I asked Jessica to add those. Maybe these new functions could just be called somewhere in an existing test, instead creating new ones?

I could include the Particles call in the `NeighborList` test. Not sure if it's the best option. I am not sure where to put the FabArray one, though.

Src/Base/AMReX_FabArray.H
jessdagostini: Just update both function signatures with the new template https://github.com/AMReX-Codes/amrex/pull/4629/commits/55d4a3489967204334bb17b4a1b0563b90826cfa

atmyers:
> > > The function is straightforward enough that I don't think we need to add tests for it.
> > 
> > 
> > I asked Jessica to add those. Maybe these new functions could just be called somewhere in an existing test, instead creating new ones?
> 
> I could include the Particles call in the `NeighborList` test. Not sure if it's the best option. I am not sure where to put the FabArray one, though.

I think that if you 1) Remove the new particle test added in this PR, in lieu of the modification to the existing NeighborList one, and 2) move/rename the MultiFab one to `amrex/Tests/HeatEquation`, then this should be good to go.

jessdagostini:
> > > > The function is straightforward enough that I don't think we need to add tests for it.
> > > 
> > > 
> > > I asked Jessica to add those. Maybe these new functions could just be called somewhere in an existing test, instead creating new ones?
> > 
> > 
> > I could include the Particles call in the `NeighborList` test. Not sure if it's the best option. I am not sure where to put the FabArray one, though.
> 
> I think that if you 1) Remove the new particle test added in this PR, in lieu of the modification to the existing NeighborList one, and 2) move/rename the MultiFab one to `amrex/Tests/HeatEquation`, then this should be good to go.

Sounds good! Did that on https://github.com/AMReX-Codes/amrex/pull/4629/commits/64e1e8479fd478d9eefbcd70a180a51968ec2152

atmyers (approved):

WeiqunZhang:
LGTM. Hopefully, one last thing. For the test to run, we need to add it to Line 128 of amrex/Tests/CMakeLists.txt.

jessdagostini:
> LGTM. Hopefully, one last thing. For the test to run, we need to add it to Line 128 of amrex/Tests/CMakeLists.txt.

Cool! Pushed here https://github.com/AMReX-Codes/amrex/pull/4629/commits/2b010978ecc8800fed546bd84e68bad3da48ef22

Tests/CMakeLists.txt
```diff
@@ -125,9 +125,9 @@ else()
    #
    # List of subdirectories to search for CMakeLists.
    #
-   set( AMREX_TESTS_SUBDIRS Amr AsyncOut CallNoinline CLZ CommType CTOParFor DeviceGlobal Enum
-                            MultiBlock MultiPeriod ParmParse Parser Parser2 ParserUserFn Reinit
-                            RoundoffDomain SmallMatrix)
+   set( AMREX_TESTS_SUBDIRS Amr AsyncOut CallNoinline CLZ CommType CTOParFor DeviceGlobal
+                            Enum HeatEquation MultiBlock MultiPeriod ParmParse Parser
```
WeiqunZhang:
```suggestion
                            Enum HeatEquation MultiBlock MultiPeriod ParmParse Parser
```

WeiqunZhang (approved):

PR #4635, opened 2025-09-01
Use `p.id().is_valid()` consistently

AlexanderSinn:
## Summary

This PR applies further cleaning to particles.

The largest change is the replacement of all numeric id comparisons with `id().is_valid()` and `id().make_invlid()`. Previously, `id() > 0`, `id() < 0`, `id() < 1` and `id() <= 0` with `id() = - id()` were used, making the treatment of id == 0 inconsistent (particle ids start at 1 because of this).

Both `ptd.m_aos[i]` and `make_particle<ConstParticleType>{}(ptd, ip);` were replaced with `ptd[i]` which is more concise and works with PureSoA particles.

The PureSoA specialization of numParticlesOutOfRange was removed because the other version also works with PureSoA particles.

## Checklist

The proposed changes:


atmyers (approved):

atmyers:
Thanks!

PR #4643, opened 2025-09-05
`AMREX_ENUM`: Cache String Map

ax3l:
## Summary

The `getEnumNameValuePairs` logic used below a lot of `AMREX_ENUM` functions is very slow for long-enough (e.g., N>70) enums.

In WarpX 1D simulations, it took up close to 50% of the runtime, as Dave Grote found.

This caches the key-value string list on first creation, because the type is known at compile-time.

A better implementation might skip the dictionary altogether for functions that just need one string, like `getEnumNameString`.

## Additional background

https://github.com/BLAST-WarpX/warpx/pull/6139

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_Enum.H
+++ b/Src/Base/AMReX_Enum.H
@@ -52,7 +52,7 @@ namespace amrex
               std::enable_if_t<ET::value,int> = 0>
     T getEnum (std::string_view const& s)
     {
-        static auto const& kv = getEnumNameValuePairs<T>();
+        static auto const kv = getEnumNameValuePairs<T>();
         auto it = std::find_if(kv.begin(), kv.end(),
                                [&] (auto const& x) -> bool
                                    { return x.first == s; });
@@ -71,7 +71,7 @@ namespace amrex
               std::enable_if_t<ET::value,int> = 0>
     T getEnumCaseInsensitive (std::string_view const& s)
     {
-        static auto const& kv = getEnumNameValuePairs<T>();
+        static auto const kv = getEnumNameValuePairs<T>();
         std::string ls = amrex::toLower(std::string(s));
         auto it = std::find_if(kv.begin(), kv.end(),
                                [&] (auto const& x) -> bool
@@ -91,7 +91,7 @@ namespace amrex
               std::enable_if_t<ET::value,int> = 0>
     std::string getEnumNameString (T const& v)
     {
-        static auto const& kv = getEnumNameValuePairs<T>();
+        static auto const kv = getEnumNameValuePairs<T>();
         auto it = std::find_if(kv.begin(), kv.end(),
                                [&] (auto const& x) -> bool
                                    { return x.second == v; });
--- a/Src/Base/AMReX_Enum.H
+++ b/Src/Base/AMReX_Enum.H
@@ -17,9 +17,11 @@ using amrex_enum_traits = decltype(amrex_get_enum_traits(std::declval<T>()));
 
 namespace amrex
 {
+namespace detail
+{
     template <typename T, typename ET = amrex_enum_traits<T>,
               std::enable_if_t<ET::value,int> = 0>
-    std::vector<std::pair<std::string,T>> getEnumNameValuePairs ()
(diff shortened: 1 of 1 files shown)


Src/Base/AMReX_Enum.H
```diff
@@ -52,7 +52,7 @@ namespace amrex
               std::enable_if_t<ET::value,int> = 0>
     T getEnum (std::string_view const& s)
     {
-        auto const& kv = getEnumNameValuePairs<T>();
+        static auto const& kv = getEnumNameValuePairs<T>();
```
ax3l: I do not get what the original `&` was aiming to achieve here, because the `getEnumNameValuePairs` function returns by value.

WeiqunZhang: https://herbsutter.com/2008/01/01/gotw-88-a-candidate-for-the-most-important-const/

ax3l: Nice thx. TIL.

I think it still needs to go now that it is static (?)

WeiqunZhang:
I think we should create a helper function to avoid multiple statics. Something like
```
    template <typename T, typename ET = amrex_enum_traits<T>,
              std::enable_if_t<ET::value,int> = 0>
    std::vector<std::pair<std::string,T>> getEnumNameValuePairs_helper ()
    {
            // the function body of current getEnumNameValuePairs
    }

    template <typename T, typename ET = amrex_enum_traits<T>,
              std::enable_if_t<ET::value,int> = 0>
    std::vector<std::pair<std::string,T>> const& getEnumNameValuePairs ()
    {
        static auto r = getNumNameValuePairs_helper<T>();
        return r;
    }
````

File: Src/Base/AMReX_Enum.H (hunk 1 above)
Comment (WeiqunZhang) [reply Src/Base/AMReX_Enum.H]: By the way, although I am not against using `const`, I use it less frequently than you. Anyway, always const is certainly wrong. Here is an example.
```
std::vector<int> make_vector()
{
     std::vector<int> const r(...);
     // code using r
     return r;
}

auto r = make_vector();
```
That `const` will prevent move, I think.

**Comment by ax3l** [comment]
Can we even create an optimized version of `getEnumNameString` that avoids creating a list of all (unused) names altogether? @WeiqunZhang

File: Src/Base/AMReX_Enum.H (hunk 1 above)
Comment (ax3l) [reply Src/Base/AMReX_Enum.H]: That is correct. There is a similar case preferring by-value arg passing (over `const &`).

**Comment by WeiqunZhang** [comment]
Depending the size, maybe using std::map or std::unordered_map is better than std::vector. But it's hard to avoid building the whole list.

**Comment by ax3l** [comment]
@WeiqunZhang ready from my end

@dpgrote can you please test if this solves your performance issue on WarpX 1D runs? :)

**Comment by WeiqunZhang** [comment]
We could also test alternative versions using std::map or std::unordered_map.

File: Src/Base/AMReX_Enum.H [hunk 2]
```diff
@@ -47,6 +49,16 @@ namespace amrex
         }
         return r;
     }
+}  // namespace detail
+
+    template <typename T, typename ET = amrex_enum_traits<T>,
+              std::enable_if_t<ET::value,int> = 0>
+    std::vector<std::pair<std::string,T>> getEnumNameValuePairs ()
```
WeiqunZhang, line 56: We should return by const&.

dpgrote:
This seems to work, thanks! I did a test comparing the time spent in the `getExtractedName` in WarpX. The original version was taking 0.8 seconds. With this PR, it is 0.1 seconds. With the PR in WarpX, it is 0.07 seconds. Not sure why that would be slightly faster.

WeiqunZhang:
Probably because the WarpX PR uses std::map.

WeiqunZhang (approved):

Src/Base/AMReX_Enum.H
WeiqunZhang: Oh, we never resolved this comment. I believe we should return by const& for performance.

ax3l: Let's do in #4655

PR #4652, opened 2025-09-09
ParmParse: Add prettyPrintUsedInputs and prettyPrintUnusedInputs

WeiqunZhang:
Add two new functions to ParmParse. The existing function prettyPrintTable prints all entries. We can now use prettyPrintUsedInputs to print used inputs, and use prettyPrintUnusedInputs to print unused inputs.

### DIFF
--- a/Src/Base/AMReX_ParmParse.H
+++ b/Src/Base/AMReX_ParmParse.H
@@ -1654,9 +1654,17 @@ public:
     //! Write the contents of the table in ASCII to the ostream.
     static void dumpTable (std::ostream& os, bool prettyPrint = false);
 
-    //! Write the table in a pretty way to the ostream. If there are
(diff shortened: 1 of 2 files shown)


Src/Base/AMReX_ParmParse.H
```diff
@@ -1654,10 +1654,18 @@ public:
     //! Write the contents of the table in ASCII to the ostream.
     static void dumpTable (std::ostream& os, bool prettyPrint = false);
 
-    //! Write the table in a pretty way to the ostream. If there are
+    //! Write all entried in the table in a pretty way to the ostream. If there are
```
atmyers:
```suggestion
    //! Write all entries in the table in a pretty way to the ostream. If there are
```

atmyers (approved):

PR #4655, opened 2025-09-11
AMREX_ENUM: Use std::unordered_map for cache

WeiqunZhang:
Use std::unordered_map instead of std::vector for caching.

Follow up to #4643


ax3l:
Thank you!

Cc @dpgrote

ax3l:
@dpgrote on Slack:
> They all work well and take about the same time, about half the time of the [#4643] version.

Issue #4656, opened 2025-09-16
Time integrator for particle types


srichers:
It seems that a change from 15 months ago (259db7cfb99e7d1d2ab4bec9b1587fdf624a138a) introduced some logic into the `initialize_stages` function of the time integrators (e.g., https://github.com/AMReX-Codes/amrex/blob/development/Src/Base/AMReX_RKIntegrator.H#L169) that involves a function (`CreateLike`) that has a different signature for MultiFabs (https://github.com/AMReX-Codes/amrex/blob/development/Src/Base/AMReX_IntegratorBase.H#L137) and ParticleContainers (https://github.com/AMReX-Codes/amrex/blob/development/Src/Base/AMReX_IntegratorBase.H#L25). Because of this, the implementation of initialize_stages only works for multifabs.

To me it seems that either the initialize_stages functions can be specialized to different types or the CreateLike function for particle containers can be given an additional dummy argument to have the same signature as the equivalent functions for multifabs. The latter seems more suitable to me.

PR #4657, opened 2025-09-16
add dummy argument to CreateLike for particlecontainers to match the …

srichers:
…signature for multifabs

## Summary
Modify the CreateLike function in IntegratorBase so all template specializations have the same function signature.

## Additional background
Used in initialize_stages called in, for example RKIntegrator.
Addresses #4656 I think.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_IntegratorBase.H
+++ b/Src/Base/AMReX_IntegratorBase.H
@@ -22,7 +22,7 @@ template<class T>
 struct IntegratorOps<T, std::enable_if_t<std::is_base_of_v<amrex::ParticleContainerBase, T> > >
 {
 
-    static void CreateLike (amrex::Vector<std::unique_ptr<T> >& V, const T& Other, bool dummy = false)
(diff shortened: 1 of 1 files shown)


Src/Base/AMReX_IntegratorBase.H
```diff
@@ -22,7 +22,7 @@ template<class T>
 struct IntegratorOps<T, std::enable_if_t<std::is_base_of_v<amrex::ParticleContainerBase, T> > >
 {
 
-    static void CreateLike (amrex::Vector<std::unique_ptr<T> >& V, const T& Other)
+    static void CreateLike (amrex::Vector<std::unique_ptr<T> >& V, const T& Other, bool dummy = false)
```
atmyers: I suggest:

```suggestion
    static void CreateLike (amrex::Vector<std::unique_ptr<T> >& V, const T& Other, bool /*dummy*/ = false)
```

to avoid unused parameter warnings.

atmyers (approved):

WeiqunZhang:
Same as #5442.

PR #4668, opened 2025-09-19
Add range-based for and ND to BoxIterator

AlexanderSinn:
## Summary

This PR extends BoxIterator to be n-dimensional. Additionally, it now supports a range-based for loops. The hope is that this can be useful to reduce the reliance on ifdefs/macros to write loops over boxes that run on CPUs, though so far I haven't found that many places where this can be used.
```C++
Box b;
...
for (IntVect iv : b.iterator()) {
    // do operations involving iv
}
```
Note that while regular iteration of the range-based version is reasonably performant,
it is not compatible with `omp parallel for` or simd autovectorization.

## Additional background

The range-based version produces good assembly when compiling with -O3, having minimal instructions in the innermost loop.
https://godbolt.org/z/MTcM8Ysaq
```C++
void g (const IntVect&);

void test (Box bx) {
    for (auto iv : bx.iterator()) {
        g(iv);
    }
}
```
```assembly
        ...
.L3:
        mov     ebx, DWORD PTR [rsp+8]
.L4:
        mov     r15d, r13d
.L6:
        mov     DWORD PTR [rsp+16], r15d
        lea     rdi, [rsp+16]
        add     r15d, 1
        mov     DWORD PTR [rsp+20], ebx
        mov     DWORD PTR [rsp+24], ebp
        call    "g(IntVectND<3> const&)"
        cmp     r15d, r12d
        jle     .L6
        add     ebx, 1
        cmp     r14d, ebx
        jge     .L4
        cmp     ebp, DWORD PTR [rsp+12]
        je      .L13
        add     ebp, 1
        jmp     .L3
        ...
```

## Checklist

The proposed changes:


WeiqunZhang (approved):

Src/Base/AMReX_BoxIterator.H
```diff
@@ -12,152 +11,179 @@
...
+        define(a_bx);
+    }
+
+    /**
+    * Associates a Box with this BoxIteratorND.
+    * Arguments:
+    * a_bx (not modified) the Box to iterate over.
+    */
+    void define (const BoxND<dim>& a_bx) noexcept {
+        if (a_bx.ok() && a_bx.smallEnd().allLE(a_bx.bigEnd())) {
```
WeiqunZhang: The second part of the test seems redundant.

WeiqunZhang (changes requested):

Src/Base/AMReX_BoxIterator.H
AlexanderSinn: It is. It was a leftover from the previous version. I removed it now.

WeiqunZhang (approved):

PR #4671, opened 2025-09-23
OpenBC FFT Solver: Fix a 2d mode bug.

WeiqunZhang:
This bug only affected 3D WarpX/ImpactX when there was only one cell in the z-direction for the solver.

The fix is that we have simplified the logic for handling domain decomposition. If it's in the 2D mode in 3D, we simply do slab decomposition.


ax3l (approved):
Yay, thank you!

PR #4681, opened 2025-09-29
Add function transposeCtoF for transposing multi-d array

WeiqunZhang:


WeiqunZhang:
We might want to using tiling for CPU implementation.

WeiqunZhang:
On my machine, the new CPU version with cache blocking is 4x faster than the original version without cache blocking.

Src/Base/AMReX_BaseFabUtility.H
```diff
@@ -95,6 +95,172 @@ void fill (BaseFab<STRUCT>& aos_fab, F const& f)
...
+    dim3 grid{unsigned((nx+tile_dim-1)/tile_dim),
+              unsigned((nz+tile_dim-1)/tile_dim), unsigned(ny)};
+
+    AMREX_LAUNCH_KERNEL(nthreads, grid, block, 0, Gpu::gpuStream(),
+    [=] AMREX_GPU_DEVICE ()
+    {
+        __shared__ T tile[tile_dim][tile_dim+1]; // +1 to avoid bank conflicts
+
+        int tx = blockIdx.y * tile_dim + threadIdx.x; // for loading z-direction
+        int ty = blockIdx.x * tile_dim + threadIdx.y; // for loading x-direction
```
AlexanderSinn: It looks correct; however, I find the variable names confusing. Maybe here tx could be k and ty i. Later ii = it + i.

WeiqunZhang: Yes, it's confusing. Maybe we don't even need to have `int ii = it + i`. It can be just `i += it;`.

WeiqunZhang: Never mind. No. that's not right.

WeiqunZhang:
```
        int k = blockIdx.y * tile_dim + threadIdx.x; // for loading z-direction                                                                                                            
        int i = blockIdx.x * tile_dim + threadIdx.y; // for loading x-direction                                                                                                            

        int j = blockIdx.z; // for y-direction                                                                                                                                             

        if (k < nz) {
            for (int it = 0; it < tile_dim; it += block_rows, i += block_rows) {
                if (i < nx) {
                    //      x               z                                                                                                                                              
                    tile[threadIdx.y+it][threadIdx.x] = pi[k + (j+i*std::size_t(ny))*nz];
                }
            }
        }
```
Better?

WeiqunZhang: https://github.com/WeiqunZhang/amrex-devtests/blob/main/transpose/main.cpp This is the test I use for correctness and performance.

AlexanderSinn: Yes

atmyers (approved):

PR #4697, opened 2025-10-06
Improve usability of amrex::Gpu::Buffer

AlexanderSinn:
## Summary

This PR expands the capabilities of `amrex::Gpu::Buffer<T>` so that it can be first default constructed, then modified by the CPU, and lastly copied over to the GPU, instead of everything happening in the constructor.

Example:
```C++
amrex::Gpu::Buffer<int> buf;

buf.resize(n);

for (int i=0; i<n; ++i) {
    buf[i] = i*i;
}

buf.copyToDeviceAsync();

int * ptr = buf.data();
// Use ptr inside ParallelFor

// optional:
// Change values of ptr inside ParallelFor
buf.copyToHost();
// Use buf.hostData() or buf[] on the CPU
```

## Additional background

Follow-up to discussion in https://github.com/AMReX-Codes/amrex/pull/4640

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_GpuBuffer.H
+++ b/Src/Base/AMReX_GpuBuffer.H
@@ -21,7 +21,7 @@ public:
     Buffer (std::initializer_list<T> init) {
         resize(init.size());
 
-        if (init.size() > 0 && init.begin() != nullptr) {
+        if (init.size() > 0) {
             std::memcpy(h_vect.data(), init.begin(), init.size()*sizeof(T));
             copyToDeviceAsync();
         }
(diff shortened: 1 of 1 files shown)


Src/Base/AMReX_GpuBuffer.H
```diff
@@ -17,87 +18,150 @@ class Buffer
...
-        std::memcpy(h_data, init.begin(), m_size*sizeof(T));
-#ifdef AMREX_USE_GPU
-        if (Gpu::inLaunchRegion())
-        {
-            d_data = static_cast<T*>(The_Arena()->alloc(m_size*sizeof(T)));
-            Gpu::htod_memcpy_async(d_data, h_data, m_size*sizeof(T));
+    Buffer (std::initializer_list<T> init) {
+        resize(init.size());
+
+        if (init.size() > 0 && init.begin() != nullptr) {
```
WeiqunZhang: The test `init.begin() != nullptr` seems redundant.

WeiqunZhang (approved):

PR #4703, opened 2025-10-07
OpenBC FFT Solver: Fix 2d mode bug

WeiqunZhang:
PR #4671 did not fix all the issues with the 2d mode OpenBC FFT solver in 3D WarpX/ImpactX runs. The scaling was still wrong.


ax3l (commented):
:tada:

ax3l (approved):

PR #4709, opened 2025-10-10
Update out of memory error message

WeiqunZhang:


AlexanderSinn:
I think it would be good to clarify the type of memory (CPU not GPU) in the message

Src/Base/AMReX_Utility.cpp
```diff
@@ -237,7 +237,7 @@ amrex::UtilRenameDirectoryToOld (const std::string &path, bool callbarrier)
 void
 amrex::OutOfMemory ()
 {
-    amrex::Error("Sorry, out of memory, bye ...");
+    amrex::Error("Terminating due to insufficient memory.");
```
WeiqunZhang:
```suggestion
    amrex::Error("Terminating due to insufficient CPU memory.");
```

atmyers (approved):

PR #4714, opened 2025-10-14
Add Build with Explicit Multi-Level Geometry for EB

jmsexton03:
## Summary
This PR adds a new `EB2::Build` variant that accepts a `Vector<Geometry>` to explicitly specify the geometry at each AMR level. Unlike the existing Build function that generates coarse levels through automatic coarsening, this version constructs EB data directly from the provided geometries at each level. This is useful for applications that need precise control over domain specification at each AMR level or when coarse levels are not simple coarsenings of the finest level.

## Checklist

The proposed changes:

### DIFF
--- a/Docs/sphinx_documentation/source/EB.rst
+++ b/Docs/sphinx_documentation/source/EB.rst
@@ -141,9 +141,12 @@ Given an implicit function object, say :cpp:`f`, we can make a
 :cpp:`EB2::IndexSpace`
 ----------------------
 
-We build :cpp:`EB2::IndexSpace` with a template function
+We build :cpp:`EB2::IndexSpace` with one of several template functions depending
+on the application needs.
 
-.. highlight: c++
+**Standard Build with Automatic Coarsening**
+
+.. highlight:: c++
 
 ::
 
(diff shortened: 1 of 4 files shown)


Docs/sphinx_documentation/source/EB.rst
```diff
@@ -173,12 +179,57 @@ ngrow` parameter specifies the number of ghost cells outside the domain on
...
+::
+
+    template <typename G>
+    void EB2::Build (const G& gshop, const Vector<Geometry>& geom,
+                     int ngrow = 4,
+                     bool extend_domain_face = ExtendDomainFace(),
+                     int num_coarsen_opt = NumCoarsenOpt());
+
+This version takes a :cpp:`Vector<Geometry>` where each element corresponds to
+the geometry of a specific AMR level, ordered from finest (index 0) to coarsest.
```
WeiqunZhang: Does the order matter?

jmsexton03: I don't think it matters in terms of 0 being finest, would it be better to not mention order here? The other Build only has the finest Geometry and then builds coarser, I think, but for this it's just using the order that comes in.

WeiqunZhang: The thing is AmrMesh stores Geometry in the opposite order. We should try to accept unordered Geometry objects. Internally, we could make a copy and then sort it.

jmsexton03: I see, that sounds like a better approach then assuming a certain order. Would you sort on `CellSize` or something else? Is there an example of sorting somewhere else (I only saw `<<` `>>` and `=` as overloaded operators)
```
    Real CellSize (int dir) const noexcept { return dx[dir]; }
```

WeiqunZhang: Make the function take `Vector<Geometry> geom` instead of `Vector<Geometry> const& geom`. Then
```
std::sort(geom.begin(), geom.end(), [] (Geometry const& a, Geometry const& b) { return a.Domain().numPts() > b.Domain().numPts(); });
```

jmsexton03: Thanks for the code and for finding that lack of generality. I've pushed what I think fixes it.

WeiqunZhang (approved):

PR #4721, opened 2025-10-20
Robustify file renaming

WeiqunZhang:
In `amrex::UtilCreateCleanDirectory`, if a directory already exists, we preserve it by renaming it to a directory whose name includes a string generated by `amrex::UniqueString()`. However, this string is not guaranteed to be unique. The behavior of `std::rename` is implementation-defined, and it may fail if the destination already exists. This happens rarely, but it does happen.

This PR addresses the issue by removing the existing conflicting directory before renaming. This makes the file renaming process more robust.

### DIFF
--- a/Src/Base/AMReX_Utility.cpp
+++ b/Src/Base/AMReX_Utility.cpp
@@ -166,12 +166,16 @@ amrex::UniqueString()
 void
 amrex::UtilCreateCleanDirectory (const std::string &path, bool callbarrier)
 {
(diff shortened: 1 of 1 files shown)


ax3l (commented):
LGTM, thank you :)

Should `TryFileOutput` have a `BL_PROFILE`, too?

ax3l (approved):

WeiqunZhang:
I thought about `TryFileOutput`, but decided not to add profiler to it. The file names there are different, and it's even less likely to have conflicts.

PR #4731, opened 2025-10-25
Reserve space in particle redistribution

WeiqunZhang:
When redistributing particles, we need to resize a number of vectors. This sometimes causes out-of-memory issues for GPU runs, because the vector resize may fragment the memory arena. To address the issue, we try to minimize the number of memory reallocation calls by reserving space before we unpack local and remote communication buffers. Previously, resize may cause memory reallocation in both unpacking operations.

We have also collected all the vectors that need to be resized into one place. This will allows to explore different strategies for how to resize a number of vectors.


Src/Particle/AMReX_ParticleTile.H
```diff
@@ -1351,6 +1351,58 @@ struct ParticleTile
...
+            }
+            for (auto* v : iv) {
+                ivs.emplace_back(v, s);
+            }
+            for (auto* v : idcpuv) {
+                idcpuvs.emplace_back(v, s);
+            }
+        }
+
+        // xxxxx TODO: Can we come up a better strategy?
```
WeiqunZhang: Maybe we can try to shrink the size first if a vector has say 2x capacity than needed. Then we can order the vectors by their future capacity, and reserve big vectors first.

WeiqunZhang: If a reserve fails, maybe we should try to defragment these vectors that are under our control. We can order them by their pointer address, and then try to move the vectors to lower addresses.

Src/Particle/AMReX_ParticleContainerI.H
```diff
@@ -1623,6 +1626,42 @@ ParticleContainer_impl<ParticleType, NArrayReal, NArrayInt, Allocator, CellAssig
...
+            }
+        }
+    }
+
+    if (plan.m_nrcvs > 0) {
+        for (int i = 0, N = int(plan.m_rcv_box_counts.size()); i < N; ++i) {
+            int copy_size = plan.m_rcv_box_counts[i];
+            int lev = plan.m_rcv_box_levs[i];
+            int gid = plan.m_rcv_box_ids[i];
+            int tid = 0; // xxxxx why is tid 0?
```
WeiqunZhang:
```suggestion
            int tid = 0; // It's always 0 because this function is for RedistributeGPU only and the tiling is off.
```

Src/Particle/AMReX_ParticleContainerI.H
```diff
@@ -1623,6 +1626,42 @@ ParticleContainer_impl<ParticleType, NArrayReal, NArrayInt, Allocator, CellAssig
...
+            int tid = mfi.LocalTileIndex();
+            auto& tile = this->DefineAndReturnParticleTile(lev, gid, tid);
+            int num_copies = plan.m_box_counts_h[this->BufferMap().gridAndLevToBucket(gid, lev)];
+            if (num_copies > 0) {
+                addsizes[&tile] += num_copies;
+            }
+        }
+    }
+
+    if (plan.m_nrcvs > 0) {
```
WeiqunZhang, line 1651:
```suggestion
    if (plan.m_nrcvs > 0) {
        AMREX_ALWAYS_ASSERT(do_tiling == false);
```

asalmgren (approved):

PR #4734, opened 2025-10-26
PODVector: Add extra capacity

WeiqunZhang:
Add some extra room when we call `PODVector::resize` and `reserve`. The extra capacity is computed as 3*sqrt(capacity) suggested by @AlexanderSinn, and is capped at 10%. This might help particle codes avoid memory re-allocation.

### DIFF
--- a/Src/Base/AMReX_PODVector.H
+++ b/Src/Base/AMReX_PODVector.H
@@ -611,7 +611,7 @@ namespace amrex
 
         [[nodiscard]] size_type capacity () const noexcept { return m_capacity; }
 
-        [[nodiscard]] bool empty () const noexcept { return m_size == 0; }
+        [[nodiscard]] bool empty () const noexcept { return m_size == 0 || m_data == nullptr; } // test m_data to avoid compiler warning
 
         [[nodiscard]] T& operator[] (size_type a_index) noexcept { return m_data[a_index]; }
 
--- a/Src/Base/AMReX_PODVector.H
+++ b/Src/Base/AMReX_PODVector.H
@@ -261,6 +261,10 @@ namespace amrex
     inline std::size_t grow_podvector_capacity (std::size_t s)
     {
         if (s <= 900) {
+            // 3*sqrt(900) = 900/10. Note that we don't need to be precise
+            // here. Even if later we change the else block to
+            // 4*std::sqrt(s), it's not really an issue to still use 900
+            // here.
             return s + s/10;
         } else {
             return s + std::size_t(3*std::sqrt(s));
--- a/Src/Base/AMReX_PODVector.H
+++ b/Src/Base/AMReX_PODVector.H
@@ -10,6 +10,7 @@
 #include <AMReX_MemPool.H>
 #include <AMReX_TypeTraits.H>
 
+#include <cmath>
 #include <iterator>
 #include <type_traits>
 #include <utility>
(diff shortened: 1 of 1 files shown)


AlexanderSinn:
I feel like there should still be a way to resize a vector to the exact capacity that is asked for. Only if the vector is used in a particle container with multiple tiles where particles are exchanged using a thermal process is this a good strategy. If there is only one tile or if the vector is used to store a grid or metadata, the extra capacity is not needed.

WeiqunZhang:
What's the downside?

Src/Base/AMReX_PODVector.H
```diff
@@ -257,6 +258,15 @@ namespace amrex
         void Initialize ();
     }
 
+    inline std::size_t grow_podvector_capacity (std::size_t s)
+    {
+        if (s <= 900) {
```
ax3l, line 263: Can you pls add an inline comment or docstring for the logic here? It is not 100% obvious on first sight why `s <= 900` caps the resize at +10% :)

AlexanderSinn:
Just a waste of (a little bit of) memory. It also could get very confusing when trying to implement custom allocation size strategies on top of PODVector.

WeiqunZhang:
Because it scales as sqrt(n), the extra memory cost is not an issue. As for implementing other growth strategies such as what are in #4735 , here is what I think. `push_back` and `insert` are rarely used and should not be used for GPU codes anyway. We probably do not care about constructors either. So our focus should be on just resize and reserve. In this PR, the growth strategy is implemented in function `grow_podvector_capacity`. It should not be hard at all to allow other strategies. In the current PR, `grow_podvector_capacity` is being used in 4 places. All we need is to have an extra parameter to the function, and the parameter can be obtained from the resize and reserve function as it's already being done in #4735. Re: ParmParse parameters in #4735, I think it's better to have them in particle containers, not in the vector class. I think that would be much more flexible and less confusing by having a fixed default in the vector and the change needs to be explicit. I think the default in vector should be that we have a little bit extra that scales as sqrt(n), because that cost is really not that much.

Src/Base/AMReX_PODVector.H
WeiqunZhang: It does not really matter if we have a number that is not consistent with 10%. Anyway, I will add some comments.

AlexanderSinn:
Sounds good. I am fine with merging this PR as is, and after that I will make a follow-up to add exact and geometric strategies to resize and reserve only, using grow_podvector_capacity. I also agree that those are the most important functions, and insert/push_back shouldn't really be used. Should the interface for resize and reserve include an optional argument or add new funcitons?

Option 1:
```C++
vec.reserve(n, amrex::GrowthStrategy::Geometric);
vec.resize(n, amrex::GrowthStrategy::Exact);
vec.resize(n /*, amrex::GrowthStrategy::Poisson*/); // Poisson is the default
```
Option 2:
```C++
vec.reserve_geometric_grow(n);
vec.resize_exact_grow(n);
vec.resize(n); // uses Poisson grow
```
Option 3:
```C++
vec.reserve_geometric(n);
vec.resize_exact(n);
vec.resize(n); // uses Poisson grow
```

WeiqunZhang:
I prefer Option 1. It's easier for the users.

ax3l (approved):
Thank you! :rocket: :sparkles:

PR #4741, opened 2025-10-30
amrex::Initialize: Add optional argument of device ID

WeiqunZhang:
In most cases, you should let AMReX select a device when multiple GPU devices are visible. However, if another library has already been initialized and assigned processes to specific devices, you may need AMReX to use a particular GPU. In that case, you can pass the desired device ID to `amrex::Initialize`.

### DIFF
--- a/Docs/sphinx_documentation/source/Basics.rst
+++ b/Docs/sphinx_documentation/source/Basics.rst
@@ -771,7 +771,7 @@ second and third versions, the user may also pass a function that adds
 parameters to the ParmParse database instead of reading from command line or
 input file.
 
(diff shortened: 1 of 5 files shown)


Docs/sphinx_documentation/source/Basics.rst
```diff
@@ -768,6 +771,16 @@ second and third versions, the user may also pass a function that adds
 parameters to the ParmParse database instead of reading from command line or
 input file.
 
+The last optional parameter, :cpp:`int a_device_id = -1`, applies only to
```
atmyers:
```suggestion
The last optional parameter, :cpp:`int a_device_id = -1`, applies to
```

atmyers (approved):

PR #4749, opened 2025-11-03
Fix cmake options table in documentation

atmyers:
This table wasn't being rendered properly in the docs.

The proposed changes:


WeiqunZhang (approved):

WeiqunZhang:
We need to catch the error in the CI to prevent this from happening again.

zingale:
we use:
```
 make SPHINXOPTS='-v -W --keep-going' NO_DOXYGEN=TRUE html
 ```

atmyers:
Let's try this: https://github.com/AMReX-Codes/amrex/pull/4751

PR #4751, opened 2025-11-03
Catch Sphinx build errors in CI

atmyers:
## Checklist

The proposed changes:

### DIFF
--- a/build_docs.sh
+++ b/build_docs.sh
@@ -16,9 +16,9 @@ cd ../..
 cd Docs/sphinx_documentation
 
 echo "Build the Sphinx documentation for Amrex."
-make PYTHON="python3" latexpdf
+make SPHINXOPTS='-v -W --keep-going' PYTHON="python3" latexpdf
 mv build/latex/amrex.pdf source/
-make PYTHON="python3" html &> make_source_html.out
+make SPHINXOPTS='-v -W --keep-going' PYTHON="python3" html &> make_source_html.out
 cd ../../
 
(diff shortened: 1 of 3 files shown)


build_docs.sh
```diff
@@ -16,9 +16,9 @@ cd ../..
 cd Docs/sphinx_documentation
 
 echo "Build the Sphinx documentation for Amrex."
-make PYTHON="python3" latexpdf
+make SPHINXOPTS='-v -W --keep-going' PYTHON="python3" latexpdf
 mv build/latex/amrex.pdf source/
-make PYTHON="python3" html &> make_source_html.out
+make SPHINXOPTS='-v -W --keep-going' PYTHON="python3" html &> make_source_html.out
```
WeiqunZhang: I think we should remove `&> make_source_html.out` so that we can actually see the output and error.

WeiqunZhang: We also need to build `html` first to treat warnings as errors, if we ignore the warnings from `latexpdf`. Or we do `make clean` before `make html`.

WeiqunZhang: `make html` was reusing the results of `make latexpdf`. That was why it did not break.

atmyers: The current code breaks. Now let me fix the table again.

WeiqunZhang:
There is `rm -rf ../Doxygen/xml` in `make clean`.

WeiqunZhang (approved):

PR #4760, opened 2025-11-05
Add information about generating particles for GPU codes to the docs

atmyers:
The proposed changes:

### DIFF
--- a/Docs/sphinx_documentation/source/Particle.rst
+++ b/Docs/sphinx_documentation/source/Particle.rst
@@ -315,10 +315,10 @@ outside their parent grid).
    vectors, assumes you have compiled AMReX for CPU execution.
    For GPU codes, one can either generate particles on the host and copy them
    to the device, or generate the particles directly on the GPU. For the first
-   approach, please see the sample code `here <https://github.com/AMReX-Codes/amrex/blob/development/Tests/Particles/Redistribute/main.cpp#L81>`_.
+   approach, please see the sample code `here <https://github.com/AMReX-Codes/amrex/blob/development/Tests/Particles/Redistribute/main.cpp#L81>`__.
    For an example of generating a variable number of particles in each cell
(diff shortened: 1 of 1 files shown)


Docs/sphinx_documentation/source/Particle.rst
```diff
@@ -309,6 +309,18 @@ particles, if the processes generate particles they don't own (for example, if
...
+.. note::
+
+   The above code snippet, which successively calls :cpp:`push_back` on the particle
+   vectors, assumes you have compiled AMReX for CPU execution.
+   For GPU codes, one can either generate particles on the host and copy them
+   to the device, or generate the particles directly on the GPU. For the first
+   approach, please see the sample code `here <https://github.com/AMReX-Codes/amrex/blob/development/Tests/Particles/Redistribute/main.cpp#L81>`_.
+   For an example of generating a variable number of particles in each cell
+   directly on the GPU, please see
+   `this <https://github.com/AMReX-Codes/amrex-tutorials/blob/main/ExampleCodes/Particles/ElectromagneticPIC/Source/EMParticleContainerInit.cpp#L48>`_
```
ax3l: Anonymous links (with a double `__`) will not create clashes (i.e. warnings) on the internal Sphinx reference system, if the link text `here` and `this` is also used in other links: 
```suggestion
   approach, please see the sample code `here <https://github.com/AMReX-Codes/amrex/blob/development/Tests/Particles/Redistribute/main.cpp#L81>`__.
   For an example of generating a variable number of particles in each cell
   directly on the GPU, please see
   `this <https://github.com/AMReX-Codes/amrex-tutorials/blob/main/ExampleCodes/Particles/ElectromagneticPIC/Source/EMParticleContainerInit.cpp#L48>`__
```

ax3l (approved):
Thank you! :)

PR #4765, opened 2025-11-07
Add ParmParse::Add for AMREX_NUM

WeiqunZhang:
This will also enable `ParmParse::queryAdd`.


ax3l (approved):
Thank you!

This adds `add`, but didn't you mean to also add `queryAdd`?

WeiqunZhang:
We already have a templated queryAdd. But it needs to call `add`. That's why by adding `add` for AMREX_NUM, we also enable `queryAdd` for AMREX_ENUM.

PR #2012, opened 2021-05-12
CMake: add support for CMake >= 3.20

mic84:
## Summary
Add support for CMake >= 3.20, with special focus on CMake handling of CUDA builds. 

## Additional background

The major change in this PR is the use of the `CUDA_ARCHITECTURES` target property for setting the arch flags.
This allows to propagate the arch flags downstream (together with our custom `setup_target_for_cuda_compilation`)
to each step of the build process, including device linking step. Consequently, `AMReX_SetupCUDA` is not required anymore.

## Checklist

The proposed changes:
- [ ] fix a bug or incorrect behavior in AMReX
- [x] add new capabilities to AMReX
- [ ] changes answers in the test suite to more than roundoff level
- [ ] are likely to significantly affect the results of downstream AMReX users
- [ ] are described in the proposed changes to the AMReX documentation, if appropriate

PR #4773, opened 2025-11-10
Modernize CMake CUDA to 3.25+ (native arch, device LTO)

### LINKED ISSUE #3948: CMake: Update CUDA Logic to 3.24+

ax3l:
## Summary

We already use features of `FindCUDAToolkit` for nvtx3 that are CMake 3.25+. Modernize and remove pre CMake 3.20 CUDA logic.

Move CUDA architecture handling to the standard CMake interface and drop the deprecated CMP0104 OLD path and the FindCUDA/select_compute_arch mangling (convert_cuda_archs / set_cuda_architectures / set_nvcc_arch_flags), which broke on Hopper-next/Blackwell (sm_100/sm_120) and CUDA 12.9. Closes https://github.com/AMReX-Codes/amrex/issues/3948.

- Default to `native`: CMake resolves it at **configure** time from the GPU(s) visible on the machine running `cmake`, and the resolved architecture is what gets built, reported and exported. If no GPU is visible and no architecture was given, configuration stops with an error pointing at `-DCMAKE_CUDA_ARCHITECTURES=...`.
- `AMReX_CUDA_LTO=ON` now performs real CUDA device LTO via `INTERPROCEDURAL_OPTIMIZATION` (requires RDC; CMake >= 3.25). It covers the AMReX libraries *and* the targets set up with `setup_target_for_cuda_compilation`, so dependent code takes part in the LTO instead of getting nvlink's *partial* LTO. Host code generation is deliberately left untouched.
- Minimum CMake is now 3.25 — for building AMReX and for consuming an installed AMReX (the installed config used to demand the exact CMake version AMReX was built with).

User transition:
- Select architectures with the standard `-DCMAKE_CUDA_ARCHITECTURES=80` (or the `CUDAARCHS` env var). Integers only (`80`, not `8.0`); `native`, `all`, `all-major`, `-real`/`-virtual` and `90a` forms are supported, as is `OFF` for packagers who pass the flags themselves.
- `-DAMReX_CUDA_ARCH` and the `AMREX_CUDA_ARCH` env var still work but warn. They map to `CMAKE_CUDA_ARCHITECTURES` / `CUDAARCHS` with unchanged precedence, and their historical spellings (`Auto`, `Common`, `Volta`, `8.0`, `7.0+PTX`, whitespace-separated lists) are translated.
- On GPU-less machines (login nodes, CI, container builds), pass an explicit architecture instead of relying on the `native` default. This includes configuring on a login node and building on a GPU node.
- Architectures below compute capability 6.0 are dropped with a message (also when they are part of what `all`/`all-major` stands for); configuration fails if nothing is left.
- `CUDA_FORWARD_UNKNOWN_FLAGS_HOST` removed (CMake >= 3.17 forwards natively).

Also removes now-dead CMake < 3.20 guards.

## Additional background

[CMake 3.25 was released in November 2022](https://github.com/Kitware/CMake/releases/tag/v3.25.0) and is widely available and very easy to install as a user, if missing.

Also a bit of prep for #4572

Close #3948
Close #4572
Close #5524

## Checklist

The proposed changes:


ax3l:
@WeiqunZhang can you please remove `SENSEI Adaptor [test]` from the required tests in CI?

WeiqunZhang:
Should we also update this? https://github.com/AMReX-Codes/amrex/blob/e7271974e848bb3a472b7230e88a1ceac9fe247a/Tools/CMake/AMReXCUDAOptions.cmake#L61

Also don't think we need to worry about CUDA 10 anymore.

WeiqunZhang (approved):

Tools/CMake/AMReXParallelBackends.cmake
```diff
@@ -82,11 +80,9 @@ if (  AMReX_GPU_BACKEND STREQUAL "CUDA"
            target_link_libraries(amrex_${D}d PUBLIC CUDA::cusparse)
        endif ()
 
-       if(CMAKE_CUDA_COMPILER_VERSION VERSION_LESS 11.2)
```
jmsexton03, line 85: https://amrex-codes.github.io/amrex/docs_html/GPU_Chapter.html

Do we need to update the docs (not that I expect someone to find cuda 11.0 or 11.1)

ax3l:
Yes, [couple of options](https://github.com/AMReX-Codes/amrex/pull/4773#issuecomment-3514160257) we need to refresh now.

E.g., add support for LTO on `AMReX_CUDA_LTO`.
Supported by CMake 3.25+, too: https://cmake.org/cmake/help/latest/module/CheckIPOSupported.html

Tools/CMake/AMReXParallelBackends.cmake
ax3l: updated the docs :+1:

ax3l:
@ChristosT is there a newer Catalyst container (we currently use `kitware/paraview:ci-catalyst-amrex-warpx-20240701`) that we can use to have a more modern CMake available? :)

Docs/sphinx_documentation/source/GPU.rst
```diff
@@ -277,26 +277,39 @@ check the :ref:`table <tab:cmakecudavar>` below.
...
+
+.. note::
+
+   The legacy ``-DAMReX_CUDA_ARCH`` option and the ``AMREX_CUDA_ARCH`` environment variable
+   are deprecated but still honored (with a warning): they map to ``CMAKE_CUDA_ARCHITECTURES``
+   and the standard ``CUDAARCHS`` environment variable, respectively, and keep their historical
+   precedence. Please migrate to the standard CMake variables.
+
+Setting ``-DAMReX_CUDA_LTO=ON`` enables CUDA device link-time optimization. This requires
+relocatable device code (``AMReX_GPU_RDC=ON``, the default) and CMake 3.25 or newer.
```
ax3l: We require CMake 3.25+ now:
```suggestion
relocatable device code (``AMReX_GPU_RDC=ON``, the default).
```

ChristosT:
> @ChristosT is there a newer Catalyst container (we currently use `kitware/paraview:ci-catalyst-amrex-warpx-20240701`) that we can use to have a more modern CMake available? :)

None that I know of but the catalyst maintainers should be able to create one. These are the source files used to build the one currently in use 

https://gitlab.kitware.com/christos.tsolakis/catalyst-amrex-docker-images

ax3l:
Thanks @ChristosT , your latest tag already bumped to 3.26 and does it! :)

Docs/sphinx_documentation/source/GPU.rst
```diff
@@ -277,26 +277,39 @@ check the :ref:`table <tab:cmakecudavar>` below.
...
-``-DAMReX_CUDA_ARCH=<target-architecture>``, where ``<target-architecture>`` can be either
-the name of the NVIDIA GPU generation, e.g., ``Turing``, ``Volta``, or ``Ampere``, or its
-`compute capability <https://developer.nvidia.com/cuda-gpus>`_, e.g., ``10.0`` or ``9.0``.
-For example, on Cori GPUs you can specify the architecture as follows:
+The target architecture(s) to build for are selected via the standard CMake option
+``-DCMAKE_CUDA_ARCHITECTURES=<arch>`` (or the ``CUDAARCHS`` environment variable), where
+``<arch>`` is one or more `CUDA compute capabilities <https://developer.nvidia.com/cuda-gpus>`_
+given as integers without the dot, e.g., ``80`` for compute capability 8.0. Multiple
+architectures can be set as a semicolon-separated list, e.g.,
+``-DCMAKE_CUDA_ARCHITECTURES="80;90"``; building for multiple architectures generally
```
```suggestion
``-DCMAKE_CUDA_ARCHITECTURES="80;90"``. Building for multiple architectures generally
```

Docs/sphinx_documentation/source/GPU.rst
```diff
@@ -277,26 +277,39 @@ check the :ref:`table <tab:cmakecudavar>` below.
...
-If no architecture is specified, CMake will default to the architecture defined in the
-*environment variable* ``AMREX_CUDA_ARCH`` (note: all caps).
-If the latter is not defined, CMake will try to determine which GPU architecture is supported by the system.
-If more than one is found, CMake will build for all of them.
-If autodetection fails, a list of "common" architectures is assumed.
-`Multiple CUDA architectures <https://cmake.org/cmake/help/latest/module/FindCUDA.html#commands>`__ can also be set manually as a semicolon-separated list, e.g., ``-DAMReX_CUDA_ARCH=7.0;8.0``.
-Building for multiple CUDA architectures will generally result in a larger library and longer build times.
+If no architecture is specified, AMReX defaults to ``native``, which builds for the GPU(s)
+installed in the machine running CMake. This requires a GPU to be visible at configuration
+time; on GPU-less machines (e.g., HPC login nodes or CI runners) an explicit architecture
```
```suggestion
time. On machines without GPUs (e.g., HPC login nodes or CI runners) an explicit architecture
```

ChristosT:
> Thanks @ChristosT , your latest tag already bumped to 3.26 and does it! :)

Glad to hear that !

WeiqunZhang:
```
$ git grep -n -i "cmake.*version\ *3\.18"
Docs/sphinx_documentation/source/BuildingAMReX_Chapter.rst:26: building with CMake, the minimal requirement is version 3.18.
Tests/CMakeTestInstall/CMakeLists.txt:6:  cmake_minimum_required(VERSION 3.18)
Tests/SpackSmokeTest/CMakeLists.txt:9:  cmake_minimum_required(VERSION 3.18)
```
There are a few places we need to update the cmake minimum version requirement. @ax3l

WeiqunZhang:
Comments from AI.

Exported AMREX_CUDA_ARCHS may now carry the literal native/all-major (clarify intent)

  AMReXParallelBackends.cmake records:
  set(AMREX_CUDA_ARCHS "${CMAKE_CUDA_ARCHITECTURES}" CACHE INTERNAL ...)
  which captures the CMAKE_CUDA_ARCHITECTURES variable — i.e. the literal native/all-major when nothing explicit is given. This is then baked into AMReXConfig.cmake.in (@AMREX_CUDA_ARCHS@) and
  applied to downstream targets' CUDA_ARCHITECTURES. Previously set_cuda_architectures resolved to concrete integers at AMReX build time, so an installed AMReX exported the exact arches it was
  compiled for. Now, an AMReX built with native on an H100 propagates the string native, which a downstream project resolves against its machine — potentially linking AMReX objects (compiled for
  sm_90) against freshly-compiled sm_80 objects. Worth confirming this is the desired semantics; if installed-library consistency matters, export the resolved arches (e.g. read the
  CUDA_ARCHITECTURES target property after creation) rather than the raw variable. (This is a behavior question, not necessarily a defect — it matches CMake's own native semantics — but it's a
  quiet change for downstream consumers that the release notes should call out.)

Minor / nits

  - _amrex_cuda_ipo lifecycle: it's set in directory scope only when AMReX_CUDA_LTO is ON and is unset() after the loop. The later if (AMReX_CUDA_LTO AND _amrex_cuda_ipo) is safe when undefined
  (evaluates false), so this is fine — just noting the guard is what keeps the LTO-off path correct.

WeiqunZhang:
More comments from AI:

• Found two actionable issues in PR 4773:

  - [P1] Preserve legacy named CUDA architectures — /tmp/amrex-pr4773-head-961f1/Tools/CMake/AMReXCUDAArchs.cmake:43 claims backward compatibility but only normalizes Auto and dotted numbers.
    Previously documented values such as Volta, Turing, and Ampere now reach CMAKE_CUDA_ARCHITECTURES unchanged and fail configuration. I reproduced this with -DAMReX_CUDA_ARCH=Volta. These
    legacy names should be mapped to their numeric architectures.

  - [P2] Export the CUDA-LTO device-link requirement — /tmp/amrex-pr4773-head-961f1/Tools/CMake/AMReXParallelBackends.cmake:133 enables IPO only on the build-tree AMReX target. After installing
    an LTO-enabled static build, a downstream consumer linking AMReX::amrex had no -dlto in its device-link command, silently discarding LTO. The exported target needs an interface device-link
    requirement for installed consumers.

ax3l:
@WeiqunZhang back from PTO now. Pushed two commits to address those.

Docs/sphinx_documentation/source/GPU.rst
```diff
@@ -277,26 +277,41 @@ check the :ref:`table <tab:cmakecudavar>` below.
    \end{center}
 
 
-The target architecture to build for can be specified via the configuration option
-``-DAMReX_CUDA_ARCH=<target-architecture>``, where ``<target-architecture>`` can be either
-the name of the NVIDIA GPU generation, e.g., ``Turing``, ``Volta``, or ``Ampere``, or its
-`compute capability <https://developer.nvidia.com/cuda-gpus>`_, e.g., ``10.0`` or ``9.0``.
-For example, on Cori GPUs you can specify the architecture as follows:
+The target architecture(s) to build for are selected via the standard CMake option
+``-DCMAKE_CUDA_ARCHITECTURES=<arch>`` (or the ``CUDAARCHS`` environment variable), where
```
ax3l, line 282: Add an explicit section about environment hints to set for CMake (all standard, but good to say here):

```bash
# For Cray/HPE machines: necessary to use CUDA-Aware MPI and run a job
export CRAY_ACCEL_TARGET=nvidia80

# example: overwrite the auto-detection to optimize CUDA compilation for A100
export CUDAARCHS=80

# example: optimize CPU microarchitecture for AMD EPYC 3rd Gen (Milan/Zen3)
export CXXFLAGS="-march=znver3"
export CFLAGS="-march=znver3"

# CMake compiler environment hints
export CC=$(which gcc)       # or $(which cc) for Cray/HPE
export CXX=$(which g++)      # or $(which CC) for Cray/HPE
export FC=$(which gfortran)  # or $(which ftn) for Cray/HPE
export CUDACXX=$(which nvcc)
export CUDAHOSTCXX=${CXX}
```

Tools/CMake/AMReXCUDAArchs.cmake
```diff
@@ -0,0 +1,86 @@
...
+      "The AMREX_CUDA_ARCH environment variable is deprecated; use the standard CUDAARCHS "
+      "environment variable instead (or -DCMAKE_CUDA_ARCHITECTURES=...).")
+   set(_amrex_cuda_archs "$ENV{AMREX_CUDA_ARCH}")
+elseif (DEFINED CMAKE_CUDA_ARCHITECTURES)
+   set(_amrex_cuda_archs "${CMAKE_CUDA_ARCHITECTURES}")
+elseif (DEFINED ENV{CUDAARCHS})
+   set(_amrex_cuda_archs "$ENV{CUDAARCHS}")
+else ()
+   set(_amrex_cuda_archs "native")
+endif ()
```
ax3l, line 184: Let us CMake-print (again?) what CUDA arch the user selected.

We can check if we can find out what `native` resolved to, but also ok if not.

Docs/sphinx_documentation/source/GPU.rst
ax3l: added

Tools/CMake/AMReXCUDAArchs.cmake
ax3l: added
image

Docs/sphinx_documentation/source/GPU.rst
ax3l: added

Docs/sphinx_documentation/source/GPU.rst
```diff
@@ -277,30 +277,68 @@ check the :ref:`table <tab:cmakecudavar>` below.
...
+relocatable device code (``AMReX_GPU_RDC=ON``, the default). Device LTO is currently
+available only with the NVIDIA ``nvcc`` compiler; with a Clang ``-x cuda`` compiler it is
+gracefully disabled with a warning.
 
 **Note that AMReX supports NVIDIA GPU architectures with compute capability 6.0 or higher and
 CUDA Toolkit version 12.2 or higher.**
 
+**Environment hints.** All of AMReX's CMake configuration can also be driven through standard
+CMake and compiler environment variables. These are especially handy on HPC systems and in CI,
+where they are typically exported in a job or module script before invoking ``cmake``:
```
```suggestion
where they are typically exported in a job or module script before invoking ``cmake``.
Example on a Cray/HPE machine:
```

Docs/sphinx_documentation/source/GPU.rst
```diff
@@ -277,30 +277,68 @@ check the :ref:`table <tab:cmakecudavar>` below.
...
+
+.. code-block:: bash
+
+   # For Cray/HPE machines: necessary to use CUDA-Aware MPI and run a job
```
ax3l, line 362:
```suggestion
   # For Cray/HPE machines: necessary to use CUDA-Aware MPI and run a job
   # Note: also sometimes already set by a loaded module.
```

Docs/sphinx_documentation/source/GPU.rst
```diff
@@ -277,30 +277,68 @@ check the :ref:`table <tab:cmakecudavar>` below.
...
+   export CRAY_ACCEL_TARGET=nvidia80
+
+   # example: overwrite the auto-detection to optimize CUDA compilation for A100
+   export CUDAARCHS=80
+
+   # example: optimize CPU microarchitecture for AMD EPYC 3rd Gen (Milan/Zen3)
```
```suggestion
   # overwrite the auto-detection to optimize CUDA compilation for A100
   export CUDAARCHS=80

   # optimize CPU (host) microarchitecture for AMD EPYC 3rd Gen (Milan/Zen3)
```

Docs/sphinx_documentation/source/GPU.rst
```diff
@@ -277,30 +277,68 @@ check the :ref:`table <tab:cmakecudavar>` below.
...
+   export CXXFLAGS="-march=znver3"
+   export CFLAGS="-march=znver3"
+
+   # CMake compiler environment hints
```
```suggestion
   # CMake compiler selection
```

WeiqunZhang:
Claude:

---
1. AMReX_CUDA_LTO=ON + AMReX_GPU_RDC=OFF silently produces a non-LTO build

Tools/CMake/AMReXParallelBackends.cmake:133-138 sets CUDA_SEPARABLE_COMPILATION ON, but Src/CMakeLists.txt:240 runs later and calls setup_target_for_cuda_compilation(), which unconditionally resets it (Tools/CMake/AMReXTargetHelpers.cmake:127-128):

-DAMReX_CUDA_LTO=ON -DAMReX_GPU_RDC=OFF   # configures clean, no warning → CUDA_FLAGS = ... "--generate-code=arch=compute_80,code=[compute_80,sm_80]"   # no lto_80 The PR says LTO "requires RDC", but nothing enforces it. This should be a FATAL_ERROR (or an explicit warn-and-disable) when AMReX_CUDA_LTO AND NOT AMReX_GPU_RDC.

2. AMReX_CUDA_LTO=ON also silently enables host LTO

  INTERPROCEDURAL_OPTIMIZATION (AMReXParallelBackends.cmake:137) is a per-target, all-language property, but check_ipo_supported(LANGUAGES CUDA ...) only validates CUDA. With -DAMReX_FORTRAN=ON -DAMReX_CUDA_LTO=ON:

  Fortran_FLAGS = -O3 -flto=auto -fno-fat-lto-objects -J../mod_files_3d -fimplicit-none
  Two consequences: host codegen changes behind the user's back, and -fno-fat-lto-objects in an installed static library breaks non-LTO downstream links. Under cmake_minimum_required(3.25) CMP0069 is NEW, so a host/Fortran compiler CMake has no IPO flags for will hard-error at generate time even though the CUDA check passed. Either add CXX/Fortran to check_ipo_supported, or drop INTERPROCEDURAL_OPTIMIZATION and apply -dlto explicitly as compile+device-link options.

  3. Multi-arch lists break downstream find_package(AMReX)

  Tools/CMake/AMReXConfig.cmake.in:283 has an unquoted CUDA_ARCHITECTURES ${AMREX_CUDA_ARCHS}. With -DCMAKE_CUDA_ARCHITECTURES="80;90" the generated lib/cmake/AMReX/AMReXConfig.cmake contains:
  set_target_properties(AMReX::amrex_3d PROPERTIES CUDA_ARCHITECTURES 80;90)
  which fails: set_target_properties called with incorrect number of arguments (reproduced standalone). The bug is pre-existing, but this PR moves that exact block and the new docs actively recommend -DCMAKE_CUDA_ARCHITECTURES="80;90", so it becomes reachable. One-line fix: quote it. CI only ever uses a single arch, so nothing catches this.

  4. native is baked into the installed AMReXConfig.cmake

  AMREX_CUDA_ARCHS:INTERNAL=native → the installed config exports set(AMREX_CUDA_ARCHS native ...). Previously Auto was resolved to concrete integers at AMReX configure
  time. Now a downstream app configuring on a login node, or on a node with a different GPU than AMReX was built on, either fails or silently compiles for a different
  architecture than the AMReX device objects it RDC-links against. Export CMAKE_CUDA_ARCHITECTURES_NATIVE instead when the value is native.

  5. "Hard-fails at configure time if no GPU is present" is not what happens

  With CUDA ≥ 11.5, CMake forwards -arch=native to nvcc rather than resolving it, so configure succeeds and the failure lands at build time. Worse, the new status line
  prints garbage:
  CUDA_VISIBLE_DEVICES=-1 cmake ...
  --    CUDA architectures: native -> No CUDA devices found.-real
  (CMAKE_CUDA_ARCHITECTURES_NATIVE is literally set to "No CUDA devices found.-real".) AMReXCUDAOptions.cmake:24-28 should sanity-check the value, and GPU.rst:299-301 plus
  the PR body should say the failure is deferred to compile time.

  6. A stale AMReX_CUDA_ARCH cache entry silently overrides the new option

  Realistic upgrade path — configure an existing build dir with development (leaves AMReX_CUDA_ARCH:STRING=Auto in the cache), then switch to this branch:
  cmake -S . -B build -DCMAKE_CUDA_ARCHITECTURES=90     # ignored
  --    CUDA architectures: native -> 120-real
  CMAKE_CUDA_ARCHITECTURES:STRING=native
  The user follows the PR's own migration advice and it does nothing, with no visible cache entry explaining why (AMReX_CUDA_ARCH is no longer a declared option). Suggest
  unset(AMReX_CUDA_ARCH CACHE) after translating it — that also makes the deprecation warning fire once instead of on every reconfigure.

  7. Docs / minor

  - Docs/sphinx_documentation/source/BuildingAMReX_Chapter.rst:26 still says "the minimal requirement is version 3.18" — needs 3.25.
  - Tests/CMakeTestInstall/CMakeLists.txt:6 and Tests/SpackSmokeTest/CMakeLists.txt:9 still declare cmake_minimum_required(VERSION 3.18).
  - Legacy forms that used to work now hard-error: -DAMReX_CUDA_ARCH="7.0+PTX" fails in CMakeCUDAArchitecturesValidate.cmake; All/Common likewise. Worth one line in the
  migration note. (Kepler/Maxwell don't matter — AMReX requires CC ≥ 6.0.)
  - AMREX_CUDA_ARCH is also the GNU Make variable (Tools/GNUMake/Make.defs:1160, documented at BuildingAMReX.rst:67,144). Anyone exporting it for make builds now gets a
  CMake warning telling them to stop. The warning text should say "for CMake builds".
  - PR description says SENSEI CI is disabled — there is no SENSEI workflow in this repo and no such change in the diff (looks copied from WarpX). The --diag-error 20092
  bullet is also stale: that landed in development via #5555. The branch is 22 commits behind and re-applies it; a rebase would clean both up.

  ---
  Items 1–4 are the ones I'd hold on. 3 and 4 are one-liners; 1 and 2 need a small decision about how strictly to couple LTO to RDC.

WeiqunZhang:
Codex:

## Finding

- **[P2] Resolve `all`/`all-major` before enabling CUDA IPO.**

  [`AMReXCUDAOptions.cmake`](https://github.com/AMReX-Codes/amrex/blob/dcdd11a59c4cad1bccb463207ccbacebe84cc92d/Tools/CMake/AMReXCUDAOptions.cmake#L22-L48) resolves `native`, but leaves CMake's documented `all` and `all-major` aliases unchanged. [`AMReXParallelBackends.cmake`](https://github.com/AMReX-Codes/amrex/blob/dcdd11a59c4cad1bccb463207ccbacebe84cc92d/Tools/CMake/AMReXParallelBackends.cmake#L126-L145) then enables IPO with that literal architecture.

  With CMake 4.2.3 and CUDA 13.3, configuring with

  ```text
  -DAMReX_CUDA_LTO=ON
  -DCMAKE_CUDA_ARCHITECTURES=all-major
  ```

  produced CUDA compile commands containing `-arch=all-major` but no `code=lto_*`. `cuobjdump --list-elf` showed ordinary `sm_75`, `sm_80`, `sm_90`, `sm_100`, `sm_110`, and `sm_120` cubins. In contrast, an explicit architecture-80 build correctly emitted `code=[compute_80,lto_80]` and its object contained no compiled cubin.

  Therefore, `AMReX_CUDA_LTO=ON` silently does not provide device LTO for `all`, `all-major`, or the legacy `All`/`Common` values that normalize to them, despite those architecture forms being documented as supported. The aliases should be expanded to explicit architectures before setting IPO, or the LTO/alias combination should be rejected with a diagnostic.

WeiqunZhang:
# Review: AMReX PR #4773 — "Modernize CMake CUDA to 3.25+ (native arch, device LTO)"

- **Repo:** AMReX-Codes/amrex
- **PR:** https://github.com/AMReX-Codes/amrex/pull/4773
- **Author:** ax3l · **Branch:** `cmake-3.25` · **Head:** `5bc76eb0f19d2a2a6925231bf7a38f5b0418d703`
- **Base:** `13221b4ab1c6bd9ecbbc2d8c77268bac5175d8b6` (`development`)
- **Size:** 19 files, +820 / −448, 25 commits · **State:** open, approved by WeiqunZhang, **30/30 CI checks green**
- **Reviewed:** 2026-08-20
- **Test host:** CUDA 13.3 (`nvcc` 13.3.73), NVIDIA GeForce RTX 3060 (sm_86), CMake 3.28.3, 16 cores

---

## Verdict

The logic holds up under real testing. **No correctness bugs found.** Eight findings
below, none blocking: one substantive gap (partial LTO), one semantics question, and
six nits / doc corrections.

This was verified by execution, not inspection: 14 configure scenarios, a full device-LTO
library build, and running the result on the GPU.

---

## What was verified

### Architecture selection

| Input | Result |
|---|---|
| (nothing) | `native → 86` |
| `-DCMAKE_CUDA_ARCHITECTURES=80` | `80` |
| `-DAMReX_CUDA_ARCH=8.0` | `80` + deprecation warning, cache entry removed after use |
| `-DAMReX_CUDA_ARCH=Volta` | `70` (then fails CMake's compiler test — CUDA 13.3 dropped sm_70; correct) |
| `-DAMReX_CUDA_ARCH=Blackwell` | `100;120` |
| `AMREX_CUDA_ARCH="7.0 8.0"` (env) | `70;80` (whitespace list handled) |
| `-DCMAKE_CUDA_ARCHITECTURES=50` | fatal error, actionable message |
| `-DCMAKE_CUDA_ARCHITECTURES="50;80"` | warns, builds `80` |
| `-DCMAKE_CUDA_ARCHITECTURES=OFF` | passes through (packager path) |
| `-DAMReX_CUDA_ARCH=` (blank) | correctly treated as unset → `native` |
| `native` with `CUDA_VISIBLE_DEVICES=""` | clear warning, defers to compiler, configure continues |

### Precedence across re-configures (the subtle part — all correct)

1. `-DCMAKE_CUDA_ARCHITECTURES=80` → `80`, remembered as `AMREX_CUDA_ARCHS_CMDLINE`
2. re-configure, no args → stays `80`, **silently** (no spurious warning)
3. `-DAMReX_CUDA_ARCH=8.9` → switches to `89`, deprecation warning, cmdline marker dropped
4. re-configure, no args → stays `89`
5. re-configure with `AMREX_CUDA_ARCH=8.6` exported → flips to `86` **and** fires
   *"The CUDA architectures of this build directory change from '89' to '86', although
   none were requested on this command line"*

That last guard is the strongest part of the design — it makes the one genuinely
surprising behavior (hints re-evaluated every configure) visible instead of silent.

### `amrex_expand_cuda_archs_alias` matches nvcc exactly

Ground truth from `nvcc -arch=all[-major] --dryrun -c e.cu` on CUDA 13.3:

- `all-major` → SASS `75, 80, 90, 100, 110, 120` + PTX `120`
- `all` → SASS `75, 80, 86, 87, 88, 89, 90, 100, 103, 110, 120, 121` + PTX `120`

AMReX produces `75-real;80-real;90-real;100-real;110-real;120` for `all-major` — identical,
including the non-obvious cases: `--list-gpu-arch` prints unsorted (`110` before `103`,
handled by `list(SORT ... COMPARE NATURAL)`), `110` counts as major, and PTX lands on
`120` rather than the newest arch `121`.

### Device LTO

- `code=[compute_86,lto_86]` on all 118 AMReX TUs
- `-dlto` at the consumer device link
- exported as `INTERFACE_LINK_OPTIONS "$<DEVICE_LINK:$<$<CUDA_COMPILER_ID:NVIDIA>:-dlto>>"`
  in the installed `AMReXTargets.cmake`
- shared-library build gets `-dlto` on its own device link (once)
- full build **104/104 clean, zero warnings**; both test executables ran correctly on the GPU
- host-IPO neutering verified: **zero** occurrences of `-flto`, `gcc-ar`, `gcc-ranlib`
  in the generated build

The directory-scope trick in `AMReXParallelBackends.cmake` is safe: it is included from
`Src/CMakeLists.txt:40`, the same scope where the `amrex_${D}d` targets are created
(`Src/CMakeLists.txt:10-13`). Subdirectories added later (`Src/Base`, `Src/AmrCore`, …)
only add sources to those existing targets, so the IPO variables are read from the
right scope.

`AMREX_CUDA_ARCHS` now exports the resolved `86` rather than the literal `native` —
the earlier review comment in the PR thread on that point is addressed.

### Configure-time guards for LTO

| Input | Result |
|---|---|
| `-DAMReX_CUDA_LTO=ON -DAMReX_GPU_RDC=OFF` | fatal (`AMReXCUDAOptions.cmake:26`) |
| `-DAMReX_CUDA_LTO=ON -DCMAKE_CUDA_ARCHITECTURES=OFF` | fatal (`AMReXCUDAOptions.cmake:133`) |
| `-DAMReX_CUDA_LTO=ON -DCMAKE_CUDA_ARCHITECTURES=all-major` | expands to concrete list |
| `-DAMReX_CUDA_LTO=ON -DAMReX_BUILD_SHARED_LIBS=ON` | works |

### Cleanup verified

No leftover references anywhere to `AMReX_SetupCUDA`, `set_nvcc_arch_flags`,
`NVCC_ARCH_FLAGS`, `convert_cuda_archs`, `set_cuda_architectures`,
`CUDA_FORWARD_UNKNOWN_FLAGS_HOST`, or `cuda_select_nvcc_arch_flags`. All
`cmake_minimum_required` sites bumped to 3.25 (`CMakeLists.txt:1`,
`Tests/CMakeTestInstall/CMakeLists.txt:6`, `Tests/SpackSmokeTest/CMakeLists.txt:9`);
the docs floor in `BuildingAMReX_Chapter.rst:26` matches.

---

## Findings

### 1. `AMReX_CUDA_LTO=ON` only ever gives users *partial* LTO, and the docs don't say so

Every consumer link prints:

```
nvlink info : requested LTO but '.../main.cpp.o' not built for LTO so doing partial LTO
```

`setup_target_for_cuda_compilation` (`Tools/CMake/AMReXTargetHelpers.cmake:125`) sets
`CUDA_ARCHITECTURES` but never `INTERPROCEDURAL_OPTIMIZATION`, so tests, tutorials and
downstream code compile straight to SASS and cannot participate in the LTO.

AMReX's own device code *is* LTO'd, so the option is not broken. But most AMReX device
code is header-instantiated in the **user's** translation units, so the practical win is
smaller than *"enables CUDA device link-time optimization"* (`GPU.rst:356`) implies.

**Suggested fix:** either propagate IPO in `setup_target_for_cuda_compilation` when
`AMReX_CUDA_LTO` is on, or state the limitation in `GPU.rst`.

### 2. `all` / `all-major` are still exported as the literal alias

When LTO is off and the toolkit has no sub-6.0 architectures, the export is
`AMREX_CUDA_ARCHS:INTERNAL=all-major`, so a downstream project re-expands it against
*its own* toolkit. The same input with `AMReX_CUDA_LTO=ON` exports the concrete list
instead.

This is the residual half of the `native` comment from the PR thread. It is defensible as
CMake-native semantics — the user explicitly asked for the alias — but it should be a
deliberate choice rather than a side effect of which branch expanded the alias.

### 3. `-dlto` is duplicated in consumer device links

`LINK_FLAGS = -dlto -dlto` for each consumer target. nvcc tolerates it; purely untidy.

### 4. Two variables one character apart meaning different things

- `_amrex_cuda_archs_cmdline_now` — about `CMAKE_CUDA_ARCHITECTURES` (`AMReXCUDAArchs.cmake:83`)
- `_amrex_cuda_arch_cmdline_now` — about `AMReX_CUDA_ARCH` (`AMReXCUDAArchs.cmake:87`)

Both appear together in the file's trickiest expressions (`:136`, `:273`). A typo-bug
waiting to happen; worth renaming one of them (e.g. `_amrex_legacy_arch_cmdline_now`).

### 5. PR description is stale on the `native` failure mode

The body says `native` *"hard-fails at configure time if no GPU is present"*. It actually
**warns and defers to the compiler**, so the failure comes at build time
(`nvcc fatal : Cannot find valid GPU for '-arch=native'`). Verified with
`CUDA_VISIBLE_DEVICES=""`.

`GPU.rst` describes this correctly — only the PR description needs a touch-up before merge.

### 6. Worth folding in: the 3.25 floor doesn't reach downstream consumers

`Tools/CMake/AMReXConfig.cmake.in:26` is:

```cmake
cmake_minimum_required(VERSION @CMAKE_VERSION@)
```

so the installed config on this machine demands **CMake 3.28.3** from consumers. Pre-existing
(introduced in #2012), but it is squarely this PR's subject: bumping the floor to 3.25 has no
effect on users of an installed AMReX as long as this pins the exact build-time CMake.

### 7. Incidental fix worth knowing about: `AMReX_CUDA_SHOW_CODELINES`

The old default was broken. `option(AMReX_CUDA_SHOW_CODELINES "..." AMReX_CUDA_SHOW_CODELINES_DEFAULT)`
passes a variable **name**, which CMake evaluates as `OFF` (confirmed with a standalone
`option()` test). The `if (CMAKE_CUDA_COMPILER_VERSION VERSION_GREATER_EQUAL 12.1)` branch
above it therefore never had any effect.

The new plain `OFF` preserves the real prior behavior, and the doc table change
`Auto → NO` is a correction, not a regression.

### 8. Coverage gap (not a defect): installed-CUDA export path is untested by CI

The new `-dlto` interface option and `AMREX_CUDA_ARCHS` export are not exercised in CI:
the CUDA jobs never run `Tests/CMakeTestInstall`, and that project is not CUDA-aware
anyway — it compiles AMReX headers with plain `g++` and dies on `cub/cub.cuh`. Export
contents were verified by hand instead.

WeiqunZhang:
I have filed some of the non-blocking findings as issues that we can address later. For finding 5, I will fix it here to make it fail loudly.

PR #4775, opened 2025-11-10
Add macro for uninstall target and fix symlink tracking

jmsexton03:
## Summary
- Add `add_uninstall_target` macro to create reusable uninstall target
- ~Track libamrex symlink in install manifest for proper cleanup:~ done via #5277
- Add `AMReX_cmake_uninstall.in` template for uninstall script
- [ ] after #5283 was merged, extend the install smoke test

## Additional background

Previously, the libamrex symlink was not tracked in the install manifest. Newer CMake versions provide better ways to reference source locations using variables instead of hardcoded paths, allowing manifest information to be stored more accurately. The new macro also prevents conflicts when multiple projects using AMReX each try to create their own uninstall targets.

## Checklist

The proposed changes:


ax3l:
Thank you for this PR and sorry for the long wait.

There is a small problem: the install prefix now lacks `libamrex.a`, while the symlink appears in the original configure-time prefix instead.

The symlink install logic around `Tools/CMake/AMReXInstallHelpers.cmake:74` still derives `symlink_name` from the configure-time prefix, so `cmake --install --prefix <new-prefix>` can silently write `libamrex.a` into the old tree instead of the requested one. So also the uninstall targets the wrong location.

ax3l:
Thanks again for putting this together, and sorry this sat for so long.

After looking at a recent PR in #5277, one can fold the symlink/install-manifest bug fix in there, mainly because that PR was already the narrower diff for the existing `cmake --install --prefix ...` failure.

The legacy link is now also recorded in `install_manifest.txt`, so manifest-based uninstall flows will remove it correctly as well. In other words, the install bookkeeping piece from your PR is covered there now, even though we did not carry over the separate uninstall target helper.

If you still want the reusable uninstall target machinery, that could still make sense to add (e.g. update this PR?). This is useful on its own and independent from the symlink bug.

ax3l:
updated PR accordingly :) does that work?

PR #4778, opened 2025-11-13
EB: First batch of Marching Cubes

WeiqunZhang:
This adds the marching cubes algorithm of "Efficient implementation of Marching Cubes' cases with topological guarantees" by Lewiner, Lopes, Vieira & Tavares, Journal of Graphics Tools 8(2): pp. 1-15 (2003). The implementation is adapted from the source code available at http://thomas.lewiner.org/publication_page.php%EF%B9%96pubkey=marching_cubes_jgt.html.

Given a signed distance function (e.g., from our STL tools), this generates a list of topologically consistent triangles. Every cut cell's EB surfaces are decomposed into up to 9 triangles. There could be multiple cuts and/or volumes in a cell. The cuts are shared faces and edges between neighbor cells are consistent.

For debugging purpose, we have also added a function that can save these triangles into a STL file.

Future work involves the computation of EB information such as volume fraction, area faction, etc. based on the triangles.


ruohai0925:
Hi @Weiqun, are there examples that I can test and play with this feature?

WeiqunZhang:
There is a test at `Tests/EB/MarchingCubes`. It reads an STL file and write out another one. The output is the re-triangulation of the EB surface within cells.

ankithadas:
Can we expect Multi Cut-Cell support with this PR sometime in the future ?

WeiqunZhang:
Currently we are not actively working on it. So we cannot promise anything.

ruohai0925:
Hi Weiqun, I have another question. Since marching cube algorithm cuts cubes based on sdf values, it is highly possible to generate very very small cut cells in real complex cases. Are there some special treatments for these tiny cells in this PR? Do you have some suggestions? Thanks!

WeiqunZhang:
No, we are not doing anything special here. I think that belongs to the next step of generating EB information based on the triangles generated with the marching cubes.

asalmgren (approved):

PR #4780, opened 2025-11-13
Add Amr::derive overloads for all levels

mirenradia:
## Summary

This allows the user to create a vector of `MultiFab`s (one per level) with the derived quantity computed on each level. The overloads match those in `AmrLevel`.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Amr/AMReX_Amr.H
+++ b/Src/Amr/AMReX_Amr.H
@@ -222,6 +222,16 @@ public:
                                       Real           time,
                                       int            lev,
                                       int            ngrow);
+    //! Compute derived data on all levels and store in the dcompth component of
+    //! each Multifab in mf
+    void derive (const std::string&       name,
+                 Real                     time,
+                 const Vector<MultiFab*>& mf,
+                 int                      dcomp);
+    //! Compute derived data on all levels and return as a vector of MultiFab pointers
+    Vector<std::unique_ptr<MultiFab>> derive (const std::string& name,
+                                              amrex::Real        time,
+                                              int                ngrow);
     //! Name of the restart chkpoint file.
     const std::string& theRestartFile () const noexcept { return restart_chkfile; }
     //! Name of the restart plotfile.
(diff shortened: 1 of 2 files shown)


Src/Amr/AMReX_Amr.H
```diff
@@ -222,6 +222,16 @@ public:
                                       Real           time,
                                       int            lev,
                                       int            ngrow);
+    //! Compute derived data on all levels and store in the dcompth component of
+    //! each Multifab in mf
+    void derive (const std::string&                 name,
+                 Real                               time,
+                 Vector<std::unique_ptr<MultiFab>>& mf,
```
WeiqunZhang: I think we should change this to `Vector<MultiFab*> const& mf`. There are two main advantages. (1) The user might have `Vector<MultiFab> mf`. It's hard for them to use the `unique_ptr` interface. (2) The ownership is unclear in the `unique_ptr` interface. The function takes `Vector<std::unique_ptr<MultiFab>> & mf`. Is the function going to resize the vector? is the function going to allocate new `MultiFabs`? The raw pointer interface give the user a clear signal, because the vector is const that prevents the modification of the raw pointers. It's up to the user to own the MultiFabs in the vector.

Note that amrex provides a helper function for converting from `Vector<std::unique_ptr<T>>` to `Vector<T*>`. https://amrex-codes.github.io/amrex/doxygen/namespaceamrex.html#a75a0cde85de080ac29b7c4510da71e16 So if you have `Vector<std::unique_ptr<MultiFab>> mf`, you can just call `derive(..., amrex::GetVecOfPtrs(mf));`.

WeiqunZhang (changes requested):

Src/Amr/AMReX_Amr.H
mirenradia: That makes sense. I've made the change to this overload. Are you happy with leaving the other overload returning `Vector<std::unique_ptr<MultiFab>>`? I guess the reasons you give don't really apply in this case.

WeiqunZhang: Yes, returning `Vector<std::unique_ptr>` is fine.

WeiqunZhang:
I merged development branch into this so that the CUDA ci can pass.

WeiqunZhang (approved):

PR #4786, opened 2025-11-15
Environment variable: AMREX_THE_ARENA_INIT_SIZE

WeiqunZhang:
We can now set the default value of amrex.the_arena_init_size with an environment variable. This is convenient for CI jobs.

### DIFF
--- a/Src/Base/AMReX_Arena.cpp
+++ b/Src/Base/AMReX_Arena.cpp
@@ -6,6 +6,7 @@
 
 #include <AMReX.H>
 #include <AMReX_BLProfiler.H>
+#include <AMReX_IParser.H>
 #include <AMReX_Print.H>
 #include <AMReX_ParallelDescriptor.H>
 #include <AMReX_ParmParse.H>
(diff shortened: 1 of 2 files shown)


Src/Base/AMReX_Arena.cpp
```diff
@@ -362,6 +362,11 @@ Arena::Initialize (bool minimal)
     the_pinned_arena_release_threshold = Gpu::Device::totalGlobalMem() / Gpu::Device::numDevicePartners() / 2L;
 #endif
 
+    // Overwrite the initial size with environment variables
+    if (char const* init_size_p = std::getenv("AMREX_THE_ARENA_INIT_SIZE")) {
+        the_arena_init_size = std::stoi(init_size_p);
```
AlexanderSinn:
```suggestion
        the_arena_init_size = std::stoll(init_size_p);
```
int could overflow here

WeiqunZhang: Absolutely

WeiqunZhang:
We can also use IParser, that will allow `10'000'000'000`and `10e9`.

asalmgren (approved):

PR #4789, opened 2025-11-16
Enable Intel in GitLab CI

WeiqunZhang:
This also adds timeout to avoid infinite loop.

Post a comment immediately after being triggered.


WeiqunZhang:
The intel GitLab CI is expected to fail. /run-hpsf-gitlab-ci

asalmgren (approved):

PR #4792, opened 2025-11-16
Re-enable some SYCL tests

WeiqunZhang:
This will cause the gitlab CI tests to fail.


asalmgren (approved):

WeiqunZhang:
/run-hpsf-gitlab-ci I think this the SYCL tests will pass.

PR #4824, opened 2025-12-01
Workaround for MPICH MPIX_Query_cuda_support assertion error

### LINKED ISSUE #4820: MPICH <4: Broken MPIX_Query_cuda_support
Calling `MPIX_Query_cuda_support` will raise an `assert` for all MPICH implementations that are:
- less than v4.0 and
- compiled w/o GPU support

Thus, we might need to introduce more logic here, since Clusters like Perlmutter (NERSC) still use MPICH 3.4.3:
https://github.com/AMReX-Codes/amrex/blob/25.11/Src/Base/AMReX_ParallelDescriptor.cpp#L1547-L1549

MPICH: https://github.com/pmodels/mpich/issues/5716#issuecomment-996078783
WarpX: https://github.com/BLAST-WarpX/warpx/pull/6389

WeiqunZhang:
For MPICH < 4, MPIX_Query_cuda_support will trigger an assertion error if its call to PMPIX_GPU_query_support returns an error code. Here, we work around it by calling MPIX_GPU_query_support directly. If it returns an error code, it is treated as not supporting CUDA.

Close #4820

### DIFF
--- a/Src/Base/AMReX_ParallelDescriptor.cpp
+++ b/Src/Base/AMReX_ParallelDescriptor.cpp
@@ -1544,8 +1544,13 @@ Initialize ()
 {
 #if defined(AMREX_USE_CUDA)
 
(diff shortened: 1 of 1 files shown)


atmyers (approved):

ax3l (approved):
Thank you!

PR #4828, opened 2025-12-03
Fix comparison with zero warning

atmyers:
This removes warnings like:

```
AMReX_Array.H(158): warning #186-D: pointless comparison of unsigned integer with zero
```

that can arise in cases where `NAR` or `NAI` is zero. 

The proposed changes:


WeiqunZhang:
Wait for #4823

WeiqunZhang (approved):

PR #4832, opened 2025-12-04
clear freeAsync buffer before allocating new memory

AlexanderSinn:
## Summary

Clear the buffer used by freeAsync every time the CArena needs to make a new memory allocation.

## Additional background

So far I don't have a test where this would be triggered.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_CArena.cpp
+++ b/Src/Base/AMReX_CArena.cpp
@@ -59,18 +59,15 @@ CArena::alloc_protected (std::size_t nbytes)
         // We need to unlock the mutex for this so that free() can be called.
         // This may invalidate free_it.
         carena_mutex.unlock();
-        const bool freed_something = Gpu::clearFreeAsyncBuffer();
+        Gpu::clearFreeAsyncBuffer();
         carena_mutex.lock();
 
-        if (freed_something) {
-            free_it = m_freelist.begin();
-            for ( ; free_it != m_freelist.end(); ++free_it) {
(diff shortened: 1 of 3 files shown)


Src/Base/AMReX_CArena.cpp
```diff
@@ -60,6 +60,28 @@ CArena::alloc_protected (std::size_t nbytes)
...
+#ifdef AMREX_USE_GPU
+    if (free_it == m_freelist.end()) {
+        // Clear Gpu::freeAsync buffer and try again.
+        // We need to unlock the mutex for this so that free() can be called.
+        // This may invalidate free_it.
+        carena_mutex.unlock();
+        const bool freed_something = Gpu::clearFreeAsyncBuffer();
+        carena_mutex.lock();
+
+        if (freed_something) {
```
WeiqunZhang: Even if nothing is freed by `Gpu:clearFreeAsyncBuffer`, we may still want to check again, because `carena_mutex` was unlocked and during that time things might have changed.

WeiqunZhang: That also means we don't really need to have `wait_list_size()`.

WeiqunZhang: And `wait_list_size()` could be outdated anyway.

AlexanderSinn: We do still need `wait_list_size()` to avoid a stream sync.

WeiqunZhang (approved):

PR #4837, opened 2025-12-09
fix: SpMV CPU version with out of bounds access

julien-fausty-gev:
## Summary

Currently, when computing matrix vector multiplication on the CPU, the `amrex::SpMV` method accesses memory of the input vector wherever the matrix has non zero column indices. This can lead to errors since accessing out of bounds memory is UB and can effectively lead to completely wrong results.

This commit proposes first a fix by checking if the column value is in the right range before proceeding to the multiplication.

This commit also proposes to put the OpenMP loop one loop higher to avoid excessive thread spawning at this local matrix multiplication.

## Checklist

The proposed changes:

### DIFF
--- a/Tests/Algebra/SpMV/main.cpp
+++ b/Tests/Algebra/SpMV/main.cpp
@@ -1,0 +1,85 @@
+#include <AMReX_GMRES_MV.H>
+#include <AMReX_SpMV.H>
+
+#include <AMReX.H>
+
+using namespace amrex;
+
+int main(int argc, char *argv[]) {
+  amrex::Initialize(argc, argv);
+  {
+    AlgVector<Real> xvec(100);
+    AlgVector<Real> bvec(xvec.partition());
+    AlgVector<Real> exact(xvec.partition());
+
+    int num_non_zeros = 2;
+    SpMatrix<Real> mat(xvec.partition(), num_non_zeros);
+
+    auto *rhs = bvec.data();
+    auto *phi = exact.data();
+    auto nrows = bvec.numLocalRows();
+    auto ib = bvec.globalBegin();
+
+    auto *matVals = mat.data();
+    auto *matRowOffsets = mat.rowOffset();
+    auto *matCols = mat.columnIndex();
+
+    // simple algebraic system:
+    //
+    //     s_n + 2.0 * s_{n+1} = (n + 1) + 2 * (n + 2)
+    //
+    // independent on each processor
+    ParallelFor(nrows - 1, [=] AMREX_GPU_DEVICE(Long lrow) {
+      auto row = lrow + ib; // global row index
+
+      rhs[lrow + 1] = static_cast<Real>((lrow + 1.0) + 2.0 * (lrow + 2.0));
+      phi[lrow + 1] = static_cast<Real>(lrow + 2);
+
+      matRowOffsets[lrow + 1] = static_cast<Long>(2 * (lrow + 1));
+      matCols[2 * (lrow + 1)] = row + 1;
+      matCols[2 * (lrow + 1) + 1] = row;
+      matVals[2 * (lrow + 1)] = static_cast<Real>(2.0);
+      matVals[2 * (lrow + 1) + 1] = static_cast<Real>(1.0);
+    });
+
+    // boundary condition: s_0 = 1.0
+    ParallelFor(1, [=] AMREX_GPU_DEVICE(Long) {
+      // global row index
+
+      rhs[0] = 1.0;
+      phi[0] = 1.0;
+
+      matRowOffsets[0] = 0;
+      matCols[0] = ib;
+      matCols[1] = ib + 1;
+      matVals[0] = 1.0;
+      matVals[1] = 0.0;
+    });
+
+    auto eps = (sizeof(Real) == 4) ? Real(1.e-5) : Real(1.e-12);
+    amrex::SpMV(xvec, mat, exact);
+
+    // Check the multiplication
+    amrex::Axpy(xvec, Real(-1.0), bvec);
+
+    Real multiplicationError = xvec.norminf();
+
+    xvec.setVal(1.0);
+
+    GMRES_MV<Real> gmres(&mat);
+    gmres.setPrecond(JacobiSmoother<Real>(&mat));
+    gmres.setVerbose(2);
+
+    gmres.solve(xvec, bvec, Real(0.0), eps);
+
+    // Check the solution
+    amrex::Axpy(xvec, Real(-1.0), exact);
+
+    auto solveError = xvec.norminf();
+    amrex::Print() << " Max norm error: multiplication = "
+                   << multiplicationError << ", solve = " << solveError << "\n";
+
+    AMREX_ALWAYS_ASSERT(multiplicationError < eps && solveError < eps);
+  }
+  amrex::Finalize();
+}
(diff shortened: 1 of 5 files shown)


julien-fausty-gev:
@WeiqunZhang please take a look at this PR if you have some time 🙂

WeiqunZhang:
Let's leave OpenMP aside for now, because the out-of-bound access would be a critical error. Do you have a test showing there is an issue? We have a test at `amrex/Tests/Algebra/GMRES` showing the proposed change in this PR does NOT work. Could you try the test on two processes?

Note that we split the matrix into a local part and a remote part. The part you touched is the local part and some of the indices there are supposed to be local.

WeiqunZhang:
The OpenMP change is certainly an improvement. Thanks!

julien-fausty-gev:
@WeiqunZhang 

Thanks for your comments!

> Let's leave OpenMP aside for now, because the out-of-bound access would be a critical error. Do you have a test showing there is an issue? We have a test at amrex/Tests/Algebra/GMRES showing the proposed change in this PR does NOT work. Could you try the test on two processes?
> 
> Note that we split the matrix into a local part and a remote part. The part you touched is the local part and some of the indices there are supposed to be local.

I will run the test and debug on my side and try to add a supplemental test that fails without the fix. I can ping you when I push the changes.

julien-fausty-gev:
@WeiqunZhang, I can reproduce the failure of the `Tests/Algebra/GMRES` in parallel with my change.

I have taken the liberty of adding a test for the `SpMV` operation however. This test fails with the previous `SpMV` implementation and passes with the proposed changes. Moreover, this test also runs a `GMRES` solve on the system with the same conclusions.

My guess is that there is something funky in the treatment of the solution vectors in the GMRES algorithm. I do not know the code base well enough to identify it exactly though. I would greatly appreciate your help here 🙂.

Tests/Algebra/SpMV/main.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+      matRowOffsets[lrow + 1] = static_cast<Long>(2 * (lrow + 1));
+      matCols[2 * (lrow + 1)] = row + 1;
+      matCols[2 * (lrow + 1) + 1] = row;
+      matVals[2 * (lrow + 1)] = static_cast<Real>(2.0);
+      matVals[2 * (lrow + 1) + 1] = static_cast<Real>(1.0);
+    });
+
+    // boundary condition: s_0 = 1.0
+    ParallelFor(1, [=] AMREX_GPU_DEVICE(Long) {
```
WeiqunZhang, line 46: This setting the local row 0 on every process, not just the global row 0.

WeiqunZhang:
Your setup of the test is incorrect. It's possible there might be a bug in the current code in development branch. However, I I have NOT seen any evidence whatsoever and I am very confident that your change is definitely wrong. My suggestion is please close this PR and submit a new PR with the OMP change **only**.

WeiqunZhang:
If I try to set your test up in a way that is independent of the number of MPI processes, it would be something like,
```
    ParallelFor(nrows, [=] AMREX_GPU_DEVICE (Long lrow)
    {
        auto row = lrow +ib; // global row index                                                                                                                                           
        if (row == 0) {
            rhs[0] = Real(1);
            phi[0] = Real(1);
            matCols[0] = 0;
            matCols[1] = 1;
            matVals[0] = Real(1);
            matVals[1] = Real(0);
        } else {
            rhs[lrow] = Real(row + 2*(row + 1));
            phi[lrow] = Real(row + 1);
            matCols[2*lrow  ] = row;
            matCols[2*lrow+1] = row-1;
            matVals[2*lrow  ] = Real(2);
            matVals[2*lrow+1] = Real(1);
        }
        matRowOffsets[lrow] = 2*lrow;
        if (lrow == nrows - 1) {
            matRowOffsets[lrow+1] = 2*(lrow+1);
        }
    });
```
Here `phi` is set to `[1 2 3 ... 100]^T`, `rhs` is set to `[1 5 8 11 14 ... 299]^T`, and the matrix is
```
[ [1 0 ... ]
  [1 2 0 ...]
  [0 1 2 0 ...]
  ...
  [0 0 ... 0 1 2]]
```
It's obvious that `rhs` is the product of M and `phi`. Then your PR branch will fail and the development branch will succeed. Hope that helps.

Tests/Algebra/SpMV/main.cpp
julien-fausty-gev: It is meant to, the system is local to each process. This case should not need any communication to succeed in fact.

julien-fausty-gev:
> If I try to set your test up in a way that is independent of the number of MPI processes, it would be something like,
> 
> ```
>     ParallelFor(nrows, [=] AMREX_GPU_DEVICE (Long lrow)
>     {
>         auto row = lrow +ib; // global row index                                                                                                                                           
>         if (row == 0) {
>             rhs[0] = Real(1);
>             phi[0] = Real(1);
>             matCols[0] = 0;
>             matCols[1] = 1;
>             matVals[0] = Real(1);
>             matVals[1] = Real(0);
>         } else {
>             rhs[lrow] = Real(row + 2*(row + 1));
>             phi[lrow] = Real(row + 1);
>             matCols[2*lrow  ] = row;
>             matCols[2*lrow+1] = row-1;
>             matVals[2*lrow  ] = Real(2);
>             matVals[2*lrow+1] = Real(1);
>         }
>         matRowOffsets[lrow] = 2*lrow;
>         if (lrow == nrows - 1) {
>             matRowOffsets[lrow+1] = 2*(lrow+1);
>         }
>     });
> ```
> 
> Here `phi` is set to `[1 2 3 ... 100]^T`, `rhs` is set to `[1 5 8 11 14 ... 299]^T`, and the matrix is
> 
> ```
> [ [1 0 ... ]
>   [1 2 0 ...]
>   [0 1 2 0 ...]
>   ...
>   [0 0 ... 0 1 2]]
> ```
> 
> It's obvious that `rhs` is the product of M and `phi`. Then your PR branch will fail and the development branch will succeed. Hope that helps.

Have you actually tried it on the development branch or is it just obvious?

julien-fausty-gev:
@WeiqunZhang thanks for your help.

I am not sure closing the PR is very conducive to open source contributions.

It is clear that the development branch has a bug.

Imagine a distribution of a matrix with 5 rows per process. Now imagine the 0th row on the 0th process as such:
```
[1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0, ...]
```
equivalent in CSR format to:
```
rowOffsets = [0, 2, ...]
columns = [0, 6, ...]
values = [1.0, 1.0]
```

The memory of the vector to multiply the matrix on the 0th rank will only have 5 values:
```
[1.0, 2.0, 3.0, 4.0, 5.0]^{T}
```
for example.

The code currently in the `development` branch looks like:
```cpp
r += mat[i] * px[col[j]];
```

Following my example above, this will lead to two operations:
```cpp
r += mat[0] * px[0];
r += mat[1] * px[6]
```

The `px[6]` value is clearly out of bounds for the vector which should only have 5 values and therefore have a maximum index of 4.

The fix here (which is rather inelegant) was to ensure that the local contribution to the matrix vector product only used vector values that are local to the process and allocated.

Unless the `AlgVector` allocates enough memory on each process to cover the entire vector (which would be counter productive in distributing the vector at all), there is clearly a risk here that the program accesses unallocated memory in the vector.

The fact that the test fails if it does not access this unallocated memory is worrying in its own right.

I guess we will just have to maintain this patch internally.

julien-fausty-gev:
I am looking into the `SpMatrix::prepare_comm` method and I am starting to see that the case I described above may not be possible

julien-fausty-gev:
@WeiqunZhang I think you can disregard my previous comments I will re-open a new PR, I believe I have found the error.

Issue #4840, opened 2025-12-09
Moving to C++20


ax3l:
Hi,

@WeiqunZhang and I would like to start the discussion when people are ready to move to require C++20 for AMReX.

We have some optional recent features, like SIMD support, that already need C++20 when used. Generally, C++20 adds new features and math functions that we would like to exploit, as usual.

From looking at other projects reporting [regressions](https://github.com/openPMD/openPMD-api/issues/1720): The usual candidate that holds us back is the CUDA Toolkit deployed on HPC machines that we run production runs on. From what I saw, I think we need to require at least CUDA/NVCC 12.4.1, even though NVCC added C++20 support already in [CUDA 12.0+](https://gist.github.com/ax3l/9489132#device-side-c-standard-support).

Community: What are your thoughts? Any important machines holding you back?

BenWibking:
We have migrated Quokka to C++20 already. The issues we've seen are related to missing `<concepts>` and `<ranges>` headers on SLES 15 since the system libstdc++ is so old. No AMReX or GPU-specific issues so far.

The one other exception to this is when using `<ranges>` with HIP and libc++ (only): https://github.com/ROCm/rocm-systems/issues/2141. This is probably very uncommon in practice.

WeiqunZhang:
@BenWibking How do you work around the missing header issue?

BenWibking:
> [@BenWibking](https://github.com/BenWibking) How do you work around the missing header issue?

@WeiqunZhang On Frontier, it doesn't seem to be an issue with these modules:
```
module load PrgEnv-cray
module load cce/20.0.0
module load rocm/6.4.2
module load cmake/4.1.0
export CC=$(which hipcc)
export CXX=$(which hipcc)
```
I haven't tested other combinations.

In our SLES container (used for CI), I added `libstdc++6-devel-gcc13` with the system package manager, and that seems to just work without changing anything else:
https://github.com/quokka-astro/quokka/blob/1dcebb5ac78db19d127a2205a5ac8e63951eba8e/.devcontainer/rocm-container/Dockerfile#L79

mirenradia:
[...]

PR #4842, opened 2025-12-10
Workaround for MSVC

WeiqunZhang:
This is needed for MSVC, because it does not seem to handle `using MLLinOpT<MF>::setLevelBC` correctly.

Additional background: https://github.com/BLAST-WarpX/warpx/issues/6406

### DIFF
--- a/Src/LinearSolvers/MLMG/AMReX_MLCellLinOp.H
+++ b/Src/LinearSolvers/MLMG/AMReX_MLCellLinOp.H
@@ -55,7 +55,7 @@ public:
                      const AMF* robinbc_b = nullptr,
                      const AMF* robinbc_f = nullptr)
     {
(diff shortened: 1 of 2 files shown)


ax3l (approved):
Thank you for digging into this!

I just reminded myself: the issue is that MSVC does a somehow no-standard/weird two-phase name lookup when dealing with C++ templates, which delays almost all template validation until instantiation...

There might be flag for modern MSVC, `/permissive-` or so, that makes it standard compliant. I think it would be possible for all our toolchains and deployments if we would require this for our code, but the devil is always in the detail, especially when it comes to working with third party include headers, too. We would have to test.
https://learn.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance?view=msvc-170
https://learn.microsoft.com/en-us/cpp/build/reference/permissive-standards-conformance?view=msvc-170

ax3l:
I saw googling somewhere that template standard compliance might be automatic in MSVC once we switch to C++20 #4840. To be seen :D

PR #4844, opened 2025-12-11
fix: `SpMatrix::prepare_comm` in case of no communication necessary

julien-fausty-gev:
## Summary

Currently, when computing matrix vector multiplication on the CPU, the `amrex::SpMV` method can accesses memory of the input vector wherever the matrix has non zero column indices if the matrix is distributed but does not need any communication to do the multiplication. This can lead to errors since accessing out of bounds memory is UB and can effectively lead to completely wrong results.

This contribution proposes first a fix by adding an else clause for the case where there are only local non-zero values in the matrix.

This commit also proposes to put the OpenMP loop one loop higher to avoid excessive thread spawning at this local matrix multiplication.

## Checklist

The proposed changes:

### DIFF
--- a/Src/LinearSolvers/AMReX_SpMV.H
+++ b/Src/LinearSolvers/AMReX_SpMV.H
@@ -192,9 +192,6 @@ void SpMV (AlgVector<T>& y, SpMatrix<T> const& A, AlgVector<T> const& x)
 
 #else
 
-    Long const rowsBegin = x.globalBegin();
-    Long const rowsEnd = x.globalEnd();
-
     Long const ny = y.numLocalRows();
 
 #ifdef AMREX_USE_OMP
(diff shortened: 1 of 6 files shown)


julien-fausty-gev:
@WeiqunZhang please review this second iteration when you have the time. The `GMRES` test should pass as should the new `SpMV` test.

WeiqunZhang:
Yes, I believe this is the correct fix. There is a minor issue in the test setup. The row offset vector has nrows+1 elements. We need to set the last element too. So I fixed the test and I also fused the two ParallelFor loops.

WeiqunZhang:
Actually we don't even need to set row offsets because they are already set correctly in the constructor of SpMatrix.

WeiqunZhang (approved):

julien-fausty-gev:
Thanks for your help @WeiqunZhang! And thanks for merging

PR #4859, opened 2025-12-18
Use PolymorphicArenaAllocator in the SOAParticle test

atmyers:
This expands the coverage to include the C++ version of the functionality.

The proposed changes:

### DIFF
--- a/Tests/Particles/SOAParticle/main.cpp
+++ b/Tests/Particles/SOAParticle/main.cpp
@@ -11,7 +11,7 @@
 
 using namespace amrex;
 
-template <typename T_PC,template<class> class Allocator=DefaultAllocator>
+template <typename T_PC,template<class> class Allocator=amrex::PolymorphicArenaAllocator>
 void addParticles ()
 {
     int is_per[AMREX_SPACEDIM];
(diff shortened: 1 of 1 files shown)


ax3l (approved):
Thank you! :)

Tests/Particles/SOAParticle/main.cpp
```diff
@@ -11,7 +11,7 @@
 
 using namespace amrex;
 
-template <typename T_PC,template<class> class Allocator=DefaultAllocator>
+template <typename T_PC,template<class> class Allocator=amrex::PolymorphicArenaAllocator>
```
ax3l, line 14: This looks like an unused template parameter to me...?
```suggestion
template <typename T_PC>
```

Tests/Particles/SOAParticle/main.cpp
```diff
@@ -37,6 +37,7 @@ void addParticles ()
     DistributionMapping dm(ba);
 
     T_PC pc(geom, dm, ba);
+    pc.SetArena(The_Arena());
```
ax3l, line 40: @atmyers what happens if someone changes the `addParticles<...>` template parameter to anything but `amrex::PolymorphicArenaAllocator` and tries to call `SetArena()`?

Just a test, but a bit funny.

AlexanderSinn: It does nothing

PR #4861, opened 2025-12-20
FFTW: Always initialize FFTW for users when OMP is on

### LINKED ISSUE #4856: AMReX FFT: FFTW w/ OpenMP Threads
In our FFTW wrapper in WarpX/ABLASTR, we support threads.
It seems AMReX does not support OpenMP/threads with FFTW yet.

ABLASTR: https://github.com/BLAST-WarpX/warpx/blob/25.12/Source/ablastr/math/fft/WrapFFTW.cpp#L44-L52
AMReX: https://github.com/AMReX-Codes/amrex/blob/25.12/Src/FFT/AMReX_FFT_Helper.H#L285-L304

This causes simulations with few/single boxes to not make use of CPU threads:
Re: https://github.com/orgs/BLAST-ImpactX/discussions/1245

WeiqunZhang:
Close #4856

### DIFF
--- a/Src/FFT/AMReX_FFT.H
+++ b/Src/FFT/AMReX_FFT.H
@@ -29,7 +29,8 @@ namespace amrex
     inline void Finalize_FFT () { amrex::Finalize_minimal(); }
 }
 
-namespace amrex::FFT
+//! Not to be called the user
+namespace amrex::FFT::detail
 {
     void Initialize ();
     void Finalize ();
--- a/Src/FFT/AMReX_FFT.H
+++ b/Src/FFT/AMReX_FFT.H
@@ -29,7 +29,7 @@ namespace amrex
     inline void Finalize_FFT () { amrex::Finalize_minimal(); }
 }
 
-//! Not to be called the user
+//! Not to be called by the user
 namespace amrex::FFT::detail
 {
     void Initialize ();
--- a/Src/FFT/AMReX_FFT.cpp
+++ b/Src/FFT/AMReX_FFT.cpp
@@ -18,7 +18,7 @@ void Initialize ()
 {
     if (s_initialized == 0)
     {
-#if defined(AMREX_USE_HIP)
+#if defined(AMREX_USE_HIP) && defined(AMREX_USE_FFT)
         AMREX_ROCFFT_SAFE_CALL(rocfft_setup());
 #elif !defined(AMREX_USE_GPU) && defined(AMREX_USE_OMP) && defined(AMREX_USE_FFT)
         fftw_init_threads();
@@ -39,7 +39,7 @@ void Finalize ()
     {
         Clear();
 
-#if defined(AMREX_USE_HIP)
+#if defined(AMREX_USE_HIP) && defined(AMREX_USE_FFT)
         AMREX_ROCFFT_SAFE_CALL(rocfft_cleanup());
 #elif !defined(AMREX_USE_GPU) && defined(AMREX_USE_OMP) && defined(AMREX_USE_FFT)
         fftw_cleanup_threads();
@@ -99,7 +99,7 @@ DistributionMapping make_iota_distromap (Long n)
     return DistributionMapping(std::move(pm));
 }
 
-#ifdef AMREX_USE_HIP
+#if defined(AMREX_USE_HIP) && defined(AMREX_USE_FFT)
 void hip_execute (rocfft_plan plan, void **in, void **out)
 {
     rocfft_execution_info execinfo = nullptr;
--- a/Src/FFT/AMReX_FFT.cpp
+++ b/Src/FFT/AMReX_FFT.cpp
@@ -1,44 +1,53 @@
 #include <AMReX_FFT.H>
 #include <AMReX_FFT_Helper.H>
+#include <AMReX_OpenMP.H>
 
 #include <map>
 
-namespace amrex::FFT
+namespace amrex::FFT::detail
 {
 
 namespace
 {
-    bool s_initialized = false;
+    long long s_initialized = 0;
     std::map<Key, PlanD> s_plans_d;
     std::map<Key, PlanF> s_plans_f;
 }
 
 void Initialize ()
 {
-    if (!s_initialized)
+    if (s_initialized == 0)
     {
-        s_initialized = true;
-
-#if defined(AMREX_USE_HIP) && defined(AMREX_USE_FFT)
+#if defined(AMREX_USE_HIP)
         AMREX_ROCFFT_SAFE_CALL(rocfft_setup());
+#elif !defined(AMREX_USE_GPU) && defined(AMREX_USE_OMP)
+        fftw_init_threads();
+        fftwf_init_threads();
+        fftw_plan_with_nthreads(amrex::OpenMP::get_max_threads());
+        fftwf_plan_with_nthreads(amrex::OpenMP::get_max_threads());
 #endif
     }
 
-    amrex::ExecOnFinalize(amrex::FFT::Finalize);
+    ++s_initialized;
+
+    amrex::ExecOnFinalize(amrex::FFT::detail::Finalize);
 }
 
 void Finalize ()
 {
-    if (s_initialized)
+    if (s_initialized == 1)
     {
-        s_initialized = false;
-
         Clear();
 
-#if defined(AMREX_USE_HIP) && defined(AMREX_USE_FFT)
+#if defined(AMREX_USE_HIP)
         AMREX_ROCFFT_SAFE_CALL(rocfft_cleanup());
+#elif !defined(AMREX_USE_GPU) && defined(AMREX_USE_OMP)
+        fftw_cleanup_threads();
+        fftwf_cleanup_threads();
 #endif
     }
+
+    if (s_initialized > 0) { --s_initialized; }
 }
 
 void Clear ()
@@ -79,11 +88,6 @@ void add_vendor_plan_f (Key const& key, PlanF plan)
 {
     s_plans_f[key] = plan;
 }
-
-}
-
-namespace amrex::FFT::detail
-{
 
 DistributionMapping make_iota_distromap (Long n)
 {
--- a/Src/FFT/AMReX_FFT.cpp
+++ b/Src/FFT/AMReX_FFT.cpp
@@ -20,7 +20,7 @@ void Initialize ()
     {
 #if defined(AMREX_USE_HIP)
         AMREX_ROCFFT_SAFE_CALL(rocfft_setup());
-#elif !defined(AMREX_USE_GPU) && defined(AMREX_USE_OMP)
+#elif !defined(AMREX_USE_GPU) && defined(AMREX_USE_OMP) && defined(AMREX_USE_FFT)
         fftw_init_threads();
         fftwf_init_threads();
         fftw_plan_with_nthreads(amrex::OpenMP::get_max_threads());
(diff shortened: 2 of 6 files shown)


Src/FFT/AMReX_FFT.cpp
```diff
@@ -1,44 +1,53 @@
...
 {
-    if (!s_initialized)
+    if (s_initialized == 0)
     {
-        s_initialized = true;
-
-#if defined(AMREX_USE_HIP) && defined(AMREX_USE_FFT)
+#if defined(AMREX_USE_HIP)
         AMREX_ROCFFT_SAFE_CALL(rocfft_setup());
+#elif !defined(AMREX_USE_GPU) && defined(AMREX_USE_OMP)
```
ax3l: I think this also needs a `defined(AMREX_USE_FFT)` guard here, otherwise AMReX compiled for CPU will require FFTW to be present, no?
```suggestion
#elif !defined(AMREX_USE_GPU) && defined(AMREX_USE_OMP) && defined(AMREX_USE_FFT)
```

Src/FFT/AMReX_FFT.cpp
```diff
@@ -1,44 +1,53 @@
...
+    if (s_initialized == 1)
     {
-        s_initialized = false;
-
         Clear();
 
-#if defined(AMREX_USE_HIP) && defined(AMREX_USE_FFT)
+#if defined(AMREX_USE_HIP)
         AMREX_ROCFFT_SAFE_CALL(rocfft_cleanup());
+#elif !defined(AMREX_USE_GPU) && defined(AMREX_USE_OMP)
```
ax3l: As above?
```suggestion
#elif !defined(AMREX_USE_GPU) && defined(AMREX_USE_OMP) && defined(AMREX_USE_FFT)
```

Src/FFT/AMReX_FFT.cpp
```diff
@@ -1,44 +1,53 @@
...
 void Initialize ()
 {
-    if (!s_initialized)
+    if (s_initialized == 0)
     {
-        s_initialized = true;
-
-#if defined(AMREX_USE_HIP) && defined(AMREX_USE_FFT)
+#if defined(AMREX_USE_HIP)
         AMREX_ROCFFT_SAFE_CALL(rocfft_setup());
```
ax3l, line 22: I think it would be awesome if for cufft/rocfft/FFTW, AMReX checks if they were already externally initialized. If yes, then AMReX does neither init nor finalize them, making AMReX more compatible in complex downstream combinations.

Tools/GNUMake/Make.defs
```diff
@@ -634,6 +634,9 @@ ifeq ($(USE_FFT),TRUE)
       LIBRARY_LOCATIONS += $(FFTW_HOME)/lib
       LIBRARIES += -Wl,-rpath,$(FFTW_HOME)/lib
     endif
+    ifeq ($(USE_OMP),TRUE)
+      LIBRARIES += -lfftw3f_omp -lfftw3_omp
```
ax3l, line 638: There are two ways in FFTW to have threading support: OpenMP and a threads implementation. I have only used the OpenMP one so far, too.

Currently, the CMake logic does treat this quite liberal: [link if found](https://github.com/AMReX-Codes/amrex/blob/26.01/Tools/CMake/FindAMReXFFTW.cmake#L71-L76).

I think we should decide: do we want to *force* an OpenMP/thread-enabled FFTW when AMReX is build with OpenMP (and otherwise error)? This is what the GNUmake logic does now.
While technically not needed, I think that could be a sensible requirement to avoid confusing users that expect OpenMP threading in their FFT computations when they enable it in AMReX. For that, we just need to add an `else()` branch with a `message(FATAL_ERROR ...)` in the above link if not found for OpenMP-enabled AMReX.

Src/FFT/AMReX_FFT.cpp
```diff
@@ -1,44 +1,53 @@
...
+#elif !defined(AMREX_USE_GPU) && defined(AMREX_USE_OMP)
+        fftw_init_threads();
+        fftwf_init_threads();
+        fftw_plan_with_nthreads(amrex::OpenMP::get_max_threads());
```
ax3l, line 26: I think we also need a path here for:
+ AMReX is on CPU
+ AMReX uses no OpenMP

and init/finalize FFTW in that case w/o threads...?

ax3l (changes requested):
Thank you! This looks great, but I have a few questions about:
- external init/finalize
- small update to third-party requirements (CMake)
- macro guards w/o FFT option enabled

Probably all quick to resolve next week :)

Src/FFT/AMReX_FFT.cpp
WeiqunZhang: We did not really need to do that because we would not include this cpp file in build if AMREX_USE_FFT is not define. But for consistency, we should add that.

Src/FFT/AMReX_FFT.cpp
WeiqunZhang: The problem is such functions do not exist.

Tools/GNUMake/Make.defs
WeiqunZhang: Either way is fine with me.

Src/FFT/AMReX_FFT.cpp
ax3l: Ah you are right, what a mess.
https://stackoverflow.com/a/50117333/2719194

Src/FFT/AMReX_FFT.cpp
WeiqunZhang: Init is not needed for FFTW on single process with a single thread, and it does not exist.

Tools/GNUMake/Make.defs
ax3l: Then let's force it, to avoid surprising users. I added a commit.

Src/FFT/AMReX_FFT.H
```diff
@@ -29,7 +29,8 @@ namespace amrex
     inline void Finalize_FFT () { amrex::Finalize_minimal(); }
 }
 
-namespace amrex::FFT
+//! Not to be called the user
```
ax3l:
```suggestion
//! Not to be called by the user
```

Src/FFT/AMReX_FFT.cpp
```diff
@@ -1,44 +1,53 @@
...
 {
-    if (!s_initialized)
+    if (s_initialized == 0)
     {
-        s_initialized = true;
-
-#if defined(AMREX_USE_HIP) && defined(AMREX_USE_FFT)
+#if defined(AMREX_USE_HIP)
         AMREX_ROCFFT_SAFE_CALL(rocfft_setup());
+#elif !defined(AMREX_USE_GPU) && defined(AMREX_USE_OMP) && defined(AMREX_USE_FFT)
```
ax3l, line 23: While updating CMake, I saw that I also add the `AMREX_FFTW_OMP` define, but this block is fine now, too.

Simpler would be
```suggestion
#elif defined(AMREX_FFTW_OMP)
```
and you need to set the same in GNUmake.

Src/FFT/AMReX_FFT.cpp
```diff
@@ -1,44 +1,53 @@
...
+    if (s_initialized == 1)
     {
-        s_initialized = false;
-
         Clear();
 
-#if defined(AMREX_USE_HIP) && defined(AMREX_USE_FFT)
+#if defined(AMREX_USE_HIP)
         AMREX_ROCFFT_SAFE_CALL(rocfft_cleanup());
+#elif !defined(AMREX_USE_GPU) && defined(AMREX_USE_OMP) && defined(AMREX_USE_FFT)
```
ax3l, line 44:
```suggestion
#elif defined(AMREX_FFTW_OMP)
```

Tools/GNUMake/Make.defs
ax3l, line 638: If you like to 100% mirror the CMake logic:
```suggestion
      DEFINES += -DAMREX_FFTW_OMP=1
      LIBRARIES += -lfftw3f_omp -lfftw3_omp
```

Tools/CMake/FindAMReXFFTW.cmake
```diff
@@ -54,7 +54,7 @@ endfunction()
 # Check if the found FFTW install location has an _omp library, e.g.,
 # libfftw3(f)_omp.(a|so) shipped and if yes, set the AMReX_FFTW_OMP=1 define.
```
ax3l: Ouch, the `e` should be uppercase in defines like `AMREX_FFTW_OMP`

ax3l: Fixed.

ax3l (approved):
LGTM, a small minor suggestion for a simpler/explicitly controlled FFTW w/ OpenMP macro.

Src/FFT/AMReX_FFT.cpp
```diff
@@ -1,44 +1,53 @@
...
 {
-    if (s_initialized)
+    if (s_initialized == 1)
     {
-        s_initialized = false;
-
         Clear();
 
-#if defined(AMREX_USE_HIP) && defined(AMREX_USE_FFT)
+#if defined(AMREX_USE_HIP)
```
ax3l: Bug? This line now always cleans up rocfft, even if `AMREX_USE_FFT` is set explictly off.

Src/FFT/AMReX_FFT.cpp
```diff
@@ -1,44 +1,53 @@
...
 
 void Initialize ()
 {
-    if (!s_initialized)
+    if (s_initialized == 0)
     {
-        s_initialized = true;
-
-#if defined(AMREX_USE_HIP) && defined(AMREX_USE_FFT)
+#if defined(AMREX_USE_HIP)
```
ax3l: Bug? This line now always inits rocfft, even if `AMREX_USE_FFT` is set explictly off.

Src/FFT/AMReX_FFT.cpp
ax3l: Correct, resolved.

Src/FFT/AMReX_FFT.cpp
WeiqunZhang: I actually was trying to remove AMREX_USE_FFT in this file, because it's not necessary. This file will not be included in compilation if AMREX_USE_FFT is not set. Anyway, I will update it to have a consistent style.

Src/FFT/AMReX_FFT.cpp
WeiqunZhang: But AMREX_FFTW_OMP is not defined in AMReX_Config.H.

ax3l: True, it would belong there, too.

ax3l (approved):

Src/FFT/AMReX_FFT.cpp
WeiqunZhang: My opinion is that we keep the old way to avoid a new macro.

ax3l: Agreed.

ax3l:
@WeiqunZhang I just realized if I pass `-DAMReX_FFTW_IGNORE_OMP=ON` then I allow non-threaded FFTWs to build, which means we will need an internal macro like `AMREX_FFTW_OMP` or `AMREX_USE_FFTW_OMP` after all to guard the symbols/init logic.

Will PR in #4941

PR #4868, opened 2025-12-26
Convert Array4 to ArrayND

ankithadas:
## Summary

This PR introduces a preliminary implementation that generalizes Array4 to an N-dimensional case, extending AMReX support for arbitrary dimensionality as outlined in #3955.

The new `ArrayND` implementation has the same size as `Array4` (64 bytes) and is intended to exhibit identical behaviour. However, this change modifies the types of `begin` and `end` and removes `ncomp`, which may break downstream code. In particular, `pyamrex` will likely need to be updated to accommodate these type changes. All internal AMReX source code that uses `begin`, `end`, or `ncomp` has been updated.

One open question concerns the behaviour of `ArrayND<dim>::operator()(idx...)` when the number of indices provided is `dim-1`. In `Array4`, this case sets the last dimension `ncomp` to zero. Should this behaviour be extended to all `dim` cases in `ArrayND` ? 

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_Array4.H
+++ b/Src/Base/AMReX_Array4.H
@@ -663,9 +663,6 @@ namespace amrex {
             constexpr_for<1, M>([&](int d) {
                 offset += (iv[d] - begin[d]) * nstride[d-1];
             });
-            constexpr_for<M, N-1>([&](int d) {
-                offset += (Long(0) - begin[d]) * nstride[d-1];
-            });
             offset += (n - begin[N-1]) * nstride[N-2];
             return offset;
         }
--- a/Src/Base/AMReX_Array4.H
+++ b/Src/Base/AMReX_Array4.H
@@ -5,7 +5,6 @@
 #include <AMReX.H>
 #include <AMReX_IntVect.H>
 #include <AMReX_GpuPrint.H>
-#include <AMReX_ConstexprFor.H>
 
 #include <iostream>
 #include <sstream>
@@ -444,11 +443,11 @@ namespace amrex {
         void index_assert (IntVectND<dim> const& iv) const
         {
             bool out_of_bounds = false;
-            constexpr_for<0, dim>([&] (auto d) {
+            for (int d = 0; d < dim; ++d) {
                 if (iv[d] < begin[d] || iv[d] >= end[d]) {
                     out_of_bounds = true;
                 }
-            });
+            }
             if (out_of_bounds) {
                 AMREX_IF_ON_DEVICE((
                     device_printf_impl(std::make_index_sequence<dim>{}, iv, begin, end);
@@ -457,15 +456,15 @@ namespace amrex {
                 AMREX_IF_ON_HOST((
                     std::stringstream ss;
                     ss << " (";
-                    constexpr_for<0, dim>([&] (auto d) {
+                    for (int d = 0; d < dim; ++d) {
                         ss << iv[d];
                         if (d + 1 < dim) ss << ",";
-                    });
+                    }
                     ss << ") is out of bound (";
-                    constexpr_for<0, dim>([&] (auto d) {
+                    for (int d = 0; d < dim; ++d) {
                         ss << begin[d] << ":" << end[d]-1;
                         if (d + 1 < dim) ss << ",";
-                    });
+                    };
                     ss << ")";
                     amrex::Abort(ss.str());
                 ))
--- a/Src/Base/AMReX_Array4.H
+++ b/Src/Base/AMReX_Array4.H
@@ -1,5 +1,6 @@
 #ifndef AMREX_ARRAY4_H_
 #define AMREX_ARRAY4_H_
+
 #include <AMReX_Config.H>
 
 #include <AMReX.H>
@@ -10,6 +11,9 @@
 #include <sstream>
 
 namespace amrex {
+
+    template<int dim>
+    class BoxND;
 
     template <typename T>
     struct CellData // Data in a single cell
@@ -167,22 +171,9 @@ namespace amrex {
         constexpr ArrayND (ArrayND<std::remove_const_t<T>, N> const& rhs) noexcept
             : p(rhs.p), nstride(rhs.nstride), begin(rhs.begin), end(rhs.end) {}
 
-        AMREX_GPU_HOST_DEVICE
-        constexpr ArrayND (T* a_p, IntVectND<N> const& a_begin, IntVectND<N> const& a_end) noexcept
-            : p(a_p), begin(a_begin), end(a_end)
-        {
-            if constexpr (N > 1) {
-                Long current_stride = 1;
-                for (int d = 0; d < N - 1; ++d) {
-                    Long len = a_end[d] - a_begin[d];
-                    current_stride *= len;
-                    nstride[d] = current_stride;
-                }
-            }
-        }
-
         // Dim3 overload for dim=4
         template <int Dim=N, std::enable_if_t<Dim==4,int> = 0>
+        AMREX_GPU_HOST_DEVICE
         constexpr ArrayND (T* a_p, Dim3 const& a_begin, Dim3 const& a_end, int a_ncomp) noexcept
             : p(a_p), begin(a_begin.x, a_begin.y, a_begin.z, 0), end(a_end.x, a_end.y, a_end.z, a_ncomp)
         {
@@ -194,10 +185,91 @@ namespace amrex {
             }
         }
 
+        // Note: BoxND constructors are not constexpr because functions in BoxND are not constexpr
+        template <int Dim>
+        AMREX_GPU_HOST_DEVICE
+        ArrayND (T* a_p, BoxND<Dim> const& box) noexcept
+            : ArrayND(a_p, box.smallEnd(), box.bigEnd() + 1)
+        {}
+
+        template <int Dim>
+        AMREX_GPU_HOST_DEVICE
+        ArrayND (T* a_p, BoxND<Dim> const& box, int ncomp) noexcept
+            : ArrayND(a_p, box.smallEnd(), box.bigEnd() + 1, ncomp)
+        {}
+
+
+        // Standard constructor for full dimension N
+        AMREX_GPU_HOST_DEVICE
+        constexpr ArrayND (T* a_p, IntVectND<N> const& a_smallend, IntVectND<N> const& a_bigend) noexcept
+            : p(a_p), begin(a_smallend), end(a_bigend)
+        {
+            if constexpr (N > 1) {
+                Long current_stride = 1;
+                for (int d = 0; d < N - 1; ++d) {
+                    Long len = a_bigend[d] - a_smallend[d];
+                    current_stride *= len;
+                    nstride[d] = current_stride;
+                }
+            }
+        }
+
+        // Reduced dimension constructor: Pads remaining dims with [0,1)
+        template <int M, std::enable_if_t<(M < N),int> = 0>
+        AMREX_GPU_HOST_DEVICE
+        constexpr ArrayND (T* a_p, IntVectND<M> const& a_smallend, IntVectND<M> const& a_bigend) noexcept
+            : p(a_p), begin(), end()
+        {
+            for (int d = 0; d < M; ++d) {
+                begin[d] = a_smallend[d];
+                end[d] = a_bigend[d];
+            }
+            for (int d = M; d < N; ++d) {
+                begin[d] = 0;
+                end[d] = 1;
+            }
+
+            if constexpr (N > 1) {
+                Long current_stride = 1;
+                for (int d = 0; d < N - 1; ++d) {
+                    Long len = end[d] - begin[d];
+                    current_stride *= len;
+                    nstride[d] = current_stride;
+                }
+            }
+        }
+
+        // Reduced dimension constructor with ncomp: Pads intermediate dims with [0,1), sets last dim to [0,ncomp)
+        template <int M, std::enable_if_t<(M < N) && (N != 1),int> = 0>
+        AMREX_GPU_HOST_DEVICE
+        constexpr ArrayND (T* a_p, IntVectND<M> const& a_smallend, IntVectND<M> const& a_bigend, int ncomp) noexcept
+            : p(a_p), begin(), end()
+        {
+            for (int d = 0; d < M; ++d) {
+                begin[d] = a_smallend[d];
+                end[d] = a_bigend[d];
+            }
+            for (int d = M; d < N-1; ++d) {
+                begin[d] = 0;
+                end[d] = 1;
+            }
+            begin[N-1] = 0;
+            end[N-1] = ncomp;
+
+            if constexpr (N > 1) {
+                Long current_stride = 1;
+                for (int d = 0; d < N - 1; ++d) {
+                    Long len = end[d] - begin[d];
+                    current_stride *= len;
+                    nstride[d] = current_stride;
+                }
+            }
+        }
+
         template <class U,
             std::enable_if_t
                   <std::is_same_v<std::remove_const_t<T>,
-                                std::remove_const_t<U>> && (N == 4),int> = 0>
+                                std::remove_const_t<U>> && (N  >= 2),int> = 0>
         AMREX_GPU_HOST_DEVICE
         constexpr ArrayND (ArrayND<U, N> const& rhs, int start_comp) noexcept
             : p((T*)(rhs.p + start_comp*rhs.nstride[N-2])),
@@ -212,7 +284,7 @@ namespace amrex {
         template <class U,
             std::enable_if_t
                   <std::is_same_v<std::remove_const_t<T>,
-                                std::remove_const_t<U>> && (N == 4),int> = 0>
+                                std::remove_const_t<U>> && (N >= 2),int> = 0>
         AMREX_GPU_HOST_DEVICE
         constexpr ArrayND (ArrayND<U, N> const& rhs, int start_comp, int num_comp) noexcept
             : p((T*)(rhs.p + start_comp*rhs.nstride[N-2])),
@@ -225,7 +297,9 @@ namespace amrex {
         }
 
         AMREX_GPU_HOST_DEVICE
-        explicit operator bool() const noexcept { return p != nullptr; }
+        explicit operator bool() const noexcept {
+            return p != nullptr && end.allGT(begin);    // Additional check for properly initialized ArrayND
+        }
 
         // This overload is enabled only when number of indices matches dimension or is one less than dimension
         template <typename... idx, class U=T,
@@ -494,6 +568,24 @@ namespace amrex {
         }
 #endif
     };
+
+    // Deduction guides for ArrayND
+    // 1: Matches ArrayND (T*, BoxND<N>) -> ArrayND<T,N>
+    template <typename T, int Dim>
+    ArrayND (T*, BoxND<Dim> const&) -> ArrayND<T, Dim>;
+
+    // 2: Matches ArrayND (T*, BoxND<N>, ncomp) -> ArrayND<T,N+1>
+    // This supports the "N+1" logic (Spatial + Component)
+    template <typename T, int Dim>
+    ArrayND (T*, BoxND<Dim> const&, int) -> ArrayND<T, Dim+1>;
+
+    // 3: Matches ArrayND (T*, IntVectND<N>, IntVectND<N>) -> ArrayND<T,N>
+    template <typename T, int N>
+    ArrayND (T*, IntVectND<N> const&, IntVectND<N> const&) -> ArrayND<T, N>;
+
+    // 4: Matches ArrayND (T*, IntVectND<N>, IntVectND<N>, int) -> ArrayND<T,N+1>
+    template <typename T, int N>
+    ArrayND (T*, IntVectND<N> const&, IntVectND<N> const&, int) -> ArrayND<T, N+1>;
 
     template<typename T>
     using Array4 = ArrayND<T,4>;
--- a/Src/Base/AMReX_Array4.H
+++ b/Src/Base/AMReX_Array4.H
@@ -71,7 +71,7 @@ namespace amrex {
         template <int N, typename... idx>
         struct ArrayNDIndexCheck_impl {
         private:
-            static constexpr int num_indicies = sizeof...(idx);
+            static constexpr int num_indices = sizeof...(idx);
 
             template <typename... Ts>
             struct AllExceptLastEnum;
@@ -95,8 +95,8 @@ namespace amrex {
                 IsValidIndexType<std::decay_t<idx>>...>::value;
 
         public:
-            static constexpr bool value = ((num_indicies == N) && index_with_enum) ||
-                                          ((num_indicies == N - 1) && index_all_values);
+            static constexpr bool value = ((num_indices == N) && index_with_enum) ||
+                                          ((num_indices == N - 1) && index_all_values);
         };
 
         template <int N, class... idx>
@@ -235,7 +235,7 @@ namespace amrex {
          * \param a_p Pointer to data.
          * \param box The domain covered by this array.
          *
-         * \TODO: Make BoxND functions constexpr to allow this constructor to be constexpr.
+         * TODO: Make BoxND functions constexpr to allow this constructor to be constexpr.
          */
         template <int M>
         AMREX_GPU_HOST_DEVICE
@@ -396,13 +396,14 @@ namespace amrex {
 
         /**
          * \brief Multi-index operator() for accessing elements.
-         * \param i... Indices for each dimension.
+         * \tparam idx Variadic template for indices.
+         * \param i Variadic list of indicies.
          * \return Reference to the element.
          *
          * This operator is only enabled when the number of indices provided matches the dimension N or N-1
          * where in the latter case the last dimension is assumed to be 0. The passed indices must be of
          * integral types that can be converted to int without narrowing. The last index can also be an enum type.
-         * \TODO: Use concepts when we move to C++20. The error message now is not user friendly.
+         * TODO: Use concepts when we move to C++20. The error message now is not user friendly.
          */
         template <typename... idx, class U=T,
             std::enable_if_t<!std::is_void_v<U> && detail::ArrayNDIndexCheck_impl_v<N, idx...>,int> = 0>
@@ -473,7 +474,8 @@ namespace amrex {
 
         /**
          * \brief Multi-index ptr() for accessing pointer to element.
-         * \param i... Indices for each dimension.
+         * \tparam idx Variadic template for indices.
+         * \param i Variadic list of indicies.
          * \return Pointer to the element.
          */
         template <typename... idx, class U=T,
--- a/Src/Base/AMReX_Array4.H
+++ b/Src/Base/AMReX_Array4.H
@@ -181,6 +181,7 @@ namespace amrex {
      * \brief A multidimensional array accessor.
      * \tparam T The type of data (e.g., Real, int).
      * \tparam N The dimension of the array (rank).
+     * \tparam last_dim_component If true, the last dimension is treated as components.
      *
      * ArrayND is a lightweight, non-owning wrapper around a pointer that provides
      * multidimensional indexing into a contiguous memory block. It is trivially
@@ -190,10 +191,11 @@ namespace amrex {
      * It stores the pointer, stride information, and the bounds (begin and end)
      * of the array.
      */
-    template<typename T, int N>
+    template<typename T, int N, bool last_dim_component = true>
     struct ArrayND
     {
         static_assert(N >= 1, "ArrayND must have at least one dimension");
+        static_assert(N > 1 || !last_dim_component, "ArrayND with N=1 cannot have last_dim_component=true");
 
         T* AMREX_RESTRICT p = nullptr;
         std::array<Long, N-1> nstride{};
@@ -562,10 +564,14 @@ namespace amrex {
          * For N!=4 case, the user is expected to keep track if they are using the last dimension
          * as components.
          */
-        template <int M=N, std::enable_if_t<M==4,int> = 0>
+        template <int M=N, std::enable_if_t<M==4 || last_dim_component,int> = 0>
         [[nodiscard]] AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
         constexpr int nComp () const noexcept {
-            return end[N-1] - begin[N-1];
+            if constexpr (last_dim_component) {
+                return end[N-1];
+            } else {
+                return end[N-1] - begin[N-1];
+            }
         }
 
         [[nodiscard]] AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
@@ -587,9 +593,16 @@ namespace amrex {
         [[nodiscard]] AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
         constexpr bool contains (IntVectND<N> const& iv) const noexcept {
             bool inside = true;
-            constexpr_for<0, N>([&](int d) {
-                inside = inside && (iv[d] >= begin[d]) && (iv[d] < end[d]);
-            });
+            if constexpr (last_dim_component == false) {
+                constexpr_for<0, N>([&](int d) {
+                    inside = inside && (iv[d] >= begin[d]) && (iv[d] < end[d]);
+                });
+            } else {
+                constexpr_for<0, N-1>([&](int d) {
+                    inside = inside && (iv[d] >= begin[d]) && (iv[d] < end[d]);
+                });
+                inside = inside && (iv[N-1] >= 0) && (iv[N-1] < end[N-1]);
+            }
             return inside;
         }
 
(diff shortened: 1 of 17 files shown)


zingale:
does `std::array` work with HIP on device?

BenWibking:
> does `std::array` work with HIP on device?

We use it. I haven't seen any issues.

Src/Base/AMReX_Array4.H
```diff
@@ -56,212 +57,373 @@ namespace amrex {
...
+            }
+        }
+
+        // Dim3 overload for dim=4
+        template <int Dim=dim, std::enable_if_t<Dim==4,int> = 0>
+        constexpr ArrayND (T* a_p, Dim3 const& a_begin, Dim3 const& a_end, int a_ncomp) noexcept
+            : p(a_p), begin(a_begin.x, a_begin.y, a_begin.z, 0), end(a_end.x, a_end.y, a_end.z, a_ncomp)
+        {
+            Long current_stride = 1;
+            constexpr_for<0, dim-1>([&] (int d) {
```
AlexanderSinn: I think it would be better to use regular for loops. Since the bounds are known at compile time, all compilers should fully unroll it anyway. constexpr_for could be a bit heavy to compile.

Src/Base/AMReX_Array4.H
```diff
@@ -56,212 +57,373 @@ namespace amrex {
...
-        template <class U=T, std::enable_if_t<!std::is_void_v<U>,int> = 0>
+        template <typename... idx, class U=T,
+            std::enable_if_t<!std::is_void_v<U> && detail::ArrayNDIndexCheck_impl_v<dim, idx...>,int> = 0>
         [[nodiscard]] AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
-        T* ptr (int i, int j, int k) const noexcept {
+        U& operator() (idx... i) const noexcept {
 #if defined(AMREX_DEBUG) || defined(AMREX_BOUND_CHECK)
-            index_assert(i,j,k,0);
+            if constexpr (sizeof...(i) == dim) {
+                index_assert(i...);
```
AlexanderSinn: You can put the `index_assert` inside `get_offset` to avoid needing so many ifdefs.

Src/Base/AMReX_Array4.H
```diff
@@ -56,212 +57,373 @@ namespace amrex {
...
+            } else if constexpr (sizeof...(i) == dim - 1) {
+                // When number of indices is one less than dimension, we assume the last index is zero
```
AlexanderSinn: In this case the last index should not be zero but effectively `begin[dim-1]`. If it were zero, the index calculation would still add instructions to subtract `begin[dim-1]`, multiply by the stride, and add the result to the offset. Currently with Array4 this is not done. To guarantee the absence of the extra instructions, there should be two versions of `get_offset` where one of them leaves out the calculation for the last index.

Src/Base/AMReX_Array4.H
```diff
@@ -56,212 +57,373 @@ namespace amrex {
...
+            return this->operator()(iv[0],0,0,0);
 #elif (AMREX_SPACEDIM == 2)
-            return this->operator()(iv[0],iv[1],0);
-#else
-            return this->operator()(iv[0],iv[1],iv[2]);
+            return this->operator()(iv[0],iv[1],0,0);
+#elif (AMREX_SPACEDIM == 3)
+            return this->operator()(iv[0],iv[1],iv[2],0);
 #endif
         }
```
AlexanderSinn: Don't these two definitions collide? Ideally, this would be one function without needing the special case.

Src/Base/AMReX_Array4.H
```diff
@@ -56,212 +57,373 @@ namespace amrex {
...
-            return this->contains(cell.x,cell.y,cell.z);
+        CellData<T> cellData (int i, int j, int k) const noexcept {
+            int ncomp = end[3] - begin[3];
+            return CellData<T>(this->ptr(i,j,k), nstride[dim-2], ncomp);
+        }
+
+
+        template <typename... idx, std::enable_if_t<sizeof...(idx) == dim,int> = 0>
+        [[nodiscard]] AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+        constexpr Long get_offset (idx... indices) const noexcept
```
AlexanderSinn: I think we can simplify this and remove this version of `get_offset`by constructing an IntVectND at the place where it is called. Then `get_impl` can also be removed.

AlexanderSinn (commented):
Currently there are a lot of special cases for `dim == 4` for both the accessors and constructors. I think in general it is useful to use an ArrayND with n-1 spatial dimensions and 1 component dimension, even if `dim != 4`. For example, I would use this for a 2+1D array. There should be these 5 accessor overloads, each for both `operator()` and `ptr()`:
- `idx... i`: This should cover both `sizeof...(idx) == dim` and `dim-1` like currently, but also replace the `int, int, int, int` version. For this, we might have to allow all types that are implicitly convertible to int for `idx`.
- `IntVectND<Dim>`:  for `Dim == dim`, `dim-1` and `DIM < dim-1`. In the last case, the remaining indexes are set to 0, except the last one, which is set to `lo[dim-1]` but not present in the offset calculation.
- `IntVectND<Dim>, int`: for `Dim == dim-1` and `DIM < dim-1`. In the second case, the IntVect is zero-extended as above. This overload won't be usable if `dim == 1`
- `Dim3` kept for backward compatibility, only works if `dim == 4`.
- `Dim3, int` kept for backward compatibility, only works if `dim == 4`.

In addition to that, I think there should be more constructors, for example, taking a BoxND as input or an ArrayND of different dimensions.

Src/Base/AMReX_Array4.H
```diff
@@ -56,212 +57,373 @@ namespace amrex {
...
 #if (AMREX_SPACEDIM == 1)
-            return this->operator()(iv[0],0,0);
+            return this->operator()(iv[0],0,0,0);
 #elif (AMREX_SPACEDIM == 2)
-            return this->operator()(iv[0],iv[1],0);
-#else
-            return this->operator()(iv[0],iv[1],iv[2]);
+            return this->operator()(iv[0],iv[1],0,0);
+#elif (AMREX_SPACEDIM == 3)
+            return this->operator()(iv[0],iv[1],iv[2],0);
```
AlexanderSinn: @WeiqunZhang  Would it be ok to put `begin` here instead of zero? That would be more consistent with the component index and would allow for a small performance optimization, as the offset resulting from that index would be zero.

Src/Base/AMReX_Array4.H
ankithadas: I have currently kept it seperate to make `get_offset` method constexpr.

ankithadas:
@AlexanderSinn  Could you provide more information regarding the types of constructors that would be beneficial for your application?

I am not convinced that `BoxND` as an input is a good idea. There is an implicit understanding that `BoxND` is related to `AMREX_SPACEDIM` of the application, whereas `ArrayND` is not necessarily tied to this. This mismatch can lead to unexpected behaviour.

AlexanderSinn:
Yes, currently lower dimensions also use Array4. I assume this is out of simplicity and to make it easy to write code that works in 1D, 2D and 3D by always passing i,j,k,n to the array. However, I think eventually it should also be possible to use the ArrayND of the dimension the code is compiled for (plus one), as this should be slightly more performant and cleaner. The constructors should allow for both ways of using the array and should essentially mirror the interface of `operator()` for what combination of inputs and dimensions are allowed. For constructors using BoxND I would suggest:

- `T*, BoxND<Dim>`:  for `Dim == dim`, `dim-1` and `DIM < dim-1`. In the last case, the remaining dimensions are set to begin=0, end=1.
- `T*, BoxND<Dim>, int ncomp`: for `Dim == dim-1` and `DIM < dim-1`. In the second case, the Box is extended with begin=0, end=1. This overload won't be usable if `dim == 1`

Additionally, the `ArrayND, ncomp` and `ArrayND, start_comp, ncomp` constructors could be made to work with `N>=2` instead of only `N==4`.

For the application I'm working on, I currently use implicit conversions from Array4 to a 2D array and a 3D array, where the 3D version removes the second to last dimension but keeps the last one. Dimensions that are removed are asserted to have size one if the constructor is called on the CPU.  This might be best left to the application, not AMReX.

ankithadas:
I believe the final issue to address is the default indices used by `ArrayND::operator()(IntVect<Dim>)` when `Dim < N`. In my opinion, the default value should be `begin` rather than zeros.

AlexanderSinn:
The forward declaration
```C++
template <typename T> struct Array4;
```
in `AMReX_BaseFwd.H` still needs to be updated.

Also in the `ArrayND, ncomp` constructors, replace `3` with `N-1`.

Src/Particle/AMReX_ParticleContainerI.H
```diff
@@ -976,9 +976,9 @@ ParticleContainer_impl<ParticleType, NArrayReal, NArrayInt, Allocator, CellAssig
             if(imf_arr(i,j,k,0)!=0)
             {
                 const auto idx = last_offset + offsets_ptr[
-                    ((i-imf_arr.begin.x)
-                    +(j-imf_arr.begin.y)*imf_arr.jstride
-                    +(k-imf_arr.begin.z)*imf_arr.kstride)
+                    ((i-imf_arr.begin[0])
+                    +(j-imf_arr.begin[1])*imf_arr.nstride[1]
```
WeiqunZhang: Should this be `nstride[0]`? A similar issue in the line below.

WeiqunZhang:
@yut23 Note that this PR will break the GDB script (https://github.com/AMReX-Codes/amrex/tree/development/Tools/GDB) contributed by you. It would be great if you can fix it when this is merged.

Src/Base/AMReX_Array4.H
```diff
@@ -56,212 +61,453 @@ namespace amrex {
         }
     };
 
-    template <typename T>
-    struct Array4
+    namespace detail {
+        /**
+         * /brief Indices checker for ArrayND
```
WeiqunZhang: If this is intended for doxygen, it should be `\brief`, not `/brief`.

Src/Particle/AMReX_ParticleContainerI.H
ankithadas: Oh yeah good catch.

Src/Base/AMReX_Array4.H
```diff
@@ -56,212 +61,453 @@ namespace amrex {
...
+        constexpr ArrayND (ArrayND<U, N> const& rhs, int start_comp) noexcept
+            : p((T*)(rhs.p + start_comp*rhs.nstride[N-2])),
               nstride(rhs.nstride),
               begin(rhs.begin),
-              end(rhs.end),
-              ncomp(rhs.ncomp-start_comp)
-        {}
+              end(rhs.end)
+        {
+            begin[3] = 0;
```
WeiqunZhang: @AlexanderSinn has already mentioned this. `3` should be replaced by `N-1`.

zingale:
one of our Castro problem setups fails to build with this PR:
http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2026-01-05-003/rad-2Tshock-3d.make.out

WeiqunZhang:
The Castro issue is that the indices passed to ArrayND are Real not integers.
```
    Real jfac = 1.0_rt;
    Real kfac = 1.0_rt;
    fine(i,j,k) = crse(clo.x, j / jfac, k / kfac) / rfac;
```

ankithadas:
> one of our Castro problem setups fails to build with this PR: http://groot.astro.sunysb.edu/Castro/test-suite/gfortran/2026-01-05-003/rad-2Tshock-3d.make.out

The error is from this line 
```(const int&, amrex::Real, amrex::Real)â€™
  327 |                 fine(i,j,k) = crse(clo.x, j / jfac, k / kfac) / rfac;
```

I believe `jfac` is of type `amex::Real`. Previously this code relied on implicit conversion from `Real` to `int` when indexing `Array4`. The changes in this PR now require all index arguments to be explicitly `int`.

Src/Base/AMReX_Array4.H
```diff
@@ -56,212 +61,453 @@ namespace amrex {
...
+         * /brief Indices checker for ArrayND
+         *
+         * 1. All indices are int
+         * 2. Number of indices matches dimension or is one less than dimension
+         */
+        template <int N, typename... idx>
+        struct ArrayNDIndexCheck_impl {
+            constexpr static bool value =
+                (sizeof...(idx) == N || sizeof...(idx) == N - 1)
+                && (Conjunction<std::is_same<int, std::decay_t<idx>>...>::value);
```
WeiqunZhang: We might want to relax this to idx is an integral type rather than it must be int.

AlexanderSinn: `IsConvertible_v<int, idx...>`  from `AMReX_TypeTraits.H` could be used for this. It will check if an implicit conversion is allowed when calling a function with that type.

WeiqunZhang: I think we should forbid implicit conversion from Real.

ankithadas: My reasoning was to prevent implicit type casting. This would prevent users from using `long` or `long long` as indexing arguments. I can change the constraint such that the index arguments have size smaller than or equal to int

ankithadas:
> I believe the final issue to address is the default indices used by `ArrayND::operator()(IntVect<Dim>)` when `Dim < N`. In my opinion, the default value should be `begin` rather than zeros.

@AlexanderSinn agrees that missing indices should default to `begin`. Any thoughts @WeiqunZhang ?

zingale:
I'm fine with modifying that part of Castro to explicitly cast to int.

Src/Base/AMReX_Array4.H
WeiqunZhang: Restrict it to signed integer with size <= sizeof(int) is fine. But we also want to support enum types. We have code doing this. I think it's a style that we should support.
```
namespace EddyDiff {
    enum {
        Mom_h = 0,
        Theta_h,
        KE_h,
        Scalar_h,
        Q_h,
        Mom_v,
        Theta_v,
        KE_v,
        Scalar_v,
        Q_v,
        Turb_lengthscale,
        NumDiffs
    };
}

a(i,j,k,EddyDiff::KE_v)
```

WeiqunZhang:
> > I believe the final issue to address is the default indices used by `ArrayND::operator()(IntVect<Dim>)` when `Dim < N`. In my opinion, the default value should be `begin` rather than zeros.
> 
> @AlexanderSinn agrees that missing indices should default to `begin`. Any thoughts @WeiqunZhang ?

Yes, I also agree.

Src/Base/AMReX_Array4.H
WeiqunZhang: We could try to make sure enum is only allowed for the last index.

ankithadas:
I am not quite sure why documentation build failed. Any insights ?

WeiqunZhang:
```
2026-01-07T00:31:56.6866927Z Generating XML output for class amrex_paralleldescriptor_/home/runner/work/amrex/amrex/Src/Base/AMReX_Array4.H:238: warning: Found unknown command '\TODO'
2026-01-07T00:31:56.6867503Z /home/runner/work/amrex/amrex/Src/Base/AMReX_Array4.H:399: warning: argument 'i...' of command @param is not found in the argument list of amrex::ArrayND< T, N >::operator()(idx... i) const
2026-01-07T00:31:56.6867798Z /home/runner/work/amrex/amrex/Src/Base/AMReX_Array4.H:405: warning: Found unknown command '\TODO'
2026-01-07T00:31:56.6868451Z /home/runner/work/amrex/amrex/Src/Base/AMReX_Array4.H:399: warning: The following parameter of amrex::ArrayND::operator()(idx... i) const is not documented:
2026-01-07T00:31:56.6868580Z   parameter 'i'
2026-01-07T00:31:56.6869111Z /home/runner/work/amrex/amrex/Src/Base/AMReX_Array4.H:476: warning: argument 'i...' of command @param is not found in the argument list of amrex::ArrayND< T, N >::ptr(idx... i) const
2026-01-07T00:31:56.6869602Z /home/runner/work/amrex/amrex/Src/Base/AMReX_Array4.H:476: warning: The following parameter of amrex::ArrayND::ptr(idx... i) const is not documented:
2026-01-07T00:31:56.6869704Z   parameter 'i'
2026-01-07T00:31:56.6869812Z module::amrex_pd_bcast
```

Src/Base/AMReX_Array4.H
```diff
@@ -56,212 +61,613 @@ namespace amrex {
...
+        /**
+         * \brief IntVectND<N> constructor.
+         * \param a_p Pointer to data.
+         * \param a_smallend Start index (inclusive).
+         * \param a_bigend End index (exclusive).
+         *
+         * This constructor specifies begin and end indices using IntVectND<N> for all N dimensions.
+         */
+        AMREX_GPU_HOST_DEVICE
+        constexpr ArrayND (T* a_p, IntVectND<N> const& a_smallend, IntVectND<N> const& a_bigend) noexcept
```
AlexanderSinn: This has the wrong name. Usually in amrex, bigend refers to an inclusive high index. Just begin and end would be better names for this.

AlexanderSinn: Or maybe it is better if this actually were an inclusive bigend

ankithadas:
@yut23 I have completed a preliminary implementation of updated GDB scripts for `ArrayND` (https://github.com/ankithadas/amrex/tree/ArrayND-GDB-Tools). This seems to work for `ArrayND::operator()(i...)`, but I'm unsure how to access template parameters for member functions. Feel free to use this as a starting point for your PR

AlexanderSinn:
Hmm, `amrex::ArrayND<T, 3>` looks to make GPU kernels 3-5% slower compared to the 3D array that I previously used. Also, one place using Array4 directly is slower with this PR. Looking at some examples in Compiler Explorer suggests that subtracting begin from the ncomp variable prevents some simplifications when having multiple array accesses.

ankithadas:
Could you share details of your 3D array implementation? A simple way to make the computational cost the same is by adding specific overload for `ArrayND<T,4>` that removes the `begin[3] - end[3]` for `ncomp`. A general approach would be adding another class template parameter asking the user if the last dimension is `ncomp`. In that case, we could simplify the offset calculations.

AlexanderSinn:
It is the Array2 and (mainly) Array3 that I replaced with ArrayND here https://github.com/Hi-PACE/hipace/pull/1336/files#diff-32376acddbd6018a475c8f547ae49c6087080c5470cc4cf42718060b2295c7b9
After doing more benchmarking, it now seems that most of the performance difference comes from the general way I calculate the offset and not just from begin[3]. I use a 64-bit offset variable that combines all the begins. However, I think this approach cannot be used in amrex as it would increase the size of Array4 (unless begin and end are removed in non-debug builds like in hipace, but these variables are sometimes used by other code) and because it could overflow the 64 bit index when using it with many levels of mesh refinement at the edge of a big 3D domain.

AlexanderSinn:
As for adding another template parameter, I would do it by reinterpreting Array4 as a 3, 1 Array. The arrays in hipace for example, would be 2, 1 and 2, 0 in that notation. Not sure if this ends up being cleaner or more complicated in the end.

WeiqunZhang:
Let's keep it simple. Having two size template parameters seems confusing.

WeiqunZhang:
I have run some tests on Perlmutter and Frontier, and have not noticed any performance difference between this branch and the development. So that's good.

AlexanderSinn:
Did you test ArrayND with or without the "Added last_dim_componet template paramter" commit?

WeiqunZhang:
I was talking about the latest version with the new bool template parameter. I also tried the version before the change. Most of the tests did not show much difference, but one test was 5-10% slower with the old version.

ankithadas:
Any idea why the Intel SYCL tests fail when using standard `for` loops instead of `constexpr_for`?

WeiqunZhang:
I think it was just some large kernels were hitting the hardware and/or software limit of Intel GPUs.

WeiqunZhang:
Looks like we will need to add `-Wno-maybe-uninitialized` to that failed gcc test.

ankithadas:
Not sure why that test fails for some commits and not others. Anyway, I have updated the test

Src/Base/AMReX_Array4.H
```diff
@@ -56,212 +61,634 @@ namespace amrex {
...
-        constexpr Array4 (Array4<std::remove_const_t<T>> const& rhs) noexcept
-            : p(rhs.p),
-              jstride(rhs.jstride),
-              kstride(rhs.kstride),
-              nstride(rhs.nstride),
-              begin(rhs.begin),
-              end(rhs.end),
-              ncomp(rhs.ncomp)
-            {}
+        constexpr ArrayND (ArrayND<std::remove_const_t<T>, N> const& rhs) noexcept
```
WeiqunZhang: I think it will be safer if we make the type of rhs be `ArrayND<std::remove_const_t<T>, N, last_dim_component> const&`.

Src/Base/AMReX_Array4.H
```diff
@@ -56,212 +61,634 @@ namespace amrex {
...
+
 
+        /**
+         * \brief Check if the ArrayND is properly initialized.
+         * \return True if initialized properly.
+         */
         AMREX_GPU_HOST_DEVICE
-        explicit operator bool() const noexcept { return p != nullptr; }
+        explicit operator bool() const noexcept {
+            return p != nullptr && end.allGT(begin);
```
WeiqunZhang: I understand it's always a tradeoff. But is the second part of the test necessary in most cases if we trust the users? Currently this function is often used like this
```
if (a && a(i,j,k) > 0) { ... }
```
It offers a cheap test. Maybe we could make `bool()` simple and add a new function called `ok` that checks the pointer and the bounds.

Src/Base/AMReX_Array4.H
```diff
@@ -56,212 +61,634 @@ namespace amrex {
...
+        /**
+         * \brief Access element by IntVectND<M>.
+         * \param iv IntVectND<M> (1 <= M <= N).
+         * \return Reference to the element.
+         * When M < N, the missing indices are set to begin values.
+         */
+        template <int M, class U=T, std::enable_if_t<!std::is_void_v<U>,int> = 0>
         [[nodiscard]] AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
-        U& operator() (int i, int j, int k, int n) const noexcept {
+        U& operator() (IntVectND<M> const& iv) const noexcept {
```
WeiqunZhang, line 436: This comment applies to all versions of `opearator()` and the constructors. Since we have the `last_dim_component` template parameter now, should we be more strict on the value of `M`? For example, should we only allow either `M == N` or `M == N-1 && last_dim_component`.

Src/Base/AMReX_Array4.H
```diff
@@ -56,212 +61,634 @@ namespace amrex {
...
 
-        template <class U=T, std::enable_if_t<!std::is_void_v<U>,int> = 0>
+        /**
+         * \brief Get number of components (only when N=4).
+         * \return Number of components.
+         *
+         * For N!=4 case, the user is expected to keep track if they are using the last dimension
+         * as components.
+         */
+        template <int M=N, std::enable_if_t<M==4 || last_dim_component,int> = 0>
```
WeiqunZhang: Do we really need this function if last_dim_component is false?

WeiqunZhang:
For discussion. Should we make last_dim_component = false by default and make it true for Array4?

Src/Base/AMReX_TypeTraits.H
```diff
@@ -275,6 +269,30 @@ namespace amrex
...
+    template <typename From, typename To>
+    inline constexpr bool IsNarrowingConversion_v = IsNarrowingConversion<From, To>::value;
+
+    template <typename From, typename To>
+    inline constexpr bool IsNonNarrowingConversion_v = !IsNarrowingConversion<From, To>::value;
+
+    //! \brief Test if type T is a valid index type. Valid index type is integral and can be safely
+    //!         converted to int without narrowing.
+    template <typename T>
+    struct IsValidIndexType {
```
WeiqunZhang: Since this is not in Array4 or namespace detail, is it a good idea to have `int` hardwired?

ankithadas:
> For discussion. Should we make last_dim_component = false by default and make it true for Array4?

I agree. Would an `enum` be better ? Something like this 

```c++
enum class DimSpec: int { Spatial = 0, Component = 1 };
```

WeiqunZhang:
`ArrayND<3,DimSpec::Component>` looks weird to me.

ankithadas:
I have added explicit overloads for `ArrayND::operator()` and `ArrayND::ptr()` covering the following three argument patterns:
1. `IntVect<N>`  
2. `IntVect<M>` where `1 <= M <= N-1` 
3. `IntVect<M>, int` where `1 <= M <= N-1`  and `last_dim_component=true`

These overloads are not required but were added to provide better error messages. Without them, the compiler errors typically report a failure to find a matching `get_offset` with the given signature, which although correct, does not clearly indicate that the provided arguments fail to meet the required constraint.
I have also added overloads for `Dim3` for `N=3` cases as well. 

Lastly, one issue I noticed is that CTAD cases that worked previously would not work in the new version. The following construction now fails
```c++
Array4 (Real *, IntVect(), IntVect(), int)
```

WeiqunZhang:
> Lastly, one issue I noticed is that CTAD cases that worked previously would not work in the new version. The following construction now fails
> 
> ```c++
> Array4 (Real *, IntVect(), IntVect(), int)
> ```
Maybe you were using C++20 previously? I get this with C++17.
```
error: alias template deduction only available with ‘-std=c++20’ or ‘-std=gnu++20’
```
In my test, using C++20 works and `ArrayND` with C++17 works as well.

Src/Base/AMReX_Array4.H
WeiqunZhang: My question still remains. Is it worth to allow code like this?
```
ArrayND<Real,4> a(p, IntVectND<2>(0), IntVectND<2>(4));
a(IntVectND<1>(0)) = 4;
```

ankithadas:
Sorry, my earlier test case was incorrect, since the `IntVect` overloads were only introduced in this branch. 
The following code fails to compile in the `ArrayND` branch with C++17, but works on the dev branch. Compiling with C++20 fixes the issue. 
```cpp
Real* nullptr_real = nullptr;
Dim3 begin = IntVect(0).dim3();
Dim3 end = IntVect(ncells).dim3();
Array4 t (nullptr_real, begin, end, 1);
```

Src/Base/AMReX_Array4.H
ankithadas: Looks like I missed this case. Sorry about that. I don’t specifically need this feature in my code, but I can see it being useful for applications with a mix of periodic and non-periodic directions, where this overload could be used to set those components.

That said, if used improperly, this feature could also lead to subtle or unexpected bugs.

WeiqunZhang: Agreed. My feeling is we should make it hard for the users to write incorrect code, even if it means they need to explicitly expand or shrink IntVectND in certain use cases. (We do have helper functions for that.)

ankithadas: Sounds good.

ankithadas:
For discussion. Should we allow this construction now that `last_dim_component` is a template parameter 
```cpp 
ArrayND<Real, 5> a(nullptr, IntVectND<1>(0), IntVectND<1>(ncells));
ArrayND<Real, 5> b(nullptr, BoxND(IntVectND<1>(0), IntVectND<1>(ncells)));
```
I find this rather confusing tbh.

WeiqunZhang:
Agreed. It's very confusing. We should only allow `M==N` and `M+1==N && last_dim_component`.

WeiqunZhang:
This now compiles.
```
constexpr Array4<int> a{};
static_assert(a.size() == 0);
```

Src/Base/AMReX_Array4.H
```diff
@@ -56,212 +61,630 @@ namespace amrex {
...
-            return this->p;
+        constexpr bool contains (IntVectND<M> const& iv) const noexcept {
+            bool inside = true;
+            constexpr int idx = (last_dim_component && M == N) ? N - 1 : M;
+            constexpr_for<0, idx>([&](int d) {
+                inside = inside && (iv[d] >= begin[d]) && (iv[d] < end[d]);
+            });
+            if constexpr (last_dim_component && M == N) {
+                inside = inside && (iv[N-1] >= 0) && (iv[N-1] < end[N-1]);
+            }
```
WeiqunZhang: Could this be simplified to
```suggestion
            constexpr_for<0, M>([&](int d) {
                inside = inside && (iv[d] >= begin[d]) && (iv[d] < end[d]);
            });
```

Src/Base/AMReX_Array4.H
```diff
@@ -270,46 +693,110 @@ namespace amrex {
...
+    template <typename T>
+    ArrayND (T*, Dim3 const&, Dim3 const&, int) -> ArrayND<T, 4, true>;
+
+    template<typename T>
+    using Array4 = ArrayND<T, 4, true>;
+
+    template <class Tto, class Tfrom, int N>
     [[nodiscard]] AMREX_GPU_HOST_DEVICE
-    Array4<Tto> ToArray4 (Array4<Tfrom> const& a_in) noexcept
+    ArrayND<Tto,N> ToArray4 (ArrayND<Tfrom,N> const& a_in) noexcept
```
WeiqunZhang: I don't think anyone is using this function. We could remove it or we limit it to Array4 only because of its name.

Src/Base/AMReX_Array4.H
```diff
@@ -270,46 +693,110 @@ namespace amrex {
...
     }
 
     template <class T>
     [[nodiscard]] AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
     Dim3 lbound (Array4<T> const& a) noexcept
+    {
+        return Dim3{a.begin[0],a.begin[1],a.begin[2]};
+    }
+
+    template <class T, int N, std::enable_if_t<(N != 4), int> = 0>
```
WeiqunZhang: I think we should remove this version of lbound and ubound because of potential confusion.

Src/Base/AMReX_Array4.H
```diff
@@ -318,19 +805,41 @@ namespace amrex {
...
-        return Dim3{a.end.x-a.begin.x,a.end.y-a.begin.y,a.end.z-a.begin.z};
+        return Dim3{a.end[0]-a.begin[0],a.end[1]-a.begin[1],a.end[2]-a.begin[2]};
     }
 
-    template <typename T>
-    std::ostream& operator<< (std::ostream& os, const Array4<T>& a) {
-        os << "((" << lbound(a) << ',' << ubound(a) << ")," << a.ncomp << ')';
+    template <class T, int N, std::enable_if_t<(N != 4), int> = 0>
+    [[nodiscard]] AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+    IntVectND<N> length (ArrayND<T,N> const& a) noexcept
```
WeiqunZhang: Remove this version of length as well to avoid confusion.

Src/Base/AMReX_Array4.H
```diff
@@ -318,19 +805,41 @@ namespace amrex {
...
+        IntVectND<N> len;
+        for (int d = 0; d < N; ++d) {
+            len[d] = a.end[d] - a.begin[d];
+        }
+        return len;
+    }
+
+    template <typename T, int N>
+    std::ostream& operator<< (std::ostream& os, const ArrayND<T,N>& a) {
+        os << "((" << lbound(a) << ',' << ubound(a) << "))";
```
WeiqunZhang:
```suggestion
    template <typename T, int N, bool C>
    std::ostream& operator<< (std::ostream& os, const ArrayND<T,N,C>& a) {
        os << "(" << a.begin << ',' << a.end-1 << ")";
```

Src/Base/AMReX_Array4.H
ankithadas: Would it be better to rename them to ubound_iv ?

WeiqunZhang: I think it's better to not having it. We added the lbound and ubound functions for Array4 because we needed them, not because we anticipated that we would need them. So maybe in the future it will become clear we need a helper function like that for non-Array4 ArrayND.

WeiqunZhang (approved):

ankithadas:
Thank you for patiently reviewing the PR so many times.

WeiqunZhang:
Thank you for the PR! You did the work! If there are still bugs/defects there, it's all my fault for not catching it. :)

zingale:
This change causes Castro when built with `DEBUG=TRUE` to run 3x slower.  Is there a way to get the performance back in debug mode?  Things only seem slightly slower when optimized.

WeiqunZhang:
Yes we should be able to get the performance back.

WeiqunZhang:
I have a version that is only 10% slower than the original version in debug mode.

WeiqunZhang:
See #4900

Issue #4867, opened 2025-12-25
Issues when building with on windows with Visual studio and CUDA, latest AMReX release 2025


horstsoft0815:
When trying to build on Windows (after generating with CMake) within Visual Studio a number of compile errors are obtained:

1) "For this host platform/dialect, an extended lambda cannot be defined inside the 'if' or 'else' block of a constexpr if statement	amrex_3d	D:\mfem_testing\amrex_clean\amrex-25.12\Src\Base\AMReX_FBI.H	1261"

I was able to fix this issue by surrounding the "if constexpr" in line 1259 of AMReX_FBI.H by sth like

#ifdef _WIN32
if (...)
#else
if constexpr(...)
#endif

2) "no instance of overloaded function "atomicAdd" matches the argument list	amrex_3d	D:\mfem_testing\amrex_clean\amrex-25.12\Src\Base\AMReX_GpuAtomic.H	143"

The problem here seems to be that the MSVSC compiler tries to compile also unused functions for the template "class BaseFab" 
(AMRex_BaseFab.H) including the atomicAdd member functions. This becomes a problem when the following instantiations are created:
i) class TagBox final : public BaseFab<char> in AMReX_TagBox.H
ii) BaseFab<GpuArray<Real,2>> in main.cpp of Tests/CommType
since for these template arguments the CUDA compiler can obviously find no specialization of atomicAdd (as mitigated in AMRex_GpuAtomic.H:143).

This issue can be solved by applying SFINAE to all the atomicAdd variants of the "class BaseFab"  via sth like:

template<class T>
class BaseFab {
....
template <RunOn run_on AMREX_DEFAULT_RUNON, typename = std::enable_if_t<std::is_arithmetic_v<T>>>
		    BaseFab<T>& atomicAdd (const BaseFab<T>& x) noexcept;
...
};

and 

template <class T>
template <RunOn run_on, typename>
BaseFab<T>&
BaseFab<T>::atomicAdd (const BaseFab<T>& x) noexcept {....}

Build configuration:

Windows version: 11 23H2 (10.0.22631 Build 22631.6060)
CUDA Toolkit: 12.8
MSVSC compiler version is 19.40.33818
Visual Studio version is 17.10.10
CMake (GUI): 3.31.10

WeiqunZhang:
@horstsoft0815 Thanks for reporting the issues! Could you please try #4869?

ax3l:
[...]

PR #4869, opened 2025-12-26
CI: Enable MPI in CUDA test on Windows

WeiqunZhang:


WeiqunZhang:
This is able to reproduce the first issue in #4867.

WeiqunZhang:
The first issue in #4867 has been fixed. The CI test did not have the second issue likely because the fix in FBI.H also worked around it. Nevertheless, we can implement a more robust fix for BaseFab::atomicAdd.

asalmgren (approved):

PR #4872, opened 2025-12-29
SpMatrix: Sort CSR by column index within each row

WeiqunZhang:
This will make matrix-matrix multiplication easier.


WeiqunZhang:
This will be folded into #4870.

PR #4880, opened 2026-01-07
SmallMatrix `operator*=` for MM

ax3l:
## Summary

Add a short-hand to multiply `SmallMatrix` with another value and assign the result.

## Additional background

Often useful in ImpactX, e.g., latest is https://github.com/BLAST-ImpactX/impactx/pull/1256

## Checklist

The proposed changes:


WeiqunZhang:
This only makes sense for square matrix, because the number of rows and cols are part of the type. So we should make that clear with enable_if when defining the function. We might also want to enforce that they have the same order and starting index.

ax3l:
Ah you are totally right. Will improve accordingly.

ax3l:
Don't need it after all.

PR #4884, opened 2026-01-08
Added define method to FabArray that makes alias of other FabArray

jjschoser:
## Summary
Created a new define method for FabArray (and its subclasses iMultiFab and MultiFab), which can be used to make an alias of an existing FabArray. This functionality was already available through a constructor, but only with this change has it become possible to define a FabArray as an alias after it has already been declared.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_FabArray.H
+++ b/Src/Base/AMReX_FabArray.H
@@ -440,6 +440,8 @@ public:
                  const FabFactory<FAB>&     factory = DefaultFabFactory<FAB>());
 #endif
 
+    void define (const FabArray<FAB>& rhs, MakeType maketype, int scomp, int ncomp);
+
     const FabFactory<FAB>& Factory () const noexcept { return *m_factory; }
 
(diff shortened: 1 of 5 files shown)


WeiqunZhang:
It's actually possible with constructors. Just do `existing_fa =
FabArray<Fab>(...)`.

WeiqunZhang:
It works because FabArray and MultiFab are move-assignable even though not
copy-assignable.

jjschoser:
I didn't think of that, thank you Weiqun!

WeiqunZhang:
The question now is whether we should add a new `define` function. It seems to me that using `=` is clearer to the user and to people who read the code, because `=` is a standard opearator, whereas `define` is not.

jjschoser:
I think it could still be worthwhile for completeness, given that the other constructors have an equivalent define function. But I understand your point!

WeiqunZhang:
After thinking about this, I am going to close it.

PR #4890, opened 2026-01-11
Remove uses of Box range version of amrex::launch

AlexanderSinn:
## Summary

There are two very different types of functions called `amrex::launch`. One gives raw access to a GPU block to be used with shared memory. The other type iterates over threads more like ParallelFor but gives a Box to the functor, not an IntVect. For GPU the box will only have one element, for CPU this is the same box that was passed into launch. This was used to quickly port amrex CPU code to GPU. However, I find this quite confusing.

This PR replaces the uses of the second type of launch inside amrex with ParallelFor and removes the section in the documentation that covered it. This avoids having to write the 4D loop inside every kernel.

TODO (in a future PR?): replace ...LAMBDA_RANGE macros used in LinearSolvers.

## Checklist

The proposed changes:


ankithadas:
There are more instances of such device launches in the linear solvers; there are several `AMREX_LAUNCH_HOST_DEVICE_LAMBDA` calls that do the same thing. I don't think we are ready to remove documentation regarding this yet.

WeiqunZhang (approved):

PR #4891, opened 2026-01-13
Remove identity Lambdas in amrex::launch

AlexanderSinn:
## Summary

I think these are unnecessary. Removing them shortens the name of the generated kernels a little bit.

Before:
```
void amrex::launch_global<(int)1024, void amrex::launch<(int)1024, void hpmg::<unnamed>::gsrb_shared<
(int)1, (bool)0,(bool)0, (bool)0>(const amrex::BoxND<(int)3> &, const amrex::Array4<double> &,
const amrex::Array4<const double> &, const amrex::Array4<const double> &, const amrex::Array4<double> &,
const amrex::Array4<const double> &, double, double)::
[lambda() (instance 1)]>(int, CUstream_st *, const T2 &)::[lambda() (instance 1)]>(T2)
```
After:
```
void amrex::launch_global<(int)1024, void hpmg::<unnamed>::gsrb_shared<
(int)1, (bool)0, (bool)0, (bool)0>(const amrex::BoxND<(int)3> &, const amrex::Array4<double> &,
const amrex::Array4<const double> &, const amrex::Array4<const double> &, const amrex::Array4<double> &,
const amrex::Array4<const double> &, double, double)::
[lambda() (instance 1)]>(T2)
```

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_GpuLaunchFunctsG.H
+++ b/Src/Base/AMReX_GpuLaunchFunctsG.H
@@ -154,7 +154,7 @@ void single_task (gpuStream_t stream, L const& f) noexcept
     auto& q = *(stream.queue);
     try {
         q.submit([&] (sycl::handler& h) {
-            h.single_task([=] () { f(); });
+            h.single_task(f);
         });
     } catch (sycl::exception const& ex) {
         amrex::Abort(std::string("single_task: ")+ex.what()+"!!!!!");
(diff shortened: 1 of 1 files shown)


ax3l (approved):
Nice, thank you! :)

WeiqunZhang (approved):

PR #4895, opened 2026-01-14
MLMG: Make the single-stream no-sync GPU mode safe and document it

WeiqunZhang:
`MLMG::setNoGpuSync(true)` runs `solve()` in a `Gpu::SingleStreamRegion` without
the implicit stream synchronizations. This PR makes that mode safe. The default
stays off because it only helps small problems with few boxes and slows down
solvers with many small per-box kernels (e.g., nodal and EB with many boxes).

* `FabArray::clear_arrays()` and `detail::destroy_par_for_boxes()` free the
  pinned mirrors and their device copies with `Gpu::freeAsync`, because the
  host-to-device copy is asynchronous. `Arena::Finalize` flushes the deferred
  frees, and `Device::freeAsync` frees directly when there are no streams.
* Temporary `MultiFab`s created inside `solve()` are allocated from
  `The_Async_Arena()` (smoothers, restriction/interpolation, bottom CG vectors,
  nodal reflux, `makeNLinOp`, HYPRE/PETSc, `average_down*`, `OverrideSync`).
  `MLLinOp::make`, `makeCoarseMG` and `makeCoarseAmr` gained `MFInfo` overloads.
* HYPRE and PETSc `loadVectors` synchronize before passing device data to the
  library.
* `solve()` uses `Gpu::SyncAtExitOnly` and `Gpu::SingleStreamRegion` as RAII
  objects, so the state is restored on exceptions and streams are synchronized
  on exit.
* Docs describe `setNoGpuSync` and the rules for temporaries in a no-sync
  region. `Tests/GPU/ExternalStream` gained an MLMG lifetime test that runs in
  both modes.


WeiqunZhang:
For temporary MultiFabs allocated in solve() and functions called by solve(), we could use The_Async_Arena(). We also need to make sure that MultiFab utility functions are safe to be used inside no-sync region.

WeiqunZhang:
# Review: PR #4895 "MLMG: Change default of do_no_sync_gpu to true"

Reviewed 2026-09-01 at PR head 517a4101 (branch `mlmg_nosync`; the local merge of
`development` on top only adds CHANGES.md). CI is green on the head commit.

## Verdict

No correctness problems found. The lifetime story is sound: every function-local
temporary reachable from `MLMG::solve()` now lives in `The_Async_Arena()` (an `SArena`
that allocates from `The_Arena` and defers the free to the allocating stream's next
sync), the `FabArray` Array4 mirrors and `ParForInfo` box tables are freed with
`Gpu::freeAsync`, and the HYPRE/PETSc `loadVectors` paths got the missing stream sync
before handing device pointers to the library.

The concern is the *default* flip. On this machine the new mode is a clear win only for
cell-centered, non-EB solves with few boxes. Nodal and EB solves get 1.3x to 2.3x
slower, and cell-centered solves with many 16^3 boxes get 1.4x slower. See the numbers
below. The cause is `SingleStreamRegion` serializing per-box `MFIter` kernels; it is not
the deferred-free cap and not arena growth (both tested).

Recommendation: do not flip the default unconditionally. Either keep it off (WarpX already
opts in via `setNoGpuSync(true)`), or make it operator-dependent (e.g. a virtual on
`MLLinOp` that is true for `MLABecLaplacian`/`MLPoisson`/`MLALaplacian` without EB and
false for nodal, EB and tensor operators until their per-box loops are fused).

## Measurements (RTX 5070, CUDA 13.3, 3D, single rank, `MLMG: Timers: Solve`)

The PR head was built in a scratch worktree; the only local change was an input knob
`mlmg.no_sync` in the three test drivers so both modes run the same binary.
`amrex.the_arena_init_size=6e9` for the multi-level cases (2e9 for the single-box ones).

| Case | no_sync=1 (new default) | no_sync=0 | ratio |
|---|---|---|---|
| ABecLaplacian_C, 1 box 32^3 | 0.0040 s | 0.0071 s | 0.56 |
| ABecLaplacian_C, 1 box 64^3 | 0.0074 s | 0.0115 s | 0.64 |
| ABecLaplacian_C, 2 lev 128^3 mgs=64 | 0.078 s | 0.083 s | 0.94 |
| ABecLaplacian_C, 2 lev 128^3 mgs=16 | 0.41 s | 0.29 s | 1.42 |
| MLPoisson (prob_type=2), 2 lev mgs=16 | 0.45 s | 0.34 s | 1.31 |
| NodalPoisson, 1 box 32^3 | 0.0108 s | 0.0129 s | 0.84 |
| NodalPoisson, 1 box 64^3 | 0.023 s | 0.025 s | 0.90 |
| NodalPoisson, 2 lev 64^3 mgs=32 | 0.097 s | 0.072 s | 1.35 |
| NodalPoisson, 2 lev 64^3 mgs=16 | 0.35 s | 0.15 s | 2.30 |
| CellEB2 flower, 2 lev 128^3 mgs=64 | 4.11 s | 1.79 s | 2.30 |

Controls: raising the 100-entry `StreamManager::free_async` cap to 1e8 changed no number;
neither did the arena pre-allocation. nsys on the 64^3 nodal case shows identical kernel
counts (1365) in both modes and 96 vs 599 `cudaStreamSynchronize` calls. The "Bottom"
timer looks 3-8x worse in no-sync mode only because the first reduction in the bottom
solve absorbs the smoother work queued ahead of it; it is a timer artifact.

Residuals are bit-identical between modes for the cell-centered solvers. The nodal
solver differs in the last digits between runs in both modes (pre-existing atomics).

## Static review notes

- `solve()`: `no_sync_region` is declared before `single_stream_region`, so the
  single-stream region is popped first and `SyncAtExitOnly` then calls
  `streamSynchronizeAll()`. Nested solves (NSolve) inherit the region and do not sync
  at exit, as intended.
- `SArena::free` goes through `Gpu::freeAsync` on the FAB's allocation stream, so a
  `The_Async_Arena` MultiFab that is created inside an `ExternalGpuStreamRegion` and
  destroyed after it would abort in `streamOrderedFreeAsync`. Every site changed in
  the PR is function-local, so this is only reachable through
  `OverrideSync_nowait`/`_finish` straddling a stream-region boundary. Worth a comment.
- With more than 100 local FABs, destroying one Async-Arena temporary forces a stream
  sync through the wait-list cap. Measured to be harmless here, but it silently
  reintroduces syncs for many-box ranks.
- Remaining default-arena temporaries in MLMG (`defineBC`, `setLevelBC`,
  `MLCurlCurl::setAlpha`, `compResidual`, `compRHS`, `compSyncResidualCoarse`,
  `apply`, `getFluxes`) are all outside `solve()`, so they are fine.
- HYPRE/PETSc `getSolution`: the Async-Arena `tmp` is safe because the library writes
  complete before `hypreSynchronize()`/`Gpu::synchronize()` and the copy back runs on
  the AMReX stream that also owns the deferred free.
- `Arena::Finalize` needs the new `clearFreeAsyncBuffer()` call because
  `FabArrayBase::Finalize` (run from the finalize-function stack) frees `ParForInfo`
  tables after the initial `streamSynchronizeAll()`.

## Nits

- Docs (GPU.rst): "undefined behavior on some backends such as SYCL" is technically
  right (`cudaFree`/`hipFree` synchronize the device before releasing; `sycl::free` does
  not wait), but it should not single out SYCL. Suggested: "Whether this is safe depends
  on the backend (CUDA and HIP force a device-wide synchronization in the release; SYCL
  does not), so it must not be relied upon."
- `Tests/GPU/ExternalStream/GNUmakefile` was not updated, so `test_mlmg_lifetime`
  only builds with CMake.
- CodeQL "unused local variable" on `[[maybe_unused]] Gpu::SyncAtExitOnly no_sync{};`
  in `setBCoeffs` is a false positive; same pattern as `ParticleContainerI.H`.

## Commands run

```
cmake -S . -B build-cuda -DAMReX_GPU_BACKEND=CUDA -DCMAKE_CUDA_ARCHITECTURES=120 \
  -DAMReX_MPI=OFF -DAMReX_SPACEDIM=3 -DAMReX_EB=ON -DAMReX_ENABLE_TESTS=ON \
  -DAMReX_TEST_TYPE=All -DCMAKE_BUILD_TYPE=Release -DAMReX_FORTRAN=OFF
cmake --build build-cuda -j20 --target Test_GPU_ExternalStream_3d \
  Test_LinearSolvers_ABecLaplacian_C_3d Test_LinearSolvers_NodalPoisson_3d
cd Tests/LinearSolvers/CellEB2 && make -j8 USE_CUDA=TRUE CUDA_ARCH=120 USE_MPI=FALSE DIM=3
```

`Test_GPU_ExternalStream_3d` (with `AMREX_TEST_LINEAR_SOLVERS`) passed 3/3 runs.

WeiqunZhang:
Maybe the default should be off, and the users can opt in. @AlexanderSinn 

@RemiLehe I will make sync vs. nosync part of auto-tuner.

WeiqunZhang:
Needed by https://github.com/BLAST-WarpX/warpx/pull/7213

atmyers (approved):

PR #4898, opened 2026-01-15
Fix a bug introduced in #4897

WeiqunZhang:
The issue was there was an out of bound error in MFIter when local size is zero.


atmyers (approved):

WeiqunZhang:
Note that the mistake was to call numLocalTiles() in Intitialize(). That function can only be used inside the loop body. Furthermore, it is actually unnecessary because the size of index_map is sufficient.

PR #4900, opened 2026-01-16
ArrayND: Add specialization for Array4

WeiqunZhang:
This significantly improves the performance of Array4 in debug mode.

In my testing using Tests/LinearSolvers/ABecLaplacian_C on CPU, this new version is still slightly slower (10-15%) in debug mode than the pre-ArrayND version. But in release mode, it's actually 5% faster than the pre-ArrayND version.

### DIFF
--- a/Src/Base/AMReX_Array4.H
+++ b/Src/Base/AMReX_Array4.H
@@ -63,6 +63,9 @@ namespace amrex {
 
     namespace detail {
 
+        template <int N> struct Stride { Long a[N] = {}; };
+        template <> struct Stride<0> {};
+
         /**
          * \brief Helper struct to check if a type is a valid index type.
          *
@@ -215,7 +218,7 @@ namespace amrex {
         static constexpr bool IsArray4_v = (N==4 && last_dim_component);
 
         T* AMREX_RESTRICT p = nullptr;
-        std::array<Long, N-1> nstride{};
+        AMREX_NO_UNIQUE_ADDRESS detail::Stride<N-1> stride{};
         IntVectND<N> begin{1};
         IntVectND<N> end{0}; // end is hi+1
 
@@ -225,7 +228,7 @@ namespace amrex {
         template <class U=T, std::enable_if_t<std::is_const_v<U>,int> = 0>
         AMREX_GPU_HOST_DEVICE
         constexpr ArrayND (ArrayND<std::remove_const_t<T>, N, last_dim_component> const& rhs) noexcept
-            : p(rhs.p), nstride(rhs.nstride), begin(rhs.begin), end(rhs.end) {}
+            : p(rhs.p), stride(rhs.stride), begin(rhs.begin), end(rhs.end) {}
 
         /**
          * \brief Constructor using a BoxND.
@@ -271,7 +274,7 @@ namespace amrex {
                 constexpr_for<0, N-1>([&](int d) {
                     Long len = a_end[d] - a_begin[d];
                     current_stride *= len;
-                    nstride[d] = current_stride;
+                    stride.a[d] = current_stride;
                 });
             }
         }
@@ -294,7 +297,7 @@ namespace amrex {
             constexpr_for<0, N-1>([&](int d) {
                 Long len = end[d] - begin[d];
                 current_stride *= len;
-                nstride[d] = current_stride;
+                stride.a[d] = current_stride;
             });
         }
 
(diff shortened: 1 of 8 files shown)


zingale:
Thanks, I can confirm that this brings the performance gap back down to about 20% of where it was previously in debug mode.

WeiqunZhang:
@zingale What about release mode?

zingale:
It seems comparable to the performance from before Array4 now.

WeiqunZhang:
It seems that in debug mode std::array is noticeably slower than C array. We use std::array to store strides.

zingale:
I guess that std::array adds bounds checking

WeiqunZhang:
In my testing, I see no performance difference in debug mode now after replacing std::array. @zingale Could you give the latest version a try?

WeiqunZhang:
I also tweaked the index assert part, the latest version should now be faster than the pre-ArrayND version.

zingale:
indeed.  It is faster now in debug mode than previously

WeiqunZhang:
It seems that the best way to compute offset is
```
#if defined(AMREX_USE_GPU) || defined(AMREX_DEBUG)
            return p[(i-begin.vect[0]) +
                     (j-begin.vect[1]) * stride.a[0] +
                     (k-begin.vect[2]) * stride.a[1]];
#else
            Long idx1 = i + j*stride.a[0] + k*stride.a[1];
            Long idx0 = begin.vect[0]
                +       begin.vect[1] * stride.a[0]
                +       begin.vect[2] * stride.a[1];
            return p[idx1-idx0];
#endif
```

Thank @ankithadas for suggesting the second approach. For CPU code in release build, the compiler is able to hoist the invariant part of the index out of the loop.

Below are results from perlmutter runs.
```
| Version    | CPU Time | GPU Time |
|------------+----------+----------|
| ArrayND-1  |      184 |     2.22 |
| ArrayND-2  |      132 |     2.35 |
| Array4-1   |      147 |     2.23 |
| Array4-2   |      129 |     2.35 |
```
Here `-1` is the approach in the `#if` part and `-2` is the approach in the `#else`. `ArrayND` is the development branch and `Array4` is this PR branch.

zingale:
that's awesome!

WeiqunZhang:
That Intel PVC failure is unrelated to this PR. It's likely an issue of the machine.

WeiqunZhang:
Some more performance data I collected.
```
* Tests/LinearSolver/ABecLaplacian_C
  // mytest.writePlotfile commented out
  Run with `inputs n_cell=512`
  Solver time
| Version        | A100-80GB | MI250X |
|----------------+-----------+--------|
| pre-arraynd    |      2.23 |   2.94 |
| arraynd        |      2.23 |   2.98 |
| array4 (#4900) |      2.23 |   2.94 |
| base           |      2.26 |   2.94 |
```
where is base a branch that stores the pointer offset invariant in Array4.

BenWibking:
The Array4 changes broke our code:
```
/Users/runner/work/quokka/quokka/extern/turbulence/plugins/AMReX/TurbGenEx.h:158:17: error: alias template 'Array4' requires template arguments; argument deduction only allowed for class templates
                amrex::Array4 xPrecomp = xPrecompFab.array();
                              ^
/Users/runner/work/quokka/quokka/extern/amrex/Src/Base/AMReX_Array4.H:777:5: note: template is declared here
    using Array4 = ArrayND<T, 4, true>;
    ^
```
Source: https://github.com/chfeder/turbulence_generator/blob/master/plugins/AMReX/TurbGenEx.h

Will this PR fix the above code?

If we did something weird in the TurbGen code, let me know and we could fix it there.

WeiqunZhang:
I don't think this PR will fix it. But if you use C++20, it should work.

BenWibking:
> I don't think this PR will fix it. But if you use C++20, it should work.

We have C++20 enabled throughout Quokka now: https://github.com/quokka-astro/quokka/blob/535802bc79c024423194597008308c53f4745911/CMakeLists.txt#L6

Are you able to build the code or a MWE? It looks like all our CI builds that failed were on Mac or ARM64 Linux :(

ankithadas:
Unfortunately, Apple Clang does not support CTAD for alias templates. https://cppreference.com/w/cpp/compiler_support/20.html

You might have to use `auto` or specify template arguments.

WeiqunZhang:
Just making cosmetic changes.

asalmgren (approved):

PR #4902, opened 2026-01-17
ArrayND: Specialize operator()(idx...) and ptr(idx...) for N <= 3

WeiqunZhang:


WeiqunZhang:
@AlexanderSinn I think Array2 and Array3 in Hipace are similar to amrex::ArrayND<T,2,false>, amrex::ArrayND<T,3,true>, respectively. Is that right? I made some changes to operator(idx...) and ptr(idx...). I wonder if this has better performance than the previous version for you. Could you please test it?

AlexanderSinn:
Yes they are simular. However Array2/3 in hipace use a global offset that is computed in the constructor of the array instead of subtracting begin from each component.

Array2/3 from arraynd_23:
```
TinyProfiler total time across processes [min...avg...max]: 10.16 ... 10.16 ... 10.16
```
Array2/3 from AMReX dev:
```
TinyProfiler total time across processes [min...avg...max]: 10.16 ... 10.16 ... 10.16
```
Array2/3 from HiPACE++:
```
TinyProfiler total time across processes [min...avg...max]: 10.11 ... 10.11 ... 10.11
```

WeiqunZhang:
Thanks for the information! It does not seem necessary to specialize for N <= 3.

PR #4903, opened 2026-01-17
MakeNewGrids: param to set number of grid iterations

BenWibking:
## Summary
This adds the ability to set the number of grid iterations in `MakeNewGrids`. The default is 4, which was the previous hard-coded value.

## Additional background
For some problems, this is required for correct grid generation. This typically is only needed with deeply nested AMR hierarchies (8+ levels).
See https://github.com/BenWibking/amrex-gridgen-test for a reproducer of the original issue.

## Checklist

The proposed changes:


WeiqunZhang:
The approach in this PR may not be necessary. Could we simply change that hardwired 4 to a parameter that has a default value of 4. I think for location based tagging, it's sufficient to have the iteration number set to max_level. During each iteration a fine level can grow to its coarse level's domain. So with max_level iterations, it should be sufficient (again assuming tagging is location based and it does not change during the iteration).

BenWibking:
@WeiqunZhang Ok, I've changed the PR to just add a runtime parameter for the max number of iterations. We can set this equal to `amr.max_level` by default in our code.

WeiqunZhang (approved):

PR #4904, opened 2026-01-19
Updated `ArrayND` offset calculation 

ankithadas:
## Summary
Changes the order of calculation for `get_offset` in `ArrayND`. For CPU builds in Release mode, the compiler is able to hoist the invariant part of the offset calculation out of the nested loops. See #4900 for some performance comparison.

I have also removed the use of `constexpr_for` as it adds a slight performance penalty. 
## Checklist

The proposed changes:


WeiqunZhang:
Let's wait on #4900

ankithadas:
Any idea why the Intel SYCL builds are failing ? It seems to work when I use `constexpr_for`

PR #4906, opened 2026-01-19
Fix bug in multi-ghost FillBoundary for the Alamo code

WeiqunZhang:
This also adds a FillBoundary call for Alamo's linear solver that uses the multi-ghost mode.


WeiqunZhang:
This PR also contains commits from #4900, because the current development breaks Alamo. So we need to get #4900 merged first.

WeiqunZhang:
@bsrunnels has reported that this appears to fix the issue in Alamo. I have also removed #4900 from this PR. So this is now ready.

bsrunnels (approved):

asalmgren (approved):

PR #4922, opened 2026-01-23
[WIP] GPU specific kernels for `MLEBABecLap`

ankithadas:
## Checklist

The proposed changes:

### DIFF
--- a/Src/EB/AMReX_EBData.H
+++ b/Src/EB/AMReX_EBData.H
@@ -34,10 +34,10 @@ struct EBData
             return m_real_data[static_cast<int>(T)](i,j,k);
         }
     }
-    
+
     template <EBData_t T>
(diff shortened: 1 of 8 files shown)


ankithadas:
For discussion: This PR introduces fairly substantial changes to `MLEBABecLap`, and I would like feedback on whether it should be split into smaller PRs. The main changes are:

1. Refactored the linop kernels into a templated form for consistency with other linops.
2. Removed the `Box` based kernels and replaced them with explicit index-based access `(i, j, k, n)`.
3. Introduced a multi-array version of `EBData`, called `EBDataArrays`.
4. Added GPU-specific kernels to enable MF fusion.

WeiqunZhang:
Yes, it's a good idea to split this into smaller PRs so that it's easier for testing correctness and performance. As for making the linop kernels into templates, unless it's necessary, I don't think we should do that.

PR #4924, opened 2026-01-27
Generalize SIMD Single Source Design

ax3l:
## Add `ParallelForSIMD<T>`

This adds another template overload to `ParallelForSIMD`.

A typical user pattern for maximum controls so far is:
```C++
#ifdef AMREX_USE_SIMD
if constexpr (amrex::simd::is_vectorized<T>) {
    amrex::ParallelForSIMD<T::simd_width>(np, pushSingleParticle);
} else
#endif
{
    amrex::ParallelFor(np, pushSingleParticle);  // GPU & non-SIMD CPU
}
```

This simplifies it to:
```C++
amrex::ParallelForSIMD<T>(np, pushSingleParticle);
```
indicating there _might_ be a SIMD path if `T` (e.g., a functor) implements it.

One can still call `ParallelForSIMD` with an explicit SIMD width (int), as before.

## Generalized Particle Load/Store

A typical SIMD user pattern for particle SoA kernels was:
```C++
SIMDParticleReal<SIMD_WIDTH> part_x;
part_x.copy_from(&m_part_x[i], stdx::element_aligned);

el.compute(part_x);

#ifdef AMREX_USE_SIMD
if constexpr (is_nth_arg_non_const<&el::compute, n>)
    part_x.copy_to(&m_part_x[i], stdx::element_aligned);
#endif
```

This simplifies it to:
```C++
decltype(auto) x = load_1d(m_part_x, i);

el.compute(x);

store_1d<&el::compute, 0>(x, m_part_x, i);
```
and can now also be used for the GPU path, where for now the load is a transparent pointer forward/deref and the store is a no-OP.

## Combined

Using these patterns together, one can now write single-source SIMD-CPU/non-SIMD-CPU/GPU kernels, e.g., https://github.com/BLAST-ImpactX/impactx/pull/1279

Follow-up to #4520

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_GpuLaunchFunctsC.H
+++ b/Src/Base/AMReX_GpuLaunchFunctsC.H
@@ -220,6 +220,25 @@ void ParallelForSIMD (N n, L const& f) noexcept
     }
 }
 
+/** ParallelFor with a SIMD Width (derived from a type w/ fallback to non-SIMD)
+ *
+ * SIMD load/Write-back operations need to be performed before/after calling this.
+ *
+ * @tparam T a type that implements amrex::simd::Vectorized (or not)
+ * @tparam N index type (integer)
+ * @tparam L function/functor to call per SIMD set of elements
+ */
+template <typename T, typename N, typename L, typename M=std::enable_if_t<std::is_integral_v<N>> >
+AMREX_ATTRIBUTE_FLATTEN_FOR
+void ParallelForSIMD (N n, L && f) noexcept
+{
+    if constexpr (amrex::simd::is_vectorized<T>) {
+        amrex::ParallelForSIMD<T::simd_width>(n, std::forward<L>(f));
+    } else {
+        amrex::ParallelForSIMD<1, N>(n, std::forward<L>(f));
+    }
+}
+
 template <typename T, typename L, typename M=std::enable_if_t<std::is_integral_v<T>> >
 void ParallelFor (Gpu::KernelInfo const&, T n, L&& f) noexcept
 {
--- a/Src/Base/AMReX_GpuLaunchFunctsC.H
+++ b/Src/Base/AMReX_GpuLaunchFunctsC.H
@@ -4,24 +4,6 @@
 
 namespace amrex {
 
-/** Helper type to store/access the SIMD width in ParallelForSIMD lambdas
- *
- * Use instead of int as the running index i. Used to pass the
- * SIMD WIDTH as compile-time meta-data into a called function/method.
- *
- * @tparam WIDTH SIMD width in elements
- * @tparam N index type (integer)
- */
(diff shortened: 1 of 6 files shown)


WeiqunZhang:
I don't think `amrex::ParallelForSIMD<T>(np, pushSingleParticle);` compiles if AMREX_USE_GPU is true, because that function only exists in AMReX_GpuLaunchFunctsC.H, which is only included when AMREX_USE_GPU is not defined.

Src/Base/AMReX_GpuLaunchFunctsC.H
```diff
@@ -220,6 +220,25 @@ void ParallelForSIMD (N n, L const& f) noexcept
...
+ *
+ * SIMD load/Write-back operations need to be performed before/after calling this.
+ *
+ * @tparam T a type that implements amrex::simd::Vectorized (or not)
+ * @tparam N index type (integer)
+ * @tparam L function/functor to call per SIMD set of elements
+ */
+template <typename T, typename N, typename L, typename M=std::enable_if_t<std::is_integral_v<N>> >
+AMREX_ATTRIBUTE_FLATTEN_FOR
+void ParallelForSIMD (N n, L && f) noexcept
```
ax3l: Need to move this and the existing `ParallelForSIMD` into a place where both `AMReX_GpuLaunchFunctsC.H` and `AMReX_GpuLaunchFunctsG.H` can find it, if we want to use it as a GPU-enabled switch yard, too.

https://github.com/AMReX-Codes/amrex/pull/4924#issuecomment-3836715234

Src/Base/AMReX_GpuLaunchFunctsSIMD.H
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * @tparam T a type that implements amrex::simd::Vectorized (or not)
+ * @tparam N index type (integer)
+ * @tparam L function/functor to call per SIMD set of elements
+ */
+template <typename T, typename N, typename L, typename M=std::enable_if_t<std::is_integral_v<N>> >
+AMREX_ATTRIBUTE_FLATTEN_FOR
+void ParallelForSIMD (N n, L && f) noexcept
+{
+#ifdef AMREX_USE_SIMD
```
ax3l, line 69: One could maybe add another "and not GPU" here, just to be super sure if someone turns on SIMD and GPU at the same time to stay with the GPU path... Not sure if relevant.

Src/Base/AMReX_GpuLaunchFunctsC.H
ax3l: fixed

WeiqunZhang (approved):

ax3l:
Tests in #4938

PR #4925, opened 2026-01-28
Reducer: Reduction class with support for runtime number of values

WeiqunZhang:
This new class adds support for reduction of a runtime number of values in addition to the existing support for a compile time number of values. The values need to be the same type. However, the reduction types can be a mixing of max, min and sum.

### DIFF
--- a/Tests/Reducer/main.cpp
+++ b/Tests/Reducer/main.cpp
@@ -1,0 +1,238 @@
+#include <AMReX.H>
+#include <AMReX_MultiFabUtil.H>
+#include <AMReX_ParmParse.H>
+#include <AMReX_Reduce.H>
+
+using namespace amrex;
+
+
+int main (int argc, char* argv[])
+{
+    amrex::Initialize(argc, argv);
+
+    int n_cell = 64;
+    int max_grid_size = 32;
+    {
+        ParmParse pp;
+        pp.query("n_cell", n_cell);
+        pp.query("max_grid_size", max_grid_size);
+    }
+
+    {
+        BoxArray ba(Box(IntVect(0),IntVect(n_cell-1)));
+        ba.maxSize(max_grid_size);
+        MultiFab mf(ba, DistributionMapping{ba}, 2, 0);
+        FillRandom(mf, 0, 2);
+        mf.plus(Real(-0.2), 0, 2);
+
+        // No need to use MPI in testing local reduce. Hence true.
+        Vector<Real> benchmark1{mf.min(0,0,true), mf.max(0,0,true),
+            mf.sum(0,true), mf.norm1(0,0,true), mf.norminf(0,0,true)};
+
+        Vector<Real> benchmark2{mf.min(1,0,true), mf.max(1,0,true),
+            mf.sum(1,true), mf.norm1(1,0,true), mf.norminf(1,0,true)};
+
+        benchmark2[0] = std::min(benchmark2[0], benchmark1[0]);
+        benchmark2[1] = std::max(benchmark2[1], benchmark1[1]);
+        benchmark2[2] += benchmark1[2];
+        benchmark2[3] += benchmark1[3];
+        benchmark2[4] = std::max(benchmark2[4], benchmark1[4]);
+
+        // eval(Box, f(iop, i,j,k))
+        {
+            Reducer<Real> reducer(Vector<ReduceOpType>{
+                    ReduceOpType::min, ReduceOpType::max, ReduceOpType::sum,
+                    ReduceOpType::sum, ReduceOpType::max});
+#ifdef AMREX_USE_OMP
+#pragma omp parallel if (Gpu::notInLaunchRegion())
+#endif
+            for (MFIter mfi(mf); mfi.isValid(); ++mfi) {
+                Box const& box = mfi.validbox();
+                auto const& a = mf.array(mfi);
+                reducer.eval(box, [=] AMREX_GPU_DEVICE (int iop, int i, int j, int k)
+                { // 0 <= iop < 5
+                    if (iop >= 0 && iop <= 2) { // min, max & sum
+                        return a(i,j,k);
+                    } else { // 1-norm & inf-norm
+                        return std::abs(a(i,j,k));
+                    }
+                });
+            }
+            Vector<Real> result = reducer.getResults();
+
+            AMREX_ALWAYS_ASSERT(benchmark1[0] == result[0] &&
+                                benchmark1[1] == result[1] &&
+                                almostEqual(benchmark1[2], result[2], 100) &&
+                                almostEqual(benchmark1[3], result[3], 100) &&
+                                benchmark1[4] == result[4]);
+        }
+
+        // eval(Box, f(iop, IntVect))
+        {
+            Reducer<Real> reducer(Vector<ReduceOpType>{
+                    ReduceOpType::min, ReduceOpType::max, ReduceOpType::sum,
+                    ReduceOpType::sum, ReduceOpType::max});
+#ifdef AMREX_USE_OMP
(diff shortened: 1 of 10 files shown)


WeiqunZhang:
@ax3l This can be used to implement the ParticleReduce feature you have requested.

WeiqunZhang:
You can do something like
```C++
Vector<ReduceOpType> ops;
ops.push_back(ReduceOpType::min);
ops.push_back(ReduceOpType::sum);
Reducer<ParticleReal> reducer(ops);
for (ParIter ...) {
    Long np = ...;
    reducer.eval(np, [=] AMREX_GPU_DEVICE (int iop, Long i)
    {
        return ...; // particle i's value for iop'th reduction.
    });
}
Vector<ParticleReal> result = reducer.getResults();
```

Tests/Reducer/main.cpp
```diff
@@ -0,0 +1,238 @@
...
+                Box const& box = mfi.validbox();
+                auto const& a = mf.array(mfi);
+                reducer.eval(box, [=] AMREX_GPU_DEVICE (int iop, int i, int j, int k)
+                { // 0 <= iop < 5
+                    if (iop >= 0 && iop <= 2) { // min, max & sum
+                        return a(i,j,k);
+                    } else { // 1-norm & inf-norm
+                        return std::abs(a(i,j,k));
+                    }
+                });
```
AlexanderSinn, line 59: I could imagine this interface being quite awkward to use if, for example, up to 20-30 different quantities need to be reduced with some disabled at runtime. This would require a big switch statement with a lookup to map `iop` to quantity equations. Additionally, any shared data like `a(i,j,k)` will need to be read in from memory for each  `iop` separately instead of being reused as registers. I think this would negate most of the memory bandwidth benefit from combining all the reductions into a single kernel, leaving only to the advantage of reduced launch latency.

ax3l:
Thank you! :sparkles: 

For background, the application is to do a more fine-grained reduction selection in https://github.com/BLAST-ImpactX/impactx/issues/1102
https://github.com/BLAST-ImpactX/impactx/blob/26.01/src/diagnostics/ReducedBeamCharacteristics.cpp#L56-L218

Tests/Reducer/main.cpp
AlexanderSinn: I like this kind of style. I wonder how the performance compares, as it does more parallel updates but uses less memory bandwidth.
```C++
amrex::ParallelFor(amrex::Gpu::KernelInfo().setReduction(true), box,
    [=] AMREX_GPU_DEVICE (int i, int j, int k, amrex::Gpu::Handler const& handler) noexcept
    {
        amrex::Real * result_ptr = rptr;
        amrex::Real value = a(i, j, k);

        if (do_min_max_sum) {
            // min, max & sum
            amrex::Gpu::deviceReduceMin(result_ptr++, value, handler);
            amrex::Gpu::deviceReduceMax(result_ptr++, value, handler);
            amrex::Gpu::deviceReduceSum(result_ptr++, value, handler);
        }

        if (do_norm) {
            // 1-norm & inf-norm
            amrex::Gpu::deviceReduceSum(result_ptr++, std::abs(value), handler);
            amrex::Gpu::deviceReduceMax(result_ptr++, std::abs(value), handler);
        }
    });
```

WeiqunZhang: Let me do some testing.

WeiqunZhang: The interface is always going to awkward no matter what we use. If the number of values is big, the performance could be a real issue.

The deviceReduce* way could be very slow for OMP. For sum, it uses omp atomic. For min and max, it uses critical region.

WeiqunZhang:
Some results on perlmutter.

```
* Perlmutter, A100 80GB
  5 reduces, https://github.com/WeiqunZhang/amrex-devtests/tree/main/reducer
  Run w/ n_cell=512, max_grid_size=128
  Vector: this PR
  Tuple: Use fixed-size GpuTuple
  deviceReduce: use deviceReduceSum etc.
  Loop: Loop over single-element tuple reduction
| Method       |   CPU |     GPU | OMP w/ 8 threads |
|--------------+-------+---------+------------------|
| Vector       | 0.380 | 0.00343 |           0.0508 |
| Tuple        | 0.124 | 0.00136 |           0.0164 |
| deviceReduce | 0.161 | 0.01097 |            97.95 |
| Loop         | 0.394 | 0.00441 |           0.0627 |

* Perlmutter, A100 80GB
  30 reduces, https://github.com/WeiqunZhang/amrex-devtests/tree/main/reducer2
  Run w/ n_cell=512, max_grid_size=128
  Vector: this PR
  Tuple: Use fixed-size GpuTuple
  deviceReduce: use deviceReduceSum etc.
  Loop: Loop over single-element tuple reduction
| Method       |   CPU |    GPU |
|--------------+-------+--------|
| Vector       | 3.455 | 0.0186 |
| Tuple        | 0.117 | 0.0117 |
| deviceReduce | 0.311 | 0.0181 |
| Loop         | 3.501 | 0.0269 |
```

We should be able to improve the CPU performance.

WeiqunZhang:
I swapped the loop order for the CPU implementation so that at least the data are more likely to be in cache. That improved the 30-reduce "Vector" run from 3.455 o 2.424. But it's still way slower than the "Tuple" (0.117) and "deviceReduce" (0.311) approaches.

WeiqunZhang:
Results on Frontier.
```
* Frontier MI250X
  5 reduces, https://github.com/WeiqunZhang/amrex-devtests/tree/main/reducer
  Run w/ n_cell=512, max_grid_size=128
  Vector v0: cc375527bf92 this PR
  Vector v1: efd4839f83c2 this PR
  Tuple: Use fixed-size GpuTuple
  deviceReduce: use deviceReduceSum etc.
  Loop: Loop over single-element tuple reduction
| Method       |    CPU |     GPU |
|--------------+--------+---------|
| Vector v0    | 0.0895 | 0.00392 |
| Vector v1    | 0.5098 | N/A     |
| Tuple        | 0.0613 | 0.00142 |
| deviceReduce | 0.0627 | 0.14124 |
| Loop         | 0.2226 | 0.00427 |

* Frontier MI250X
  30 reduces, https://github.com/WeiqunZhang/amrex-devtests/tree/main/reducer2
  Run w/ n_cell=512, max_grid_size=128
  Vector: this PR
  Tuple: Use fixed-size GpuTuple
  deviceReduce: use deviceReduceSum etc.
  Loop: Loop over single-element tuple reduction
| Method       |    CPU |     GPU |
|--------------+--------+---------|
| Vector v0    | 0.4297 | 0.02054 |
| Vector v1    | 2.4189 |     N/A |
| Tuple        | 0.3310 | 0.00536 |
| deviceReduce | 0.3448 | 0.14935 |
| Loop         | 1.3521 | 0.02529 |
```

WeiqunZhang:
One thing is clear. The tuple reduce approach is the best for both GPU, CPU and CPU with OMP threads.

The deviceReduce approach seems very bad for AMD GPUs (because of atomics?).

WeiqunZhang:
I will try a different approach. If we make Reducer work with a single reduce type, it could simplify a lot of codes. Although one would need to call Reducer up to three times, it might still be a win.

WeiqunZhang:
An idea for providing runtime options in ImpacX. Maybe we could add a runtime mask to the existing tuple based reduction.

ax3l:
Can you tell me more about the mask approach?
I thought about zero-reducing unused moment terms, but I assumed the cost is in data movement and not much in the small addition per value.

WeiqunZhang:
Looks like the CPU compiler on perlmutter cheated. The time for 30-tuple reduce cannot be trusted because the compiler figured out they all have the same value. Anyway, the bottom line is still that the approach in this PR does not work well.

PR #4926, opened 2026-01-28
Add amrex::LaunchRaw

AlexanderSinn:
## Summary

This PR aims to provide a unified interface to be able to write kernels using shared memory and __syncthreads for CUDA, HIP and SYCL without the need to use ifdefs.

The number of threads per block is always a compile-time known 1D value, while the number of blocks can be 1d, 2d or 3d using the build-in platform indexes like blockIdx.y etc.

- [x] Perf testing for some existing kernels
- [x] porting/simplifying existing kernels to use this
- [x] Write documentation
- [x] add tests

## Additional background

Example of an `amrex::LaunchRaw` kernel, which fuses a transpose operation in shared memory with data preprocess and postprocess stencils. (Only works on GPUs due to threads_per_block > 1)
```C++
constexpr int tile_dim_x = 16;
constexpr int tile_dim_x_ex = 34;
constexpr int tile_dim_y = 32;
constexpr int block_rows_x = 8;
constexpr int block_rows_y = 16;

const int nx = n_half + 1;
const int ny = n_data;

const int nx_sin = n_data;
const int ny_sin = n_batch;

const int num_blocks_x = (nx + tile_dim_x - 1)/tile_dim_x;
const int num_blocks_y = (ny + tile_dim_y - 1)/tile_dim_y;

amrex::LaunchRaw<tile_dim_x*block_rows_y, amrex::Real>(
    amrex::IntVectND<2>{num_blocks_x, num_blocks_y}, tile_dim_x_ex * tile_dim_y,
    [=] AMREX_GPU_DEVICE(auto lh) noexcept
    {
        const auto [block_x, block_y] = lh.blockIdxND();

        const int tile_begin_x = 2 * block_x * tile_dim_x - 2;
        const int tile_begin_y = block_y * tile_dim_y;

        const int tile_end_x = tile_begin_x + tile_dim_x_ex;
        const int tile_end_y = tile_begin_y + tile_dim_y;

        Array2<amrex::Real> shared{{lh.shared_memory(),
                                    {tile_begin_x, tile_begin_y, 0},
                                    {tile_end_x, tile_end_y, 1}, 1}};

        {
            const auto [thread_y, thread_x] =
                lh.template threadIdxND<tile_dim_y, block_rows_x>();

            for (int tx = thread_x; tx < tile_dim_x_ex; tx += block_rows_x) {
                const int i = tile_begin_x + tx;
                const int j = tile_begin_y + thread_y;

                if (j < nx_sin && i < ny_sin && i >= 0 ) {
                    shared(i, j) = transpose_to_sine(i, j);
                }
            }
        }

        lh.syncthreads();

        {
            const auto [thread_x, thread_y] =
                lh.template threadIdxND<tile_dim_x, block_rows_y>();

            for (int ty = thread_y; ty < tile_dim_y; ty += block_rows_y) {
                const int i = block_x * tile_dim_x + thread_x;
                const int j = tile_begin_y + ty;

                if (i < nx && j < ny) {
                    out(i, j) = to_complex(shared, i, j, n_half, n_batch);
                }
            }
        }
    });
```

## Checklist

The proposed changes:


AlexanderSinn:
I don't understand why the HYPRE and SUNDIALS tests keep failing. Maybe it is a CUDA compiler bug?
```
[ 12%] Building CUDA object Src/CMakeFiles/amrex_3d.dir/Base/AMReX_ParallelContext.cpp.o
cd /home/runner/work/amrex/amrex/build/Src && ccache /usr/local/cuda/bin/nvcc -forward-unknown-to-host-compiler -DAMREX_SPACEDIM=3 -Damrex_3d_EXPORTS --options-file CMakeFiles/amrex_3d.dir/includes_CUDA.rsp -O3 -DNDEBUG -std=c++17 "--generate-code=arch=compute_80,code=[compute_80,sm_80]" -Xcompiler=-fPIC --expt-relaxed-constexpr --expt-extended-lambda -Xcudafe --diag_suppress=esa_on_defaulted_function_ignored -Xcudafe --diag_suppress=implicit_return_from_non_void_function -maxrregcount=255 -Xcudafe --display_error_number --Wext-lambda-captures-this --use_fast_math --generate-line-info -MD -MT Src/CMakeFiles/amrex_3d.dir/Base/AMReX_ParallelContext.cpp.o -MF CMakeFiles/amrex_3d.dir/Base/AMReX_ParallelContext.cpp.o.d -x cu -rdc=true -c /home/runner/work/amrex/amrex/Src/Base/AMReX_ParallelContext.cpp -o CMakeFiles/amrex_3d.dir/Base/AMReX_ParallelContext.cpp.o
/home/runner/work/amrex/amrex/Src/Base/AMReX_Reduce.H(679): error: no instance of function template "amrex::Reduce::detail::call_f_intvect_n" matches the argument list
            argument types are: (lambda [](int, int, int)->ReduceTuple, amrex::IntVectND<3>, int)
          detected during:
            instantiation of "void amrex::ReduceOps<Ps...>::eval_box(I, const amrex::BoxND<dim> &, int, D &, const F &) [with Ps=<amrex::ReduceOpSum>, I=amrex::Reduce::detail::iterate_box, dim=3, D=amrex::ReduceData<amrex::Real>, F=lambda [](int, int, int)->ReduceTuple]" 
(753): here
            instantiation of "void amrex::ReduceOps<Ps...>::eval(const amrex::BoxND<dim> &, D &, const F &) [with Ps=<amrex::ReduceOpSum>, D=amrex::ReduceData<amrex::Real>, F=lambda [](int, int, int)->ReduceTuple, dim=3]" 
/home/runner/work/amrex/amrex/Src/Base/AMReX_BaseFab.H(3749): here
            instantiation of "T amrex::BaseFab<T>::sum<run_on>(const amrex::Box &, amrex::DestComp, amrex::NumComps) const noexcept [with T=amrex::Real, run_on=amrex::RunOn::Device]" 
/home/runner/work/amrex/amrex/Src/Base/AMReX_BaseFab.H(2735): here
            instantiation of "T amrex::BaseFab<T>::sum<run_on>(const amrex::Box &, int, int) const noexcept [with T=amrex::Real, run_on=amrex::RunOn::Device]" 
/home/runner/work/amrex/amrex/Src/Base/AMReX_DistributionMapping.cpp(1708): here

1 error detected in the compilation of "/home/runner/work/amrex/amrex/Src/Base/AMReX_DistributionMapping.cpp".
```

Edit: Fixed by changed the input type of the device lambda from `auto` to the concrete type.

AlexanderSinn:
Performance test with HiPACE++.
The functions marked with `<-` contain LaunchRaw/launch kernels that use shared memory.
PR with amrex::LaunchRaw:
```powershell
TinyProfiler total time across processes [min...avg...max]: 35.24 ... 35.24 ... 35.24

-------------------------------------------------------------------------------------------------------
Name                                                    NCalls  Excl. Min  Excl. Avg  Excl. Max   Max %
-------------------------------------------------------------------------------------------------------
hpmg::MultiGrid::solve1()                                 1000      12.51      12.51      12.51  35.50% <-
ExplicitDeposition()                                      1000      5.037      5.037      5.037  14.29% <-
FFTPoissonSolverDirichletQuick::SolvePoissonEquation()    3000      4.778      4.778      4.778  13.56% <-
AdvancePlasmaParticles()                                  1000      3.628      3.628      3.628  10.30%
DepositCurrent_PlasmaParticleContainer()                  1001      3.005      3.005      3.005   8.53% <-
AdvanceBeamParticlesSlice()                               1000      1.646      1.646      1.646   4.67%
DepositCurrentSlice_BeamParticleContainer()               2000     0.9694     0.9694     0.9694   2.75% <-
Fields::SolvePoissonPsiExmByEypBxEzBz()                   1000     0.9331     0.9331     0.9331   2.65%
Fields::ShiftSlices()                                     1000     0.8349     0.8349     0.8349   2.37%
```
dev with amrex::launch:
```powershell
TinyProfiler total time across processes [min...avg...max]: 35.36 ... 35.36 ... 35.36

-------------------------------------------------------------------------------------------------------
Name                                                    NCalls  Excl. Min  Excl. Avg  Excl. Max   Max %
-------------------------------------------------------------------------------------------------------
hpmg::MultiGrid::solve1()                                 1000      12.55      12.55      12.55  35.49% <-
ExplicitDeposition()                                      1000       5.08       5.08       5.08  14.37% <-
FFTPoissonSolverDirichletQuick::SolvePoissonEquation()    3000       4.78       4.78       4.78  13.52% <-
AdvancePlasmaParticles()                                  1000       3.63       3.63       3.63  10.27%
DepositCurrent_PlasmaParticleContainer()                  1001      3.029      3.029      3.029   8.57% <-
AdvanceBeamParticlesSlice()                               1000      1.646      1.646      1.646   4.66%
DepositCurrentSlice_BeamParticleContainer()               2000     0.9731     0.9731     0.9731   2.75% <-
Fields::SolvePoissonPsiExmByEypBxEzBz()                   1000     0.9333     0.9333     0.9333   2.64%
Fields::ShiftSlices()                                     1000     0.8351     0.8351     0.8351   2.36%
```

WeiqunZhang:
Codex:

• The new LaunchRaw infrastructure introduces a compile-time error for all SYCL builds via LaunchHandler::sycl_item, so the patch cannot be considered correct.

  Review comment:

  - [P1] Return value of LaunchHandler::sycl_item is wrong type — /home/wqzhang/mygitrepo/amrex/Src/Base/AMReX_GpuTypes.H:304-307
    The new Gpu::LaunchHandler::sycl_item() implementation (src/Base/AMReX_GpuTypes.H lines 304‑307) returns m_item, but m_item is a sycl::nd_item<dim> const*. There is no implicit conversion from sycl::nd_item* to sycl::nd_item, so every SYCL build will now fail to compile (return m_item; cannot bind to the function’s sycl::nd_item<dim> return type). The method should dereference the stored pointer (or return a pointer) instead of returning it directly.

WeiqunZhang:
Codex:
```
diff --git a/Src/Particle/AMReX_ParticleUtil.H b/Src/Particle/AMReX_ParticleUtil.H
index c2278f8004..cdefebb929 100644
--- a/Src/Particle/AMReX_ParticleUtil.H
+++ b/Src/Particle/AMReX_ParticleUtil.H
@@ -822,7 +822,7 @@ void PermutationForDeposition (Gpu::DeviceVector<index_type>& perm, index_type n
                 }

                 index_type num_particles_block =
-                    Gpu::blockReduceSum<gpu_block_size>(num_particles_thread);
+                    Gpu::blockReduceSum(num_particles_thread AMREX_IF_SYCL(, lh.handler()));

                 if (lh.threadIdx1D() == 0) {
                     global_idx_start = Gpu::Atomic::Add(pglobal_idx, num_particles_block);
```

AlexanderSinn:
If this still gives the `amrex::Abort::0::LaunchRaw: The local size dimension of submitted nd_range doesn't match the required work-group size dimension!!!!!` error on PVC, then we will have to add one overload for each number of dimensions for SYCL.

WeiqunZhang:
Codex:

• The new LaunchRaw implementation introduces backend regressions: CPU-only builds cannot use normal GPU-sized block counts, and the SYCL path drops existing safeguards for large kernel captures and shared-memory alignment. Those issues make the new API unsafe to treat as fully portable.

  Full review comments:

  - [P2] Remove the CPU-only MT == 1 restriction from LaunchRaw — /home/wqzhang/mygitrepo/amrex/Src/Base/AMReX_GpuLaunchFunctsC.H:139-139
    In non-GPU builds this static_assert makes every LaunchRaw<MT> instantiation with MT != 1 fail at compile time, so code written against the new API is no longer source-portable across backends the way launch<MT> and ParallelFor are. That already leaks into this patch via the new test, which has to special-case CPU block sizes; any real CPU-only use of LaunchRaw<64>/<256> will hit the same build break.
  - [P2] Preserve the SYCL big-kernel workaround in LaunchRaw — /home/wqzhang/mygitrepo/amrex/Src/Base/AMReX_GpuLaunchFunctsG.H:913-919
    On SYCL every existing launch wrapper first routes large functors through detail::SyclKernelDevPtr, because kernels with captures above roughly 1792 bytes are not reliably accepted when captured directly. LaunchRaw skips that path and captures f inline, so any large-capture raw launch now regresses on SYCL; that is especially visible here because AMReX_Reduce.H was switched over to LaunchRaw, and reduction functors can easily exceed that size.
  - [P2] Keep SYCL shared memory aligned independently of T — /home/wqzhang/mygitrepo/amrex/Src/Base/AMReX_GpuLaunchFunctsG.H:945-945
    This changes the SYCL scratch allocation from an unsigned long long buffer to local_accessor<T>, so lh.shared_memory() is now only aligned as T. Several converted call sites placement-new a larger type on top of that storage (for example amrex::fill(BaseFab<STRUCT>) constructs STRUCT inside a T* buffer), which becomes misaligned whenever alignof(STRUCT) > alignof(T) but was still safe with the old 8-byte backing store. A SYCL build using an alignas(8) struct with value_type = float is enough to trigger undefined behavior here.

WeiqunZhang:
The changes suggested by Codex.

[uncommitted-changes.patch](https://github.com/user-attachments/files/25998329/uncommitted-changes.patch)

AlexanderSinn:
Using it on a CPU with MT > 1 is not going to work for stuff that uses syncthreads. The CPU backend version is only for testing in CI in case code can be written to work with MT == 1. The only other thing would be to spawn tons of CPU threads to emulate a GPU for testing, but that would be very slow and involved.  

The large kernel thing should indeed be added.

WeiqunZhang:
```
diff --git a/Tests/LaunchRaw/GNUmakefile b/Tests/LaunchRaw/GNUmakefile
index 173eef4c67..b23330bb62 100644
--- a/Tests/LaunchRaw/GNUmakefile
+++ b/Tests/LaunchRaw/GNUmakefile
@@ -1,4 +1,4 @@
-AMREX_HOME = ../../../
+AMREX_HOME = ../../
 
 DEBUG  = FALSE
 ```

**Comment by WeiqunZhang** [comment]
Codex:

• The new LaunchRaw API is not fully usable on SYCL for its advertised 2D/3D cases, and the newly added GNUmake test cannot be built from its checked-in path. Those issues make the patch incorrect as submitted.

  Full review comments:

  - [P2] Preserve 2D/3D SYCL support in LaunchHandler::handler() — /home/wqzhang/mygitrepo/amrex/Src/Base/AMReX_GpuTypes.H:290-292
    Under SYCL, handler() unconditionally forwards m_item into Gpu::Handler, but Gpu::Handler only has a constructor taking sycl::nd_item<1> const*. That means any new 2D or 3D LaunchRaw kernel that calls lh.handler()—for example to use Gpu::blockReduce* or another helper that still expects a Gpu::Handler—will fail to compile, even though LaunchRaw advertises 1D/2D/3D block support.
  - [P2] Point the new GNUmake test at the repository root — /home/wqzhang/mygitrepo/amrex/Tests/LaunchRaw/GNUmakefile:1-1
    From Tests/LaunchRaw, AMREX_HOME = ../../../ resolves one level above the checkout, so make -C Tests/LaunchRaw cannot find Tools/GNUMake/Make.rules and the new test does not build with the GNUmake-based test flow. The other tests use ../.. here.

**Comment by WeiqunZhang** [comment]
For the SYCL issue handle above, Codex suggests,

[launchraw-current.patch](https://github.com/user-attachments/files/25999194/launchraw-current.patch)

**Comment by WeiqunZhang** [comment]
The work-group-size runtime issue still exists. Codex suggests the following. Note that it's the diff against your branch with all the previous Codex changes.


[launchraw-workgroups-fix.patch](https://github.com/user-attachments/files/25999239/launchraw-workgroups-fix.patch)

**Comment by WeiqunZhang** [comment]
Re: MT > 1 on CPU, could you add a message to static_assert?

**Comment by AlexanderSinn** [comment]
Added. For SYCL it is getting a bit more complicated than I expected. Maybe we could just use a 1D range and split the block index manually using FastDivmodU64?

**Comment by WeiqunZhang** [comment]
Okay

**Comment by WeiqunZhang** [comment]
Codex:

• The patch introduces backend regressions: CPU-only builds now use an incomplete IntVectND type in LaunchHandler.

  Full review comments:

  - [P1] Include the full IntVectND definition before storing it in LaunchHandler — /home/wqzhang/mygitrepo/amrex/Src/Base/AMReX_GpuTypes.H:401-403
    In non-GPU builds this class now stores IntVectND<dim> members by value, but AMReX_GpuTypes.H only pulls in AMReX_BaseFwd.H, which forward-declares IntVectND. AMReX_GpuLaunch.H includes this header before AMReX_Box.H/AMReX_IntVect.H, so CPU-only configurations now see an incomplete type here and fail to compile as soon as they include the launch headers.

**Comment by AlexanderSinn** [comment]
Can you start GPU CI again?

**Comment by WeiqunZhang** [comment]
It's one of the known minor issues that we define gpuStream_t for CUDA/HIP in AMReX_Control.H and for SYCL in AMReX_GpuTypes.H. We probably should move them to AMReX_GpuTypes.H (and add appropriate CUDA/HIP headers).

Initially Arena only included GpuControl.H, which in turn included GpuTypes.H. But now GpuTypes.H is removed from GpuControl.H. If we move gpuStream_t to GpuTypes.H, Arena.H will no longer need to include GpuControl.H since you just added GpuTypes.H. There is another minor issue (found by AI recently) that can be fixed. GpuTypes.H uses macros defined in AMReX_Qualifiers.H. I was planning to fix it. But maybe you can just fix it in this PR.

File: Tests/LaunchRaw/GNUmakefile [hunk 1]
```diff
@@ -0,0 +1,15 @@
+AMREX_HOME = ../../../
```
WeiqunZhang: We need to update the path to AMREX_HOME to `../../`.

PR #4938, opened 2026-02-06
Test: SIMD

ax3l:
## Summary

Needed to cover SIMD features and generic SIMD/non-SIMD patterns.
Bravely vibe coded, but now reviewed and improved for sensibility.

## Additional background

#4924 #4607 #4600  #4520

## Checklist

The proposed changes:

### DIFF
--- a/.github/workflows/gcc.yml
+++ b/.github/workflows/gcc.yml
@@ -263,6 +263,8 @@ jobs:
       # We could put -Wnull-dereference back once this uses GCC >= 15.
       # Right now, vector::resize in packIODataCpu triggers a false positive warning.
       # https://gcc.gnu.org/bugzilla/show_bug.cgi?id=108860
+      #
+      # `-fno-operator-names` is not used here because the SIMD headers use `and` and `or`.
       env: {CXXFLAGS: "-Werror -Wall -Wextra -Wpedantic -Wfloat-conversion -Wshadow -Woverloaded-virtual -Wextra-semi -Wno-array-bounds -Wnon-virtual-dtor -Wlogical-op -Wmisleading-indentation -Wduplicated-cond -Wduplicated-branches -Wmissing-include-dirs"}
       run: |
         export CCACHE_COMPRESS=1
--- a/.github/workflows/gcc.yml
+++ b/.github/workflows/gcc.yml
@@ -234,7 +234,7 @@ jobs:
 
   # Build libamrex and all tests
   tests_cxx20:
-    name: GNU@11 C++20 OMP [tests]
+    name: GNU@12 C++20 OMP SIMD [tests]
     runs-on: ubuntu-24.04
     needs: check_changes
     if: needs.check_changes.outputs.has_non_docs_changes == 'true'
@@ -242,7 +242,7 @@ jobs:
     - uses: actions/checkout@v6
     - name: Dependencies
       run: |
-        .github/workflows/dependencies/dependencies_gcc.sh 11
+        .github/workflows/dependencies/dependencies_gcc.sh 12
         .github/workflows/dependencies/dependencies_clang-tidy-apt-llvm.sh 17
         .github/workflows/dependencies/dependencies_ccache.sh
     - name: install vir-simd
@@ -260,7 +260,7 @@ jobs:
         restore-keys: |
              ccache-${{ github.workflow }}-${{ github.job }}-git-
     - name: Build & Install
-      env: {CXXFLAGS: "-Werror -Wall -Wextra -Wpedantic -Wnull-dereference -Wfloat-conversion -Wshadow -Woverloaded-virtual -Wextra-semi -Wnon-virtual-dtor -Wlogical-op -Wmisleading-indentation -Wduplicated-cond -Wduplicated-branches -Wmissing-include-dirs"}
+      env: {CXXFLAGS: "-Werror -Wall -Wextra -Wpedantic -Wnull-dereference -Wfloat-conversion -Wshadow -Woverloaded-virtual -Wextra-semi -Wno-array-bounds -Wnon-virtual-dtor -Wlogical-op -Wmisleading-indentation -Wduplicated-cond -Wduplicated-branches -Wmissing-include-dirs"}
       run: |
         export CCACHE_COMPRESS=1
         export CCACHE_COMPRESSLEVEL=10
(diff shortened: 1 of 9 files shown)


.github/workflows/gcc.yml
```diff
@@ -245,6 +245,13 @@ jobs:
         .github/workflows/dependencies/dependencies_gcc.sh 10
         .github/workflows/dependencies/dependencies_clang-tidy-apt-llvm.sh 17
         .github/workflows/dependencies/dependencies_ccache.sh
+    - name: install vir-simd
+      run: |
```
ax3l, line 249: I think I need to pick a newer GCC entry. This one, GCC 10, still has `-std=c++2a`:
```
/home/runner/work/amrex/amrex/Src/Particle/AMReX_Particle.H: In function ‘void amrex::particle_impl::make_invalid(T_SIMD&, const T_Mask&)’:
/home/runner/work/amrex/amrex/Src/Particle/AMReX_Particle.H:111:43: error: ‘cvt’ is not a member of ‘amrex::simd::stdx’
  111 |             simd::stdx::where(simd::stdx::cvt(mask), idcpu) &= ~(T_SIMD{1} << 63);
      |                                           ^~~
```
so I think this block:
```C++
#   if __cplusplus >= 202002L
        using vir::cvt;
#   endif
```
in `AMReX_SIMD.H` does not work yet. (We need proper C++20 for our SIMD implementation for the mask converters to work properly.)

Tests/SIMD/GNUmakefile
```diff
@@ -0,0 +1,20 @@
...
+DIM	= 3
+COMP    = gcc
+
+USE_MPI   = FALSE
+USE_OMP   = FALSE
+USE_CUDA  = FALSE
+
+TINY_PROFILE = FALSE
+
+CXXSTD = c++20
```
ax3l, line 13: Is this ok here or will this infect all other GNUmake tests?

Tests/SIMD/main.cpp
```diff
@@ -0,0 +1,430 @@
+// Test coverage for AMReX SIMD single-source design:
+//   SIMDindex, ParallelForSIMD, load_1d, store_1d,
+//   SIMDInt, Vectorized, is_vectorized, is_nth_arg_non_const
```
ax3l: Removed this one:
```suggestion
//   Vectorized, is_vectorized, is_nth_arg_non_const
```

ax3l:
@WeiqunZhang can you please remove the `GNU@10.1 C++20 OMP [tests]` test from required in the AMReX branch protection settings? I will need to modify this one to be named `GNU@12 C++20 OMP SIMD [tests]` for clarity if that is ok.

ax3l:
Oh wow, GCC has so many false positives. I might bump to 13...

WeiqunZhang:
Sometimes, it's not entirely the compiler's fault. They always assume the worse case scenario. But if we also do that, it will affect the code's performance.

ax3l:
Yeah it seems a bit unhappy in GCC 11 & 12 with the ParticleIO routines. I could disable that nullptr check, but since I try to build a first SIMD runner, I would like to not turn that one off to keep this covered for purely SIMD code paths.

Let me know if I shall add a purely new runner, there might be a reason to keep GCC 10 around as a runner as well for now.

Tests/SIMD/GNUmakefile
ax3l: @WeiqunZhang quick GNUmake question here

WeiqunZhang: Yes, this is okay.

.github/workflows/gcc.yml
```diff
@@ -253,7 +260,7 @@ jobs:
         restore-keys: |
              ccache-${{ github.workflow }}-${{ github.job }}-git-
     - name: Build & Install
-      env: {CXXFLAGS: "-fno-operator-names -Werror -Wall -Wextra -Wpedantic -Wnull-dereference -Wfloat-conversion -Wshadow -Woverloaded-virtual -Wextra-semi -Wnon-virtual-dtor -Wlogical-op -Wmisleading-indentation -Wduplicated-cond -Wduplicated-branches -Wmissing-include-dirs"}
```
WeiqunZhang, line 256: Why is `-fno-operator-names` removed?

Tests/SIMD/main.cpp
```diff
@@ -0,0 +1,439 @@
...
+            nerrors += err;
+            Print() << "Position-dependent load_1d+store_1d: "
+                    << (err == 0 ? "PASSED" : "FAILED") << "\n";
+        }
+
+        // ================================================================
+        // Final report
+        // ================================================================
+        if (nerrors > 0) {
+            amrex::Finalize();
```
WeiqunZhang, line 432: Don't think we need to call amrex::Finalize here.

.github/workflows/gcc.yml
ax3l: In CI, I saw that in some of the transitive header of the SIMD libs they use `and` and `or`.

Tests/SIMD/main.cpp
ax3l: Looked a bit cleaner in the error case when I ran it locally.

.github/workflows/gcc.yml
WeiqunZhang: Could we make them belong to system headers?

Tests/SIMD/main.cpp
WeiqunZhang: Maybe because `Abort` was able to print out more information after amrex::Finalize? But I think usually we want more information.

ax3l: The "unclean" exit before finalize showed some pointer/memory issues, which did not look too good.

.github/workflows/gcc.yml
WeiqunZhang: Looks like the simd libs is installed at /usr/local and `-I/usr/local/include` is used.

WeiqunZhang:
The `null-dereference` is due to this GCC bug. https://gcc.gnu.org/bugzilla/show_bug.cgi?id=108860 I think we have to add `-Wno-null-dereference`.

WeiqunZhang (approved):

Src/Particle/AMReX_ParticleIO.H
```diff
@@ -616,7 +616,7 @@ ParticleContainer_impl<ParticleType, NArrayReal, NArrayInt, Allocator, CellAssig
         which[grid] = fnum;
         where[grid] = VisMF::FileOffset(ofs);
 
-        if (count[grid] == 0) { continue; }
+        if (count[grid] <= 0) { continue; }
```
ax3l, line 619: :tada: Thank you, Weiqun!

ax3l:
Thank you for your help, it works :tada: 

And fixes a warning :tada:

PR #4941, opened 2026-02-06
Fix: AMReX w/ OpenMP and Ignore FFTW OMP

ax3l:
## Summary

In some corner cases, we sometimes work around broken FFTW installs: for some (temporary) reason we cannot have any threading in FFTW, neither with OpenMP nor threads.

In those cases, when we still want to thread the rest of AMReX with OpenMP, we can set our AMReX transitive FFTW OpenMP requirements explicitly off, e.g., via CMake `-DAMReX_FFTW_IGNORE_OMP=ON`. But this is not yet implemented consistently and since #4861 exposes undefined FFTW symbols when used.

This enables this work-around.

- [x] macro
- [x] CMake
- [x] GNUmake
- [x] `AMReX_Config.H`

## Additional background

https://github.com/AMReX-Codes/amrex/pull/4861#issuecomment-3863046390

## Checklist

The proposed changes:

### DIFF
--- a/Src/FFT/AMReX_FFT.cpp
+++ b/Src/FFT/AMReX_FFT.cpp
@@ -21,7 +21,7 @@ void Initialize ()
     {
 #if defined(AMREX_USE_HIP) && defined(AMREX_USE_FFT)
         AMREX_ROCFFT_SAFE_CALL(rocfft_setup());
-#elif !defined(AMREX_USE_GPU) && defined(AMREX_USE_OMP) && defined(AMREX_USE_FFT)
+#elif !defined(AMREX_USE_GPU) && defined(AMREX_USE_MULTI_THREADED_FFTW)
         fftw_init_threads();
         fftwf_init_threads();
         fftw_plan_with_nthreads(amrex::OpenMP::get_max_threads());
--- a/Tools/CMake/AMReX_Config_ND.H.in
+++ b/Tools/CMake/AMReX_Config_ND.H.in
@@ -42,6 +42,7 @@
 #cmakedefine BL_FORT_USE_UPPERCASE
 #cmakedefine BL_NO_FORT
 #cmakedefine AMREX_USE_FFT
+#cmakedefine AMREX_USE_MULTI_THREADED_FFTW
 #cmakedefine AMREX_USE_SENSEI_INSITU
 #cmakedefine AMREX_NO_SENSEI_AMR_INST
 #cmakedefine AMREX_USE_CONDUIT
--- a/Tools/CMake/FindAMReXFFTW.cmake
+++ b/Tools/CMake/FindAMReXFFTW.cmake
@@ -46,20 +46,20 @@ mark_as_advanced(AMReX_FFTW_IGNORE_OMP)
 set(AMReX_FFTW_OMP_SUFFIX omp CACHE STRING "FFTW3's Thread Support variant (omp/threads)")
 mark_as_advanced(AMReX_FFTW_OMP_SUFFIX)
 
-# Set the AMREX_FFTW_OMP=1 define on AMReX::FFTW if TRUE and print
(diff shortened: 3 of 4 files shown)


Src/FFT/AMReX_FFT.cpp
```diff
@@ -21,7 +21,7 @@ void Initialize ()
     {
 #if defined(AMREX_USE_HIP) && defined(AMREX_USE_FFT)
         AMREX_ROCFFT_SAFE_CALL(rocfft_setup());
-#elif !defined(AMREX_USE_GPU) && defined(AMREX_USE_OMP) && defined(AMREX_USE_FFT)
+#elif !defined(AMREX_USE_GPU) && defined(AMREX_FFTW_OMP)
```
ax3l: @WeiqunZhang shall we:
- name this `AMREX_USE_FFT_OMP` or so?
- move this to our usual Config headers?

WeiqunZhang:
What about the call to `fftw_plan_with_nthreads`? If we are not using OMP, how many threads to use?

WeiqunZhang:
Okay this is confusing. Could you summarize what we are trying to support? AMReX may be built with or without OMP. FFTW on the system may or may not support threading. If it does support threading, it may or may not be OMP. In all the combinations, we do we do?

ax3l:
Oh yes, `fftw_plan_with_nthreads` needs a guard as well

ax3l:
> Okay this is confusing. Could you summarize what we are trying to support? AMReX may be built with or without OMP. FFTW on the system may or may not support threading. If it does support threading, it may or may not be OMP. In all the combinations, we do we do?

In #4861 we decided it might be ok to default to require that FFTW supports threading if AMReX is build with OpenMP support, because other behavior is [confusing](https://github.com/BLAST-ImpactX/impactx/issues/1249) for [users](https://github.com/orgs/BLAST-ImpactX/discussions/1245).

I now found a corner case on [conda-forge](https://github.com/conda-forge/fftw-feedstock/issues/102) and in a CI runner where I cannot have, _temporarily_ (I hope), FFTW with OpenMP nor threads. To now build I am forced to either disable all FFT support or all OpenMP support of AMReX.

In CMake, I had anticipated this a bit in the past by adding an opt-in `AMReX_FFTW_IGNORE_OMP` option, saying: I, the user, bulding AMReX with OpenMP, know that my FFTs will be slow and unthreaded, but do it nonetheless. But I forgot that we need to add support for this control in GNUmake and macros when we merged #4861 .

Src/FFT/AMReX_FFT.cpp
WeiqunZhang: Yes to both.

WeiqunZhang: Neither `AMREX_FFTW_OMP` nor `AMREX_USE_FFT_OMP` seems appropriate. The issue is FFTW might be using non-OpenMP threads. Maybe we should call it `AMREX_USE_MULTI_THREADED_FFTW`.

Or maybe we should drop the non-OpenMP FFTW threading support completely, because we don't even know how many threads to use. So it will be up to the user anyway. They could link to whatever FFT library they want. And they can still initialize the threading support in their code even if we are not doing anything.

ax3l: I rather go with `AMREX_USE_MULTI_THREADED_FFTW`. The use case is to enable serial FFTW as a fallback with OpenMP enabled AMReX.

The default is to use threaded FFTW with the same number of thread as we use OpenMP threads in AMReX.

ax3l: @WeiqunZhang ok, I think I have everything but GNUmake support set up now :)

WeiqunZhang (approved):

PR #4942, opened 2026-02-07
Fix GCC array-bounds warning

WeiqunZhang:


WeiqunZhang:
It seems hopeless to make GCC happy.
```
error: array subscript [0, 2] is outside array bounds of ‘int [3]’
```

PR #4944, opened 2026-02-10
CI: Fix clang-tidy cache

WeiqunZhang:
ccache 4.12 has a slight new different format in its log. So We have to update our python script processing and generating makefile for clang-tidy.

Fix new warnings from clang-tidy-21. We missed them because we switched to ccache-4.12 before switching to clang-tidy-21.

Also enable clang-tidy for the C++20 test.

### DIFF
--- a/Src/Amr/AMReX_Amr.cpp
+++ b/Src/Amr/AMReX_Amr.cpp
@@ -2678,7 +2678,7 @@ Amr::regrid (int  lbase,
     // Reclaim all remaining storage for levels > new_finest.
     //
     for(int lev = new_finest + 1; lev <= finest_level; ++lev) {
-        amr_level[lev].reset();
+        amr_level[lev] = nullptr;
         this->ClearBoxArray(lev);
         this->ClearDistributionMap(lev);
     }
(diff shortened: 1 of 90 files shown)


Src/Amr/AMReX_Amr.cpp
```diff
@@ -2678,7 +2678,7 @@ Amr::regrid (int  lbase,
     // Reclaim all remaining storage for levels > new_finest.
     //
     for(int lev = new_finest + 1; lev <= finest_level; ++lev) {
-        amr_level[lev].reset();
+        amr_level[lev] = nullptr;
```
ax3l, line 2681: Or did you mean `release` instead of `reset`?
```suggestion
        amr_level[lev].release();
```
https://en.cppreference.com/w/cpp/memory/unique_ptr/release.html

WeiqunZhang: No, that would be a memory leak. It's not a bug. It's a readability issue. The issue is both `amr_level[lev].reset()` and `amr_level[lev]->reset()` are legal code because the former calls `std::unique_ptr::reset`, whereas the latter calls `AmrLevel::reset`. We do want to call the former, but it can be confusing. So clang-tidy does not like it.

ax3l: Ah wild, got it.

asalmgren (approved):

PR #4949, opened 2026-02-12
SENSEI: Allocator

ax3l:
## Summary

Add the `Allocator` template, so we can use this with polymorphic PCs (and generally any PC that does not use the default allocator).

## Additional background

Same as  #4776 #4948 but for SENSEI.

## Checklist

The proposed changes:


ax3l:
@WeiqunZhang @atmyers ready for review/merge :)

atmyers (approved):

PR #4960, opened 2026-02-21
Doxygen Src/Base

WeiqunZhang:
With help from Claude Code and Codex.

### DIFF
--- a/Src/Base/AMReX_Arena.H
+++ b/Src/Base/AMReX_Arena.H
@@ -162,35 +162,44 @@ public:
     [[nodiscard]] virtual void* alloc (std::size_t sz) = 0;
 
     /**
-     * \brief Allocate memory in-place if possible
+     * \brief Allocate memory in-place if possible.
      *
      * In this implementation of the virtual function, a new allocation is
      * always performed, and it returns the new pointer and allocation size.
-     *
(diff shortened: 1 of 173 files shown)


Src/Base/AMReX_Arena.H
```diff
@@ -330,14 +349,11 @@ protected:
     static void out_of_memory_abort (std::string const& memory_type, std::size_t nbytes,
                                      std::string const& error_msg);
 
+    /// \cond DOXYGEN_IGNORE
     struct ArenaProfiler {
-        //! If this arena is profiled by TinyProfiler
```
AlexanderSinn, line 334: Can these comments be kept in a way that won't show up in Doxygen?

WeiqunZhang: `/// \cond DOXYGEN_IGNORE`  will make doxygen ignore the comments.

WeiqunZhang: I will tell it to not remove comments that are inside `\cond`. I told it not worry about non-public interface. It might have taken it to mean removing doxygen comments that will be ignored by doxygen.

WeiqunZhang:
There are more that I have not pushed. Eventually I am going to have to split this into many much smaller PRs.

PR #4969, opened 2026-02-23
prevent auto-converting const char * to bool in ParmParse::add

charlesgwaldman:
## Summary
`ParmParse::add("key", "value")` will result in adding a boolean value (1) due to C++ converting pointer to bool.  This is almost certainly not intended.  Users must write
`add("key", std::string("value")` which is cumbersome and easy to forget.  Adding explicit `const char * -> std::string` overload removes this pitfall 

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_ParmParse.H
+++ b/Src/Base/AMReX_ParmParse.H
@@ -614,6 +614,8 @@ public:
     //! Add a key 'name' with value 'val' to the end of the PP table.
     void add (std::string_view name, const std::string& val);
 
+    //! Convert const char * to string, to prevent auto-conversion to bool
+    void add(const char* name, const char* val) { add(name, std::string(val)); }
 
     //! Similar to get() but store the whole line including empty spaces in
     //! the middle if present as a single string. For example, `foo = a b c`
--- a/Src/Base/AMReX_ParmParse.H
+++ b/Src/Base/AMReX_ParmParse.H
@@ -617,12 +617,6 @@ public:
     //! Convert const char * to string, to prevent auto-conversion to bool
     void add (std::string_view name, const char* val) { add(name, std::string(val)); }
 
-    //! Special overload for 0 to avoid issues with add("key", 0) calling
-    //! the const char* version.
-    void add (std::string_view name, std::integral_constant<int, 0>) {
-        add(name, int(0));
-    }
-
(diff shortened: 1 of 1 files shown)


Src/Base/AMReX_ParmParse.H
```diff
@@ -614,6 +614,8 @@ public:
     //! Add a key 'name' with value 'val' to the end of the PP table.
     void add (std::string_view name, const std::string& val);
 
+    //! Convert const char * to string, to prevent auto-conversion to bool
+    void add(const char* name, const char* val) { add(name, std::string(val)); }
```
WeiqunZhang: This breaks `add("key", 0)`. The compilers seem to choose this new overload over the int version. I think we should do this.

```suggestion                                     
    void add (std::string_view name, const char* val) { add(name, std::string(val)); }

    //! Special overload for 0 to avoid issues with add("key", 0) calling                                                                                                    
    //! the const char* version.                                                                                                                                             
    void add (std::string_view name, std::integral_constant<int, 0>) {
        add(name, int(0));
    }
```

WeiqunZhang: Unfortunately, it does not solve the original problem.

WeiqunZhang (changes requested):

Src/Base/AMReX_ParmParse.H
WeiqunZhang: Sorry, I think we can actually remove the integral_constant version. I was testing different combinations and compiler flags, and what I suggested seemed to work. But what really made the difference was `const char* name` -> `std::string_view name`.

With your original version in this PR, we have
```
void add (std::string_view name, int val);
void add (const char* name, const char* val);
```
That makes the call to `add("name", 0)` ambiguous because the first version matches better on the second argument whereas the second version matches better on the first argument.

When we have,
```
void add (std::string_view name, int val);
void add (std::string_view name, const char* val);
```
The first argument is now the same. And the first version is now a better match for `add("name", 0)`.

charlesgwaldman:
Thanks @WeiqunZhang  - that make sense.  Clearly I should have used `string_view` in the first place.  I appreciate you taking the time to review this.

WeiqunZhang (approved):

PR #4976, opened 2026-02-24
CI: Use C++20

WeiqunZhang:
Prepare for switching to C++20.


WeiqunZhang:
See also #4840

ax3l (approved):
Looks great!

Not sure if you like to merge this one or bump the build system so not all CI scripts get a manual entry.

WeiqunZhang:
I plan to merge this after the 26.03 release so that we can be sure that 26.03 still fully works with C++17.

WeiqunZhang:
Trying to remove some `-Wno-` flags. CIs may fail. If they do and we still cannot fix it, I will put them back.

atmyers (approved):

PR #5070, opened 2026-02-25
Doxygen HYPRE

WeiqunZhang:


asalmgren:
I'd suggest we be consistent about capitalization of hypre --  it looks like the LLNL folks themselves use both "hypre" and "HYPRE" so we can choose either but I'd suggest our comments be consistent with the capitalization so it's easier to search.  Maybe "Hypre" in the function names and "hypre" or "HYPRE" in the comments?

WeiqunZhang:
I have updated the comments in Src/Extern/HYPRE to use HYPRE consistently. As for Src/LinearSolvers/, to avoid conflicts, I will do it in a different PR. (I have a branch with a huge amount of changes, and I am splitting it into much smaller chunks.)

WeiqunZhang:
I will also update sphinx doc in another PR.

asalmgren (approved):

PR #5074, opened 2026-02-25
Fix restart where you lose a level and NReaders > NProcs()

atmyers:
The proposed changes:

### DIFF
--- a/Src/Extern/HDF5/AMReX_ParticleHDF5.H
+++ b/Src/Extern/HDF5/AMReX_ParticleHDF5.H
@@ -1321,7 +1321,7 @@ ParticleContainer_impl<ParticleType, NArrayReal, NArrayInt, Allocator, CellAssig
 
             const int rank = ParallelDescriptor::MyProc();
             const int NReaders = MaxReaders();
-            if (rank >= NReaders) { return; }
+            if (rank >= NReaders) { continue; }
 
             const int Navg = ngrids[lev] / NReaders;
             const int Nleft = ngrids[lev] - Navg * NReaders;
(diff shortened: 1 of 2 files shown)


WeiqunZhang:
Do we need to fix Extern/HDF5/AMReX_ParticleHDF5.H as well?
```
$ git grep -n ">=\s*NReaders"
Extern/HDF5/AMReX_ParticleHDF5.H:1324:            if (rank >= NReaders) { return; }
Particle/AMReX_ParticleIO.H:855:            if (rank >= NReaders) { return; }
```

Src/Extern/HDF5/AMReX_ParticleHDF5.H
```diff
@@ -1321,7 +1321,7 @@ ParticleContainer_impl<ParticleType, NArrayReal, NArrayInt, Allocator, CellAssig
 
             const int rank = ParallelDescriptor::MyProc();
             const int NReaders = MaxReaders();
-            if (rank >= NReaders) { return; }
+            if (rank >= NReaders) { continue; }
```
WeiqunZhang: There are memory leaks here. Before `continue`, we need to call `H5*close` just like how it's done at the end of the loop.

WeiqunZhang (changes requested):

Src/Extern/HDF5/AMReX_ParticleHDF5.H
```diff
@@ -1321,7 +1321,12 @@ ParticleContainer_impl<ParticleType, NArrayReal, NArrayInt, Allocator, CellAssig
 
             const int rank = ParallelDescriptor::MyProc();
             const int NReaders = MaxReaders();
-            if (rank >= NReaders) { return; }
+            if (rank >= NReaders) {
+                H5Dclose(int_dset);
+                H5Dclose(real_dset);
+        H5Gclose(grp);
```
WeiqunZhang:
```suggestion
                H5Gclose(grp);
```

WeiqunZhang (approved):

Issue #5056, opened 2026-02-24
Dual-grid restart loops index vectors by `finestLevel()` after sizing by file levels


WeiqunZhang:
- **Type**: Correctness / Memory safety
- **Severity**: High
- **Component**: Particle restart dual-grid handling
- **Location**:
  - `Src/Particle/AMReX_ParticleIO.H:766`
  - `Src/Particle/AMReX_ParticleIO.H:787`
  - `Src/Particle/AMReX_ParticleIO.H:812`
  - `Src/Particle/AMReX_ParticleIO.H:933`

## Problem
`old_dms`, `old_bas`, and `particle_box_arrays` are sized with `finest_level_in_file + 1`, but several loops run to `finestLevel()`.

If runtime finest level is greater than levels present in the file, those vectors are indexed out of bounds.

## Impact
- Potential out-of-bounds access and undefined behavior during restart.
- Fragile behavior for mismatched level-count restart workflows.

## Suggested patch
Bound dual-grid bookkeeping loops to the common level count.

```diff
--- a/Src/Particle/AMReX_ParticleIO.H
+++ b/Src/Particle/AMReX_ParticleIO.H
@@
-        for (int lev = 0; lev <= finestLevel(); lev++)
+        const int nlev_common = std::min(finestLevel(), finest_level_in_file);
+        for (int lev = 0; lev <= nlev_common; lev++)
         {
@@
-        for (int lev = 0; lev <= finestLevel(); lev++) {
+        for (int lev = 0; lev <= nlev_common; lev++) {
@@
-        for (int lev = 0; lev <= finestLevel(); lev++) {
+        for (int lev = 0; lev <= nlev_common; lev++) {
```

Prepared by Codex

BenWibking:
We have seen this in practice. Currently, we've hard-coded an assert to avoid this case, because it causes a segfault otherwise.

BenWibking:
See here: https://github.com/quokka-astro/quokka/blob/9b0321556e51623e36e8d387fe2ea59e96da8727/src/simulation.hpp#L4830

atmyers:
I don't think the proposed fix here is correct as it would skip reading particles in the case that we lose a level on restart. However, a fix for the bug and a test that would have caught this are [here](https://github.com/AMReX-Codes/amrex/pull/5075).

PR #5075, opened 2026-02-25
Add new test and fix dual grid restart bug

atmyers:
This fixes Issue #5056. Note, the test included here also would have caught the NReaders bug fixed in #5074.

The proposed changes:


WeiqunZhang (approved):

WeiqunZhang:
A similar change might also be needed for HDF5?

atmyers:
Yes, I will fix the HDF5 in a follow-up PR.

PR #5086, opened 2026-02-26
Fix GPU memory issue in the ReduceToPlanePatchy test

WeiqunZhang:
ParallelDescriptor::ReduceSum does not support GPU device memory.


WeiqunZhang:
The failure was due to the Nvidia CI machine issue. HIP and Intel GPU tests worked.

asalmgren (approved):

WeiqunZhang:
I will merge this one right now, because we don't to want to have a know ctest failure in the 26.03 release. But for the other approved PRs, I will wait until after the next release.

PR #4649, opened 2025-09-08
[WIP] OpenBC SIMD

ax3l:
## Summary

Enable SIMD acceleration (CPU) in `OpenBCSolver<T>::setGreensFunction`.

- [ ] debug
- [ ] generalize

## Additional background

See https://github.com/BLAST-WarpX/warpx/issues/6090

## Checklist

The proposed changes:
- [ ] fix a bug or incorrect behavior in AMReX
- [ ] add new capabilities to AMReX
- [ ] changes answers in the test suite to more than roundoff level
- [ ] are likely to significantly affect the results of downstream AMReX users
- [ ] include documentation in the code and/or rst files, if appropriate

PR #5095, opened 2026-02-26
SIMD: Add where and ternary operator

ax3l:
## Summary

To write more portable code, add a scalar fallback for `std::simd::where` for non-SIMD code. Expose both as `amrex::simd::stdx`.

Add Doxygen string to existing scalar `amrex::simd::std::any_of` fallback.

Add a new implementation for a portable ternary operator, inspired by Kokkos [condition](https://kokkos.org/kokkos-core-wiki/ProgrammingGuide/SIMD.html#ternary-operator) as `amrex::simd::stdx::select`. (We call it `select` because that is the likely ISO C++26 name: https://en.cppreference.com/w/cpp/numeric/simd.html )

Demonstrate usage in doc strings and new test cases.

## Additional background

These are patterns I commonly see needed in [BLAST-ImpactX](https://github.com/BLAST-ImpactX/impactx), e.g. [here](https://github.com/BLAST-ImpactX/impactx/blob/26.02/src/elements/mixin/spintransport.H#L68-L79).

This will help to write more complex SIMD code like #4649 in a cleaner way, too.

Follow up to #4520 #4938 #4924

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_SIMD.H
+++ b/Src/Base/AMReX_SIMD.H
@@ -31,8 +31,57 @@ namespace amrex::simd
 #else
         // fallback implementations for functions that are commonly used in portable code paths
 
+        /** True if the boolean value is true (scalar identity fallback for simd any_of)
+         *
+         * Example:
+         * ```cpp
+         * // Works for both simd_mask and scalar bool:
+         * auto mask = a > b;
+         * if (amrex::simd::stdx::any_of(mask)) { ... }
+         * ```
+         *
+         * @see https://en.cppreference.com/w/cpp/experimental/simd/any_of
+         */
         AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
         bool any_of (bool const v) { return v; }
+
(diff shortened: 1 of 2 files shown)


ax3l:
@WeiqunZhang oh dang, I force pushed away your last commit. Sorry for this! If you still have it locally feel free to push. Just cleaning up santitizer warnings rn

WeiqunZhang:
No problems. Very simple change.

WeiqunZhang:
And your change was actually I wanted to do. But I was not 100 percent sure. So I did the nolint way. If your change works, it's better than nolint.

WeiqunZhang (approved):

ax3l:
Oh no, small CI breakage for 2 entries because I waited too long with the merge: 12f9a543b2e73d680e364a2a30240972af4e40e0

ax3l:
Ah, the CUDA build ran out of disk... The other one is a clang tidy modernize, will post a PR: #5448

PR #5149, opened 2026-03-09
ParmParse: Support query/get a specific element from TOML array.

WeiqunZhang:
See also #5147.


BenWibking (approved):
LGTM. It works in our code.

asalmgren (approved):

PR #5150, opened 2026-03-10
ParmParse: accept true/false t/f for int keys

charlesgwaldman:
## Summary
ParmParse complains about settings like
```
amrex.abort_on_out_of_gpu_memory = true
```
with a message
```
amrex::Error::0::ParmParse: failed to parse true due to unknown symbol true !!!
```
According to our MFIX-Exa users this is a recent regression.  The key documented as an integer but previous AMReX versions accepted true/false.  I'm not completly sure when this behavior changed.  The proposed patch reverts the behavior to accept "true" (case-insensitive) or "t" for the "1" value and "false/f" for 0.
 
## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_ParmParse.cpp
+++ b/Src/Base/AMReX_ParmParse.cpp
@@ -133,6 +133,19 @@ template <class T>
 bool
 is (const std::string& str, T& val)
 {
+    if constexpr (std::is_integral_v<T>) {
+        auto const lo_str = amrex::toLower(str);
+        if ( lo_str == "true" || lo_str == "t" )
+            {
+                val = 1;
+                return true;
+            }
+        else if ( lo_str == "false" || lo_str == "f" )
+            {
+                val = 0;
+                return true;
+            }
+    }
     return isT(str, val);
 }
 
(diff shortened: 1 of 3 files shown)


charlesgwaldman:
Test passes

charlesgwaldman:
Failing check is error in CI infrastructure (failure to pull ccache from github), not a problem with the code

WeiqunZhang:
`amrex.abort_on_out_of_gpu_memory` has always been a bool. So if there is a regression, it's the other way around. That is ParmParse is broken for bool. But I just tested, it seems to work fine. We accept `true`, `false`, `t`, `f`, 0 and 1 for bool.

Maybe what happened is MFIX-Exa was trying to add it as int?

That being said, it might still be a good idea to accept true`, `false`, `t`, and `f` as int because a lot of our boolean flags are stored as int.

AlexanderSinn:
In HiPACE++ we add true (=1) and false (=0) as variables to the AMReX parser so that those can also be used in parser-evaluated ParmParse expressions.

WeiqunZhang:
When convert a string to bool, only lower case strings are accepted. But in this PR, the conversion from string to int is case insensitive. I think we should make it consistent. Either

Could we also add reading bool keys to the test?

WeiqunZhang:
You can also reduce the duplication in the specialization of `is`.

```
@@ -133,6 +133,19 @@ template <class T>
 bool
 is (const std::string& str, T& val)
 {
+    if constexpr (std::is_integral_v<T>) {
+        auto const lo_str = amrex::toLower(str);
+        if ( lo_str == "true" || lo_str == "t" )
+        {
+            val = 1;
+            return true;
+        }
+        else if ( lo_str == "false" || lo_str == "f" )
+        {
+            val = 0;
+            return true;
+        }
+    }
     return isT(str, val);
 }
```

charlesgwaldman:
> When convert a string to bool, only lower case strings are accepted.

I don't think this is correct, look at line 165 of the original code

```
template <>
bool
is (const std::string& str, bool& val)
{
    auto const lo_str = amrex::toLower(str);
    if ( lo_str == "true" || lo_str == "t" )
```

I have updated the template specialization and added more tests per your suggestion.
Thank you.

WeiqunZhang:
Oh, you are right. Somehow I missed it.

WeiqunZhang (approved):

PR #5180, opened 2026-03-11
ParmParse: Fix printing TOML array

WeiqunZhang:
When ParmParse reads a TOML array from inputs, it inserts a special string "$$ARR" as an indication for TOML array. The special string should be removed before printing.


WeiqunZhang:
There is also an issue with printing strings. But that is an existing issue, #5181.

asalmgren (approved):

PR #5189, opened 2026-03-12
Parser: fix self assignment of locals

WeiqunZhang:
When a local variable appeared on both sides of an assignment (e.g., `r = x*x + y*y; r = sqrt(r);`), the parser would compile the RHS while the new `r` slot was already active, so the RHS saw the uninitialized version of `r`. We now compile the RHS first and only add/update the local slot after, so self-referential assignments evaluate correctly.


asalmgren (approved):

phsikorski-source (commented):
Tested.

PR #5191, opened 2026-03-12
Add missing include

iljah:
Fixes call to `WriteSingleLevelPlotfileHDF5` in my program:
```
.../amrex/Src/Extern/HDF5/AMReX_PlotFileUtilHDF5.cpp: In function ”void amrex::WriteSingleLevelPlotfileHDF5(const std::string&, const MultiFab&, const Vector<std::__cxx11::basic_string<char> >&, const Geometry&, Real, int, const std::string&, const std::string&, const std::string&, const std::string&, const Vector<std::__cxx11::basic_string<char> >&)”:
.../amrex/Src/Extern/HDF5/AMReX_PlotFileUtilHDF5.cpp:1297:5: virhe: ”WriteMultiLevelPlotfileHDF5” was not declared in this scope; did you mean ”WriteMultiLevelPlotfile”?
 1297 |     WriteMultiLevelPlotfileHDF5(plotfilename, 1, mfarr, varnames, geomarr, time, level_steps, ref_ratio,
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~
      |     WriteMultiLevelPlotfile
make: *** [.../amrex/Tools/GNUMake/Make.rules:262: tmp_build_dir/o/3d.gnu.DEBUG.EXE/AMReX_PlotFileUtilHDF5.o] Error 1
```

The proposed changes:

### DIFF
--- a/Src/Extern/HDF5/AMReX_PlotFileUtilHDF5.cpp
+++ b/Src/Extern/HDF5/AMReX_PlotFileUtilHDF5.cpp
@@ -1,6 +1,7 @@
 #include <AMReX_VisMF.H>
 #include <AMReX_AsyncOut.H>
 #include <AMReX_PlotFileUtil.H>
+#include <AMReX_PlotFileUtilHDF5.H>
 #include <AMReX_FPC.H>
 #include <AMReX_FabArrayUtility.H>
 


WeiqunZhang:
I think the issue is in your build system. We cannot guarantee that the code would compile if AMREX_USE_HDF5 is not defined. How are you building your code?

https://amrex-codes.github.io/amrex/docs_html/BuildingAMReX_Chapter.html
https://amrex-codes.github.io/amrex/docs_html/IO.html#hdf5-plotfile
https://amrex-codes.github.io/amrex/docs_html/BuildingAMReX.html#tab-cmakevar

iljah:
Usually N.cpp would include N.H...

My makefile is
```make
AMREX_HOME ?= $(HOME)/amrex
DEBUG = TRUE
COMP = gnu
CXXSTD = c++20
USE_CUDA = FALSE
USE_HDF5 = TRUE
USE_HIP = FALSE
USE_MPI = FALSE
USE_OMP = FALSE
USE_PARTICLES = TRUE
PRECISION = DOUBLE
include $(AMREX_HOME)/Tools/GNUMake/Make.defs
CEXE_sources += main.cpp
include $(AMREX_HOME)/Src/AmrCore/Make.package
include $(AMREX_HOME)/Src/Base/Make.package
include $(AMREX_HOME)/Src/Boundary/Make.package
include $(AMREX_HOME)/Src/Extern/HDF5/Make.package
include $(AMREX_HOME)/Src/Particle/Make.package
include $(AMREX_HOME)/Tools/GNUMake/Make.rules
test_dir = $(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
test_name = $(shell basename $(test_dir))
...
```

iljah:
If I use `USE_HDF5 = FALSE` without above patch I get same result. With above patch (and `USE_HDF5 = FALSE`) I get linking errors such as `AMReX_PlotFileUtilHDF5.cpp:1280:(.text+0x56e9): undefined reference to H5Fclose`. With above patch and `USE_HDF5 = TRUE` it builds.

WeiqunZhang:
Maybe you need to run `make clean` first to clean up the previous configuration.

The `USE_HDF5` flag does not affect the name of temporary build directory and the executable, by default. So when you switch between USE_HDF5=FALSE and TRUE, you need to clean up with `make clean`. Some flags like DEBUG, USE_MPI, etc. do not need that because they are part of the build directory name. Nevertheless, if you have a problem with `make`, you always want to try `make clean` just in case. That being said, you can make hdf5 part of your directory and executable name with this in your makefile.
```
USERSuffix = .HDF5
```

iljah:
Cleaning before building seems to have helped, thanks.

PR #5196, opened 2026-03-12
ParmParse::add: Improve TOML compatibility

### LINKED ISSUE #5182: ParmParse::add: Compatibility with TOML
We need to make sure that the entries added by ParmParse::add are compatible with TOML so that when the ParmParse database is printed out it can be read as TOML. Note that the goal is not to turn TOML incompatible entries in a ParmParse file into TOML compatible, because that is not always possible even for things as simple as `a = x y z`, because ParmParse may read it as 3 strings or 3 ints or 3 doubles or even a single string. But for ParmParse::add, the type information is specified.

We need to make sure we print floating point numbers as TOML format (e.g., 1.0, not 1 or 1.). We also need to use `[]` for arrays and `""` for strings. For bool, we need to print either `true` or `false`, not 1 or 0.

WeiqunZhang:
Make sure a string scalar with white spaces does not get split, when it's printed out by `prettyPrintTable`.

For bool, we save them as `true` or `false`, instead of 1 or 0.

For floating point numbers, we use TOML-compatible format (e.g., 1.0 not 1).

Note that `addarr` remains incompatible with TOML, because ParmParse array differs from TOML array.

Close #5182

### DIFF
--- a/Src/Base/AMReX_ParmParse.H
+++ b/Src/Base/AMReX_ParmParse.H
@@ -1707,10 +1707,13 @@ public:
     static void SetVerbose (int v);
 
     //! Write the contents of the table in ASCII to the ostream (diagnostic only, not valid ParmParse syntax).
+    //! Diagnostic view of the internal table. This text is not valid ParmParse syntax
+    //! and is meant for debugging only; use prettyPrintTable() for canonical, re-readable output.
     static void dumpTable (std::ostream& os, bool prettyPrint = false);
 
     //! Write all entries in ParmParse syntax suitable for re-reading. If
-    //! there are duplicates, only the last one is printed.
+    //! there are duplicates, only the last one is printed; the output can be fed
+    //! back to ParmParse via addfile/Initialize.
     static void prettyPrintTable (std::ostream& os);
 
     //! Write unused inputs in a pretty way to the ostream. If there are
--- a/Src/Base/AMReX_ParmParse.cpp
+++ b/Src/Base/AMReX_ParmParse.cpp
@@ -9,6 +9,7 @@
 #include <AMReX_Utility.H>
 
 #include <algorithm>
+#include <cmath>
 #include <cctype>
 #include <cstdlib>
 #include <iostream>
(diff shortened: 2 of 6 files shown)


asalmgren (approved):

WeiqunZhang:
Codex says,

• Programmatically added bool parameters are now serialized as the strings "true"/"false" but the floating-point parsing path still only accepts numeric literals, so querying those parameters as Real values now aborts. This regression can break existing usage of ParmParse where bools are treated as numeric toggles.

  Review comment:

  - [P2] Serializing bools as "true" breaks Real queries — /home/wqzhang/mygitrepo/amrex/Src/Base/AMReX_ParmParse.cpp:1360-1391
    The new to_toml_value/saddval block now formats every programmatically-added bool with std::boolalpha, so entries added via pp.add("flag", true) are stored as the literal strings "true"/"false" (lines 1360‑1391). The floating-point path (is(const std::string&, double&) → is_floating_point) still only accepts numeric literals, so attempting to read the same entry into a float/double now fails with a type-mismatch/Abort because std::istringstream cannot parse the word “true”. This used to work because bools were serialized as 0/1. Any existing code that queries a bool as a Real (e.g. to treat it as a 0/1 switch) will now crash as soon as the bool was populated via pp.add. Either keep the numeric serialization for bools or teach the floating-point converter to interpret true/false as 1/0 to avoid this regression.

WeiqunZhang:
I guess it's possible to fix it.

WeiqunZhang:
Codex:

  - is_literal_bool now sits ahead of both the integral and floating-point branches, so a value added as "true"/"false" will once again be read successfully as either an integer toggle or a Real (Src/Base/AMReX_ParmParse.cpp:160-210). I don’t see any remaining path where Real queries bypass that helper.
  - The bool specialization now just delegates to is_literal_bool before falling back to numeric parsing, keeping backward compatibility for strings like "1" or "0" and covering the new TOML-style literals at the same time (Src/Base/AMReX_ParmParse.cpp:223-241).
  - No additional issues turned up in this diff. For confidence, you might run the ParmParse tests (e.g., Tests/ParmParse) with cases that pp.add("flag", true) followed by pp.get("flag", dbl).

  Everything else looked consistent with the intended behavior.

PR #5199, opened 2026-03-13
Add external GPU stream override support

WeiqunZhang:
Expose APIs to install and reset a user-managed GPU stream, document the new behavior, and add a GPU test covering the external-stream execution path.

### DIFF
--- a/Src/Base/AMReX_GpuDevice.cpp
+++ b/Src/Base/AMReX_GpuDevice.cpp
@@ -770,6 +770,16 @@ Device::setExternalStream (gpuStream_t s)
 {
     AMREX_ALWAYS_ASSERT_WITH_MESSAGE(!OpenMP::in_parallel(),
         "Gpu::setExternalGpuStream is not supported inside OpenMP parallel regions.");
+#ifdef AMREX_USE_SYCL
+    AMREX_ALWAYS_ASSERT_WITH_MESSAGE(s.queue != nullptr,
+        "Gpu::setExternalGpuStream: null SYCL queue is not supported.");
+    AMREX_ALWAYS_ASSERT_WITH_MESSAGE(s.queue->get_context() == Device::syclContext(),
+        "Gpu::setExternalGpuStream: external SYCL queue must use AMReX's SYCL context.");
+    AMREX_ALWAYS_ASSERT_WITH_MESSAGE(s.queue->get_device() == Device::syclDevice(),
+        "Gpu::setExternalGpuStream: external SYCL queue must use AMReX's SYCL device.");
+    AMREX_ALWAYS_ASSERT_WITH_MESSAGE(s.queue->has_property<sycl::property::queue::in_order>(),
+        "Gpu::setExternalGpuStream: external SYCL queue must be in-order.");
+#endif
     // External stream overrides intentionally do not try to synchronize
     // OpenACC stream state. AMReX does not support mixing OpenACC work into
     // these regions.
--- a/Src/Base/AMReX_GpuDevice.cpp
+++ b/Src/Base/AMReX_GpuDevice.cpp
@@ -102,6 +102,9 @@ Vector<StreamManager>   Device::gpu_stream_pool;
 Vector<int>             Device::gpu_stream_index;
 gpuDeviceProp_t         Device::device_prop;
 int                     Device::memory_pools_supported = 0;
+#ifdef AMREX_USE_GPU
+Vector<Device::ExternalStream> Device::external_stream_stack;
+#endif
 
 constexpr int Device::warp_size;
 
@@ -143,7 +146,12 @@ namespace {
 }
 
 [[nodiscard]] gpuStream_t&
-StreamManager::get () {
+StreamManager::getStream () {
+    return m_stream;
+}
+
+gpuStream_t const&
+StreamManager::getStream () const {
     return m_stream;
 }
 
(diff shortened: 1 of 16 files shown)


WeiqunZhang:
Codex:

• The external-stream override does not remain transparent to existing stream-switching and per-stream bookkeeping paths. In particular, stream-ordered frees can be issued on the wrong stream, and nested external reductions can alias the same storage slot.

  Full review comments:

  - [P1] Honor setStream() while an external override is active — /home/wqzhang/mygitrepo/amrex/Src/Base/AMReX_GpuDevice.H:86-89
    With this change, gpuStream() always returns external_stream_stack.back() whenever any external stream is active, so temporary calls to Device::setStream()/setStreamIndex() no longer affect code that queries Gpu::gpuStream(). That breaks existing paths like BaseFab::clear() (alloc_stream is saved in AMReX_BaseFab.H:1470 and restored via setStream(alloc_stream) at :1710) before freeing stream-ordered memory: if a BaseFab allocated on another stream is destroyed inside an ExternalGpuStreamRegion, PArena::free()/SArena::free() will enqueue the free on the current external stream instead of the allocation stream. For cudaFreeAsync/Gpu::freeAsync, that can reuse the buffer before the original stream has finished.
  - [P2] Give external streams distinct streamIndex() values — /home/wqzhang/mygitrepo/amrex/Src/Base/AMReX_GpuDevice.cpp:717-717
    Returning 0 for every non-pool stream makes all external streams share the same per-stream slot in ReduceData (AMReX_Reduce.H:497-510). If a Reducer/ReduceData is reused across nested external-stream scopes—one of the new supported scenarios here—the outer and inner kernels both write m_device_tuple + 0*m_max_blocks with no synchronization when the inner scope is entered, so overlapping work on the two streams can corrupt the reduction buffer and m_nblocks[0] bookkeeping.

WeiqunZhang:
Re: Codex's first comment, The behavior of `Gpu::gpuStream` is intentional. However, we should try to fix the BaseFab::clear issue even though that kind of usage should be avoided.

WeiqunZhang:
The first issue raised by Codex has been addressed. BaseFab::clear is now more robust. We also can document the caveat.

WeiqunZhang:
As for the second issue raised by Codex, we document and assert the reduction limitation.

WeiqunZhang:
Review Round 2 by Codex:

• The external-stream support introduces at least one concrete regression in stream-ordered memory frees for PArena-backed BaseFabs, and it leaves OpenACC interop on the wrong stream in mixed CUDA/OpenACC applications. Those cases can break correctness in supported GPU configurations.

  Full review comments:

  - [P1] Handle PArena when freeing stream-ordered BaseFabs — /home/wqzhang/mygitrepo/amrex/Src/Base/AMReX_BaseFab.H:1710-1711
    This special-cases SArena, but PArena is also a stream-ordered arena. If a BaseFab/FArrayBox is allocated from PArena and later resized or destroyed on a different GPU stream, this now falls back to this->free() on the current stream instead of alloc_stream; PArena::free() will then enqueue cudaFreeAsync/hipFreeAsync or the fallback Elixir on the wrong stream, so work still running on the allocation stream can observe a premature free. The previous code switched to alloc_stream for all arenas, so this is a regression even without using the new external-stream API.
  - [P2] Keep OpenACC stream state in sync with external overrides — /home/wqzhang/mygitrepo/amrex/Src/Base/AMReX_GpuDevice.cpp:751-756
    In CUDA+AMREX_USE_ACC builds, AMReX normally mirrors stream changes into OpenACC with acc_set_cuda_stream/amrex_set_acc_stream (see setStreamIndex()), but setExternalGpuStream() does not. When an application enters an external-stream region and then launches Fortran/OpenACC work with async(acc_stream), those kernels continue on the previously selected AMReX-managed stream while AMReX C++ kernels run on the external stream, so the expected ordering inside the region is lost.

WeiqunZhang:
Let's forbid the use of PArena with BaseFab to address the first issue in the second review by Codex. As for the OpenACC concern, we can add a comment and simply ignore it.

WeiqunZhang:
Review Round 3 by Codex:

• The external-stream support is close, but it introduces at least two correctness regressions: a new unconditional abort for PArena-backed BaseFabs on fallback backends, and incomplete guarding of ReduceData when an external stream is popped before value() is evaluated.

  Full review comments:

  - [P2] Avoid aborting BaseFab on fallback PArena backends — /home/wqzhang/mygitrepo/amrex/Src/Base/AMReX_BaseFab.H:1468-1469
    This unconditional abort breaks any BaseFab/FArrayBox created with a PArena even on GPU builds where PArena is only a thin wrapper around The_Arena() (for example SYCL builds, CUDA < 11.2, or runtimes where memoryPoolsSupported()==false). In those configurations there is no stream-ordered free issue to protect against, so existing callers that passed a PArena will now fail during define() for no functional reason.
  - [P2] Guard Reducer::value() after leaving an external stream region — /home/wqzhang/mygitrepo/amrex/Src/Base/AMReX_Reduce.H:516-521
    This new slot-0 check only runs when the reduction writes per-stream storage. If a reduction is launched inside ExternalGpuStreamRegion(..., ExternalStreamSync::No) and value() is called after that guard has popped, ReduceOps::value() still runs its final copy/reduction on the current AMReX stream (usually stream 0) with no dependency on the external stream that filled slot 0. The documented "unsupported" cross-stream case therefore produces stale results instead of tripping the new assertion.

WeiqunZhang:
Let's ignore the PArena issue. It's not really used anymore and amrex will no longer support CUDA 11 once we switch to C++20. For the scond, we can add an assertion.

WeiqunZhang:
Codex is finally happy.

• I did not find any correctness, performance, or safety regressions introduced by the changes. The new external stream support and associated documentation and tests appear consistent with the existing design.

Src/Base/AMReX_GpuDevice.cpp
```diff
@@ -796,22 +837,23 @@ Device::usingExternalStream () noexcept
...
         }
     }
     for (auto& s : gpu_stream_pool) {
-        try {
-            s.getStream().queue->wait_and_throw();
-        } catch (sycl::exception const& ex) {
-            amrex::Abort(std::string("synchronize: ")+ex.what()+"!!!!!");
-        }
+        s.drainFreeList();
     }
```
AlexanderSinn: There is a slight bug here. Between the calls to cudaDeviceSynchronize and drainFreeList, another thread could launch a GPU function and free_async some memory related to it. That execution order would cause the GPU function to not be synchronized, but the memory to be freed already. That is why in the normal sync function, it swaps the m_free_wait_list first and then synchronizes. The way it currently is, both drainFreeList and Device::synchronize are only guaranteed to be correct outside OMP parallel regions.

AlexanderSinn: Hmm, seems like github was showing me something outdated.

WeiqunZhang: Yes, your comment about race condition was correct. It has been fixed now.

WeiqunZhang:
I changed the get function name in StreamManager from `get` to `getStream`, because it was very confusing when we have `std::unique_ptr<StreamManager>`: `manager.get()` vs. `manager->get()`.

BenWibking:
It appears to work end-to-end on Frontier.

asalmgren (approved):

PR #5200, opened 2026-03-14
IWYU with help by Codex

WeiqunZhang:
The goal is to apply some of the IWYU ideas and make us more IWYU friendly, but not to make us fully compliant with the IWYU tool.

We have added many IWYY export and private pragmas and have done some cleanup.


WeiqunZhang:
@lucafedeli88

asalmgren (approved):

WeiqunZhang:
My conversation with Codex.

Me:
› From now on, we can call this primary-header rule. That is headers include by foo.H are implicityly exported to foo.cpp.
  
Codex:  
• Understood.
  
  I’ll use primary-header rule to mean:
  
  - if foo.cpp includes its primary header foo.H
  - then anything included by foo.H is treated as implicitly exported to foo.cpp
  - so foo.cpp should not repeat those includes unless there is a special reason

  I’ll apply that rule going forward.

Me:
› Let me ask you what you think of this rule? Is it a reasonable rule or is it against the spirit of IWYU?

Codex:
• I think it’s a reasonable rule.

  For headers, IWYU’s core idea still matters: headers should include what they need and be self-sufficient.

  For implementation files, your primary-header rule is pragmatic and defensible:

  - foo.cpp is tightly coupled to foo.H
  - foo.H is part of foo.cpp’s intended interface boundary
  - repeating all of foo.H’s includes in foo.cpp usually adds noise, not clarity

  So I’d frame it like this:

  - for .H: strict IWYU-style self-sufficiency
  - for .cpp: allow transitive dependence through the primary header

  That is slightly looser than “pure” IWYU, because strict IWYU often wants every translation unit to include what it directly uses. But in real C++ codebases, your rule is common and sensible. It preserves the most important property: if foo.H changes in a way that breaks foo.cpp, that is usually an appropriate coupling, not an accidental one.

  I think the rule is especially good for AMReX because:

  - many primary headers are real public surfaces, not tiny forward-declare shells
  - implementation files often naturally use the same vocabulary as their primary headers
  - it reduces redundant include churn without weakening header hygiene

  So my answer is:
  
  - yes, it is a reasonable engineering rule
  - yes, it is looser than strict IWYU
  - no, I don’t think it violates the spirit in a harmful way, as long as you keep headers themselves disciplined

WeiqunZhang:
amrex_base_export_map

Export map for Src/Base/.

WeiqunZhang:
So I think now if you include `AMReX_MultiFab.H` and use `amrex::IntVect` and `amrex::Real`, the IWYU tool should no longer try to include `AMReX_REAL.H` and `AMReX_IntVect.H`.

@jrood-nrel

WeiqunZhang:
AMReX_MultiFab.H could be regarded as a facade that provides most (not all of course) of the classes and functions you normally need for MultiFab/FAB/Array4 related operations. We did not really add any new headers to AMReX_MultiFab.H, but we now promise that it will provide common types like Array4 and Real via IWYU export pragma. At the i-j-k kernel level, including AMReX_Array4.H will provide you with types like Array4, Real, IntVect, Dim3, and macros like AMREX_GPU_DEVICE and AMREX_FORCE_INLINE. You might still need to include AMReX_Box.H, if you use them in kernels.

asalmgren (approved):

asalmgren (approved):

WeiqunZhang:
I am going to close this PR because there is no evidence that this makes the compilation faster. See below for more details. It might be more "robust". But given that this contains so many changes and it has not been tested by application codes at all, it's likely less "robust" with this PR.

I did a test without particle, mpi, gpu and fortran. I compared three versions: development, ai_iwyu (this PR branch), iwyu_iwyu. The last one was based on the ai_iwyu branch, and it was "fixed" by the IWYU tool. The tool broke the compilation, so I fixed some changes to make it compile. I used a single process for compiling without using ccache. It took about 3.5 minute to compile. The difference between the three branches was less than 3 seconds.

I also compared development with this branch (but not with the IWYU tool because it broke the code too much) with mpi, particle, fortran and cuda enabled. The difference was still only a few seconds.

jrood-nrel:
Thanks for looking into this. I agree this seems like a lot of change for not really any benefit. I think it's worth it to know that this has been investigated and proven to not be worthwhile.

PR #5201, opened 2026-03-16
CMake: Remove hipcc deprecation warning

BenWibking:
## Summary
`hipcc` will be replaced with a symlink to `amdclang`, so this warning is unnecessary.

## Checklist

The proposed changes:


WeiqunZhang:
@ax3l I think it's safe to remove the warning.

ax3l (approved):
Ack, ty!

PR #5215, opened 2026-03-20
Avoid invalid in AMReX_MFInterp_3D_C.H

dpgrote:
## Summary

This PR is here to avoid a false positive invalid flag being raised in `AmrCore/AMReX_MFInterp_3D_C.H. The change ensures that dumax is never zero for the divides. This avoids the invalid signal from being triggered when if-conversion optimization is done, where this code is predictively executed even though dumax is zero.

Several checks demonstrate that the change does not break vectorization. Also, the change should have no effect on results since under normal conditions, the divide by dumax would never happen if dumax is zero.

## Checklist

The proposed changes:

### DIFF
--- a/Src/AmrCore/AMReX_MFInterp_3D_C.H
+++ b/Src/AmrCore/AMReX_MFInterp_3D_C.H
@@ -233,7 +233,7 @@ void mf_cell_cons_lin_interp_mcslope (int i, int j, int k, int ns, Array4<Real> 
         // This avoids the invalid signal from being triggered when
         // if-conversion optimization is done, where this code is
         // predictively executed even though dumax is zero.
-        dumax = dumax == 0. ? 1. : dumax;
+        dumax = dumax == Real(0.) ? Real(1.) : dumax;
         if (dumax * alpha > (umax - u(i,j,k,nu))) {
             alpha = (umax - u(i,j,k,nu)) / dumax;
         }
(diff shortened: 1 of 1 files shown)


Src/AmrCore/AMReX_MFInterp_3D_C.H
```diff
@@ -229,6 +229,11 @@ void mf_cell_cons_lin_interp_mcslope (int i, int j, int k, int ns, Array4<Real>
             umin = amrex::min(umin, u(i+ioff,j+joff,k+koff,nu));
             umax = amrex::max(umax, u(i+ioff,j+joff,k+koff,nu));
         }}}
+        // This ensures that dumax is never zero for the divides below.
+        // This avoids the invalid signal from being triggered when
+        // if-conversion optimization is done, where this code is
+        // predictively executed even though dumax is zero.
+        dumax = dumax == Real(0.) ? Real(1.) : dumax;
```
WeiqunZhang, line 239: Could you add to the comments something like below?
```
In normal execution, dumax is always positive because at least one of the three terms in its definition is positive, so the following statement is effectively a no-op.
```
I initially thought the added guard needed to be moved into the if block because dumax is used in the condition and modifying it could change the outcome of the test. I had to go back more than 10 lines to realize that it could never happen.

WeiqunZhang (changes requested):

WeiqunZhang (approved):

PR #5222, opened 2026-03-23
Fix Wrong `noexcept` Placements

ax3l:
## Summary

We use AMReX interactively with `amrex.throw_exception = true` to be able to handle user errors or convergence issues and act on them.

This fixes functions marked as `noexcept` where there could actually be an exception been thrown. Introduces some checks for invalid code paths in hot loops to be only added for debug builds, as in common to avoid straining compiler optimizations and bloating registers.

## To Do

- [x] rebase after #5219 was merged
- [ ] add [new clang-tidy rules](https://github.com/AMReX-Codes/amrex/pull/5219#issuecomment-4112748213)?

## Additional background

Follow-up to #5219 and bugs like https://github.com/BLAST-ImpactX/impactx/issues/1348

## Checklist

The proposed changes:


WeiqunZhang:
> This fixes functions marked as noexcept where there could actually be an exception been thrown.

I don't think we should make it a rule that `noexcept` can only be used on functions that do not have exceptions. It's about "we really mean it and it's a hard error if there is an exception."

ax3l:
The question is if a user should be able to trigger this at runtime and if yes, if we want to be able to recover from that.

In interactive workflows, the answer to that is usually: yes, do not _abort_ the process (Python interpreter). We might just stop the simulation. In those workflows, AMReX is a gear in a larger machinery, it cannot pull down the whole process without being disruptive.

Example: maybe you run 1000s of AMReX sims in a scipy.optimize or BO loop without manual supervision. Just because one did not converge or ran into some invalid state does not mean we do not want to continue optimizing for other parameters.

Thus, the philosophy I follow is more:
Can a user observe an invalid state with their runtime parameters: errors should be recoverable for at least the interpreter (we might still need to re-init AMReX).

ax3l:
@WeiqunZhang does this look good to go?

We can play with the clang tidy rule in a separate PR and see if we like it or not.

WeiqunZhang:
Most of the changes are fine. But in some places, amrex::Abort is replaced by AMREX_ASSERT(false, "..."), and AMREX_ASSERT is replaced by AMREX_ALWAYS_ASSERT. They are not equivalent.

ax3l:
Yes, I replaced a few `amrex::Abort` because I wanted to have the option to throw there, which the macros provide.

WeiqunZhang:
I think there is a misunderstanding. `amrex::Abort` throws too as long as signal handler is off and throw_exception is on. This is the same as AMREX_ASSERT (if assertion is enabled).
```

    void Abort (const char * msg = nullptr) {
        ...
        AMREX_IF_ON_HOST((Error_host("Abort", msg);))
    }

void
amrex::Error_host (const char* type, const char * msg)
{   
...
    if (system::error_handler) {
        system::error_handler(msg);
    } else if (system::throw_exception) {
        throw RuntimeError(msg);
    } else {
        ...
        ParallelDescriptor::Abort();
    }
}
```

WeiqunZhang:
We should try to improve the documentation. https://github.com/AMReX-Codes/amrex/issues/5235

ax3l:
Thank you, makes total sense.

Also discussed in person: remove the new macro `AMREX_NOEXCEPT_IF_NDEBUG` - there is no exception anyway on GPU, so we expect `noexcept` to be snakeoil in those places / not decrease performance.

Src/Base/AMReX_GpuUtility.H
```diff
@@ -202,9 +202,9 @@ namespace Gpu {
         [[nodiscard]] bool isValid () const noexcept { return m_i < m_n; }
 
 #if !defined(AMREX_USE_GPU)
-        void operator++ () noexcept { ++m_i; }
+        void operator++ () { ++m_i; }
```
ax3l: Well we could leave it for CPU, but I doubt it matters:
```suggestion
        void operator++ () noexcept { ++m_i; }
```

ax3l:
I think this should be cleaned up now @WeiqunZhang :)

Src/Base/AMReX_Array4.H
```diff
@@ -102,7 +102,7 @@ namespace amrex {
         template <class U=T,
                   std::enable_if_t<!std::is_void_v<U>,int> = 0>
         [[nodiscard]] AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
-        U& operator[] (int n) const noexcept {
+        U& operator[] (int n) const {
```
ax3l, line 105: One of the few locations where I would have left the `noexcept` for _non-debug_ modes is here:
```suggestion
        U& operator[] (int n) const {
#if defined(AMREX_DEBUG) || defined(AMREX_BOUND_CHECK)
```

Allows the compiler to mark the path as `nounwind` . But in practice, if the compiler can already see the body and prove it does not unwind on the hot path, the gain is usually zero or very small. Here the code in release builds is: `return p[n*stride];` so the `noexcept` should not add any benefit.

I think there is no loss here but a huge gain to actually catch out-of-bounds as exceptions, if desired.

Src/Particle/AMReX_Particle.H
```diff
@@ -96,7 +96,7 @@ namespace amrex {
 
     template<typename T_SIMD, typename T_Mask>
     AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
-    void make_invalid (T_SIMD & idcpu, T_Mask const & mask) noexcept {
+    void make_invalid (T_SIMD & idcpu, T_Mask const & mask) {
```
ax3l: Here actually the `AMREX_ALWAYS_ASSERT_WITH_MESSAGE` is only for really ill-configured/programmed paths as a defensible construct, not as a production-level assert we want. I would keep the noexcept here as all valid code paths are hot:
```suggestion
    void make_invalid (T_SIMD & idcpu, T_Mask const & mask) noexcept {
```

big config issues like this do not need to be recoverable, they are not normal operational state with runtime parameters that should be throw-y.

Src/Particle/AMReX_Particle.H
```diff
@@ -123,7 +123,7 @@ namespace amrex {
 
     template<typename T_SIMD, typename T_Mask>
     AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
-    void make_valid (T_SIMD & idcpu, T_Mask const & mask) noexcept {
+    void make_valid (T_SIMD & idcpu, T_Mask const & mask) {
```
ax3l: As above:
```suggestion
    void make_valid (T_SIMD & idcpu, T_Mask const & mask) noexcept {
```

ax3l:
I think this should be cleaned up now @WeiqunZhang :)

Final manual self-review completed.

WeiqunZhang (approved):

PR #5224, opened 2026-03-23
Geometry: Better assertion message

WeiqunZhang:


atmyers (approved):

Src/Base/AMReX_Geometry.cpp
```diff
@@ -520,15 +520,29 @@ Geometry::growPeriodicDomain (int ngrow) const noexcept
...
         dx[k] = prob_domain.length(k)/(Real(domain.length(k)));
         inv_dx[k] = 1.0_rt/dx[k];
+        valid_domain = valid_domain && (prob_domain.length(k) > 0) &&
+            (domain.length(k) > 0);
     }
 
+    std::stringstream ss;
+    ss << std::setprecision(std::numeric_limits<double>::max_digits10)
+       << prob_domain << " " << domain << " sizeof Real: "
+       << sizeof(Real) << " sizeof ParticleReal: " << sizeof(ParticleReal);
```
ax3l, line 536: Due to the many string conversions here, I would have probably wrapped this part in `if(!valid_domain)` :D

WeiqunZhang: We reused it for assertion later in the iteration loop.

PR #5225, opened 2026-03-25
FFT: Fix one-d mode in 2D & 3D builds

WeiqunZhang:
* Previously it did not handle the single process case correctly.

* Fix twod_mode=false && oned_mode=true in 3D. Previously, we only support twod_mode=true && oned_mode=true.

* Fix scaling factor for the one-mode.

* Add tests.


WeiqunZhang:
The GitLab CI failure was a machine issue.

ajnonaka (approved):
DIM=2 builds work with oned_mode (serial and parallel)
DIM=3 builds work with twod_mode (serial and parallel)

Issue #4179, opened 2024-10-02
Possible RedistributeGPU optimizations


atmyers:
@AlexanderSinn has suggested a few optimizations to try for RedistributeGPU:

- [ ] 1. When doing the handshake step, we could use new single-sided communication features in MPI-3 to avoid doing MPI_Reduce_Scatter. 
- [ ] 2. When doing the above, we communicate the number of bytes of data as well as the connectivity. 
- [ ] 3. We could reduce the number of `MPI_Send`  calls by using `MPI_Probe` to get the number of bytes. 
- [x] 4. We could use `MPI_isend` instead of `MPI_Send`.

atmyers:
Point 4 was done here: https://github.com/AMReX-Codes/amrex/pull/4257
Points 1 and 2 is implemented here: https://github.com/AMReX-Codes/amrex/pull/5227

PR #5227, opened 2026-03-25
Add an implementation of ParticleCopyPlan::doHandShake that uses one-sided communication from MPI-3

atmyers:
This is one of the possible optimizations raised in Issue #4179. 

This passes all tests locally and on Perlmutter. However, the one-sided version of doHandShake doesn't seem to be a performance win in the tests I ran.

The new handshake is only needed for the global redistribution path. To trigger this path, as a first test I made a particle redistribution benchmark where 1% of the particles jump to a random location in the domain. For this communication pattern, the one-sided handshake is basically the same as reduce/scatter on up to 128 nodes, then slower after that.
   
scaling_comparison_global

Then I thought, the above communication pattern has each rank on average sending some particles to every other rank, basically an all-to-all, which is a bad case for the one-sided version. So I did another test modeled on what we do in load balancing in WarpX. Here, there are 2 boxes per GPU, so on 1024 ranks there would be 2048 boxes. Instead of moving the particles, each step I change the distribution map randomly, so that on average each rank will send particles to 2 other ranks and receive from 2. This kind of sparse distribution pattern should be a good case for the one-sided version. However, even in this case the performance is basically the same in both cases. 

scaling_comparison_regrid

None of these runs included the fix in PR #5260, which would improve the overall redistribute scaling but not effect the handshake time.

Overall I think the regime in which on-sided would be expected to win is relatively small. It would need to trigger the global redistribution path but with a sparse comm pattern. And even in the case I didn't see a win on the test I did.

However, since the one-sided method is off by default I think we should merge this as an option, since the timings are dependent on how good the RMA support is for specific MPI implementations and maybe this will work better on other systems. 

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_ParticleCommunication.cpp
+++ b/Src/Particle/AMReX_ParticleCommunication.cpp
@@ -272,6 +272,7 @@ void ParticleCopyPlan::doHandShake (const ParticleContainerBase& pc,
 #if defined(BL_USE_MPI3)
         doHandShakeOneSided(pc, Snds, Rcvs);
 #else
+        amrex::ignore_unused(pc);
         amrex::Abort("ParticleCopyPlan::doHandShake: particles.do_one_sided_comms=1 requires MPI-3");
 #endif
     }
--- a/Src/Particle/AMReX_ParticleCommunication.cpp
+++ b/Src/Particle/AMReX_ParticleCommunication.cpp
@@ -1,4 +1,5 @@
 #include <AMReX_ParticleCommunication.H>
+#include <AMReX_ParticleContainerBase.H>
 #include <AMReX_ParallelDescriptor.H>
 
 namespace amrex {
@@ -45,7 +46,9 @@ void ParticleCopyPlan::clear ()
     m_rcv_box_levs.clear();
 }
 
-void ParticleCopyPlan::buildMPIStart (const ParticleBufferMap& map, Long psize) // NOLINT(readability-convert-member-functions-to-static)
+void ParticleCopyPlan::buildMPIStart (const ParticleContainerBase& pc,
+                                      const ParticleBufferMap& map,
+                                      Long psize) // NOLINT(readability-convert-member-functions-to-static)
 {
     BL_PROFILE("ParticleCopyPlan::buildMPIStart");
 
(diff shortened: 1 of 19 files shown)


Src/Particle/AMReX_ParticleCommunication.cpp
```diff
@@ -259,11 +260,21 @@ void ParticleCopyPlan::buildMPIFinish (const ParticleBufferMap& map) // NOLINT(r
...
-void ParticleCopyPlan::doHandShake (const Vector<Long>& Snds, Vector<Long>& Rcvs) const // NOLINT(readability-convert-member-functions-to-static)
+void ParticleCopyPlan::doHandShake (const ParticleContainerBase& pc,
+                                    const Vector<Long>& Snds,
+                                    Vector<Long>& Rcvs) const // NOLINT(readability-convert-member-functions-to-static)
 {
     BL_PROFILE("ParticleCopyPlan::doHandShake");
     if (m_local) { doHandShakeLocal(Snds, Rcvs); }
-    else         { doHandShakeGlobal(Snds, Rcvs); }
+    else if (m_do_one_sided_comms) {
+#if defined(BL_USE_MPI3)
```
WeiqunZhang: I think can remove this macro. It was added when MPI-3 was not widely available. But now it's likely unnecessary. Furthermore, we don't even have a CMake build option for MPI3.

atmyers: This is done now, and the one-sided version is tested in one the CI inputs.

Src/Particle/AMReX_ParticleContainerBase.H
```diff
@@ -284,6 +292,13 @@ protected:
     mutable amrex::Vector<int> neighbor_procs;
     mutable ParticleBufferMap m_buffer_map;
 
+#if defined(AMREX_USE_MPI)
+    mutable MPI_Win m_particle_handshake_win = MPI_WIN_NULL;
+    mutable Long* m_particle_handshake_ptr = nullptr;
+    mutable int m_particle_handshake_nprocs = 0;
+    mutable MPI_Comm m_particle_handshake_comm = MPI_COMM_NULL;
+#endif
```
WeiqunZhang, line 317: I would put these into a struct internal to the ParticleContainerBase class and use std::unique_ptr to hold the struct, say ParticleHandshakeWindow. The advantage is then you only to write a destructor for ParticleHandshakeWindow. The ParticleContainerBase can still have default destructor, move construction and move assignment operator. That would be more maintainable.

WeiqunZhang (changes requested):

Src/Particle/AMReX_ParticleContainerBase.H
atmyers: See recent changes

WeiqunZhang (approved):

PR #5252, opened 2026-03-28
added general divergence preserving interpolation of face centered fields

jbbel:
The refinement ratio in each direction may be 2 or 4 (for example, `IntVect(2,2,2)` or `IntVect(2,4,4)`).


WeiqunZhang:
@BenWibking Could you give it a try?

cgilet (approved):
This works for incflo.
I tested with the isentropic convected vortex problem (the problem from PeleLMeX that showed the previous issue in the interpolation). I didn't test RZ, or mixed ref_ratio_vect (because the nodal solver doesn't allow). Also did not check that answers for ref_ratio=2 are the same because I assume that was done already.

WeiqunZhang:
I will fix the conflicts.

PR #5204, opened 2026-03-16
Implement pure-SoA HDF5 particle IO

atmyers:
This closes #5152. Although it's a bit worse than that issue indicated, HDF5 particle IO wasn't implemented for pure SoA at all. This PR does so.

The proposed changes:
- [x] fix a bug or incorrect behavior in AMReX
- [ ] add new capabilities to AMReX
- [ ] changes answers in the test suite to more than roundoff level
- [ ] are likely to significantly affect the results of downstream AMReX users
- [ ] include documentation in the code and/or rst files, if appropriate

PR #5253, opened 2026-03-28
TrackedVector

ax3l:
## Summary

This adds a helper class for synchronizing a pair of host & device vectors (w/o using managed memory).

In BLAST codes, I find this to cause significant boilerplate in physics-focused parts of the code and this little helper class cuts down on lines and mental book-keeping.

Additionally, especially in interactive pyAMReX and ImpactX workflows, this construct is a building block for enabling user-friendly and even multi-simulation-spanning user data (e.g., ImpactX element data) by enabling users to define their inputs even before AMReX was initialized and being able to reuse their input classes across many thousands of simulations, e.g., in optimization loops, even with AMReX device arenas being shut down/recreated in between.

## Additional background

The unit test shows the most common usage patterns & needs.

https://github.com/BLAST-ImpactX/impactx/pull/1368

## Checklist

The proposed changes:
- [x] CPU build: write the optimized fallback path
- [x] GPU build: use AMReX-session agnostic pinned memory for host OR do synchronous copies
- [x] is covered by unit tests

### DIFF
--- a/Docs/sphinx_documentation/source/GPU.rst
+++ b/Docs/sphinx_documentation/source/GPU.rst
@@ -770,8 +770,8 @@ under **Gpu Vectors** above (``The_Managed_Arena()``).  Like :cpp:`Gpu::Buffer`
 it can only be used while AMReX is initialized / a GPU device context exists.
 
 :cpp:`Gpu::TrackedVector` exposes a host ``std::vector`` via ``host()`` /
-``host_const()`` and a device :cpp:`Gpu::DeviceVector` via ``device()`` /
-``device_const()`` (GPU builds only).  Every accessor **automatically
+``host_const()`` and a device :cpp:`Gpu::NonManagedDeviceVector` via
+``device()`` / ``device_const()`` (GPU builds only).  Every accessor **automatically
 synchronizes** the other side when needed: if the host was modified and you
 call ``device()`` or ``device_const()``, a synchronous host-to-device copy
 runs first (and vice versa).  Writable accessors (``host()``, ``device()``)
@@ -802,7 +802,7 @@ dirty state unnecessarily.
      - Only between ``amrex::Initialize/Finalize()``
      - Anytime and cross-session, GPU part only between ``amrex::Initialize/Finalize()``
    * - **Usage**
-     - ``operator[]`` etc.   , explicit ``copyToDeviceAsync`` /
+     - ``operator[]`` etc., explicit ``copyToDeviceAsync`` /
        ``copyToHost``
      - Single ``data()`` like :cpp:`amrex::Vector`
      - Separate ``host()`` / ``device()`` (and ``*_const``)
@@ -855,7 +855,7 @@ A minimal :cpp:`Gpu::Buffer` pattern (host fill, async upload, kernel pointer):
     // mv[i] now accessible on host with updated values
 
 :cpp:`Gpu::TrackedVector` example:
-On GPU builds, you can create this type at any time, even before `amrex::Initialize()`.
+On GPU builds, you can create this type at any time, even before ``amrex::Initialize()``.
 ``amrex::Finalize()`` releases device storage for the vector but
 keeps the host ``std::vector``, so a later ``Initialize()`` session
 can access ``device_const()`` or ``device()`` and the host data is
--- a/Docs/sphinx_documentation/source/GPU.rst
+++ b/Docs/sphinx_documentation/source/GPU.rst
@@ -753,6 +753,144 @@ constructed inside of an MFIter loop with GPU kernels and great care should
 be used when accessing :cpp:`Gpu::ManagedVector` data on GPUs to avoid race
 conditions.
 
+.. _sec:gpu:buffer_managed_tracked:
+
+Gpu::Buffer, Gpu::ManagedVector, and Gpu::TrackedVector
+-------------------------------------------------------
+
+:cpp:`Gpu::Buffer` (``AMReX_GpuBuffer.H``) and :cpp:`Gpu::TrackedVector`
+(``AMReX_TrackedVector.H``) pair a host allocation with a device mirror.
+
+:cpp:`Gpu::Buffer` uses :cpp:`Gpu::PinnedVector` on the host and
+``copyToDeviceAsync()`` / ``copyToHost()`` for transfers.  Use it for
+**frequent, performance-oriented** async copies during a normal GPU run.
+
+:cpp:`Gpu::ManagedVector` is the arena-backed unified-memory vector introduced
+under **Gpu Vectors** above (``The_Managed_Arena()``).  Like :cpp:`Gpu::Buffer`
+it can only be used while AMReX is initialized / a GPU device context exists.
+
+:cpp:`Gpu::TrackedVector` exposes a host ``std::vector`` via ``host()`` /
+``host_const()`` and a device :cpp:`Gpu::DeviceVector` via ``device()`` /
+``device_const()`` (GPU builds only).  Writable accessors mark the mirror
+out-of-date; ``to_device()``/``to_host()`` perform **synchronous** copies when
+needed.  You may populate the host **before** :cpp:`amrex::Initialize()`. Device
+memory is only valid while AMReX is initialized.  On :cpp:`amrex::Finalize()`,
+AMReX clears device storage via ``release_gpu()`` and leaves the host copy for
+reuse, which supports **Python / pyAMReX** and other workflows
+that cross multiple AMReX initialize/finalize cycles.  Use read-only
+``host_const()`` / ``device_const()`` when you are not writing, so the object
+does not flip to a dirty state unnecessarily.
+
+.. _tab:gpu:buffer_managed_tracked:
+
+.. list-table::
+   :widths: 12 28 28 28
+   :header-rows: 1
+
+   * -
+     - :cpp:`Gpu::Buffer`
+     - :cpp:`Gpu::ManagedVector`
+     - :cpp:`Gpu::TrackedVector`
+   * - **Lifetime**
+     - Only between ``amrex::Initialize/Finalize()``
+     - Only between ``amrex::Initialize/Finalize()``
+     - Anytime and cross-session, GPU part only between ``amrex::Initialize/Finalize()``
+   * - **Usage**
+     - ``operator[]`` etc.   , explicit ``copyToDeviceAsync`` /
+       ``copyToHost``
+     - Single ``data()`` like :cpp:`amrex::Vector`
+     - Separate ``host()`` / ``device()`` (and ``*_const``)
+   * - **Synchronization**
+     - explicit
+     - implicit
+     - explicit, but tracks status
+   * - **Performance**
+     - Best: pinned host enables asynchronous transfers
+     - Implicit memory migration can add latency
+     - Synchronous copy adds latency
+   * - **Best for**
+     - hot copy loops inside a run
+     - maximum simplicity
+     - interactive and cross-AMReX session usage, e.g., in pyAMReX for user inputs that do not change often
+
+
+A minimal :cpp:`Gpu::Buffer` pattern (host fill, async upload, kernel pointer):
+
+.. highlight:: c++
+
+::
+
+    amrex::Initialize(argc, argv);
+
+    Gpu::Buffer<int> buf(n);
+    for (int i = 0; i < n; ++i) { buf[i] = i; }
+
+    int* dp = buf.copyToDeviceAsync();
+    // launch kernels using dp, then optionally:
+    buf.copyToHost();
+
+:cpp:`Gpu::ManagedVector` example (unified memory, accessible from both host and device):
+
+.. highlight:: c++
+
+::
+
+    amrex::Initialize(argc, argv);
+
+    Gpu::ManagedVector<int> mv(n);
+    for (int i = 0; i < n; ++i) { mv[i] = i; }
+
+    int* dp = mv.data();
+    amrex::ParallelFor(n, [=] AMREX_GPU_DEVICE (int i) {
+        dp[i] *= 2;  // access on device
+    });
+
+    Gpu::streamSynchronize();
+    // mv[i] now accessible on host with updated values
+
+:cpp:`Gpu::TrackedVector` example:
+On GPU builds, you can create this type at any time, even before `amrex::Initialize()`.
+``amrex::Finalize()`` releases device storage for the vector but
+keeps the host ``std::vector``, so a later ``Initialize()`` can call
+``to_device()`` again to rebuild the device copy.
+
+.. highlight:: c++
+
+::
+
+    // Host data before AMReX init; GPU available after Initialize().
+    amrex::Gpu::TrackedVector<int> cross_session;
+    cross_session.host() = {7, 8, 9};
+
+    // ... a lot of other interactive user code, e.g., to set up
+    // complex input data, optimization libraries or ML frameworks
+    // in multi-simulation workflows ...
+
+    amrex::Initialize(argc, argv);
+    {
+        cross_session.to_device();
+        // Host and device match; use host_const() / device_const() for reads.
+    }
+    amrex::Finalize();
+
+    // cross_session.device() is not available now and will throw,
+    // but you can keep using cross_session.host() / .host_const()
+
+    amrex::Initialize(argc, argv);
+    {
+        cross_session.to_device();
+        // Device buffer is re-created; kernels may read via
+        // device_const().data() or write via device().data()
+    }
+    amrex::Finalize();
+
+Optional: Call ``release_gpu()`` when you need to free device memory while
+keeping the host ``std::vector`` for later (unless already released,
+``amrex::Finalize()`` clears device storage registered for the object).
+
+Generally, after device kernels, call :cpp:`Gpu::streamSynchronize()`
+(or equivalent ordering) before relying on host data, as for any other device work.
+
 MultiFab Reductions
 -------------------
 
--- a/Docs/sphinx_documentation/source/GPU.rst
+++ b/Docs/sphinx_documentation/source/GPU.rst
@@ -771,15 +771,21 @@ it can only be used while AMReX is initialized / a GPU device context exists.
 
 :cpp:`Gpu::TrackedVector` exposes a host ``std::vector`` via ``host()`` /
 ``host_const()`` and a device :cpp:`Gpu::DeviceVector` via ``device()`` /
-``device_const()`` (GPU builds only).  Writable accessors mark the mirror
-out-of-date; ``to_device()``/``to_host()`` perform **synchronous** copies when
-needed.  You may populate the host **before** :cpp:`amrex::Initialize()`. Device
-memory is only valid while AMReX is initialized.  On :cpp:`amrex::Finalize()`,
-AMReX clears device storage via ``release_gpu()`` and leaves the host copy for
-reuse, which supports **Python / pyAMReX** and other workflows
-that cross multiple AMReX initialize/finalize cycles.  Use read-only
-``host_const()`` / ``device_const()`` when you are not writing, so the object
-does not flip to a dirty state unnecessarily.
+``device_const()`` (GPU builds only).  Every accessor **automatically
+synchronizes** the other side when needed: if the host was modified and you
+call ``device()`` or ``device_const()``, a synchronous host-to-device copy
+runs first (and vice versa).  Writable accessors (``host()``, ``device()``)
+additionally mark the returned side as dirty so that the next access to the
+opposite side triggers a copy back.  Read-only accessors
+(``host_const()``, ``device_const()``) leave the status as ``up_to_date``
+after syncing.  You may populate the host **before**
+:cpp:`amrex::Initialize()`. Device memory is only valid while AMReX is
+initialized.  On :cpp:`amrex::Finalize()`, AMReX clears device storage via
+``release_gpu()`` and leaves the host copy for reuse, which supports
+**Python / pyAMReX** and other workflows that cross multiple AMReX
+initialize/finalize cycles.  Use read-only ``host_const()`` /
+``device_const()`` when you are not writing, so the object does not flip to a
+dirty state unnecessarily.
 
 .. _tab:gpu:buffer_managed_tracked:
 
@@ -803,7 +809,7 @@ does not flip to a dirty state unnecessarily.
    * - **Synchronization**
      - explicit
      - implicit
-     - explicit, but tracks status
+     - automatic on access, tracks status
    * - **Performance**
      - Best: pinned host enables asynchronous transfers
      - Implicit memory migration can add latency
@@ -851,8 +857,9 @@ A minimal :cpp:`Gpu::Buffer` pattern (host fill, async upload, kernel pointer):
 :cpp:`Gpu::TrackedVector` example:
 On GPU builds, you can create this type at any time, even before `amrex::Initialize()`.
 ``amrex::Finalize()`` releases device storage for the vector but
-keeps the host ``std::vector``, so a later ``Initialize()`` can call
-``to_device()`` again to rebuild the device copy.
+keeps the host ``std::vector``, so a later ``Initialize()`` session
+can access ``device_const()`` or ``device()`` and the host data is
+automatically copied to the device.
 
 .. highlight:: c++
 
@@ -868,8 +875,9 @@ keeps the host ``std::vector``, so a later ``Initialize()`` can call
 
     amrex::Initialize(argc, argv);
     {
-        cross_session.to_device();
-        // Host and device match; use host_const() / device_const() for reads.
+        // device access auto-syncs host data to device.
+        int const* dp = cross_session.device_const().data();
+        // use dp in kernels ...
     }
     amrex::Finalize();
 
@@ -878,9 +886,10 @@ keeps the host ``std::vector``, so a later ``Initialize()`` can call
 
     amrex::Initialize(argc, argv);
     {
-        cross_session.to_device();
-        // Device buffer is re-created; kernels may read via
-        // device_const().data() or write via device().data()
+        // Device buffer is re-created automatically on access.
+        int const* dp = cross_session.device_const().data();
+        // kernels may read via device_const().data()
+        // or write via device().data()
     }
     amrex::Finalize();
 
--- a/Src/Base/AMReX_TrackedVector.H
+++ b/Src/Base/AMReX_TrackedVector.H
@@ -41,106 +41,74 @@ namespace amrex::Gpu
         using value_type      = T;
         using size_type       = std::size_t;
 
-    private:
-        void register_finalize () {
-#ifdef AMREX_USE_GPU
-            if (*m_finalize_registered) { return; }
-            *m_finalize_registered = true;
-
-            std::weak_ptr<std::vector<T>> weak_host = m_host;
-            std::weak_ptr<amrex::Gpu::DeviceVector<T>> weak_device = m_device;
-            std::weak_ptr<Status> weak_status = m_status;
-            std::weak_ptr<bool> weak_registered = m_finalize_registered;
-
-            amrex::ExecOnFinalize([weak_host, weak_device, weak_status, weak_registered]() {
-                // see: release_gpu()
-                auto host   = weak_host.lock();
-                auto device = weak_device.lock();
-                auto status = weak_status.lock();
-                if (host && device && status
-                    && *status == Status::device_dirty
-                    && !device->empty())
-                {
-                    // see: to_host()
-                    host->resize(device->size());
-                    amrex::Gpu::copy(amrex::Gpu::deviceToHost,
-                        device->begin(), device->end(), host->begin());
-                }
-                if (device) {
-                    device->clear();
-                    device->shrink_to_fit();
-                }
-                if (status) {
-                    *status = Status::host_dirty;
-                }
-                if (auto reg = weak_registered.lock()) {
-                    *reg = false;
-                }
-            });
-#endif
-        }
-    public:
+        enum class Status {
+            up_to_date,    //!< host and device data are in sync
+            device_dirty,  //!< host data needs an update
+            host_dirty     //!< device data needs an update
+        };
 
         TrackedVector () = default;
 
         explicit TrackedVector (size_type a_size)
-            : m_host(std::make_shared<std::vector<T>>(a_size))
         {
+            m_data->host.resize(a_size);
 #ifdef AMREX_USE_GPU
-            *m_status = Status::host_dirty;
+            m_data->status = Status::host_dirty;
 #endif
         }
 
         TrackedVector (size_type a_size, value_type const & a_value)
-            : m_host(std::make_shared<std::vector<T>>(a_size, a_value))
         {
+            m_data->host.assign(a_size, a_value);
 #ifdef AMREX_USE_GPU
-            *m_status = Status::host_dirty;
+            m_data->status = Status::host_dirty;
 #endif
         }
 
         TrackedVector (std::initializer_list<T> a_initializer_list)
-            : m_host(std::make_shared<std::vector<T>>(a_initializer_list))
         {
+            m_data->host = a_initializer_list;
 #ifdef AMREX_USE_GPU
-            *m_status = Status::host_dirty;
+            m_data->status = Status::host_dirty;
 #endif
         }
 
         TrackedVector (std::vector<T> a_vector)
-            : m_host(std::make_shared<std::vector<T>>(std::move(a_vector)))
         {
+            m_data->host = std::move(a_vector);
 #ifdef AMREX_USE_GPU
-            *m_status = Status::host_dirty;
+            m_data->status = Status::host_dirty;
 #endif
         }
 
         TrackedVector (TrackedVector const & a_vector)
-            : m_host(std::make_shared<std::vector<T>>(*a_vector.m_host))
+            : m_data(std::make_shared<Data>())
         {
-            *m_status = *a_vector.m_status;
+            m_data->status = a_vector.m_data->status;
+            m_data->host = a_vector.m_data->host;
 #ifdef AMREX_USE_GPU
-            *m_device = *a_vector.m_device;
+            m_data->device = a_vector.m_data->device;
+            if (amrex::Initialized()) {
+                register_finalize();
+            }
 #endif
         }
 
         /** Swap data of this and a_vector */
         TrackedVector (TrackedVector && a_vector) noexcept
         {
-            std::swap(m_status, a_vector.m_status);
-            std::swap(m_host, a_vector.m_host);
-#ifdef AMREX_USE_GPU
-            std::swap(m_device, a_vector.m_device);
-            std::swap(m_finalize_registered, a_vector.m_finalize_registered);
-#endif
+            std::swap(m_data, a_vector.m_data);
         }
 
         TrackedVector& operator= (TrackedVector const & a_vector) {
             if (this != &a_vector) {
-                *m_status = *a_vector.m_status;
-                *m_host = *a_vector.m_host;
+                m_data->status = a_vector.m_data->status;
+                m_data->host = a_vector.m_data->host;
 #ifdef AMREX_USE_GPU
-                *m_device = *a_vector.m_device;
+                m_data->device = a_vector.m_data->device;
+                if (amrex::Initialized()) {
+                    register_finalize();
+                }
 #endif
             }
             return *this;
@@ -149,25 +117,14 @@ namespace amrex::Gpu
         /** Swap the data of this and a_vector */
         TrackedVector& operator= (TrackedVector && a_vector) noexcept {
             if (this != &a_vector) {
-                std::swap(m_host, a_vector.m_host);
-                std::swap(m_status, a_vector.m_status);
-#ifdef AMREX_USE_GPU
-                std::swap(m_device, a_vector.m_device);
-                std::swap(m_finalize_registered, a_vector.m_finalize_registered);
-#endif
+                std::swap(m_data, a_vector.m_data);
             }
             return *this;
         }
 
         ~TrackedVector () = default;
 
-        enum class Status {
-            up_to_date,    //!< host and device data are in sync
-            device_dirty,  //!< host data needs an update
-            host_dirty     //!< device data needs an update
-        };
-
-        [[nodiscard]] Status status () const { return *m_status; }
+        [[nodiscard]] Status status () const { return m_data->status; }
 
         /** Return writable host data
          *
@@ -177,12 +134,12 @@ namespace amrex::Gpu
         [[nodiscard]] std::vector<T> &
         host () {
 #ifdef AMREX_USE_GPU
-            if (*m_status == Status::device_dirty) {
+            if (m_data->status == Status::device_dirty) {
                 to_host();
             }
-            *m_status = Status::host_dirty;
+            m_data->status = Status::host_dirty;
 #endif
-            return *m_host;
+            return m_data->host;
         }
 
         /** Return read-only host data
@@ -192,11 +149,11 @@ namespace amrex::Gpu
         [[nodiscard]] std::vector<T> const &
         host_const () const {
 #ifdef AMREX_USE_GPU
-            if (*m_status == Status::device_dirty) {
+            if (m_data->status == Status::device_dirty) {
                 to_host();
             }
 #endif
-            return *m_host;
+            return m_data->host;
         }
 
 #ifdef AMREX_USE_GPU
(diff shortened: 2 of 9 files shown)


WeiqunZhang:
Note that we already have a similar data container called Gpu::Buffer.

Src/Base/AMReX_HostDeviceVector.H
```diff
@@ -0,0 +1,106 @@
...
+     * This object is primarily for input handling, allowing to initialize
+     * data even before AMReX was initialized and enabling workflows crossing
+     * AMReX init/finalize cycles.
+     *
+     * Usage contract: when manipulating the host data, you must call mark_dirty().
+     */
+    template <class T>
+    struct HostDeviceVector
+    {
+        std::vector<T> host;  // TODO: make pinned w/o using the PinnedArena!
```
ax3l: TODO: make pinned w/o using the PinnedArena!

We do not want to use the AMReX PinnedArena because that would clear out host data on AMReX finalize. But we want an object that can truly be used before, after and across AMReX init/finalize cycles.

ax3l:
> Note that we already have a similar data container called Gpu::Buffer.

Ah that is cool, I forgot about that one.

After checking again `Gpu::Buffer`, there are subtly different needs in my implementation: In particular, I am not bound to AMReX being initialized to start writing (host) data into it, which is the driving need I have in pyAMReX/ImpactX to have more user-friendly initialization and lifetimes of complex simulation data.

We could extend/rewrite `Gpu::Buffer` or keep this separate with a distinct enough name & docs.

WeiqunZhang:
We could make Gpu::Buffer template on container type.

Src/Base/AMReX_HostDeviceVector.H
```diff
@@ -0,0 +1,106 @@
...
+{
+    /** Base for element data with dynamic host vectors mirrored lazily to GPU.
+     *
+     * Provides versioning, sync helpers, and data release APIs.
+     *
+     * This object is primarily for input handling, allowing to initialize
+     * data even before AMReX was initialized and enabling workflows crossing
+     * AMReX init/finalize cycles.
+     *
+     * Usage contract: when manipulating the host data, you must call mark_dirty().
```
ax3l:
```suggestion
     * Usage contract:
     *   - When manipulating the host data, you must call `mark_dirty()`.
     *   - Device data can only be allocated after AMReX initialize.
     *   - You MUST call `release_gpu()` before AMReX finalize.
```

ax3l:
Yes, good idea. Let's finish the tests and impl. for this one and then we can investigate if/how we merge them?

Src/Base/AMReX_HostDeviceVector.H
```diff
@@ -0,0 +1,109 @@
...
+        enum class Status {
+            up_to_date,
+            host_outdated,
+            device_outdated
+        };
+
+        //! Incremented when host data changes; triggers re-upload
+        int host_version = 1;
+        //! Tracks last-uploaded version; stale if < host_version
+        int device_version = 0;
```
AlexanderSinn: I think it is sufficient to store a single instance of the Status enum and not have the version numbers, unless you have external code reading the version number. 
host_dirty(): status = device_outdated;
device_dirty(): status = host_outdated;
needs_copy(): return status;
ensure_same_async(): copy; status = up_to_date;
This also removes the risk of the version numbers overflowing.

ax3l: That is true. The int was mostly for curiosity how many version we had, but this is safer.

Tests/Base/HostDeviceVector/main.cpp
```diff
@@ -0,0 +1,154 @@
...
+    v.host_dirty();
+    v.ensure_same_async();
+    Gpu::streamSynchronize();
+    verify_host_device_match(v);
+    AMREX_ALWAYS_ASSERT(v.needs_copy() == Status::up_to_date);
+
+    v.release_gpu();
+    AMREX_ALWAYS_ASSERT(v.device.empty());
+    AMREX_ALWAYS_ASSERT(v.device_version == 0);
+    AMREX_ALWAYS_ASSERT(v.host.size() == 3u);
```
ax3l:
```suggestion
    AMREX_ALWAYS_ASSERT(v.host.size() == 3U);
```

Tests/Base/HostDeviceVector/main.cpp
```diff
@@ -0,0 +1,154 @@
...
+        Gpu::streamSynchronize();
+        verify_host_device_match(cross_session);
+        AMREX_ALWAYS_ASSERT(cross_session.host[0] == 7);
+        cross_session.release_gpu();
+    }
+    amrex::Finalize();
+
+    amrex::Initialize(argc, argv);
+    {
+        AMREX_ALWAYS_ASSERT(cross_session.host.size() == 3u);
```
ax3l:
```suggestion
        AMREX_ALWAYS_ASSERT(cross_session.host.size() == 3U);
```

Tests/Base/HostDeviceVector/main.cpp
```diff
@@ -0,0 +1,154 @@
...
+}
+
+} // namespace
+
+int main (int argc, char* argv[])
+{
+    HDVec cross_session;
+    cross_session.host = {7, 8, 9};
+    cross_session.host_dirty();
+
```
ax3l:
```suggestion

#ifdef AMREX_USE_MPI
    MPI_Init(&argc, &argv);
#endif

```

Tests/Base/HostDeviceVector/main.cpp
```diff
@@ -0,0 +1,154 @@
...
+        cross_session.host[1] = 99;
+        cross_session.host_dirty();
+        cross_session.ensure_same_async();
+        Gpu::streamSynchronize();
+        verify_host_device_match(cross_session);
+        AMREX_ALWAYS_ASSERT(cross_session.host[1] == 99);
+        cross_session.release_gpu();
+    }
+    amrex::Finalize();
+
```
ax3l:
```suggestion

#ifdef AMREX_USE_MPI
    MPI_Finalize();
#endif

```

Src/Base/AMReX_HostDeviceVector.H
ax3l: improved :)

Src/Base/AMReX_HostDeviceVector.H
ax3l: @WeiqunZhang I am not 100% sure from reading the implementation: is an `amrex::PinnedVector` its own memory pool in AMReX that gets freed during AMReX finalize? Or does it always survive (in CPU and GPU builds)?
Technically it is either system-allocated and/or via HIP/CUDA/SYCL host calls that are independent of any GPU context), but I do not get if AMReX still owns the memory heap for pinned arena (or just tracks how much is stored there).

ax3l: Hm, cursor says for GPU builds it is a pre-allocated and managed pinned memory heap by AMReX and freed. That indeed does not support the need we have here (see test variable `cross_session`).

In that case I should to be safe either a) use a pinned allocator with std::vector, eg., from thrust or b) not to async copies...

For b), to be fully honest, the cases here would likely only copy very few times (on init, when interactively a user adds new data) so one could consider doing sync copies and avoid fragmentation from having some potentially small pinned memory pages randomly allocated.

That said, I have not encountered this as an issue in practice yet, so we could just support a).

ax3l:
> The following tests FAILED:
	 93 - Particles_CheckpointRestartDualGridHDF5SOA_2d (Failed)
	 94 - Particles_CheckpointRestartDualGridHDF5SOA_3d (Failed)

Only on H100. Unrelated

Src/Base/AMReX_HostDeviceVector.H
ax3l: Hm, the only safe way to do a) without any context/runtime is using POSIX `mlock`/`munlock`.

Maybe I just do b) for now... the idea is not to use this as a hot-loop host->device updater but as an initial and on-demand option changer.

That said, most people will use it as tight block of:
```
async copy
async copy
amrex::Gpu::streamSynchronize();
```
which has a very low risk even with unpinned memory (yes it is technically UB, but I carried bugs like this for years with no issues).

The important feature is pre-amrex allocation and lazy GPU allocation once we have a context/session.

ax3l:
Still unrelated: only on H100:

> The following tests FAILED:
	 93 - Particles_CheckpointRestartDualGridHDF5SOA_2d (Failed)
	 94 - Particles_CheckpointRestartDualGridHDF5SOA_3d (Failed)
	 96 - Particles_CheckpointRestartDualGridSOA_2d (Failed)
	 97 - Particles_CheckpointRestartDualGridSOA_3d (Failed)

From #5204?

Src/Base/AMReX_HostDeviceVector.H
ax3l: Haha, https://docs.nvidia.com/cuda/cuda-programming-guide/02-basics/asynchronous-execution.html#launching-memory-transfers-in-cuda-streams
> cudaMemcpyAsync() will function correctly if host memory which is not pinned and page-locked is used, but it will revert to a synchronous behavior which will not overlap with other work ...

Src/Base/AMReX_HostDeviceVector.H
```diff
@@ -0,0 +1,98 @@
...
+         * This enables use outside of and across AMReX init/finalize cycles.
+         */
+        void release_gpu ()
+        {
+#ifdef AMREX_USE_GPU
+            device.clear();
+            device.shrink_to_fit();
+            status = Status::device_outdated;
+#endif
+        }
```
AlexanderSinn: This should do device to host copy first if needed

Docs/sphinx_documentation/source/GPU.rst
```diff
@@ -753,6 +753,67 @@ constructed inside of an MFIter loop with GPU kernels and great care should
...
+    Real* dp = v.device.data();
+    // use dp in GPU kernels (read)
+
+    // if you wrote on device, you can refresh the host after device writes:
+    v.device_dirty();
+    v.ensure_same();
+    // read v.host on the CPU
+
+After device kernels, call :cpp:`Gpu::streamSynchronize()` (or equivalent
+ordering) before relying on host data, as for any other device work.
```
AlexanderSinn: In this example, the API seems cumbersome to use. If `v.host/device_dirty(); v.ensure_same();` are always called in succession, then it is better to replace them with single copyTo... functions like in Gpu::Buffer. I think what you really want are ensureHostUpToDate() and ensureDeviceUpToDate() functions such that, e.g., if multiple GPU-only functions are modifying the data in succession, no copy needs to be performed.
Finally, these functions should be combined with functions to retrieve the host/device vectors to make it impossible to use incorrectly:
```
ReadHost:  ensureHostUpToDate(); return const host vector;
ReadDevice: ensureDeviceUpToDate(); return const device vector;
ReadWriteHost:  ensureHostUpToDate(); host_dirty(); return host vector;
ReadWriteDevice:  ensureDeviceUpToDate(); device_dirty(); return device vector;
```
Not sure if these should be included; it would be for fully overwriting the old data:
```
WriteHost: host_dirty(); return host vector;
WriteDevice: device_dirty(); return device vector;
```

Src/Base/AMReX_HostDeviceVector.H
```diff
@@ -0,0 +1,98 @@
...
+     * For AMReX CPU builds, the host and device members point to the same memory
+     * and status will always be up to date.
+     *
+     * Usage contract:
+     *   - When manipulating the host data, you must call `mark_dirty()`.
+     *   - Device data can only be allocated after AMReX initialize.
+     *   - You MUST call `release_gpu()` before AMReX finalize.
+     */
+    template <class T>
+    struct HostDeviceVector
```
AlexanderSinn: Since this data structure is not going to use Pinned memory and thus will not have optimal performance, I would give this a more specialized name than HostDeviceVector. A true `amrex::Gpu::HostDeviceVector` should have (almost) the same interface as `amrex::Gpu::DeviceVector` and `amrex::Gpu::HostVector`.

ax3l: I think so too, maybe TrackedVector or MarkedVector or so?

Docs/sphinx_documentation/source/GPU.rst
ax3l: There is one important detail here: we want to lazily update the GPU.

In practice, that means a user does a whole bunch of host changes even before we have a GPU context available and we also manipulate again after finalize of amrex. A host dirty does NOT mean we want or even can sync to device.

Maybe the example needs an update, it is misleading.

Src/Base/AMReX_HostDeviceVector.H
ax3l: I wonder if we can also call this implicitly in AMReX finalize. I would love if the user can call this but does not have to keep track. There is some callback mechanics for this in AMReX already.

The challenge will be that the HostDevice variable itself can be user deallocated as well before AMReX finalize is called.

Docs/sphinx_documentation/source/GPU.rst
ax3l: I was wondering if implicitly tracking with const/non-const accessors is the right way or leads to accidentally marked dirty states...

Maybe `host_data()`/`host_cdata()` are safer accessors instead of a overload.

Even with those though, the user can safe pointers to this and we technically mark dirty too early. But we can document that they should not.

AlexanderSinn: With the current API if a function gets a vector as an input to modify its state on the CPU but does not know what was called before it, it has no choice but to call `ensure_same` (and later `host_dirty`). In case the previous function was also modifying the CPU state, this will result in an unnecessary CPU-to-GPU transfer. If instead it called `ensureHostUpToDate` or `ReadWriteHost` no transfer needs to happen in that case.

AlexanderSinn: Because the D2H/H2D sync also involves a resize, it is not safe for users to store a pointer to the data. I don't think overloading const / vs. non-const accessors is a good idea, as it would require every member variable to be declared mutable, because a D2H/H2D sync is still sometimes required for just reading data.

ax3l: Sorry, I meant to write: protect the members, expose via `host()`/`host_const()` as a ref vector

Src/Base/AMReX_HostDeviceVector.H
ax3l: Ah, I could store the data members as shared part and expose to AMReX as weak ref

ax3l:
~I'll push a major refactor later, have local changes.~ done, still need to update docs later.

Docs/sphinx_documentation/source/GPU.rst
ax3l: @AlexanderSinn I did a major rewrite, let me know what you think :)

Docs/sphinx_documentation/source/GPU.rst
```diff
@@ -753,6 +753,67 @@ constructed inside of an MFIter loop with GPU kernels and great care should
 be used when accessing :cpp:`Gpu::ManagedVector` data on GPUs to avoid race
 conditions.
 
+.. _sec:gpu:buffer_hostdevice:
+
+Gpu::Buffer and Gpu::TrackedVector
```
ax3l: Todo: extend docs by adding comparison also to Vectors on the managed arena

Docs/sphinx_documentation/source/GPU.rst
AlexanderSinn: Still the same issue with `ensure_same()`. If a function calls `ensure_same()` at the beginning before accessing the data on the GPU (which I believe is the correct way to use this), one of three things is true:
* The data is already synchronized, and no copy is done.
* The data was always previously modified by the CPU. A host-to-device copy will always be performed. In this case the internal tracking of the status is somewhat unnecessary.
* It is possible that the data was previously modified by the GPU. `ensure_same()` will then copy  the data to the CPU unnecessarily.

The third point can be fixed by having functions that sync only one way, as I described above.

Src/Base/AMReX_StdPinnedAllocator.H
```diff
@@ -0,0 +1,128 @@
...
+/**
+ * \brief Host page-lock via `std::malloc` + `mlock` (POSIX) or `std::malloc`
+ *        alone (Windows).
+ *
+ * Does not call CUDA, HIP, SYCL, or AMReX GPU/device APIs. Safe to use before
+ * `amrex::Initialize()` and without any GPU context. Matches the non-GPU
+ * pinned path in `Arena::allocate_system` when host memory is locked with
+ * `mlock` (see `AMReX_Arena.cpp`).
+ */
+[[nodiscard]] inline void* std_pinned_alloc_bytes (std::size_t nbytes)
```
ax3l: Remove this file

Src/Base/AMReX_TrackedVector.H
```diff
@@ -0,0 +1,233 @@
...
+        }
+
+        TrackedVector (TrackedVector const & a_vector)
+            : m_host(a_vector.m_host)
+        {
+            register_finalize();
+            host_dirty();
+        }
+
+        TrackedVector (TrackedVector && a_vector) noexcept
```
ax3l: Move construction and assignments can be defaulted

Src/Base/AMReX_TrackedVector.H
```diff
@@ -0,0 +1,233 @@
...
+        }
+
+        TrackedVector (std::initializer_list<T> a_initializer_list)
+            : m_host(a_initializer_list)
+        {
+            register_finalize();
+            host_dirty();
+        }
+
+        TrackedVector (TrackedVector const & a_vector)
```
ax3l, line 77: Check again what to do for copy construction and assignment, handle device if amrex is initialized

Docs/sphinx_documentation/source/GPU.rst
ax3l: I am ok to add one way syncs, which themselves do not need status tracking. I think the auto tracking will be useful for complex nested edits, where we want the last one to win.

Src/Base/AMReX_HostDeviceVector.H
ax3l: done.

Docs/sphinx_documentation/source/GPU.rst
ax3l: I added the explicit APIs, safer for a lot of use cases.

Src/Base/AMReX_StdPinnedAllocator.H
ax3l: done.

Src/Base/AMReX_TrackedVector.H
ax3l: fixed.

Tests/Base/TrackedVector/main.cpp
```diff
@@ -0,0 +1,316 @@
...
+#else
+    AMREX_ALWAYS_ASSERT(v.status() == Status::up_to_date);
+#endif
+    AMREX_ALWAYS_ASSERT(v.host_const().size() == 3);
+
+    v.to_device();
+    AMREX_ALWAYS_ASSERT(v.status() == Status::up_to_date);
+
+    // potential write (device): must mark dirty
+    [[maybe_unused]] auto* dp = v.device().data();
```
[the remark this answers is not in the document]
ax3l: well done, sherlock ;-) It is nearly like `[[maybe_unused]]` told you.

Tests/Base/TrackedVector/main.cpp
```diff
@@ -0,0 +1,316 @@
...
+    v.ensure_same();
+    AMREX_ALWAYS_ASSERT(v.status() == Status::up_to_date);
+
+    // read-only (host): must not mark dirty
+    auto first = v.host_const().at(2);
+    AMREX_ALWAYS_ASSERT(first == 42);
+    AMREX_ALWAYS_ASSERT(v.status() == Status::up_to_date);
+
+    // read-only (device): must not mark dirty
+    [[maybe_unused]] auto const* dcp = v.device_const().data();
```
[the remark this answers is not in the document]
ax3l: yes indeed, see `[[maybe_unused]]`

Src/Base/AMReX_TrackedVector.H
ax3l: They cannot, but I fixed it so even the moved-from object is safe.

ax3l:
@WeiqunZhang @atmyers @AlexanderSinn I am very happy with this design now, thank you for your feedback!

The easiest way to review this is first look at the user-facing docs, then the test cases I added, then the actual class.

Let me know what you think :)

Src/Base/AMReX_TrackedVector.H
```diff
@@ -0,0 +1,294 @@
...
+        /** Conditionally synchronize host data to device
+         *
+         * If either the host or device side are dirty, this will
+         * copy the host data over the device data.
+         *
+         * @param[in] force  If true, force a copy even if the status is up to date.
+         */
+        void to_device (bool force=false) {
+#ifdef AMREX_USE_GPU
+            if (status() != Status::up_to_date || force) {
```
AlexanderSinn:
```suggestion
            if (status() == Status::host_dirty || force) {
```
If the status is device_dirty, then calling to_device would just overwrite the recent changes to the device side, but not when compiling for CPU and the vectors are shared.

Src/Base/AMReX_TrackedVector.H
```diff
@@ -0,0 +1,294 @@
...
+            register_finalize();
+            host_dirty();
+        }
+
+        TrackedVector (std::initializer_list<T> a_initializer_list)
+            : m_host(a_initializer_list)
+        {
+            register_finalize();
+            host_dirty();
+        }
```
ax3l: I need one more constructor for aggregate initialization in ImpactX:

```suggestion
        }

        TrackedVector (std::vector<T> a_vector)
            : m_host(std::move(a_vector))
        {
            register_finalize();
            host_dirty();
        }
```
Note: must not be marked `explicit`.

Src/Base/AMReX_TrackedVector.H
ax3l: I intentionally choose that: if you ask `to_device()` and the device is not what the host has (for any reason) it gets overwritten.

When compiling for CPU, host and device are in sync by definition/API contract, using the exact same memory. The status is never not `Status::up_to_date` for a CPU build.

Src/Base/AMReX_TrackedVector.H
ax3l: fixed & test covered

ax3l:
Frank issue, CUDA GPU was busy:
```
builds/amrex/amrex/Src/Base/AMReX_GpuDevice.cpp line 331: CUDA-capable device(s) is/are busy or unavailable !!!
SIGABRT
```

Other GPUs and run before passed, so this is good (also, local GPU on my laptop passed for CUDA).

Docs/sphinx_documentation/source/GPU.rst
```diff
@@ -753,6 +753,144 @@ constructed inside of an MFIter loop with GPU kernels and great care should
 be used when accessing :cpp:`Gpu::ManagedVector` data on GPUs to avoid race
 conditions.
 
+.. _sec:gpu:buffer_tracked_managed:
+
+Gpu::Buffer, Gpu::TrackedVector, and Gpu::ManagedVector
```
ax3l: Order changed in the narrative to:
```suggestion
Gpu::Buffer, Gpu::ManagedVector, and Gpu::TrackedVector
```

Docs/sphinx_documentation/source/GPU.rst
```diff
@@ -753,6 +753,144 @@ constructed inside of an MFIter loop with GPU kernels and great care should
...
+-------------------------------------------------------
+
+:cpp:`Gpu::Buffer` (``AMReX_GpuBuffer.H``) and :cpp:`Gpu::TrackedVector`
+(``AMReX_TrackedVector.H``) pair a host allocation with a device mirror.
+
+:cpp:`Gpu::Buffer` uses :cpp:`Gpu::PinnedVector` on the host and
+``copyToDeviceAsync()`` / ``copyToHost()`` for transfers.  Use it for
+**frequent, performance-oriented** async copies during a normal GPU run.
+
+:cpp:`Gpu::TrackedVector` exposes a host ``std::vector`` via ``host()`` /
```
ax3l, line 772: Order in table:
```suggestion
:cpp:`Gpu::ManagedVector` is the arena-backed unified-memory vector introduced
under **Gpu Vectors** above (``The_Managed_Arena()``).  Like :cpp:`Gpu::Buffer`
it can only be used while AMReX is initialized / a GPU device context exists.

:cpp:`Gpu::TrackedVector` exposes a host ``std::vector`` via ``host()`` /
```

Docs/sphinx_documentation/source/GPU.rst
```diff
@@ -753,6 +753,144 @@ constructed inside of an MFIter loop with GPU kernels and great care should
...
+memory is only valid while AMReX is initialized.  On :cpp:`amrex::Finalize()`,
+AMReX clears device storage via ``release_gpu()`` and leaves the host copy for
+reuse, which supports **Python / pyAMReX** and other workflows
+that cross multiple AMReX initialize/finalize cycles.  Use read-only
+``host_const()`` / ``device_const()`` when you are not writing, so the object
+does not flip to a dirty state unnecessarily.
+
+:cpp:`Gpu::ManagedVector` is the arena-backed unified-memory vector introduced
+under **Gpu Vectors** above (``The_Managed_Arena()``).  Like :cpp:`Gpu::Buffer`
+it can only be used while AMReX is initialized / a GPU device context exists.
```
ax3l: Order in table
```suggestion
```

Docs/sphinx_documentation/source/GPU.rst
```diff
@@ -753,6 +753,144 @@ constructed inside of an MFIter loop with GPU kernels and great care should
 be used when accessing :cpp:`Gpu::ManagedVector` data on GPUs to avoid race
 conditions.
 
+.. _sec:gpu:buffer_tracked_managed:
```
```suggestion
.. _sec:gpu:buffer_managed_tracked:
```

Docs/sphinx_documentation/source/GPU.rst
```diff
@@ -753,6 +753,144 @@ constructed inside of an MFIter loop with GPU kernels and great care should
...
+out-of-date; ``to_device()``/``to_host()`` perform **synchronous** copies when
+needed.  You may populate the host **before** :cpp:`amrex::Initialize()`. Device
+memory is only valid while AMReX is initialized.  On :cpp:`amrex::Finalize()`,
+AMReX clears device storage via ``release_gpu()`` and leaves the host copy for
+reuse, which supports **Python / pyAMReX** and other workflows
+that cross multiple AMReX initialize/finalize cycles.  Use read-only
+``host_const()`` / ``device_const()`` when you are not writing, so the object
+does not flip to a dirty state unnecessarily.
+
+.. _tab:gpu:buffer_tracked_managed:
```
```suggestion
.. _tab:gpu:buffer_managed_tracked:
```

Src/Base/AMReX_TrackedVector.H
ax3l: @AlexanderSinn let me know if this resolves your comment :)

AlexanderSinn: No. The reason the state is in device_dirty and one would want to do an H2D copy anyway is that the vector was somehow used incorrectly and a call to host_dirty is missing. The fix for this should be to add a call to host_dirty and not to force-overwrite the device. Nevertheless, the force overwrite would still be possible with `to_device(true)`.

With the current API, it is easy to use it wrong and to get different results depending on if it is compiled for GPU or CPU (untested):
```C++
amrex::Gpu::TrackedVector<int> tv(3, 1);
tv.to_device();
auto dp = tv.device().data();
amrex::ParallelFor(3, [=] AMREX_GPU_DEVICE (int i) {
    dp[i] += 1;
});
tv.to_device(); // ups, called to_device when tv is in the incorrect state.
// To use the API correctly we should not need to know the state of tv since it tracks it itself.
auto dp2 = tv.device().data();
amrex::ParallelFor(3, [=] AMREX_GPU_DEVICE (int i) {
    dp2[i] += 1;
});
tv.to_host();
std::cout << tv.host_const()[0] << std::endl; // GPU: 2, CPU: 3
```
If to_device and to_host would do the proper check and if device/host/device_const/host_const would call to_device and to_host, it could look like this and work correctly:
```C++
amrex::Gpu::TrackedVector<int> tv(3, 1);
auto dp = tv.device().data(); // calls to_device
amrex::ParallelFor(3, [=] AMREX_GPU_DEVICE (int i) {
    dp[i] += 1;
});
auto dp2 = tv.device().data(); // already on device, does nothing 
amrex::ParallelFor(3, [=] AMREX_GPU_DEVICE (int i) {
    dp2[i] += 1;
});
std::cout << tv.host_const()[0] << std::endl; // calls to_host, always 3
```

ax3l: ~I think you miss something: there is no way in a **CPU build** that the status is ever marked dirty. I intentionally have it always marked up-to-date `.device()` and `.host()` never change the status to anything else in a CPU build.~

_Update: dang, I still saw this from a CPU perspective. After-hour coding at its best..._

ax3l: Ah I get it in the 2nd example: we can omit explicit copys altogether because we can check on `.device()`/`.host()` and act accordingly... Good idea!

AlexanderSinn: The bug I am talking about happens on GPU builds but not on CPU build. That is why in the example it gives different output for CPU vs GPU. Please take the time to understand it, I have made 5+ comments pointing out the same bug.

ax3l: Yeah, I admittedly thought initially that allow to over-write is a cool feature to have, not a bug.

ax3l: I removed all explicit syncs now, way cleaner that way. Thanks for the suggestion!

Sorry I did not understand it earlier, I was seeing this somehow from a host-authorative POV...

Docs/sphinx_documentation/source/GPU.rst
```diff
@@ -753,6 +753,153 @@ constructed inside of an MFIter loop with GPU kernels and great care should
...
+    int* dp = mv.data();
+    amrex::ParallelFor(n, [=] AMREX_GPU_DEVICE (int i) {
+        dp[i] *= 2;  // access on device
+    });
+
+    Gpu::streamSynchronize();
+    // mv[i] now accessible on host with updated values
+
+:cpp:`Gpu::TrackedVector` example:
+On GPU builds, you can create this type at any time, even before `amrex::Initialize()`.
```
```suggestion
On GPU builds, you can create this type at any time, even before ``amrex::Initialize()``.
```

WeiqunZhang:
Could we add a GNU Makefile to the test?

Codex says,

• TrackedVector is advertised as surviving multiple AMReX initialize/finalize cycles, but the new implementation only releases and invalidates its device storage during the first Finalize(). Later sessions reuse stale GPU pointers and leak memory, so the feature does not work as documented.

  Review comment:

  - [P1] Re-arm finalize hook for cross-session TrackedVector use — /home/wqzhang/git/amrex-codes/amrex/Src/Base/AMReX_TrackedVector.H:45-59 register_finalize() (lines 45‑59) pushes a lambda onto amrex::ExecOnFinalize only at construction, but that stack is emptied inside amrex::Finalize(). After the first shutdown the callback has run and been popped, so there is no handler left for subsequent initialize/finalize cycles even though the class is advertised (Docs/GPU.rst) and tested for cross-session reuse. A Gpu::TrackedVector that survives into a third amrex::Initialize() will therefore still report Status::up_to_date, keep the old DeviceVector allocation, and skip copying host data even though the GPU context has been torn down; the stale pointer can now reference freed memory and the allocation is leaked until process exit. Please re-register the finalize hook every time GPU storage is released (e.g. schedule register_finalize() again via ExecOnInitialize or call it from release_gpu()) so that each AMReX session reliably invalidates and frees the device mirror.

ax3l:
Oh that's a good call, `ExecOnInitialize` stacks themselves are also popped during AMReX session finalize. That is indeed an issue for N>2 sessions, and I only tested this on 2 sessions so far.

I could add a state flag if we currently are registered for finalize and otherwise re-register into `ExecOnInitialize` in the next to-device copy.

ax3l:
There is another bug: if device is dirty during `ExecOnFinalize` or `release_gpu()` we will drop its state. Will copy to host now.

ax3l:
@WeiqunZhang for GitLab CI on HPSF, I see we run with:
```
    - export AMREX_THE_ARENA_INIT_SIZE=1e9
    - ctest -j 2 --test-dir build --output-on-failure
```
in `.gitlab/hpsf-gitlab-ci.yml` this might still be fragile for large tests...?

WeiqunZhang:
Minor nitpicking first. The class has the following access specifiers: default, private, public and then private. I think it would cleaner if we just have public and private.

I found that the behavior of TrackVector's copy/move constructor and assignment is very confusing. Codex points out:

`TrackedVector::register_finalize()` is the only place that queues the lambda which copies device data back to the host and drops the GPU allocation during `amrex::Finalize()`. The new class calls it from `device()` / `device_const()` and the move operations swap the bookkeeping pointer, so long-lived objects re-arm correctly. However, the copy constructor and copy assignment operator simply deep-copy the `DeviceVector` (lines 90‑116 & 138‑154) without scheduling a handler for the new object. If you copy an already-synchronized `TrackedVector` and never touch the copy’s `device()` before the next `Finalize()`, nothing drains its GPU mirror. After AMReX tears down the device context, the copy still reports `Status::up_to_date`, still holds the old `DeviceVector` pointer, and the next AMReX session will happily hand that stale pointer to kernels. This contradicts the advertised cross-session semantics and leaks the device allocation permanently.

WeiqunZhang:
Instead of multiple shared_ptrs, maybe one shared_ptr holding an internal struct that in turn holds the data could work.

ax3l:
Thanks, those are good ideas to simplify.

The new lifetime bug came in during refactoring, I'll add test coverage.

ax3l:
I fixed, clarified, DRY-ed, refactored/simplified the code. Added more test coverage for the buggy lifetime corner cases.

ax3l:
A codex ran again.

Memos to myself:
> One subtle runtime property that matters a lot: AMReX marks itself
  uninitialized before running ExecOnFinalize callbacks, but it still tears down arenas
  only after those callbacks. I’m checking whether TrackedVector ever relies on the wrong
  side of that boundary or documents guarantees that are stronger than the implementation
  can actually provide.


and

> • Findings

  1. High: release_gpu() / finalize drop a legitimate device-side state change when the
     caller shrinks the device vector to zero. In Src/Base/AMReX_TrackedVector.H:215,
     drain_device() only syncs back when status == device_dirty && !d.device.empty(). But
     device() exposes a mutable Gpu::DeviceVector in Src/Base/AMReX_TrackedVector.H:138, so
     v.device().clear(), v.device().resize(0), or any path that makes the device vector
     empty is a valid mutation of the authoritative side. With the current code,
     release_gpu() and the finalize callback will skip the host resize/copy, leave the old
     host contents intact, then clear device storage and mark host_dirty. That breaks the
     documented cross-session persistence contract in Docs/sphinx_documentation/source/
     GPU.rst:772 and Docs/sphinx_documentation/source/GPU.rst:857. The fix is to treat
     device_dirty as authoritative regardless of size: always host.resize(device.size()),
     and only guard the actual Gpu::copy(...) on size > 0.
  2. Medium: the new test only exercises device-side element writes through raw pointers,
     so it misses the bug above and leaves the most fragile part of the API unverified. The
     coverage in Tests/Base/TrackedVector/main.cpp:51, Tests/Base/TrackedVector/
     main.cpp:116, and the multi-session cases starting at Tests/Base/TrackedVector/
     main.cpp:323 never call mutating DeviceVector operations like clear(), resize(0), or
     assign(...) on the object returned by device(). Given that the API explicitly returns
     the whole vector, those are not edge abuse cases; they are part of the supported
     surface area and need dedicated coverage across both release_gpu() and
     amrex::Finalize().

  Aside from the issue above, the lifetime model is mostly coherent: ExecOnFinalize runs
  before arena and GPU teardown in Src/Base/AMReX.cpp:846, so the weak-pointer finalize
  registration and the copy/move ownership logic look sound.

Src/Base/AMReX_TrackedVector.H
```diff
@@ -0,0 +1,335 @@
...
+            }
+            m_data->status = Status::up_to_date;
+#endif
+        }
+
+        /** Synchronize device data to host */
+        void to_host () const {
+#ifdef AMREX_USE_GPU
+            require_amrex("to_host");
+            amrex::Gpu::streamSynchronize();
```
ax3l, line 345: @WeiqunZhang not sure what our conventions are in AMReX here. This can still pull stale data if the caller mutated `device()` data from another AMReX/raw stream. `amrex::Finalize()` does more safely use `Gpu::streamSynchronizeAll()` first, but in regular user code I usually only see flows through `amrex::Gpu::streamSynchronize();`. What do you think?

AlexanderSinn: The way I see it, if a user changes streams manually and launches something, then the user is also responsible to add an appropriate stream sync to ensure correctness.

WeiqunZhang: Agreed.

ax3l:
Ok, made sure intrusive resizes on device are also handled gracefully. Definitely not what I have this in mind for to use, but makes the APIs as safe as possible.

Now Codex and Claude cannot spot any defects or uncovered memory states anymore on my end, too.

Technically, the move constructor can abort on allocation exceptions from allocations `m_data`, but a true `noexcept` move that keeps the moved-from object valid is quite verbose to write. So it is safe, and unlikely to be disruptive as is.

Future: If we want to apply this pattern more often for many potentially short-lived in-session AMReX containers (not the case for the advertised use here), we could keep the stack of function on `ExecOnFinalize` cleaner by writing a registry that also allows us to pop no-ops (expired weak ptrs from earlier object deallocations). Not a biggy at this point.

Src/Base/AMReX_TrackedVector.H
ax3l: That is what I thought as well, thanks!

ax3l:
I feel like I should stop fiddling on this now. All observed state is covered, code is DRY and defensive. Solves the urgent use case we have for interactive input data handling.

WeiqunZhang:
```
    - You can call `release_gpu()` anytime, but we will call it during AMReX finalize                                                                                   
      to invalidate device().
```
Codex is not happy with that promise.

• The new Gpu::TrackedVector::release_gpu() function is documented as safe to call outside AMReX sessions, yet it unconditionally executes GPU synchronization routines that touch torn-down global state. Calling it before amrex::Initialize() or after Finalize() therefore crashes, breaking the advertised cross-session workflow.

  Review comment:

  - [P1] Guard release_gpu outside AMReX session — /home/wqzhang/git/amrex-codes/amrex/Src/Base/AMReX_TrackedVector.H:178-189
    Gpu::TrackedVector::release_gpu() (lines 178‑189) advertises that it can be called “anytime”, but it unconditionally enters Data::drain_device() which immediately executes Gpu::streamSynchronize() and indexes the global GPU stream pool (lines 223‑229). When AMReX is not initialized—before the first amrex::Initialize() or after amrex::Finalize() has torn the streams down—those globals are empty and streamSynchronize() dereferences gpu_stream_pool[gpu_stream_index[tid]], causing an abort or segfault. pyAMReX-style workflows that finalize and then call release_gpu() to free device buffers therefore crash despite the documentation promising it is safe. Please skip the synchronization/copy when !amrex::Initialized() (e.g., early return in release_gpu()/drain_device()), so the method is truly usable across sessions.

ax3l:
Lol, why doesn't my Codex catch this. Are you running `High` (me) or `Extra high` reasoning effort? :D

ax3l:
Yeah I think that review is wrong though in practice: `drain_device` only does GPU sync and copies if `device_dirty` is true. That is not possible before Initialize or after Finalize with the logic we encoded.

I am ok to program this part more defensible for future API changes, but practically speaking this state cannot be reached by a user.

ax3l:
I will add a debug-only assert in `drain_device` with a comment, so LLMs stop being confused and future readers know what is up.

ax3l:
Actually, I cannot add even a pure debug:
```C++
             static void drain_device (Data& d) {
+                // by design, status is never device_dirty outside an AMReX session
+                AMREX_ASSERT_WITH_MESSAGE(amrex::Initialized() || d.status != Status::device_dirty,
+                    "TrackedVector::drain_device reached device_dirty outside AMReX session");
```
because `amrex::Finalize()` switches to `false` before unrolling the ExecOnFinalize stack.

Let us leave it as is, this is safe.

WeiqunZhang (approved):

PR #5255, opened 2026-03-30
Fix particle dual-grid test

WeiqunZhang:
For floating-point number comparison, use amrex::almostEqual instead of ==.

For int data, cast them to amrex::Long instead of amrex::Real (which might be float) before reduction.

Follow-up to #5204


atmyers (approved):

ax3l:
Thank you! :)

PR #5258, opened 2026-03-30
Reduce memory usage in ReduceToPlaneMF2Patchy

BenWibking:
## Summary
Avoid calling `simplify()` on the BoxList. This avoids making boxes that are too big for any given MPI rank, and avoids out of memory situations with very large runs.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_MultiFabUtil.H
+++ b/Src/Base/AMReX_MultiFabUtil.H
@@ -1498,7 +1498,7 @@ ReduceToPlaneMF2Patchy (int direction, Box const& domain, FA const& mf, F const&
     }
 
     BoxList bl_unique = ba_patch2d.boxList();
-    bl_unique.simplify();
+    //bl_unique.simplify();
     BoxArray ba2d(std::move(bl_unique));
     ba2d.removeOverlap();
     DistributionMapping dm2d(ba2d);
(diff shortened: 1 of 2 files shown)


WeiqunZhang:
@BenWibking Did you try adding `maxSize` after `removeOverlap` instead of removing `simplify()`? It might be much more efficient for the required `removeOverlap`.

BenWibking:
> @BenWibking Did you try adding `maxSize` after `removeOverlap` instead of removing `simplify()`? It might be much more efficient for the required `removeOverlap`.

@WeiqunZhang I updated the PR to do that instead. It works at scale. Can you re-review?

WeiqunZhang:
I think we want to keep the old approach of how ba2d is created initially. The fact that there was a memory issue means `bl_unique` contains big boxes after `simplify()`. So we should keep it so that `removeOverLap` does not need to work too hard. After the following code block, we then use `maxSize` to break it up into smaller pieces.
```
    BoxList bl_unique = ba_patch2d.boxList();
    bl_unique.simplify();
    BoxArray ba2d(std::move(bl_unique));
    ba2d.removeOverlap();
```

BenWibking:
@WeiqunZhang Fixed. Can you re-review?

WeiqunZhang (approved):

PR #5262, opened 2026-03-31
Add physical search radius to NeighborParticleContainer

desjardi:
## Summary

Add fillNeighbors(Real search_radius) and buildNeighborList(check_pair, Real bin_size) overloads for efficient short-range particle interactions when the interaction radius is much smaller than the mesh cell size. fillNeighbors(Real search_radius):
- Stores m_search_radius; computes m_num_neighbor_cells from it
- getNeighborTags: position-based early exit for interior particles; per-level grow via ceil(search_radius / dx[lev])
- GetCommTagsBox: per-level grow via ceil(search_radius / dx[lev]) buildNeighborList(check_pair, Real bin_size):
- Spatial binning at bin_size resolution, independent of mesh dx
- Searches +/-1 bin when bin_size ~ interaction radius fillNeighbors() modified to reset m_search_radius to zero. When m_search_radius == 0, all existing code paths are unchanged.

## Additional background

Ref: https://github.com/AMReX-Codes/amrex/discussions/5243

## Checklist

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_NeighborParticles.H
+++ b/Src/Particle/AMReX_NeighborParticles.H
@@ -250,6 +250,14 @@ public:
     void fillNeighbors ();
 
     ///
+    /// Fill neighbor buffers, but only communicate particles within
+    /// physical distance search_radius of tile boundaries.
+    /// Reduces communication volume when search_radius << cell size.
+    /// CPU build only; the GPU path is not yet implemented.
+    ///
+    void fillNeighbors (amrex::Real search_radius);
+
+    ///
     /// This does an "inverse" fillNeighbors operation, meaning that it adds
     /// data from the ghost particles to the corresponding real ones.
     ///
@@ -273,6 +281,14 @@ public:
     void buildNeighborList (CheckPair const& check_pair, bool sort=false);
 
     ///
+    /// Build a Neighbor List for each tile, using custom bin_size
+    /// instead of the mesh cell size. Useful when the interaction
+    /// radius is much smaller than dx.
+    ///
+    template <class CheckPair>
+    void buildNeighborList (CheckPair const& check_pair, amrex::Real bin_size, bool sort=false);
+
+    ///
     /// Build a Neighbor List for each tile
     ///
     template <class CheckPair, class OtherPCType>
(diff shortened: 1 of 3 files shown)


atmyers:
Hi @desjardi,

Thank you for adding this capability! There are two issues I'm concerned about with the current PR:

1. As it stands, if a user calls the version of `fillNeighbors` with the physical cutoff argument, the default `m_num_neighbor_cells` will be over-ridden and used for subsequent calls to the version of `fillNeighbors` without that argument. I think that if we want to allow the flexibility of the physical cutoff distance changing from call-to-call,  we should treat the number of cutoff cells to use and the cutoff distance as specific to that call, rather than modifying `m_num_neighbor_cells`. Alternatively, if we don't want to allow that flexibility, then I think putting the cutoff distance in the constructor makes more sense.
2. Since this only effects the CPU code path for now, could you add an assertion using `AMREX_ALWAYS_ASSERT_WITH_MESSAGE` that fails if a user tries to call the new version of `fillNeighbors` while `AMREX_USE_GPU` is defined?

Thanks,
Andrew

desjardi:
Thanks for the feedback @atmyers! All fair - I'll work on an updated pull request as soon as possible.

desjardi:
Hi Andrew,

Sorry it took me so long to get back to this PR. Here's an update addressing both points:

1. `search_radius` is now threaded through as a defaulted parameter instead of being stored. `m_search_radius` and the mutation of `m_num_neighbor_cells` are gone. Default callers use the original cell-based path unchanged.

2. Added `AMREX_ALWAYS_ASSERT_WITH_MESSAGE` for the GPU build path.
Let me know if that works for you, and thanks again for your time and patience!

Olivier

desjardi:
I rebased onto current development and squashed into one commit. While doing so I also:

- Added areMasksValid(search_radius) so a previously-cached mask gets invalidated if a later fillNeighbors(radius) needs a larger grow (avoids a silent undersizing when calls are mixed).

- Added a bin_size variant of the cross-container buildNeighborList for symmetry with the single-container one.

atmyers (approved):

atmyers:
Thanks!

PR #5263, opened 2026-03-31
Add unset capability for ParmParse

baperry2:
## Summary

For ParmParse, once a variable is set in an input file, there is no way to remove it completely from the command line or subsequently in the input file. For example, you may have specified `amr.keyword = 5` in `inputs.inp` but then want to override it on the command line so a check on `pp.contains("amr.keyword")` comes up as false. Right now, using command line input overrides you can set it to a null value but not completely remove it from the table (pp.remove() can remove from the table in source code). 

This PR adds an UNSET directive to ParmParse that removes a previously-defined keyword from the parse table. This enables input files to conditionally suppress parameters — for example, an included file can cancel a key set by its parent.
```
amr.keyword = 5
UNSET = amr.keyword
# equivalent to never having set amr.keyword
```

Multiple keys can be removed in a single directive:
```
UNSET = key1 key2 key3
```

## Additional background

Motivated by discussion in https://github.com/Exawind/amr-wind/pull/1897. Code generated by Copilot.

Mirrors the design of the existing FILE directive. The new `ParmParse::UnsetKeyword = "UNSET"` static is exposed in the header for symmetry with FileKeyword. The directive is handled in addDefn() by erasing each listed key from the table; keys not present are silently ignored.

Tests added to Tests/ParmParse/inputs and Tests/ParmParse/main.cpp cover single-key unset, multi-key unset, and verifying that non-listed keys are unaffected.

## Checklist

The proposed changes:


WeiqunZhang:
One potential confusion is how to deal with TOML tables. For example,
```
[x]
a = 1  # Same as x.a = 1 at the root level

UNSET = x.a
UNSET = a # Which one do we prefer?
```

Another minor issue is whether we should add `build_test` and `build_test2` to `.gitignore`.

WeiqunZhang:
Re: table local name vs. full name, I like the current behavior of treating the name as the full name, because that's more consistent with the traditional behavior. But we probably should make it clear in the documentation.

baperry2:
Ok, I updated the docs for TOML table behavior as you suggested. 

I also removed the updates to `.gitignore`, as those are definitely irrelevant to this PR, not sure why Copilot decided to update that.

WeiqunZhang (approved):

PR #5265, opened 2026-04-01
C++20 is now the minimum required version

WeiqunZhang:
Supported compilers are:
- GCC 11 or newer (for both host builds and CUDA host compilers).
- LLVM Clang 14 or newer, including AppleClang 14 on macOS.
- Microsoft Visual Studio 2022 (MSVC 19.34 / 17.4) or newer.
- NVIDIA CUDA Toolkit 12.2 or newer.
- AMD ROCm/HIP 6.0 or newer.
- Intel oneAPI DPC++ 2025.2 or newer. (Aurora has two version only, 2025.2 and 2025.3.)


WeiqunZhang:
The supported compilers were suggested by ChatGPT. We need to verify it.

BenWibking:
There may be some GCC issues with the standard library with GCC 11. Some standard library features are not implemented until GCC 12 or 13: https://gcc.gnu.org/onlinedocs/libstdc++/manual/status.html#status.iso.2020

zingale:
We require GCC 13 for Castro now.  I don't think GCC 11 will be enough.

WeiqunZhang:
What features you guys have used require gcc 12 or 13? Right now, we are not using any in AMReX yet.

zingale:
std::numbers
std::format
some range library stuff
and concepts (see https://github.com/AMReX-Astro/Microphysics/blob/1ffc24e3c10dc0ede4d88d257deade627c28f28b/interfaces/eos_type.H#L168)

BenWibking:
`<concepts>` does not compile without gcc 13. We use that in Quokka.

zingale:
once you turn on std=c++20, clang-tidy will start recommending lots of new modernizations.

WeiqunZhang:
We have explicitly disabled them.
```
# We will try to modernize this after switching to C++20
# -modernize-use-constraints
# -modernize-use-designated-initializers
# -modernize-use-std-numbers
# -modernize-use-ranges
# -modernize-use-integer-sign-comparison
# -modernize-use-starts-ends-with
# -readability-container-contains
```

zingale:
the `modernize-use-std-numbers` is a pain -- if there is a constant that is even close to pi or sqrt(2), it will suggest replacing the entire constant.  Messes up with tabulated rate data.

WeiqunZhang:
Since setting a lower minimum in amrex will not prevent application codes from using new features, we might want to start with a lower version, and see that what features we really want to use in amrex that are not supported by old compilers.

zingale:
yeah, that's a good approach. GCC 11 is getting old, but 12 is probably still around a lot.  16 comes out in a few weeks.

WeiqunZhang:
• - GCC 11 – First stable release (11.1) landed on 27 April 2021, kicking off the GCC 11 series. (gnu.org (https://www.gnu.org/software/gcc/releases.html?utm_source=openai))
      - Where it’s default today:
          - Ubuntu 22.04 LTS “Jammy Jellyfish” ships GCC 11 as the gcc default (the gcc-defaults package in Jammy tracks version 11). (documentation.ubuntu.com
            (https://documentation.ubuntu.com/ubuntu-for-developers/reference/availability/gcc/?utm_source=openai))
          - Red Hat Enterprise Linux 9 also uses GCC 11 as the system compiler that builds the distribution and is supported for the full RHEL 9 lifecycle. (docs.redhat.com
            (https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html-single/developing_c_and_cpp_applications_in_rhel_9/index?utm_source=openai))
      - Implication: if you target GCC 11 you automatically cover the mainstream enterprise RHEL9 base and the still-supported Ubuntu 22.04 LTS line, which together account for a large
        slice of HPC/server installs.
  - GCC 12 – First stable release (12.1) arrived on 6 May 2022, and 12.2 followed on 19 Aug 2022; 12.x has been in maintenance since. (gnu.org
    (https://www.gnu.org/software/gcc/releases.html?utm_source=openai))
      - Where it’s default today:
          - Debian 12 “Bookworm” makes GCC 12.2 the unversioned /usr/bin/gcc, via the gcc (gcc-defaults) package pointing to 12.2.0-3. (packages.debian.org
            (https://packages.debian.org/bookworm/gcc?utm_source=openai))
          - Fedora 36 introduced GCC 12 as part of its GNUToolchain system-wide update, so any Fedora release ≥36 (including current Fedora 39/40 onward) inherits GCC 12 or newer by defa
            ult. (fedoraproject.org (https://fedoraproject.org/wiki/Changes/GNUToolchainF36?utm_source=openai))
      - Implication: GCC 12 support is effectively guaranteed on the latest Debian stable and on modern Fedora-based environments, so treating GCC 12 as the “new baseline” aligns you with
        developers on Bookworm, Fedora, and downstream derivatives (e.g., Rocky/Alma betas tracking RHEL 10).

  Practical takeaway: Document GCC 11 support when you need to cover RHEL 9 / Ubuntu 22.04-era systems, but for “current” distros expect GCC 12 as the default toolchain—especially
  wherever Debian 12 or Fedora 36+ is in play.

atmyers (approved):

ax3l (approved):
🚢 ✔️

PR #5267, opened 2026-04-03
Fix insufficient buffer when projecting grids during regrid

WeiqunZhang:
AmrMesh::MakeNewGrids now grows TagBoxes enough to keep intermediate levels properly nested after projecting fine grids downward, preventing the under-tagging that previously led to improperly nested grids.


BenWibking:
Although we currently regrid very frequently, this is something we also might run into with our typical number of levels. Could a test be added for this?

WeiqunZhang:
Yesterday, I had Codex to try to reproduce the issue, but it failed. It also failed to figured out the issue. But now I think I understand the issue and I should be able to come up with a small test to reproduce the issue.

Issue #4892, opened 2026-01-13
RedistributeCPU with 1 Core 1 Thread 1 Box


ax3l:
`RedistributeCPU` should be a no-OP if run with 1 MPI rank and 1 OpenMP thread.

In a benchmarked of Warp vs WarpX shown by @dpgrote today, it does in fact still take ~20% of the time of the 2D simulation loop in WarpX.

Am I overlooking something? :)

atmyers:
Are there still multiple Boxes and tiles in the simulation?

dpgrote:
The simulation had one box.

atmyers:
If `particles.do_tiling = 1`, which I believe is the default in WarpX when compiling with OpenMP, there will be multiple tiles even on that one Box, so Redistribute will still have some work to do. We will try to optimize this, but to have it truly be a no-op, you have to set `particles.do_tiling = 0`.

dpgrote:
In my test case, if I set `particles.do_tiling = 0`, it doesn't affect the timings. RedistributeCPU takes the same amount of time.

atmyers:
Can you post the `inputs` here so I can run this?

dpgrote:
This is the same input I used here https://github.com/BLAST-WarpX/warpx/issues/6487

atmyers:
This PR: https://github.com/AMReX-Codes/amrex/pull/5268

Includes a number of improvements to redistribute. When I run @dpgrote's example on 1 MPI / 1 thread, I get:

That PR branch:

```
ParticleContainer::Redistribute_impl()                                   200      2.774      2.774      2.774  12.70%
```

Dev:

```
ParticleContainer::Redistribute_impl()                                   200      5.265      5.265      5.265  21.12%
```

I think we can get some further improvement by special-casing the particle location and periodic boundary enforcement for the case of no mesh refinement. But, the current time spent in `Redistribute` is about the same as `WarpXParticleContainer::ApplyBoundaryConditions()` and I think about what you'd expect from making a pass over all the particles.  

The above example had tiling enabled, but as @dpgrote found this example isn't super sensitive to number of tiles per grid. The main cost comes from making a pass over all the particles to 1. remove the invalid ones, and 2) apply periodic boundaries, which needs to happen even in the case with 1 box, 1 tile, and no MPI or OpenMP.

atmyers:
Another optimization idea could be only removing invalid particles sometimes instead of every time step, and skipping them in the PIC loop.

[...]

PR #5268, opened 2026-04-03
Merge RedistributeCPU and RedistributeGPU into one implementation

atmyers:
This merges RedistributeCPU and RedistributeGPU into one shared implementation that works for both, improving the maintainability of the code base. It also restructures the way OpenMP parallelism works in Redistribute on the CPU, resulting in better OpenMP performance and scaling. Another consequence is that particle tiling is now supported on GPU (although probably not desirable in most cases).

Performance results on a redistribute benchmark with 2 MPI ranks on a Perlmutter CPU node as a function of the number of OpenMP threads. "run" is this branch, "dev" is development. The new version is always an improvement and, at high thread counts, is >~ 2x faster. 

When compiled for CPU with USE_OMP=FALSE, the new implementation is about 25% faster on that same benchmark, mostly owing to a new early exit in the partition step. The difference is more dramatic is cases with more particles per cell, like [this](https://github.com/AMReX-Codes/amrex/issues/4892) example from WarpX.

redist_opt_plot

Todo: 
- [x] Add support for tiling to RedistributeGPU
- [x] Rename RedistributeGPU to Redistribute_impl and make both CPU and GPU code paths go through it
- [x] Add support for OpenMP to Redistribute_impl.
- [x] Remove RedistributeCPU
- [x] Performance tests
- [x] Clean up / finalize
- [x] Run full application regtest suite.

The proposed changes:


atmyers:
Comparing the performance on redistribute benchmark with 2 MPI ranks on a Perlmutter CPU node. "run" is this branch, "dev" is development.

redist_opt_plot

atmyers:
Note: the regression test for AMReX apps on gaira, garuda, and biollante look good, so long as these PRs are merged along with this one:
https://github.com/AMReX-Astro/Nyx/pull/109
https://github.com/erf-model/ERF/pull/3123

atmyers:
I also confirmed that the GPU performance has not regressed, despite doing a little bit more work to support tiling:

This PR:
```
ParticleContainer::Redistribute_impl()          503      2.878       2.95      2.997  58.06%
```

Dev:
```
ParticleContainer::RedistributeGPU()            503      2.896      2.939      2.984  57.94%
```

atmyers:
I have added back in an assertion that tiling is off on the GPU if neighbor particles are used. I will remove this limitation and add a better neighbor particles test in a follow-up PR.

WeiqunZhang (approved):

PR #5269, opened 2026-04-05
Refactor AmrMesh::MakeNewGrids

WeiqunZhang:
The refactor addresses two issues.

First, there was a bug when projecting grids during regrid. TagBoxArray might not have enough ghost cells to guarantee proper nesting when the new fine level grows by a lot.

Second, it could be very inefficient to use the previous approach of adding extra ghost cells to TagBoxArray for proper nesting. In some users' cases, it may need more than 30 ghost cells.

In the new approach, instead of using extra ghost cells, we explicitly add the projected tags outside the TagBoxArray's domain.

### DIFF
--- a/Src/AmrCore/AMReX_AmrMesh.cpp
+++ b/Src/AmrCore/AMReX_AmrMesh.cpp
@@ -801,6 +801,7 @@ AmrMesh::MakeNewGrids (int lbase, Real time, int& new_finest, Vector<BoxArray>& 
 
         if (!tags_proj.empty()) {
             tagvec.insert(tagvec.end(), tags_proj.data(), tags_proj.data()+tags_proj.size());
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
See Discussions at #5261 and PR #5267. This supersedes #5267.

asalmgren (approved):

WeiqunZhang:
@desjardi I have convinced myself this is the right fix. The other PR also works, but at a greater cost of using too many ghost cells. Could you give this a test? Thanks!

WeiqunZhang:
@zingale @BenWibking Could you fire up your regression tests? Thanks! :pray:

zingale:
I ran the Castro regression suite, and there are no diffs from this change.

PR #5270, opened 2026-04-05
Fix `PolymorphicArenaWrapper` Use-After-Free

ax3l:
## Summary

- [x] Fix `PolymorphicArenaWrapper` use-after-free on `Arena` destruction

`PolymorphicArenaWrapper` caches a raw `Arena*` that becomes dangling after `Arena::Finalize()` deletes the `CArena`. Unlike the non-polymorphic wrappers (`ArenaWrapper`, `DeviceArenaWrapper`, ...) which call getter functions that fall back to `The_Null_Arena()`, the polymorphic variant bypassed this safety net.

Expose the existing file-static `initialized` flag via `Arena::IsInitialized()` and check it in `PolymorphicArenaWrapper::arena()`. When arenas have been torn down, the wrapper now falls back to `The_Arena()` -> `The_Null_Arena()`, whose `free()` is a no-op.

## Additional background

- [x] Test: A `PODVector` backed by `PolymorphicArenaAllocator` that outlives `amrex::Finalize()` currently segfaults on GPU builds, because the allocator holds a dangling raw `Arena*` after `Finalize` deletes the `CArena`. This test captures the bug.

Same happens for more complex types, like `ParticleContainer`. See, e.g.,: https://github.com/BLAST-ImpactX/impactx/pull/1383

Related to https://github.com/AMReX-Codes/pyamrex/issues/81

## Checklist

The proposed changes:


ax3l:
/run-hpsf-gitlab-ci

With only the reproducer (new unit test), this fails for me locally, let's see that it also fails here.

ax3l:
Unrelated to this PR, HPSF ROCm build fails to configure:
```
$ export ROCM_PATH=/opt/rocm-7.2
$ export PATH=$ROCM_PATH/bin:$ROCM_PATH/llvm/bin:$PATH
$ export LD_LIBRARY_PATH=$ROCM_PATH/lib:$ROCM_PATH/lib64:$LD_LIBRARY_PATH
$ export HIP_PATH=$ROCM_PATH
$ export HIP_CLANG_PATH=$ROCM_PATH/llvm/bin
$ export CMAKE_PREFIX_PATH=$ROCM_PATH:$CMAKE_PREFIX_PATH
$ cmake -S . -B build                              \ # collapsed multi-line command
CMake Deprecation Warning at CMakeLists.txt:34 (cmake_policy):
  The OLD behavior for policy CMP0104 will be removed from a future version
  of CMake.
  The cmake-policies(7) manual explains that the OLD behaviors of all
  policies are deprecated and that a policy should be set to OLD only under
  specific short-term circumstances.  Projects should be ported to the NEW
  behavior and not rely on setting a policy to OLD.
-- The C compiler identification is unknown
-- The CXX compiler identification is unknown
CMake Error at CMakeLists.txt:67 (project):
  No CMAKE_C_COMPILER could be found.
  Tell CMake where to find the compiler by setting either the environment
  variable "CC" or the CMake cache entry CMAKE_C_COMPILER to the full path to
  the compiler, or to the compiler name if it is in the PATH.
CMake Error at CMakeLists.txt:67 (project):
  No CMAKE_CXX_COMPILER could be found.
  Tell CMake where to find the compiler by setting either the environment
  variable "CXX" or the CMake cache entry CMAKE_CXX_COMPILER to the full path
  to the compiler, or to the compiler name if it is in the PATH.
```

ax3l:
Great, Intel-PVC and Nvidia-H100 SP fail as they should with a segfault on the new unit test / reproducer:
```
The following tests FAILED:
	195 - GPU_ArenaLifetime_1d (SEGFAULT)
	196 - GPU_ArenaLifetime_2d (SEGFAULT)
	197 - GPU_ArenaLifetime_3d (SEGFAULT)
```

ax3l:
/run-hpsf-gitlab-ci

Besides the unrelated ROCm bug on CI, this now passes locally for me and should here for Intel-PVC and Nvidia H100, too.

ax3l:
Awesome, works: 
image

(AMD CI is broken, unrelated to this PR)

Src/Base/AMReX_Arena.H
```diff
@@ -324,6 +324,12 @@ public:
     /// implementation of the virtual function.
     virtual void ResetMaxUsageCounter () {}
 
+    /** Whether the Arena subsystem is initialized.
+     *  Used by PolymorphicArenaWrapper to fall back to The_Null_Arena
+     *  when arenas have been torn down (e.g., after amrex::Finalize
+     *  in language bindings with non-deterministic destruction order). */
```
ax3l: We can simplify this, maybe too verbose what this is used for. IDEs can resolve that.
```suggestion
    /** Whether the Arena subsystem is initialized. */
```

Src/Base/AMReX_GpuAllocators.H
```diff
@@ -103,7 +103,12 @@ namespace amrex {
         explicit constexpr PolymorphicArenaWrapper (Arena* a_arena)
```
ax3l, line 103: A cleaner API would probably store via shared & invalidate via weak ptrs, but I found that change a bit too intrusive given that the fix for now can be simple (2nd commit in this PR has it still for interest).

Maybe `m_arena` should be private for safety, because the dangling pointer after finalize never gets cleaned up to be again a `nullptr`.

Src/Base/AMReX_GpuAllocators.H
```diff
@@ -103,7 +103,12 @@ namespace amrex {
...
             : m_arena(a_arena) {}
         [[nodiscard]] Arena* arena () const noexcept {
-            return (m_arena) ? m_arena : The_Arena();
+            //! After Arena::Finalize() the cached pointer is stale;
+            //! fall back to The_Arena() which returns The_Null_Arena().
+            if (m_arena && Arena::IsInitialized()) {
```
WeiqunZhang: Suppose m_arena is The_Pinned_Arena and initialized is false, this will return The_Arena() because the if-test is false. It does not really make a big difference now. But it sounds weird.

ax3l: Isn't initialized only false when AMReX is not initialized, so in that case it would also be the Null Arena?

Should we just explicitly return the null arena in the non init case?

ax3l:
Ugh, runner GPU was used while we ran.

Src/Base/AMReX_GpuAllocators.H
WeiqunZhang: It might be better that we explicitly use The_Null_Arena() if initialized is false.

Also I think neither The_Null_Arena() nor Arena::IsInitialized is completely safe with static objects in another translation unit, if that static object is created before amrex is initialized and bool initialized is created.

The only safe approach might be we intentionally have a memory leak. We put all the amrex objects that need to have a long life into the heap and we never delete them.

WeiqunZhang: This is not just theoretical. I have seen The_Null_Arena() failed.

WeiqunZhang:
We might need to merge the development branch into this to get the HIP CI fix.

Src/Base/AMReX_GpuAllocators.H
ax3l: I'll make that change to return the null arena explicitly.

Src/Base/AMReX_GpuAllocators.H
```diff
@@ -103,7 +103,12 @@ namespace amrex {
...
+                return m_arena;
+            }
+            return The_Arena();
```
ax3l:
```suggestion
            return The_Null_Arena();
```

Src/Base/AMReX_GpuAllocators.H
ax3l: Done 👍🏻

ax3l: Hm, needs another include or fwd def or so. Something I cannot commit from the playground 😅

Src/Base/AMReX_GpuAllocators.H
```diff
@@ -103,7 +103,12 @@ namespace amrex {
         explicit constexpr PolymorphicArenaWrapper (Arena* a_arena)
             : m_arena(a_arena) {}
         [[nodiscard]] Arena* arena () const noexcept {
-            return (m_arena) ? m_arena : The_Arena();
+            //! After Arena::Finalize() the cached pointer is stale;
+            //! fall back to The_Arena() which returns The_Null_Arena().
+            if (m_arena && Arena::IsInitialized()) {
+                return m_arena;
+            }
+            return The_Null_Arena();
```
WeiqunZhang: Oh, The_Null_Arena() is not exposed.

Src/Base/AMReX_GpuAllocators.H
WeiqunZhang: I will fix it.

WeiqunZhang: In the future, if we truly want it to be safe for python, we would need to do something like
```
struct AMReXStatics {
    bool arena_initialized = false;
   NullArena the_null_arena{};
};

AMReXStatics* amrex_statics = new AMReXStatics{};
```
Then we need to save a copy of `AMReXStatics* amrex_statics ` in PolymorphicArenaWrapper.

WeiqunZhang: And if we want avoid the one-time benign memory leak, we can use a shared_ptr.

ax3l:
Your patch LGTM!

Let us chat next week about follow-up options beyond double free protection. I wrote a brain dump here:
https://github.com/AMReX-Codes/pyamrex/issues/81#issuecomment-4189384763
(I also did some LLM back and forth scoping, and saved the resulting design ideas in a comment above that.)

WeiqunZhang (approved):

PR #5277, opened 2026-04-08
Create symbolic link during install

nischild:
## Summary
Setting the install directory in the CMake install step gives an error in the creation of the symlink for `libamrex.a`.
The proposed fix creates the path for the symlink during the install step, allowing to change `CMAKE_INSTALL_PREFIX` in the install step.

Basically the fix moves the creation of the path-variable into `cmake_install.cmake`.
The previous version created the path in CMake configure step and the link during the CMake install step.

## Reproducer
```
git clone https://github.com/AMReX-Codes/amrex.git
cmake -S . -B build
cmake --build build
cmake --install build --prefix install
```
Gives the error
```
CMake Error at build/cmake_install.cmake:420 (file):                                                                                                                                                                                                         
  file Copy failed: No such file or directory     
```

### DIFF
--- a/Tools/CMake/AMReXInstallHelpers.cmake
+++ b/Tools/CMake/AMReXInstallHelpers.cmake
@@ -71,11 +71,13 @@ function (install_amrex_targets)
 
        # legacy symlink for: libamrex.[so|a] / amrex.[dll.lib]
        #   escape spaces for generated cmake_install.cmake file
-       file(TO_CMAKE_PATH "${CMAKE_INSTALL_PREFIX}/lib" ABS_INSTALL_LIB_DIR)
-       install(CODE "file(CREATE_LINK
-           $<TARGET_FILE_NAME:amrex_${AMReX_SPACEDIM_LAST}d>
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
See also #4775 Are the two related?

nischild:
Thanks for pointing out the MR.
Yes, it will also solve the above bug.
Are there any plans to merge #4775 for the coming releases of AMReX?

ax3l:
After a quick test: Your patch fixes the `cmake --install --prefix ...` failure.

I added one more fix for [staged installs](https://cmake.org/cmake/help/latest/envvar/DESTDIR.html#envvar:DESTDIR), which are commonly used by package builders and installers. I was not super aware of that workflow either and found it as an issue w/ Codex: `DESTDIR` is used where the install is written into a temporary root first instead of directly into the final location. For example:

```bash
DESTDIR=/tmp/amrex-stage cmake --install build --prefix /usr/local
```

That should place everything under: `/tmp/amrex-stage/usr/local/...`

Before the follow-up, the normal AMReX install rules did that correctly, but the legacy `libamrex.a` symlink step still tried to create the link under `/usr/local/...` instead of inside the staged tree. So packaging installs could fail, or produce an incomplete staged install.

The follow-up makes the legacy symlink behave like the rest of the install, and it also records that link in `install_manifest.txt`.

ax3l (approved):

PR #5278, opened 2026-04-08
Align Arena allocations to 256 bytes

AlexanderSinn:
## Summary

This PR increases the alignment of Arena allocations from 16 to 256 bytes. This gives a small speedup when running on GPU. 
In this test the alignment affects only the particle data, while the field elements are only 8-byte aligned due to there being no padding for the 2D arrays.
```
alignment: runtime in seconds of 3 runs
512: 8.443, 8.461, 8.444
256: 8.449, 8.459, 8.446
128: 8.473, 8.466, 8.491
64:  8.506, 8.508, 8.511
32:  8.49,  8.491, 8.483
16:  8.491, 8.489, 8.486
8:   misaligned address !!!
```
This could also help with SIMD on CPU

## Additional background

It was not clear from the documentation what the alignment of memory allocated by hipMalloc is. I assume it is 256, the same as cudaMalloc.

## Checklist

The proposed changes:


ax3l:
Oh nice, maybe I can [vector align the CPU SIMD loads](https://github.com/AMReX-Codes/amrex/blob/26.04/Src/Base/AMReX_SIMD.H#L193-L198) and [stores](https://github.com/AMReX-Codes/amrex/blob/26.04/Src/Base/AMReX_SIMD.H#L242-L246) then, too.

I would have otherwise added alignas to the SoA arrays, but I guess this ends up to be the same.

ax3l:
With `AMReX_SIMD` enabled, you can even do a compile-time check for future architectures, e.g.,
```C++
  template <class Simd>
  constexpr bool needs_more_than_256B =
      (std::memory_alignment_v<Simd> > 256);

  static_assert(!needs_more_than_256B<SIMDReal>);
  static_assert(!needs_more_than_256B<SIMDParticleReal>);
  static_assert(!needs_more_than_256B<SIMDInt>);
  static_assert(!needs_more_than_256B<SIMDIdCpu>);
```

ax3l (approved):
This is great, thank you!

If you like, you can update `AMReX_SIMD.H` as well for `load_1d`/`store_1d`.

AlexanderSinn:
Since that receives a general pointer, it is not necessarily aligned. E.g., for field data or particle data that was allocated through std::allocator<double> instead of the Arena. There would need to be additional load_1d_aligned etc. functions, which I won't add in this PR.

WeiqunZhang:
To be safe, we should add an assertion on the alignment after hipMalloc.

AlexanderSinn:
Even if hipMalloc does not align to 256 bytes, things should work with this PR (besides the assert).
Unlike CUDA GPUs, gfx90a is happy to read fully unaligned memory. For example, to read 31 bytes, the compiler generates two 16-byte read instructions where the second one has an offset of 15 bytes (CUDA does 31 1-byte reads). There could still be a performance difference, however.
With this PR, memory allocated through AMReX will have the same alignment guarantee as through the vendor API, up to 256 bytes. So the vendors determine the alignment that is good for their specific GPU.

WeiqunZhang:
This is what I found on ROCm's github repos. There are many layers, here is a line that calls malloc with required alignment.
https://github.com/ROCm/rocm-systems/blob/c7d0197a5118a6006c2cb72366edeb2fc0ed8c77/projects/clr/hipamd/src/hip_vm.cpp#L114
`dev_info.memBaseAddrAlign_` is the number of bits defined here
https://github.com/ROCm/rocm-systems/blob/c7d0197a5118a6006c2cb72366edeb2fc0ed8c77/projects/clr/rocclr/device/rocm/rocdevice.cpp#L1333
It's at least 256 bytes, and it's even more for debug build.
https://github.com/ROCm/clr/blob/41709ee843e3715ebec72db854e0be1119ef620a/rocclr/utils/flags.hpp#L37

AlexanderSinn:
Maybe I will add an assert for all backends, then we can run HPSF CI, and finally I will take out both asserts so it doesn't crash for users.

WeiqunZhang:
Thinking more about this. You are absolutely right that even if hipMalloc does not align to 256 bytes, things should work with this PR (besides the assert). So maybe we should remove the assert. Then add a comment in Arena.H explaining that 256 is only forced in certain situations. For cudaMalloc, it's guaranteed in the their document to be 256. For hipMalloc, it's 256 in practice, etc. I am not sure about cuda/hip host memory

Src/Base/AMReX_Arena.cpp
```diff
@@ -335,14 +345,15 @@ Arena::allocate_system (std::size_t nbytes) // NOLINT(readability-make-member-fu
...
-                    msg = "sycl::malloc_device returned nullptr"
+                    msg = "sycl::aligned_alloc_device returned nullptr"
                 );
                 out_of_memory_abort("GPU device memory", nbytes, msg);
             }
         }
     }
 #endif
     AMREX_ALWAYS_ASSERT(p != nullptr);
+    AMREX_ALWAYS_ASSERT(is_aligned(p, align_size)); // for testing
```
AlexanderSinn:
```suggestion
```

Src/Base/AMReX_Arena.H
```diff
@@ -306,19 +306,9 @@ public:
...
-        long double ld;
-        void (*f) ();
-    };
-    static const std::size_t align_size = sizeof(Word);
-#endif
-
     /// The alignment of allocated memory
-    static const std::size_t align_size = 16;
+    /// 256 bytes is the default alignment that cudaMalloc uses
+    static const std::size_t align_size = 256;
```
AlexanderSinn:
```suggestion
    /// 256 bytes is the default alignment that cudaMalloc uses.
    /// If the native allocation function were to give memory that
    /// has a smaller alignment than align_size, memory from
    /// the Arena will also only have this smaller alignment.
    /// In practice, native allocation functions have >= 256 bytes
    /// of alignment. In all cases, memory should be at least
    /// aligned to std::max_align_t, usually 16.
    static const std::size_t align_size = 256;
```

AlexanderSinn:
Managed memory should be aligned to 4k page sizes; otherwise, the explicit prefetch to host / device would move unrelated allocations. CUDA/HIP host memory probably is also aligned to 4k pages, again poorly documented.

WeiqunZhang (approved):

PR #5281, opened 2026-04-08
avoid duplicate find_package(Conduit)

charlesgwaldman:
## Summary

Building MFIX-Exa with Conduit and Ascent enabled (`cmake -S . -B build -DAMReX_CONDUIT=ON -DAMReX_ASCENT=ON`) fails with:

```
CMake Error at /nfs/home/3/cwaldman/Libs/RelWithDebInfo/lib/cmake/conduit/conduit_setup_targets.cmake:65 (add_library):
  add_library cannot create imported target "conduit::conduit" because
  another target with the same name already exists.
--
CMake Error at /nfs/home/3/cwaldman/Libs/RelWithDebInfo/lib/cmake/conduit/conduit_setup_targets.cmake:101 (add_library):
  add_library cannot create imported target "conduit::conduit_mpi" because
  another target with the same name already exists.
```

This MR adds a guard against duplicate `find_package(Conduit)` calls.

The `if (NOT TARGET ...)` guard is the standard CMake idiom for this (used extensively in VTK, HDF5, etc.). If the target already exists, we use it; if not, we find it. The `target_link_libraries` calls below are unchanged either way since they just reference the same imported target name.

## Checklist

The proposed changes:

### DIFF
--- a/Tools/CMake/AMReXThirdPartyLibraries.cmake
+++ b/Tools/CMake/AMReXThirdPartyLibraries.cmake
@@ -117,7 +117,7 @@ endif ()
 #
 #  Catalyst
 #
-if (AMReX_CATALYST) 
+if (AMReX_CATALYST)
     find_package(Catalyst REQUIRED PATHS "$ENV{CATALYST_IMPLEMENTATION_PATHS}")
     foreach(D IN LISTS AMReX_SPACEDIM)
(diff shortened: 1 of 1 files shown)


ax3l (approved):
While repeated `find_packages` calls to the same package are usually cached, this can happen in complex superbuild logic. LGTM.

PR #5297, opened 2026-04-11
Gpu::Buffer: Add comments about GPU launch region

WeiqunZhang:


Src/Base/AMReX_GpuBuffer.H
```diff
@@ -132,7 +132,7 @@ public:
 
     T* copyToDeviceAsync () noexcept {
 #ifdef AMREX_USE_GPU
-        if (useDVect() && !h_vect.empty())
+        if (!h_vect.empty())
```
AlexanderSinn: Once amrex detects unified memory to avoid having two separate vectors (the commented-out `&& !use_unified_gpu_memory` check in useDVect), then this will need to skip the copy.

WeiqunZhang: Okay. I will revert the change and and a comment instead.

asalmgren (approved):

PR #5304, opened 2026-04-11
Update AGENTS.md

WeiqunZhang:
- reorganize AGENTS.md into a tighter field-manual-style structure
  - consolidate build/test instructions into one reference section
  - convert operating guidance into concise hard rules with AI safety notes
  - keep task playbooks but point them to the shared build reference

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -43,6 +43,10 @@ Use this guide whenever you orchestrate explorers/workers inside the AMReX repos
 - Testing accountability: every substantive change must state which tests ran (and command lines) in the PR/issue.
 - Historical hotspots: mirrored kernels and dimension-specific paths often drift—compare siblings whenever you touch them.
 
+## GPU Lambda Safety
+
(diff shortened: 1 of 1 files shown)


asalmgren (approved):

BenWibking:
For our code, I found that agents often get confused about the AMReX GPU lambda abstraction, so I added this section to Quokka's AGENTS.md to avoid common bugs:
```
## GPU Lambda Safety
Avoid capturing host pointers inside `AMREX_GPU_DEVICE` lambdas.
Prefer device-safe value types like `amrex::GpuArray` (`geom.ProbLoArray()`, 
`geom.CellSizeArray()`, `geom.InvCellSizeArray()`). Do not pass `Geometry::ProbLo()/CellSize()` 
raw pointers into device lambdas; use the array forms instead. Never capture raw pointers from 
`GeometryData` inside GPU lambdas. Avoid accessing `GeometryData` directly; this is almost 
never required.
```
If this appears useful, feel free to borrow this text. I saw this about three months ago with Codex. I'm not sure if this is still a common failure mode.

PR #5329, opened 2026-04-14
EdgeFluxRegister: add plus() member function

BenWibking:
## Summary
Adds `plus(rhs)` to `EdgeFluxRegister`. (`FluxRegister` already has this operator.)

## Additional background
This allows the user to add a temporary EdgeFluxRegister (say, for a given hydro step) to a simulation-wide EdgeFluxRegister. Then, for example, if a hydro step is rejected, the temporary EdgeFluxRegister (which may have been incremented during each stage for a multi-stage integrator) for that step can simply be discarded and implicitly deallocated.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Boundary/AMReX_EdgeFluxRegister.H
+++ b/Src/Boundary/AMReX_EdgeFluxRegister.H
@@ -87,6 +87,12 @@ public:
     //! Reset accumulated state at the start of a coarse step.
     void reset ();
 
+    //! Accumulate another register with identical layout into this one.
+    EdgeFluxRegister& operator+= (EdgeFluxRegister const& rhs);
+
+    //! Convenience alias for operator+=.
+    EdgeFluxRegister& plus (EdgeFluxRegister const& rhs);
+
 #if (AMREX_SPACEDIM == 3)
 
     /**
--- a/Src/Boundary/AMReX_EdgeFluxRegister.H
+++ b/Src/Boundary/AMReX_EdgeFluxRegister.H
@@ -88,9 +88,6 @@ public:
     void reset ();
 
     //! Accumulate another register with identical layout into this one.
-    EdgeFluxRegister& operator+= (EdgeFluxRegister const& rhs);
-
-    //! Convenience alias for operator+=.
     EdgeFluxRegister& plus (EdgeFluxRegister const& rhs);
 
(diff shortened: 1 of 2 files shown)


WeiqunZhang:
I found the meaning of `+=` ambiguous, especially because it does not have an `operator+`. For me, I have to think about whether `+=` means append like for std::string. I think `plus` is better.

BenWibking:
> I found the meaning of `+=` ambiguous, especially because it does not have an `operator+`. For me, I have to think about whether `+=` means append like for std::string. I think `plus` is better.

Ok, I can remove the `operator+=` and just keep the `plus` method.

This is what the code actually does. Here,`fr_as_crse` and `fr_as_fine` are `FluxRegister` objects.
```
			if (fr_as_crse != nullptr) {
				*fr_as_crse += *step_fr_as_crse;
			}
			if (fr_as_fine != nullptr) {
				*fr_as_fine += *step_fr_as_fine;
			}
			if constexpr (Physics_Traits<problem_t>::is_mhd_enabled) {
				if (emf_as_crse != nullptr) {
					emf_as_crse->plus(*step_emf_as_crse);
				}
				if (emf_as_fine != nullptr) {
					emf_as_fine->plus(*step_emf_as_fine);
				}
			}
```
Would you recommend a different pattern here?

WeiqunZhang:
For EdgeFluxRegister, CrseAdd and FineAdd work on fabs, not multifabs. So the new function in this PR is needed for what you do. But for FluxRegister, do you really need to use a temporary FluxRegister?

BenWibking:
> For EdgeFluxRegister, CrseAdd and FineAdd work on fabs, not multifabs. So the new function in this PR is needed for what you do. But for FluxRegister, do you really need to use a temporary FluxRegister?

We have a set of flux multifabs for each RK stage, and currently, also a set of flux multifabs to store the sum of the fluxes across all stages. To save memory, I would like to eliminate the latter flux accumulation multifabs. Although whether this actually makes sense depends on the storage used by the additional temporary FluxRegister.

WeiqunZhang:
FluxRegister's implementation is more memory efficient. It only stores the data at the box boundaries.

WeiqunZhang (approved):

PR #5330, opened 2026-04-14
`SmallMatrix`: `Omega`

ax3l:
## Summary

Add alongside the `Identity` and `Zero` matrix the skew-symmetric matrix Omega:
https://en.wikipedia.org/wiki/Symplectic_matrix#The_matrix_%CE%A9

## Additional background

Used in https://github.com/BLAST-ImpactX/impactx/pull/1396

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_SmallMatrix.H
+++ b/Src/Base/AMReX_SmallMatrix.H
@@ -232,24 +232,25 @@ namespace amrex {
         /** Returns the skew-symmetric matrix Omega (symplectic form)
          *
          * With 2x2 block-diagonal structure [[0,1],[-1,0]].
-         * Requires a square matrix of even dimension.
+         * Requires a square matrix of even dimension and a floating-point
+         * element type.
          *
          * @see https://en.wikipedia.org/wiki/Symplectic_matrix
          */
         template <int MM=NRows, int NN=NCols,
-                  std::enable_if_t<MM==NN && MM%2==0, int> = 0>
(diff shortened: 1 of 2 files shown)


WeiqunZhang (approved):
A minor comment is whether it's worth checking `T` is floating point. It certainly does not make sense if T is unsigned integral type.

Src/Base/AMReX_SmallMatrix.H
```diff
@@ -229,6 +229,29 @@ namespace amrex {
...
+         * @see https://en.wikipedia.org/wiki/Symplectic_matrix
+         */
+        template <int MM=NRows, int NN=NCols,
+                  std::enable_if_t<MM==NN && MM%2==0, int> = 0>
+        static constexpr
+        AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+        SmallMatrix<T,NRows,NCols,ORDER,StartIndex>
+        Omega () noexcept {
+            static_assert(StartIndex == 0 || StartIndex == 1);
+            SmallMatrix<T,NRows,NCols,ORDER,StartIndex> O{};
```
WeiqunZhang: I almost forgot. I think a single letter `O` as the variable name is hard to read. It's too close to `0`. We should make something like `omega` or `omg`.

ax3l:
Thanks for the comments, good points! Addressed the floating point restriction, changed the internal variable name and added a little test to the existing tests.

WeiqunZhang (approved):

PR #5335, opened 2026-04-15
Align capacity to 256 bytes in ParticleTileRT

AlexanderSinn:
## Summary

This PR makes sure that all components of the 2D particle array in ParticleTileRT are aligned to the memory alignment used by `amrex::Arena` (256 bytes). This is achieved by rounding the capacity up to a multiple of 64 particles.

## Checklist

The proposed changes:


Src/Particle/AMReX_ParticleTileRT.H
```diff
@@ -805,6 +808,15 @@ private:
         return static_cast<int>(std::distance(name_list->begin(), pos));
     }
 
+    static size_type align_capacity (size_type capacity) {
+        const auto aligment_in_bytes = Arena::align_size;
+        const auto aligment_in_elements = aligment_in_bytes / std::gcd(
+            std::gcd(aligment_in_bytes, sizeof(uint64_t)),
+            std::gcd(sizeof(RType), sizeof(IType))
+        );
+        return (capacity + aligment_in_elements - 1) / aligment_in_elements * aligment_in_elements;
```
WeiqunZhang: A few minor comments. There is a function, amrex::aligned_size(requirement, size) we can use. But I can understand the argument for not using an obscure function for readability.

The compilers will probably do the optimization. But it won't hurt to use `constexpr auto`.

Could we include `<numeric>` in this header?

WeiqunZhang (approved):

PR #5336, opened 2026-04-15
Call HYPRE_DeviceInitialize after HYPRE_Init if using GPUs

jrood-nrel:
## Summary

Without this, I was experiencing errors during init on a machine where MPI ranks can see multiple GPUs:
```
CUDA ERROR (code = 1, invalid argument) at /whatever/hypre/src/utilities/memory.c:136
```

Setting `export CUDA_VISIBLE_DEVICES=0` was able to avoid the error, but the actual fix seems to be to include the call to `HYPRE_DeviceInitialize()` after `HYPRE_Init()` when using GPUs.

I can verify this solves the error I was seeing when using hypre with AMReX in AMR-Wind.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX.cpp
+++ b/Src/Base/AMReX.cpp
@@ -754,6 +754,7 @@ amrex::Initialize (int& argc, char**& argv, bool build_parm_parse,
         hypre_HandleDefaultExecPolicy(hypre_handle()) = HYPRE_EXEC_DEVICE;
         hypre_HandleSpgemmUseCusparse(hypre_handle()) = 0;
 #endif
+        HYPRE_DeviceInitialize();
 #endif
 
         if (system::verbose > 0) {
--- a/Src/Base/AMReX.cpp
+++ b/Src/Base/AMReX.cpp
@@ -754,7 +754,10 @@ amrex::Initialize (int& argc, char**& argv, bool build_parm_parse,
         hypre_HandleDefaultExecPolicy(hypre_handle()) = HYPRE_EXEC_DEVICE;
         hypre_HandleSpgemmUseCusparse(hypre_handle()) = 0;
 #endif
+#if (HYPRE_RELEASE_NUMBER >= 23100)
+        // Discussion in https://github.com/AMReX-Codes/amrex/pull/5336
         HYPRE_DeviceInitialize();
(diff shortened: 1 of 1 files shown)


Src/Base/AMReX.cpp
```diff
@@ -754,6 +754,7 @@ amrex::Initialize (int& argc, char**& argv, bool build_parm_parse,
         hypre_HandleDefaultExecPolicy(hypre_handle()) = HYPRE_EXEC_DEVICE;
         hypre_HandleSpgemmUseCusparse(hypre_handle()) = 0;
 #endif
+        HYPRE_DeviceInitialize();
```
WeiqunZhang, line 757: Could we move this line right after `HYPRE_init`, in case it needs to be called before functions like HYPRE_SetGPUMemoryPoolSize?

jrood-nrel: Copilot (Opus) claimed it should be at this location. I'm going to convert this to a draft anyway, because it seems like there is more to this I think I need to understand.

WeiqunZhang:
I failed to reproduce the issue with an amrex test on perlmutter. It just worked without calling HYPRE_DeviceInitialize. I ran on 4 MPI processes. Each process can see all four GPUs. Anyway calling it also did not cause any issues for my test. So I am fine with adding it. But I think the function does not exist before v2.31.0. So a guard is needed around the call.

WeiqunZhang:
How did you build hypre? Maybe the issue is device pool?

WeiqunZhang:
With device pool, I can reproduce what you saw. In fact, even running on a single process could fail, if amrex sets device id to non-zero. It appears that hypre device memory pool will use device 0 if `HYPRE_DeviceInitialize` is not called. If it is called, it will call `cudaGetDevice` first to get the device id, and call `hypre_SetDevice(device_id, handle)`. Since before calling `HYPRE_DeviceInitialize`, we have already called `cudaSetDevice`, `hypre_SetDevice` will get the correct ID instead of its default value of 0.

WeiqunZhang:
We are good to go after adding a version guard.

WeiqunZhang (approved):

PR #5345, opened 2026-04-17
Add option to fill ghost cells in `average_cellcenter_to_face`

BenDavisonPetch:
## Summary

Adds an optional parameter `ngrow` to `average_cellcenter_to_face` and two new overloads that take an array of `AMREX_SPACEDIM` `IntVect`s. `average_cellcenter_to_face` fills this many ghost cells in the face-centred `MultiFab`s.

## Checklist

The proposed changes:


Src/Base/AMReX_MultiFabUtil.cpp
```diff
@@ -277,14 +305,15 @@ namespace amrex
             MultiFab foo(amrex::convert(cc.boxArray(),IntVect(1)), cc.DistributionMap(), 1, 0,
                          MFInfo().SetAlloc(false));
             IntVect ng = -cc.nGrowVect();
+            GpuArray<IntVect,AMREX_SPACEDIM> ngv{AMREX_D_DECL(ng_vects[0], ng_vects[1], ng_vects[2])};
             ParallelFor(foo, IntVect(0), ncomp,
```
WeiqunZhang: We need to something like,
```
IntVect ng_foo(0);
for (auto const& iv : ng_vects) {
    ng_foo.max(iv);
}
```
This gives the maximum ghost face indices. We need use this to build MultiFab foo and in ParallelFor. Otherwise the `i,j,k` provided ParallelFor will not include ghost faces.

WeiqunZhang:
Also, let's not worry about the versions using `Vector<MultiFab*>`. We have that for backward compatibility, otherwise we would have removed them. The Array version is better because it has explicit size.

Src/Base/AMReX_MultiFabUtil.cpp
BenDavisonPetch: Ok I've constructed `ng_foo` as you've suggested and given it to `foo` - will what I've got there work? I haven't used that version of `ParallelFor` very much, and I can't build/run the GPU version on the machine I've got right now.

BenDavisonPetch:
> Also, let's not worry about the versions using `Vector<MultiFab*>`. We have that for backward compatibility, otherwise we would have removed them. The Array version is better because it has explicit size.

Makes sense; I've removed the new `Vector<MultiFab*>` overload.

Src/Base/AMReX_MultiFabUtil.cpp
BenDavisonPetch: Ah ok never mind I've figured out how this version works. What I've got should work now, hopefully.

WeiqunZhang (approved):

PR #5347, opened 2026-04-18
Test getEBFluxes in CellEB2

eebasso:
Modify the CellEB2 test to also check the output of getEBFluxes and compare with the exact theoretical flux.


Tests/LinearSolvers/CellEB2/MyTest.cpp
```diff
@@ -368,13 +419,15 @@ MyTest::initData ()
...
                          Array4<Real> const& bz_arr = bcoef[ilev][2].array(mfi););
+            Array4<Real> const& feb_ex_arr = fluxeb_exact[ilev].array(mfi);
 
             auto fabtyp = flags[mfi].getType(bx);
             if (FabType::covered == fabtyp) {
                 amrex::ParallelFor(bx, [=] AMREX_GPU_DEVICE (int i, int j, int k) noexcept
                 {
                     phi_ex_arr(i,j,k) = 0.0;
                     phi_eb_arr(i,j,k) = 0.0;
+                    feb_ex_arr(i,j,k) = 0.0;
```
WeiqunZhang, line 430: Need to do this for FabType::regular too.

WeiqunZhang: What I was talking about is this branch, not about regular cells in else branch.
```
            } else if (FabType::regular == fabtyp) {
                amrex::ParallelFor(nbx, [=] AMREX_GPU_DEVICE (int i, int j, int k) noexcept
                {
                    mytest_set_phi_reg(i,j,k,phi_ex_arr,rhs_arr,
                                       AMREX_D_DECL(bx_arr,by_arr,bz_arr),
                                       dx, lprob_type, bx);
                });
```
It never touches feb_ex_arr.

WeiqunZhang: BTW, you can remove this comment ` // feb_ex_arr(i,j,k) = 0.0;`.

eebasso: Okay, should be done now. Would using `fluxeb_exact.setVal(0.0)` at the same place where `fluxeb_sol.setVal(0.0)` occurs be appropriate? I noticed that this was not done for phiexact

WeiqunZhang: We only need to do it in one place. In the test, either way is fine.

WeiqunZhang (approved):

PR #5351, opened 2026-04-18
TagParallelFor: Return early in case the number of warps is 0

WeiqunZhang:


atmyers:
Under what conditions would the number of warps be 0 but the number of tags not be?

WeiqunZhang:
It's just codex nit-picking. The total number of warps is the sum of warps for each tag, which computed with
```
            auto nw = (detail::get_tag_size(tag) + Gpu::Device::warp_size-1) /
                Gpu::Device::warp_size;
```
Basically it's saying what if you have 10 empty boxes. Then the number of tags is 10, but the total number of warps will be zero. It's unlikely to happen. But in principle it could because we have a public function taking user-defined TagType that we don't have direct controls.

atmyers (approved):

PR #5360, opened 2026-04-20
Reimplement IsConvertible with std::conjunction<std::is_convertible>

### LINKED ISSUE #5354: `AMReX_TypeTraits::IsConvertible` incorrectly requires source types to be default-constructible
`amrex::IsConvertible<T, Args...>` in `Src/Base/AMReX_TypeTraits.H` does not behave like `std::is_convertible`.

The current implementation checks convertibility via an overload using `U{}`:

```cpp
namespace detail {
    template<typename T>
    constexpr bool is_convertible(T) { return true; }

    template <typename T, typename U, typename Enable = void>
    struct IsConvertibleImp : std::false_type {};

    template <typename T, typename U>
    struct IsConvertibleImp<T, U, std::enable_if_t<is_convertible<T>(U{})>>
        : std::true_type {};
}
```

Because this uses `U{}`, it adds an unintended requirement that each source type `U` be default-constructible. That causes false negatives for types that are implicitly convertible to `T` but do not have a default constructor.

Minimal reproducer:

```cpp
#include <type_traits>
#include <AMReX_TypeTraits.H>

struct NoDefault {
    NoDefault() = delete;
    operator double() const { return 1.0; }
};

static_assert(std::is_convertible_v<NoDefault, double>);
static_assert(amrex::IsConvertible_v<double, NoDefault>, "unexpected false");
```

Today the second assertion fails.

This matters because `IsConvertible_v` is used in SFINAE for constructors/utilities such as `IntVect`, `RealVect`, and `IndexType`, so valid call sites can be rejected just because an argument type lacks a default constructor.

I think the trait should delegate directly to the standard one. For example:

```cpp
template <typename T, typename... Args>
using IsConvertible = std::conjunction<std::is_convertible<Args, T>...>;

template <typename T, typename... Args>
inline constexpr bool IsConvertible_v = IsConvertible<T, Args...>::value;
```

This removes the accidental default-constructor requirement and matches the semantics of the standard trait.

One caution: replacing the current logic with a `static_cast<T>(std::declval<U>())`-based detector would fix the default-constructor problem, but it would not exactly match `std::is_convertible`, because it would also accept explicit conversions. Delegating to `std::is_convertible` directly appears to be the cleanest fix.

WeiqunZhang:


Src/Base/AMReX_TypeTraits.H
```diff
@@ -253,24 +253,9 @@ namespace amrex
...
-
     //! \brief Test if all the types Args... are automatically convertible to type T
     template <typename T, typename... Args>
-    struct IsConvertible {
-        static constexpr bool value = (... && detail::IsConvertibleImp<T, Args>::value);
-    };
+    using IsConvertible = std::conjunction<std::is_convertible<Args, T>...>;
 
     template <typename T, typename... Args>
     inline constexpr bool IsConvertible_v = IsConvertible<T, Args...>::value;
```
WeiqunZhang:
```suggestion
    inline constexpr bool IsConvertible_v = IsConvertible<T, Args...>::value; // NOLINT(modernize-type-traits)
```

atmyers (approved):

PR #5361, opened 2026-04-20
FaceDivFree interpolater: Fix component

WeiqunZhang:
Fix component indices for non-zero starting index.


atmyers:
Does it make sense to modify `Tests/DivFreePatch` to cover this case?

WeiqunZhang:
Yes, definitely.

WeiqunZhang:
The test is actually not included in CTest and it does not compile with GPU. I will modify the test and add it CTest.

atmyers (approved):

PR #5376, opened 2026-04-23
Support pure SoA for Neighbor particles

atmyers:
Adds a pure SoA version of NeighborParticleContainer, allowing that layout to be used in MFiX-Exa, FHDeX, etc. This is also needed in order to drop Python support for the FHDeX particle container type. Also adds a new test for pure SoA neighbor particles and expands coverage of the existing NeighborParticles test to cover the sumNeighbors function.

After this is merged, the next step is to update FHDeX to use pure SoA + polymorphic memory allocation.

This will also allow MFiX-Exa to use the pure SoA layout - currently it does a transposition of the particle positions.

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_NeighborList.H
+++ b/Src/Particle/AMReX_NeighborList.H
@@ -39,6 +39,19 @@ namespace detail
 
     template <typename F,
               typename SrcData, typename DstData,
+              typename N1, typename N2,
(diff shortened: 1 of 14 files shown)


atmyers:
Nightly regression tests pass with this PR.

WeiqunZhang (approved):

ax3l:
Thanks, @atmyers :tada:

PR #5403, opened 2026-04-29
Fix OMP issue in BoxArray

WeiqunZhang:
An `omp flush` is missing for the reader. Without it a thread reads has_hashmap=true may still not see the constructed hash map, even though the writer constructs the hash map, flushes and then writes has_hasmap=true.

### DIFF
--- a/Src/Base/AMReX_BoxArray.H
+++ b/Src/Base/AMReX_BoxArray.H
@@ -103,12 +103,17 @@ struct BARef
 #endif
 
     [[nodiscard]] bool HasHashMap () const {
-        bool r;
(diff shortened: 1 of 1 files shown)


ax3l:
Wow, I don't know if I ever touched this :D
We use the has maps in distribution mapping, I assume? :)

ax3l (approved):
LGTM

WeiqunZhang:
In BoxArray/Box intersection.

WeiqunZhang:
The issue appears to be `omp atomic read` uses a relaxed order, thus it's more error prone. Whereas the C++ std::atomic's load is `memory_order_seq_cst` by default.

PR #5404, opened 2026-04-29
Fix for race condition when using getConstParticleTileData in threaded region.

atmyers:
The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_ParticleTile.H
+++ b/Src/Particle/AMReX_ParticleTile.H
@@ -11,8 +11,10 @@
 #include <AMReX_RealVect.H>
 
 #include <array>
+#include <atomic>
 #include <string>
 #include <type_traits>
+#include <utility>
 #include <vector>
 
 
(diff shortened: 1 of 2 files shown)


WeiqunZhang:
With C++20, I think you could use std::atomic_ref. Then the two flags can be normal bools and you don't need to write move ctor and operator=.

WeiqunZhang:
I guess omp atomic is just like std::atomic_ref.

Src/Particle/AMReX_ParticleTile.H
```diff
@@ -1483,6 +1457,124 @@ struct ParticleTile
...
+        }
+    }
+
+    void refreshConstRuntimePtrCaches () const
+    {
+        if (!m_runtime_cptrs_dirty.load(std::memory_order_acquire)) {
+            return;
+        }
+#if defined(AMREX_USE_OMP)
+#pragma omp critical (amrex_particle_tile_data_cache)
```
WeiqunZhang: This one should have a different name from the non-const one.

atmyers:
In my tests this caching mechanism does not meaningfully to the "packBuffer" time on a redistribute benchmark on 16 threads.

WeiqunZhang:
Codex says,

• Found two issues in PR 5404.

  - Src/Particle/AMReX_ParticleTile.H:767: default move assignment now carries the cache arrays and the m_runtime_*_dirty flags together. That is only safe when all particle data buffers are pointer-stolen. If any member move falls back to copying, or on _WIN32 where copy is not deleted under the existing guard, a clean cache can be transferred with raw pointers still targeting the source tile. The new fast path then skips rebuild and returns stale runtime component pointers. I would force both dirty flags true after move/copy-like transfer paths.
  - Src/Particle/AMReX_ParticleTile.H:1292: collectVectors still exposes mutable SoA vectors without invalidating the runtime pointer caches. The internal reserve(addsizes) path now invalidates before calling it, but the public method itself can hand out runtime vectors; mutating/reserving those after a clean get*ParticleTileData() leaves dirty false. This should invalidate like GetStructOfArrays() or stop being a public mutable escape hatch.

Me: The first issue is very interesting. If true, it is an existing bug.

WeiqunZhang:
More from codex.

• I agree it is a non-Windows risk in the PR, but I would not call it an existing development bug.

  The reason: in development, getParticleTileData() and getConstParticleTileData() rebuild m_runtime_*ptrs unconditionally on every call. So even if move assignment copied stale pointer values into the cache arrays, the next getter overwrote them from m_soa_tile.Get*Data(...).dataPtr() before returning.

  PR 5404 changes that contract: after transfer, if m_runtime_*dirty is false, the getter returns early and reuses the transferred cache. That makes the existing default move-assignment behavior newly observable. For non-Windows, the risky case is move assignment, not ordinary move construction: PODVector::operator=(PODVector&&) copies instead of stealing when allocators differ, while the dirty flag can still be copied as clean.

  So I’d phrase it as: existing transfer semantics are fragile, but PR 5404 turns that fragility into a real stale-cache bug. The narrow fix is to make ParticleTile transfer operations explicitly set both dirty flags true after move/copy-like transfer paths.

WeiqunZhang:
Me: How about we define a special struct to store an std::atomic<bool> that has a copy and move constructor and copy and move assignment that always make the new flag dirty?

Codex: Yes, that’s a good local pattern for this PR.

  I’d prefer it over relying on std::atomic_ref<bool> plus defaulted transfer semantics, because it makes the cache invariant part of the member type: after any copy/move of ParticleTile, the cache is dirty by construction.

  Shape would be something like:

```
  struct RuntimePtrCacheFlag {
      RuntimePtrCacheFlag () noexcept = default;

      RuntimePtrCacheFlag (RuntimePtrCacheFlag const&) noexcept
          : flag(true) {}
      RuntimePtrCacheFlag (RuntimePtrCacheFlag&&) noexcept
          : flag(true) {}

      RuntimePtrCacheFlag& operator= (RuntimePtrCacheFlag const&) noexcept {
          flag.store(true, std::memory_order_release);
          return *this;
      }

      RuntimePtrCacheFlag& operator= (RuntimePtrCacheFlag&&) noexcept {
          flag.store(true, std::memory_order_release);
          return *this;
      }

      void set_dirty () noexcept {
          flag.store(true, std::memory_order_release);
      }

      void set_clean () noexcept {
          flag.store(false, std::memory_order_release);
      }

      bool is_dirty () const noexcept {
          return flag.load(std::memory_order_acquire);
      }

      bool is_dirty_relaxed () const noexcept {
          return flag.load(std::memory_order_relaxed);
      }

  private:
      mutable std::atomic<bool> flag {true};
  };
```

  Then ParticleTile can keep:

```
  RuntimePtrCacheFlag m_runtime_ptrs_dirty;
  mutable RuntimePtrCacheFlag m_runtime_cptrs_dirty;
```

  This handles non-Windows move construction/assignment and the _WIN32 copy fallback in one place. I would still keep explicit invalidateRuntimePtrCaches() calls for actual storage mutations; the wrapper only solves transfer safety. The one detail I’d watch is that defaulted ParticleTile move assignment remains okay only because the dirty flag members are assigned after the cache vectors in member order, so they end the assignment dirty.

WeiqunZhang (approved):

PR #5406, opened 2026-04-30
Use std::atomic instead omp flush

WeiqunZhang:
This makes thread sanitizer happy because it does not appear to understand omp flush.


ax3l (approved):
`std::atomic` is nice, too :)

PR #5429, opened 2026-05-02
gfx12xx (RDNA4) regex update

wiersema1:
## Summary
Updated regex in Tools/Cmake/AMReXSetDefines.cmake to catch gfx12xx devices and use the correct wavefront size.

## Additional background
My build environment includes
- ROCm 6.2.4 / HIP 6.2.41134 / AMD Clang 18.0.0
- GCC 13.3.0 (Ubuntu 24.04)
- CMake 3.28.3

## Checklist

The proposed changes:


Tools/CMake/AMReXSetDefines.cmake
```diff
@@ -65,7 +65,7 @@ endif()
 if (AMReX_HIP)
    add_amrex_define( AMREX_USE_HIP NO_LEGACY )
    add_amrex_define( NDEBUG )  # This address a bug that causes slow build times
-   if (${AMReX_AMD_ARCH} MATCHES "gfx1[01].*")
+   if (${AMReX_AMD_ARCH} MATCHES "gfx1[012].*")
```
WeiqunZhang:
```suggestion
   if (${AMReX_AMD_ARCH} MATCHES "gfx1???")
```
Could you tell me if this works?

wiersema1: `"gfx1???"` gives an error during compilation.

> CMake Error at Submodules/AMReX/Tools/CMake/AMReXSetDefines.cmake:68 (if):
>   if given arguments:
> 
>     "gfx1201" "MATCHES" "gfx1???"
> 
>   Regular expression "gfx1???" cannot compile
> Call Stack (most recent call first):
>   Submodules/AMReX/Src/CMakeLists.txt:45 (include)

AlexanderSinn: Can you try
```
if (${AMReX_AMD_ARCH} MATCHES "^gfx1..[a-z0-9]($|:)")
```
this should also cover the case of extra options like `gfx90a:sramecc+:xnack+`

https://godbolt.org/z/baGYGaG8o

WeiqunZhang: Didn't know CMake's regex does not support `?`.

AlexanderSinn: It does, but it is a modifier for the previous character (can match 0 or 1) and there can only be one question mark in a row.

WeiqunZhang: I got confused with shell.

wiersema1: @AlexanderSinn your suggestion worked on my system and I've pushed that change into this PR.

WeiqunZhang (approved):

PR #5433, opened 2026-05-05
Parser: Update jn and yn

WeiqunZhang:
For host functions on Windows, we now either use _jn and _yn to fix deprecation warnings or use std functions to implement them.

For SYCL device functions, we now use Intel's extension if available.

Run ctest in Windows CIs and fix some tests.


ax3l:
Thank you for this!

I would maybe remove these function again: We intentionally added them a while ago for laser modeling, but decided to follow a more general, file-based approach with [LASY](https://github.com/LASY-org/lasy/) now that also supports measured and modeled data.

Since jn/yn are quite costly functions, removing them could help reduce register/performance bloat for every function that uses the AMReX parser.

ax3l (approved):
The structural changes look good, but I would probably remove the two functions in question from the parser again. I suspect they are not used in production.

Issue #5375, opened 2026-04-22
C++20 Features To Prioritize For AMReX


WeiqunZhang:
## Summary
Adopt C++20 features that improve template constraints, diagnostics, and low-level utility code without fighting AMReX's GPU model or replacing existing core abstractions like `Array4`, `GpuArray`, `MFIter`, and particle tile views.

The best practical-now priorities are:
- `concepts` / `requires`
- `std::span` in host-side APIs only
- `<bit>` utilities such as `std::bit_cast`, `std::popcount`, `std::has_single_bit`, `std::bit_ceil`
- `constinit` for global/static state that must never be dynamically initialized
- selective use of `std::ranges`, `std::erase_if`, `std::ssize`, and `std::cmp_*`
- `[[no_unique_address]]` for small policy/empty-member optimization cases

### 1. Concepts / requires
Highest-value upgrade for AMReX.

Why it fits:
- AMReX is full of SFINAE, traits, `enable_if`, and callable-shape dispatch.
- Better diagnostics matter a lot in user-facing templates and backend-specialized code.

Use first in:
- public template entry points in particles, reductions, linear solvers, and array accessors
- trait-heavy helpers in `AMReX_TypeTraits.H`
- callable-based APIs where AMReX currently probes multiple valid forms

Guidance:
- use named concepts for domain ideas like MultiFab-like, ParticleContainer, valid index type, callable on PTD/particle
- keep `if constexpr` for real behavior branching
- do not convert all metaprogramming helpers at once

### 2. `std::span` for host-facing contiguous views
Useful, but not as a replacement for core AMReX views.

Good targets:
- host-only helper APIs that currently take pointer-plus-length
- metadata and buffer utilities
- test/helper code and non-kernel plumbing

Do not use as blanket replacements for:
- `Array4`
- `GpuArray`
- particle tile/device views
- APIs that must be obviously device-callable across CUDA/HIP/SYCL

Rule of thumb:
- `span` for host-side contiguous metadata and buffers
- existing AMReX views remain the kernel/data-path abstraction

### 3. `<bit>` utilities
Good fit for low-level numeric and systems code.

Most useful:
- `std::bit_cast`
- `std::popcount`
- `std::has_single_bit`
- `std::bit_ceil`
- `std::bit_floor`
- `std::rotl` / `std::rotr` where relevant

Good targets:
- descriptor/packing code
- alignment and rounding helpers
- power-of-two checks and bucket sizing
[...]

PR #5372, opened 2026-04-22
Clang-Tidy: modernize-use-constraints

WeiqunZhang:
This PR migrates `std::enable_if_t` to C++20 concepts and constraints. To make the compilers and doxygen happy, (1) we need to write `requires (...)`, instead of bare `requires expr`; (2) we need to place `requires` immediately after `template <...>`.

This is expected to break pyamrex, which will be fixed later.

PR #5434, opened 2026-05-06
Concepts trial

ThomasHowarth:
I've had a go at adding in a couple of trial `concepts` into AMReX, specifically for the `FillPatchUtil` by adding in `IsFabArrayType`, `IsBaseFabType` and `IsSpecificInterp` concepts. The large number of changes mostly comes from the formatter (does anyone have a suggestion for that?).

A first step in #5375

### DIFF
--- a/Src/AmrCore/AMReX_FillPatchUtil.H
+++ b/Src/AmrCore/AMReX_FillPatchUtil.H
@@ -32,28 +32,23 @@
 #include <limits>
 #include <map>
 
-namespace amrex
-{
-    template <typename MFFAB>
-    struct NullInterpHook
-    {
-        template <typename T = MFFAB>
+namespace amrex {
+    template<typename MFFAB>
+    struct NullInterpHook {
+        template<typename T = MFFAB>
             requires IsBaseFabType<T>
-        void operator()(MFFAB& /*fab*/, const amrex::Box& /*bx*/, int /*icomp*/, int /*ncomp*/) const
-        {
+        void operator()(MFFAB & /*fab*/, const amrex::Box & /*bx*/, int /*icomp*/, int /*ncomp*/) const {
         }
 
-        template <typename T = MFFAB>
+        template<typename T = MFFAB>
             requires IsBaseFabType<T>
-        void operator()(amrex::Array<MFFAB*, AMREX_SPACEDIM> /*fab*/, const amrex::Box& /*bx*/, int /*icomp*/,
-                        int /*ncomp*/) const
-        {
+        void operator()(amrex::Array<MFFAB *, AMREX_SPACEDIM> /*fab*/, const amrex::Box & /*bx*/, int /*icomp*/,
+                        int /*ncomp*/) const {
         }
 
-        template <typename T = MFFAB>
+        template<typename T = MFFAB>
             requires IsFabArrayType<T>
-        void operator()(MFFAB& /*mf*/, int /*icomp*/, int /*ncomp*/) const
-        {
+        void operator()(MFFAB & /*mf*/, int /*icomp*/, int /*ncomp*/) const {
         }
     };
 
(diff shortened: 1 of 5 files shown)


ThomasHowarth:
I realise that most of these changes have also been suggested in #5372 - perhaps listing the concepts in AMReX_Concepts could make things more readable, however?

WeiqunZhang:
Adding AMReX_Concepts is not a bad idea, but the unnecessary  white space changes are unacceptable.

> The large number of changes mostly comes from the formatter (does anyone have a suggestion for that?).

Don't use it.

Src/Base/AMReX_Concepts.H
```diff
@@ -0,0 +1,27 @@
...
+#include <AMReX_Config.H>
+#include <AMReX_TypeTraits.H>
+#include <concepts>
+
+namespace amrex {
+/**
+ * \brief Checks if a type is derived from amrex::BaseFab.
+ */
+template <typename T>
+concept IsBaseFabType = amrex::IsBaseFab<T>::value;
```
WeiqunZhang: As we know, naming is always the hardest thing. I don't think we should start with `Is`, because we might write code like below, which sounds awkward.
```
template <IsBaseFabType FAB> void f (FAB& fab);
```
Maybe we should use either `BaseFabType` or `BaseFabLike`.

Anyone who has an opinion, please feel free to chime in.

Src/Base/AMReX_Concepts.H
```diff
@@ -0,0 +1,27 @@
...
+ * \brief Checks if a type is an amrex::FabArray (like MultiFab).
+ */
+template <typename T>
+concept IsFabArrayType = amrex::IsFabArray<T>::value;
+
+/**
+ * \brief Checks if a Interpolation is not generic MFInterpolater.
+ */
+template <typename Interp>
+concept IsSpecificInterp = !std::is_same_v<Interp, amrex::MFInterpolater>;
```
WeiqunZhang: I don't think this concept is useful, because its meaning is very unclear. I think we should remove this.

Src/AmrCore/AMReX_FillPatchUtil.H
```diff
@@ -88,8 +91,8 @@ namespace amrex
      * \param bcfcomp starting component for physbcf
      */
     template <typename MF, typename BC>
-    std::enable_if_t<IsFabArray<MF>::value>
-    FillPatchSingleLevel (MF& mf, IntVect const& nghost, Real time,
+    requires IsFabArrayType<MF>
```
WeiqunZhang: I think it would look cleaner if we do `template <FabArrayType/FabArrayLike MF, typename BC>` instead of `requires`.

Src/Base/AMReX_Concepts.H
WeiqunZhang: ^ @atmyers @ax3l @AlexanderSinn  @zingale @BenWibking

WeiqunZhang: Or FabType or FabLike?

AlexanderSinn: I agree with not having "Is" in the name. The standard library also doesn't have it.
```
template< class T >
concept floating_point = std::is_floating_point_v<T>;
```
Since BaseFAB is already taken, I think `BaseFabType` or `FabType` are fine.

ThomasHowarth: I've gone for BaseFabType/FabArrayType throughout

Src/Base/AMReX_Concepts.H
ThomasHowarth: Removed. It is something that does seem to appear throughout - should it be left as is, or just a clearer concept?

Src/AmrCore/AMReX_FillPatchUtil.H
ThomasHowarth: Used this formatting throughout; it is indeed much cleaner.

WeiqunZhang (approved):

PR #5437, opened 2026-05-07
SIMD: `store_1d` Add `ForceWriteback`

ax3l:
## Summary

The `ForceWriteback` flag forces the writeback regardless of the push method's argument constness. Use it for slots that may be modified by something other than the named push method itself (e.g., a pre/post step in the caller), where the push method's signature is not the authoritative source of truth for whether the value changed.

## Additional background

First needed in ImpactX: https://github.com/BLAST-ImpactX/impactx/pull/1455 via https://github.com/BLAST-ImpactX/impactx/pull/1289
Follow-up to #4924

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_SIMD.H
+++ b/Src/Base/AMReX_SIMD.H
@@ -218,8 +218,15 @@ namespace amrex::simd
      * themselves, but we better help a little for robustness. Background:
      * https://github.com/AMReX-Codes/amrex/pull/4520#issuecomment-3064064215
      *
+     * The `ForceWriteback` flag forces the writeback regardless of the push
+     * method's argument constness. Use it for slots that may be modified by
+     * something other than the named push method itself (e.g., a pre/post
+     * step in the caller), where the push method's signature is not the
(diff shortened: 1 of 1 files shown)


Src/Base/AMReX_SIMD.H
```diff
@@ -218,16 +218,24 @@ namespace amrex::simd
...
+     * @tparam ForceWriteback if true, write back regardless of P_Method's argument constness
      * @tparam T data type
      * @tparam IndexType int or SIMD index
      * @tparam ValType the type of the value to store
      * @param val the value to store
      * @param ptr pointer to contiguous 1D array data
      * @param i index or SIMD index
      */
-    template <auto P_Method, int N, typename T, typename IndexType, typename ValType>
+    template <auto P_Method, int N, bool ForceWriteback = false,
```
ax3l, line 237: Today I learned that function templates allow mid-list defaults, if the rest that follows can be deducted.

WeiqunZhang (approved):

PR #5440, opened 2026-05-08
Fix dual grid restart when particles only on finest level

### LINKED ISSUE #5438: Dual-grid particle restart: dual_grid flag not set when some levels lack Particle_H and box arrays differ
## Description

In `ParticleContainer::Restart()`, when a checkpoint has particles only at the finest AMR level (so coarser levels have no `Particle_H` files), the `dual_grid` flag is not set for those coarser levels even when their box arrays differ between the checkpoint and the restart run. This causes the read loop to use the wrong-sized dummy MultiFab for those levels, producing out-of-bounds access into the `which`/`count`/`where` vectors — resulting in garbage DATA file indices and a crash.

## Affected file

`Src/Particle/AMReX_ParticleIO.H`

## Reproduction conditions

1. Write a checkpoint with particles **only at the finest AMR level** (e.g. `max_level=1`, all particles at Level_1). Level_0 will have no `Particle_H` file in the checkpoint.
2. Restart with the **same number of MPI ranks** but a different `blocking_factor` or `max_grid_size`, so that the level-0 box arrays differ from the checkpoint (e.g. `blocking_factor=32` at write time, `blocking_factor=8` at restart).
3. Crash with garbage DATA file indices such as `DATA_1302810847`.

## Root cause

The dual-grid detection loop in `Restart()`:

```cpp
if (have_pheaders)
{
    for (int lev = 0; lev <= finestLevel(); lev++)
    {
        ...
        if (! amrex::FileExists(phdr_name)) { continue; }  // <-- silently skips

        particle_box_arrays[lev].readFrom(phdr_file);
        if (! particle_box_arrays[lev].CellEqual(ParticleBoxArray(lev))) { dual_grid = true; }
    }
}
```

When `Particle_H` does not exist for a level (no particles at that level in the checkpoint), the level is silently skipped. The `dual_grid` flag is never set for that level, even when the current particle box array at that level differs from the single-domain-box fallback that the dual-grid setup would use.

## Proposed fix

When `Particle_H` is absent for a level, compare the single-domain-box fallback (used in the dual-grid setup path) against the current particle box array. If they differ, force `dual_grid = true`:

```cpp
if (! amrex::FileExists(phdr_name)) {
    // No particles at this level in the checkpoint. The dual-grid setup
    // substitutes BoxArray(Geom(lev).Domain()) for this level. If that
    // single-box domain array does not match the current particle box
    // array, we need dual-grid mode so the read loop uses the correct
    // (1-box) dummy MF.
    if (! BoxArray(Geom(lev).Domain()).CellEqual(ParticleBoxArray(lev))) {
        dual_grid = true;
    }
    continue;
}
```

## Notes

- This is distinct from the bug fixed by PR #5075 (vectors sized to `finest_level_in_file` instead of `max(finest_level_in_file, finestLevel())`).
- The original crash reported in [quokka-astro/quokka#1544](https://github.com/quokka-astro/quokka/issues/1544) involved **different MPI rank counts**. In that case the finest-level `Particle_H` also differs between checkpoint and restart, so `dual_grid = true` is set correctly and the fix in PR #2276 handles the level-0 fallback. This edge case only manifests with the **same number of ranks** but different `blocking_factor`/`max_grid_size`.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

atmyers:
Closes #5438. 

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_ParticleIO.H
+++ b/Src/Particle/AMReX_ParticleIO.H
@@ -794,7 +794,17 @@ ParticleContainer_impl<ParticleType, NArrayReal, NArrayInt, Allocator, CellAssig
             phdr_name = amrex::Concatenate(phdr_name, lev, 1);
             phdr_name += "/Particle_H";
 
-            if (! amrex::FileExists(phdr_name)) { continue; }
+            if (! amrex::FileExists(phdr_name)) {
+                // No particles were checkpointed at this level. Dual-grid
+                // restart will use a single-box domain BoxArray as the
+                // temporary layout for this level, so detect that mismatch
+                // here as well.
+                particle_box_arrays[lev] = BoxArray(Geom(lev).Domain());
+                if (! particle_box_arrays[lev].CellEqual(ParticleBoxArray(lev))) {
+                    dual_grid = true;
+                }
+                continue;
+            }
 
             Vector<char> phdr_chars;
             ParallelDescriptor::ReadAndBcastFile(phdr_name, phdr_chars);
--- a/Tests/Particles/CheckpointRestartDualGrid/main.cpp
+++ b/Tests/Particles/CheckpointRestartDualGrid/main.cpp
@@ -46,12 +46,12 @@ MeshData build_mesh (int ncells, int nlevs, int max_grid_size, int fine_cells = 
     MeshData m;
 
     // Level-0 domain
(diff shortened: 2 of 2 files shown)


chongchonghe:
@atmyers I tested and confirmed that this PR fixed the issue #5438

WeiqunZhang (approved):

PR #5441, opened 2026-05-11
Fix index ordering convention in DenseBins and SparseBins.

### LINKED ISSUE #5432: DenseBins Uses Unconventional Index Order
cc @atmyers @WeiqunZhang 

Follow-up to https://github.com/BLAST-WarpX/warpx/pull/6827

atmyers:
Currently, in the particle code - particularly in `DenseBins`, `NeighborList`, and `NeighborParticleContainer`, we use a z-fastest index ordering convention. This is not consistent with elsewhere in amrex, for example `Box::atOffset`. This PR changes the particle code to use the same x-fastest convention as elsewhere in the code. This should be less surprising for users, for example see this PR: https://github.com/BLAST-WarpX/warpx/pull/6827

Closes #5432. Supersedes PR #3195. 

Note: on the older version of this PR, I verified that this change is performance-neutral, as expected.

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_NeighborParticlesI.H
+++ b/Src/Particle/AMReX_NeighborParticlesI.H
@@ -1091,10 +1091,10 @@ selectActualNeighbors (CheckPair const& check_pair, int num_cells)
                 // it adds a call to check_pair. Thus the maximum number value of "loc" below is the
                 // number of neighbors minus 1, and the buffer pointed to by p_boundary_particle_ids
                 // will not be overwritten.
-                for (int ii = amrex::max(ix-num_cells, 0); ii <= amrex::min(ix+num_cells, nx-1); ++ii) {
+                for (int kk = amrex::max(ix-num_cells, 0); kk <= amrex::min(ix+num_cells, nx-1); ++kk) {
                     for (int jj = amrex::max(iy-num_cells, 0); jj <= amrex::min(iy+num_cells, ny-1); ++jj) {
-                        for (int kk = amrex::max(iz-num_cells, 0); kk <= amrex::min(iz+num_cells, nz-1); ++kk) {
-                            if (isActualNeighbor) { break; }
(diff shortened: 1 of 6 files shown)


Src/Particle/AMReX_NeighborParticlesI.H
```diff
@@ -1091,11 +1091,11 @@ selectActualNeighbors (CheckPair const& check_pair, int num_cells)
...
                 // number of neighbors minus 1, and the buffer pointed to by p_boundary_particle_ids
                 // will not be overwritten.
-                for (int ii = amrex::max(ix-num_cells, 0); ii <= amrex::min(ix+num_cells, nx-1); ++ii) {
+                for (int kk = amrex::max(iz-num_cells, 0); kk <= amrex::min(iz+num_cells, nz-1); ++kk) {
                     for (int jj = amrex::max(iy-num_cells, 0); jj <= amrex::min(iy+num_cells, ny-1); ++jj) {
-                        for (int kk = amrex::max(iz-num_cells, 0); kk <= amrex::min(iz+num_cells, nz-1); ++kk) {
-                            if (isActualNeighbor) { break; }
-                            int nbr_cell_id = (ii * ny + jj) * nz + kk;
+                        for (int ii = amrex::max(ix-num_cells, 0); ii <= amrex::min(ix+num_cells, nx-1); ++ii) {
+                            if (isActualNeighbor) break;
```
WeiqunZhang: We have a clang-tidy check that forces the use of `{}`. It did not get triggered probably because this is instantiated in testing. But it might get triggered in the future. So it should be fixed it by reversing back to the old form.

WeiqunZhang (changes requested):

Src/Particle/AMReX_NeighborParticlesI.H
atmyers:
```suggestion
                            if (isActualNeighbor) { break; }
```

WeiqunZhang (approved):

PR #5446, opened 2026-05-13
Runtime `amrex::Math::powi`

ax3l:
## Summary

After C++11, `std::pow` overloads are a bit of a nightmare: https://en.cppreference.com/cpp/numeric/math/pow

A call to `std::pow(float, int)` promotes to `std::pow(double, double)` and there are no `int` overloads at all after C++17.

Now, if we just call into `std::pow(float, float)` by casting the variable of the exponent, then the compiler backend (e.g., in NVCC) likely misses that the 2nd variable is an int and will not replace to the cheaper int power math.

But especially SYCL/HIP libraries have int overloads (`powi`/`pown`) that are optimized for whole integer exponents.
Interestingly, [CUDA](https://docs.nvidia.com/cuda/cuda-math-api/cuda_math_api/group__CUDA__MATH__DOUBLE.html) seems to have no int overload advertised, but it seems to exist in device headers... ah [here they are](https://docs.nvidia.com/cuda/libdevice-users-guide/)

We already have a compile-time int in `amrex::Math::powi` and this adds a runtime overload as well that we can use on hot code paths.

## Additional background

Triggered investigation in https://github.com/BLAST-ImpactX/impactx/pull/1465

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_Math.H
+++ b/Src/Base/AMReX_Math.H
@@ -250,6 +250,60 @@ constexpr T powi (T x) noexcept
     }
 }
 
+} // namespace amrex::Math
+
+#if defined(AMREX_USE_CUDA)
+// Forward-declare libdevice integer-exponent intrinsics. They are part of
+// NVIDIA's libdevice and linked at PTX time but are not declared in standard
+// CUDA cmath headers outside __CUDACC_RTC__ / _LIBCPP_VERSION blocks.
+// See https://docs.nvidia.com/cuda/libdevice-users-guide/__nv_powif.html
+extern "C" __device__ float  __nv_powif (float,  int);
+extern "C" __device__ double __nv_powi  (double, int);
+#endif
+
+namespace amrex::Math {
+
+//! Return pow(x, n) with a runtime integer exponent.
+//!
+//! On GPU backends this routes to the integer-exponent intrinsic
+//! (libdevice __nv_powif / __nv_powi on CUDA, __ocml_pown_f32 /
+//! __ocml_pown_f64 on HIP, sycl::pown on SYCL), which is faster than the
+//! general pow and avoids the C++11 mixed-arg promotion that would
+//! otherwise call the slower double-precision pow(double, double) and
+//! narrow on return. On host it casts the int exponent to the
+//! floating-point type and calls std::pow.
+AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+float powi (float x, int n) noexcept
+{
+#if defined(AMREX_USE_SYCL)
+    return sycl::pown(x, n);
+#elif defined(AMREX_USE_HIP)
+    AMREX_IF_ON_DEVICE(( return ::powif(x, n); ))
+    AMREX_IF_ON_HOST(( return std::pow(x, static_cast<float>(n)); ))
+#elif defined(AMREX_USE_CUDA)
+    AMREX_IF_ON_DEVICE(( return __nv_powif(x, n); ))
+    AMREX_IF_ON_HOST(( return std::pow(x, static_cast<float>(n)); ))
+#else
+    return std::pow(x, static_cast<float>(n));
+#endif
+}
+
+AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+double powi (double x, int n) noexcept
+{
+#if defined(AMREX_USE_SYCL)
+    return sycl::pown(x, n);
+#elif defined(AMREX_USE_HIP)
+    AMREX_IF_ON_DEVICE(( return ::powi(x, n); ))
+    AMREX_IF_ON_HOST(( return std::pow(x, n); ))
+#elif defined(AMREX_USE_CUDA)
+    AMREX_IF_ON_DEVICE(( return __nv_powi(x, n); ))
+    AMREX_IF_ON_HOST(( return std::pow(x, n); ))
+#else
+    return std::pow(x, n);
+#endif
+}
+
 #if defined(AMREX_INT128_SUPPORTED)
 AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
 std::uint64_t umulhi (std::uint64_t a, std::uint64_t b)


Src/Base/AMReX_Math.H
```diff
@@ -250,6 +250,35 @@ constexpr T powi (T x) noexcept
...
+//! would otherwise call the slower double-precision `pow(double,double)`.
+//! On host (and on backends without a faster path) it falls back to `std::pow`.
+AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+float powi (float x, int n) noexcept
+{
+#if defined(AMREX_USE_SYCL)
+    return sycl::pown(x, n);
+#else
+    AMREX_IF_ON_DEVICE(( return ::powif(x, n); ))
+    AMREX_IF_ON_HOST(( return std::pow(x, n); ))
```
ax3l:
```suggestion
    AMREX_IF_ON_HOST(( return std::pow(x, static_cast<float>(n)); ))
```

Src/Base/AMReX_Math.H
```diff
@@ -250,6 +250,35 @@ constexpr T powi (T x) noexcept
...
+}
+
+AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+double powi (double x, int n) noexcept
+{
+#if defined(AMREX_USE_SYCL)
+    return sycl::pown(x, n);
+#else
+    AMREX_IF_ON_DEVICE(( return ::powi(x, n); ))
+    AMREX_IF_ON_HOST(( return std::pow(x, n); ))
```
ax3l, line 298:
```suggestion
    AMREX_IF_ON_HOST(( return std::pow(x, static_cast<double>(n)); ))
```

ax3l: optional... ends up here anyway for C++

Src/Base/AMReX_Math.H
ax3l: fixed, comment added

Src/Base/AMReX_Math.H
```diff
@@ -250,6 +250,60 @@ constexpr T powi (T x) noexcept
...
+} // namespace amrex::Math
+
+#if defined(AMREX_USE_CUDA)
+// Forward-declare libdevice integer-exponent intrinsics. They are part of
+// NVIDIA's libdevice and linked at PTX time but are not declared in standard
+// CUDA cmath headers outside __CUDACC_RTC__ / _LIBCPP_VERSION blocks.
+// See https://docs.nvidia.com/cuda/libdevice-users-guide/__nv_powif.html
+extern "C" __device__ float  __nv_powif (float,  int);
+extern "C" __device__ double __nv_powi  (double, int);
+#endif
```
ax3l, line 262: This could be too much of a hack :D

ax3l: I mean they're documented and ABI-stable. The references only appear inside AMREX_IF_ON_DEVICE so the host pass doesn't see them.

ax3l:
GPU CI failure seems unrelated :tada:

WeiqunZhang:
Claude says,

---

Thanks for the PR! I have some concerns about using `__nv_powif` / `__nv_powi` directly, and would like to suggest an alternative approach.

### 1. Portability

The `__nv_*` symbols live in NVIDIA's libdevice bitcode and are not part of the public CUDA Math API. The documented, stable entry points are `powf`/`__powf`/`pow`/etc.; the `__nv_*` names are internal implementation symbols that NVCC happens to lower to. NVIDIA has presumably chosen not to expose these in the public API for good reason — most likely so they retain the freedom to rename, re-signature, remove, or change the semantics of these functions across toolkit versions without breaking user code. By calling them directly, we're taking on a compatibility risk that NVIDIA explicitly hasn't promised to manage for us.

### 2. Fast Math Interaction

We enable `-use_fast_math` by default in CUDA release builds, which rewrites `powf(x, y)` to `__powf(x, y)` ≈ `__exp2f(y * __log2f(x))`. For larger integer exponents, this constant-cost log/exp approach is likely *faster* than the `O(log n)` repeated-squaring that `__nv_powif` performs — so using `__nv_powif` may actually be a pessimization in release builds for non-tiny exponents (and it doesn't help in debug builds where performance matters less anyway). The win for `__nv_powif` is really only for small `|n|`.

### 3. Suggested Alternative: Roll Our Own

Given (1) and (2), I'd suggest we just implement `amrex::Math::powi` ourselves. It's a handful of lines, fully portable, works on host and device with no `#ifdef` gymnastics, and the compiler will fully unroll it when the exponent is a compile-time constant.

```cpp
namespace amrex::Math {

// Integer power via exponentiation by squaring.
// Best suited for small |n| (say, |n| <~ 20). For larger exponents,
// std::pow / powf (especially under fast math, which lowers to
// __powf = exp2(y*log2(x))) is generally faster, though less accurate
// and undefined for negative bases.
// Unlike __powf, this works correctly for negative bases.
template <typename T>
constexpr T powi (T x, int n) noexcept
{
    static_assert(std::is_floating_point_v<T>);
    T base = (n < 0) ? T(1)/x : x;
    unsigned int e = (n < 0) ? static_cast<unsigned int>(-n)
                             : static_cast<unsigned int>(n);
    T result = T(1);
    while (e) {
        if (e & 1U) { result *= base; }
        e >>= 1;
        if (e) { base *= base; }
    }
    return result;
}

} // namespace amrex::Math
```

### Benchmark Suggestion

Before merging, it'd be worth running a quick microbenchmark on a representative GPU comparing:

- `amrex::Math::powi(x, n)` (the template above)
- `__nv_powif(x, n)` / `__nv_powi(x, n)`
- `powf(x, (float)n)` with and without `-use_fast_math`

across `n ∈ {2, 3, 5, 10, 20, 50}`. My expectation is:

- For `n ≤ ~8`, the template wins or ties (compiler unrolls and strength-reduces).
- For moderate `n`, `__nv_powi*` is competitive with the template.
- For larger `n` under fast math, `__powf` beats both.

If that holds, the template covers the use case `__nv_powi*` is actually good at, with none of the portability baggage.

WeiqunZhang:
> GPU CI failure seems unrelated 🎉

#5449.

WeiqunZhang:
I showed Claude some code from ROCm's github repo. It says,

Nice find — this is from OCML's `pownF.cl` (or equivalent), implementing `pown(float, int)`. Let me walk through it.

## High-Level Strategy

Despite the integer exponent, this implementation does **not** use exponentiation by squaring. It computes `pow(x, n)` via the standard `exp(n * log(|x|))` formulation, then patches up the sign and edge cases afterward. The integer-ness of `n` is exploited only for:

1. Determining the sign of the result (odd vs. even exponent).
2. Splitting `n` into high/low parts for an extended-precision multiply with `log(|x|)`.

So this is essentially the same algorithm as the general `pow`, with integer-aware fixups. That's a bit surprising if you expected repeated squaring — apparently AMD decided the log/exp path is faster on GCN/RDNA for the general case than a loop with a data-dependent trip count (which would hurt SIMT execution).

## `MATH_MANGLE(pown)` — The Entry Point

```c
if (ny == 0)
    x = 1.0f;
```

Handles `pow(x, 0) = 1` by forcing the base to 1 (so `log(1)*0 = 0` and `exp(0) = 1` falls out naturally without a special branch later). This works even for `x = 0` or `x = NaN`, matching the IEEE convention that `pow(anything, 0) = 1`.

```c
float expylnx = compute_expylnx_int(x, ny);
return pown_fixup(x, ny, expylnx);
```

Compute the magnitude via `exp(ny * ln(x))`, then apply sign and edge-case fixups.

## `compute_expylnx_int` — The Core Computation

```c
if (UNSAFE_MATH_OPT())
    return fast_expylnx(x, (float)ny);
```

When unsafe/fast math is enabled (analog of `-use_fast_math`), it takes the fast path: just cast `ny` to float and use the fast `exp(y*log(x))` routine. Same shortcut as NVIDIA's `__powf`.

```c
float ax = BUILTIN_ABS_F32(x);
```

Take `|x|` because `log` requires a positive argument; sign is restored later in `pown_fixup`.

```c
int nyh = ny & 0xffff0000;
float2 y = fadd((float)nyh, (float)(ny - nyh));
```

This is the clever bit. A 32-bit integer can have up to ~31 significant bits, but a `float` only has 24 bits of mantissa, so converting a large `ny` to `float` loses precision. To avoid that, `ny` is split into:

- `nyh` — the high 16 bits (low 16 zeroed out).
- `ny - nyh` — the low 16 bits.

Each half fits exactly in a `float`, and `fadd` packs them into a `float2` representing the sum as an unevaluated high+low pair (a "double-float" or "two-float" extended-precision number). This preserves all of `ny`'s precision going into the multiply.

```c
return MATH_PRIVATE(expep)(omul(y, MATH_PRIVATE(epln)(ax)));
```

Reading inside-out:

- `epln(ax)` — extended-precision natural log of `|x|`, returning a `float2` (high+low pair). Higher accuracy than a plain `logf`.
- `omul(y, ...)` — extended-precision multiply of the two `float2` values (`ny * ln|x|`).
- `expep(...)` — extended-precision exponential of that product, returning a regular `float`.

So the precise path is: split `ny` exactly → multiply by extended-precision `ln(x)` → take extended-precision `exp`. This gives much better accuracy than the naive `expf(ny * logf(x))`, which would lose bits at both the conversion and the multiply.

## `pown_fixup` — Sign and Edge Cases

```c
bool is_odd_y = ny & 1;
```

Parity of the exponent decides whether negative bases produce negative results.

```c
float ret = BUILTIN_COPYSIGN_F32(expylnx, is_odd_y ? x : 1.0f);
```

`copysign(magnitude, source)` returns `magnitude` with the sign of `source`. So:

- Odd exponent: take the sign of `x` (negative base → negative result).
- Even exponent: take the sign of `+1.0` (always positive result).

This is how the function correctly handles `pow(-2.0f, 3) = -8.0f` even though `log(-2)` would be NaN — the magnitude came from `log(|x|)` and the sign is restored here.

```c
if (BUILTIN_ISINF_F32(x) || x == 0.0f)
    ret = BUILTIN_COPYSIGN_F32((x == 0.0f) ^ (ny < 0) ? 0.0f : PINF_F32,
                               is_odd_y ? x : 0.0f);
```

Special cases for `x = 0` or `x = ±∞`, where the log/exp path doesn't give the right answer. The magnitude logic via XOR:

- `x == 0` and `ny > 0` → false XOR false → false → `+∞`? Wait, let me re-read… the result is `(x==0) ^ (ny<0) ? 0.0f : PINF_F32`.
  - `x = 0`, `ny > 0` (so `ny<0` false): `true ^ false = true` → `0.0f`. ✓ (`0^positive = 0`)
  - `x = 0`, `ny < 0`: `true ^ true = false` → `+∞`. ✓ (`0^negative = ∞`, division by zero)
  - `x = ∞`, `ny > 0`: `false ^ false = false` → `+∞`. ✓
  - `x = ∞`, `ny < 0`: `false ^ true = true` → `0.0f`. ✓ (`∞^negative = 0`)

Then `copysign(magnitude, is_odd_y ? x : 0.0f)` applies the sign. For odd exponents the sign of `x` propagates (giving `-0`, `-∞` for negative-zero or negative-infinity inputs); for even exponents it's forced positive via `copysign(_, +0)`.

## Takeaways Relevant to the amrex PR

A few things stand out that reinforce the earlier review:

1. **AMD's `pown` is essentially `exp(n * log(x))` with integer-aware sign handling and an extended-precision multiply for large `n`.** It is *not* exponentiation by squaring. That means on AMD hardware, calling the vendor `pown` for small integer exponents is likely *slower* than a hand-rolled `powi` that does 2–4 multiplies.

2. **Under fast math (`UNSAFE_MATH_OPT`), it collapses to the same `fast_expylnx` path as a general `pow`.** So once again, for small `n` and fast math, a hand-rolled squaring loop will beat the vendor function on AMD too.

3. **It handles negative bases correctly** (unlike NVIDIA's `__powf`), thanks to the `copysign` trick. A portable `amrex::Math::powi` using repeated squaring also handles negative bases naturally.

4. **The portability argument gets stronger.** NVIDIA does `__nv_powi*` one way (squaring, presumably), AMD does it another way (log/exp with extended precision). If amrex calls vendor symbols directly, the accuracy and performance characteristics will differ silently between backends. A unified hand-rolled implementation gives consistent behavior everywhere.

WeiqunZhang:
Me: How do you know how Nvidia implements __nv_powi*?

Claude: You caught me — I don't actually know that. I was speculating, and I shouldn't have stated it as if it were fact.

WeiqunZhang:
Me: Also your claim of NVIDIA's __powf does not handle negative base correctly is most most most likely wrong.

Claude: You're right to push back — I shouldn't have stated that confidently either.

WeiqunZhang:
I think we should ignore most of the comments from Claude. My only concern is the portability of `__nv_powi*`.

WeiqunZhang:
Since `__nv_powi` and `__nv_powif` have existed for 15+ years, I guess it's pretty safe to use it.

WeiqunZhang (approved):

ax3l:
I would be curious to micro benchmark this... Unfortunately still in proposal rush for 2 more weeks.

PR #5447, opened 2026-05-16
Sphinx build: Stabilize PDF metadata across builds

WeiqunZhang:
Setting SOURCE_DATE_EPOCH before make latexpdf so pdfTeX produces byte-identical amrex.pdf when the documentation source hasn't changed, avoiding a new ~7 MB blob in git on every deploy.  The fixed epoch (441763200 = 1984-01-01 UTC) pins CreationDate, ModDate, and /ID.

### DIFF
--- a/build_docs.sh
+++ b/build_docs.sh
@@ -27,6 +27,9 @@ cd ..
 # Build sphinx
 cd Docs/sphinx_documentation
 echo "Build the Sphinx documentation for Amrex."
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
Note that in AMReX-Codes.github.io's git repo, there are more than 5000 versions of amrex.pdf, mainly because of the metadata difference.

asalmgren (approved):

PR #5449, opened 2026-05-17
CTest: Disable CheckpointRestartDualGrid for GPU builds

WeiqunZhang:
This new test introduced in #5440 needs to be fixed for GPU builds.


ax3l (approved):

ax3l:
LGTM, let's track as a bug though so we don't forget. Cc @atmyers

PR #5450, opened 2026-05-18
generalize constants with Real() or _rt

asalmgren:
## Checklist

The proposed changes:


Src/EB/AMReX_EB_Slopes_3D_K.H.new
WeiqunZhang, line 1: Was this accidentally committed?

PR #5453, opened 2026-05-18
Doxygen: No need to copy most of the xml files.

WeiqunZhang:
These files could allow amrex doxygen to be inlined into other sites, but no other sites are currently use these xml files. By not providing them, we could make the AMReX_Codes.github.io git repo much smaller.


WeiqunZhang:
See also #5447.

asalmgren (approved):

PR #5454, opened 2026-05-18
Curl-Curl Solver: add support for inhomogeneous Dirichlet boundary

WeiqunZhang:
Previously we assume it's always homogeneous at the Dirichlet boundary.


WeiqunZhang:
@JustinRayAngus Please give this a try.

JustinRayAngus:
@WeiqunZhang Same result as before.

WeiqunZhang:
Okay. I think the issue is in how this is coupled in WarpX with GMRES, which is a **linear** solver, but the inhomogeneous Dirichlet problem is not linear. So if this is used as a preconditioner for GMRES, the system needs to be linear. Because the issue is probably not in the curl curl solver, let's move our discussion to slack and email.

That being said, this is still useful when this is used as a standalone solver. So I will keep this open.

PR #5457, opened 2026-05-19
Sphinx: Fix copyright year.

WeiqunZhang:
In #5447, we use a fixed date for SOURCE_DATE_EPOCH to avoid duplicated pdf files. But one issue is that the copyright year becomes 1984. In this PR, we change it the current year. The meta-data in pdf files will still be stable enough, because the date only changes once a year.


WeiqunZhang:
Let's try to fix the pdf duplication issue another way.

PR #5462, opened 2026-05-28
Fix cross-level rank lookup in sumNeighborsCPU

desjardi:
## Summary

`sumNeighborsCPU` looks up the owning rank with `ParticleDistributionMap(lev)`, but `dst_grid = tag.src_grid` indexes the **owner** level's BoxArray (= `tag.src_level`). For same-level entries this is harmless because `lev == tag.src_level`; cross-level entries read the wrong DM and can OOB when DM sizes differ across levels. `GetParticles(dst_level)` a few lines below already uses `dst_level` for the destination tile: this just makes the rank lookup consistent with it.

## Additional background

Found while running a multi-level `NeighborParticleContainer` with `setEnableInverse(true)`; without the fix `sumNeighbors` aborts in debug.

The proposed changes:


atmyers (commented):
Good catch!

atmyers (approved):

PR #5468, opened 2026-06-02
C++20: Remove artificial member-template parameters

WeiqunZhang:
Replace member function templates whose template parameters exist solely to mirror enclosing class-template parameters with constrained non-template members. For the few cases where real call-time template parameters remain alongside artificial ones, drop only the artificial parameters.

The pattern before C++20 required a dummy template parameter (e.g., template<int MM=NRows>) just to use std::enable_if_t. C++20 allows trailing requires on non-template members, making these helpers unnecessary.

Co-Authored-By: deepseek-v4-pro[1m]

### DIFF
--- a/.github/workflows/apps.yml
+++ b/.github/workflows/apps.yml
@@ -124,10 +124,8 @@ jobs:
     - name: Checkout pyamrex
       uses: actions/checkout@v6
       with:
-        # TODO: revert to AMReX-Codes/pyamrex @ development once the constraints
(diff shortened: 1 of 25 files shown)


ax3l (approved):
Neat!! LGTM, way cleaner.

ax3l:
Why did the Ascent & Catalyst CI runners not run? o.0

WeiqunZhang:
That's a good question.

PR #5471, opened 2026-06-02
OpenMP: tolerate invalid amrex.omp_threads values

### LINKED ISSUE #5470: `amrex.omp_threads` with a non-numeric value throws an uncaught exception
- **Type**: `correctness`
- **Severity**: `medium`
- **Component**: `OpenMP (amrex::OpenMP::Initialize)`
- **Location**: `Src/Base/AMReX_OpenMP.cpp:159-188`

## Problem
`Initialize()` parses the `amrex.omp_threads` ParmParse value and is clearly
designed to handle a bad value gracefully — the fallback branch prints a warning
("has an unknown value ... try system, nosmt, or a positive integer"), and the
helper returns a `std::optional<int>` as if to signal parse failure:

```cpp
auto to_int = [](std::string const & str_omp_threads) {
    std::optional<int> num = std::stoi(str_omp_threads);   // throws on non-numeric input
    return num;
};
...
} else {
    std::optional<int> num_omp_threads = to_int(omp_threads);
    if (num_omp_threads.has_value() && num_omp_threads.value() > 0) {
        omp_set_num_threads(num_omp_threads.value());
    }
    else {
        if (amrex::system::verbose > 0) {
            amrex::Print() << "amrex.omp_threads has an unknown value: "
                           << omp_threads
                           << " (try system, nosmt, or a positive integer)\n";
        }
    }
}
```

But `std::stoi` does not report failure through the optional — it **throws**
`std::invalid_argument` on a non-numeric string (and `std::out_of_range` on an
overflowing one). There is no `try`/`catch` in `Initialize()`, so a value such as
`amrex.omp_threads = nosmtt` (a typo) or any non-numeric string throws an
uncaught exception during initialization and aborts the run, instead of emitting
the intended warning and falling back to the default thread count.

The `std::optional` is effectively dead: when `stoi` succeeds, `has_value()` is
always true, so the warning branch is only reachable for parseable-but-non-positive
inputs like `0` or `-4`. Purely non-numeric values never reach it.

## Impact
- A typo or non-numeric `amrex.omp_threads` aborts the application at startup
  rather than producing the documented warning and using the default.
- The intended graceful-fallback path is unreachable for the most common kind of
  bad input (non-numeric strings).

## Suggested patch
```diff
--- a/Src/Base/AMReX_OpenMP.cpp
+++ b/Src/Base/AMReX_OpenMP.cpp
@@
-        auto to_int = [](std::string const & str_omp_threads) {
-            std::optional<int> num = std::stoi(str_omp_threads);
-            return num;
-        };
+        auto to_int = [](std::string const & str_omp_threads) -> std::optional<int> {
+            try {
+                return std::stoi(str_omp_threads);
+            } catch (...) {
+                return std::nullopt;
+            }
+        };
```

With this change a bad value yields an empty optional, so the existing
`else` branch prints the warning and AMReX keeps the default thread count.

Prepared by Claude (Opus 4.8)

WeiqunZhang:
std::stoi throws std::invalid_argument for non-numeric input and std::out_of_range for overflowing input. Because OpenMP::Initialize did not catch those exceptions, invalid amrex.omp_threads values could abort startup before reaching the existing warning/fallback path.

Parse amrex.omp_threads with stream extraction and return nullopt when conversion fails. This lets non-numeric or overflowing values reach the existing warning path and preserves the default OpenMP thread count instead of throwing during Initialize.

Use stream parsing instead of try/catch around std::stoi so this code remains compatible with builds that disable C++ exception handling. It also keeps input validation in the existing optional/fallback flow.

Close #5470.

### DIFF
--- a/Src/Base/AMReX_OpenMP.cpp
+++ b/Src/Base/AMReX_OpenMP.cpp
@@ -156,9 +156,17 @@ namespace amrex::OpenMP
         std::string omp_threads = "system";
         pp.queryAdd("omp_threads", omp_threads);
 
-        auto to_int = [](std::string const & str_omp_threads) {
-            std::optional<int> num = std::stoi(str_omp_threads);
-            return num;
+        auto to_int = [](std::string const & str_omp_threads) -> std::optional<int> {
+            int num = 0;
+            std::istringstream iss(str_omp_threads);
+            if (iss >> num) {
+                // Require a complete integer token; otherwise "4xxxx" would parse as 4.
+                char c;
+                if (!(iss >> c)) {
+                    return num;
+                }
+            }
+            return std::nullopt;
         };
 
         if (omp_threads == "system") {


ax3l (approved):
Looks good to me and Codex :D

PR #5483, opened 2026-06-09
Add Fortran Interface to CI

WeiqunZhang:
In our CMake build system, the AMReX_FORTRAN_INTERFACES option is off by default. None of the CIs was testing the Fortran interface. So we enable it in tests that Fortran language is enabled already.


WeiqunZhang:
`Tests/FortranInterface/Advection_F` is disabled for now if particle is enabled. See #5485.

asalmgren (approved):

PR #5487, opened 2026-06-09
Make memory management safer in Fortran APIs

WeiqunZhang:
Add destroy() calls before overwriting C++ pointers in _build, _assign, _install, _clone, and _move methods across the Fortran interfaces.  The destroy functions are no-ops when no C++ object is owned, so they are safe to call on freshly constructed or already-destroyed objects.

This guards against leaks if users call _build or similar methods multiple times without an intervening _destroy.


WeiqunZhang:
Let's wait for #5483 so that we can run some CI tests.

asalmgren (approved):

PR #5496, opened 2026-06-17
Fix dual_grid restart for particle containers constructed with AmrCore*

atmyers:
The problem, which as exposed by #5440, is that when we have constructed a ParticleContainer by giving it an AmrCore* so that it always tracks the fluid grids, calling `SetParticleBoxArray` the way we do in AMReX_ParticleIO.H unexpectedly breaks this tracking behavior. This bug was caught by a failing test in NyX. In this PR, we detect whether the ParticleContainer is currently tracking an AmrCore and clear the temporary BoxArrays and DistributionMappings instead of overriding.

For an alternative but less defensive fix, see: https://github.com/AMReX-Codes/amrex/pull/5497

The proposed changes:

### DIFF
--- a/Src/AmrCore/AMReX_AmrParGDB.H
+++ b/Src/AmrCore/AMReX_AmrParGDB.H
@@ -76,6 +76,11 @@ public:
     //! Clear the cached particle Geometry on level \p level.
     void ClearParticleGeometry (int level) override;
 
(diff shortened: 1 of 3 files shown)


atmyers:
Closing this, as we have opted to merge PR #5497 instead.

PR #5497, opened 2026-06-17
Remove dual_grid path from particle restart

atmyers:
See #5496 for more background.

It is not clear to me that we need the dual_grid restart option at all. The caller of the restart function should not care what grids are actually stored in the restart file. The restart function should always use whatever grid layout is stored in the checkpoint file to read the particles, but not change the ParticleContainer grids at all. Then, Redistribute() will move the particles to the proper place on the actual grids used by simulation.

The proposed changes:


Src/Particle/AMReX_ParticleIO.H
atmyers, line 865: Note, we may want to restore the behavior of the old reader in parallel here. But first let's gauge the correctness of this approach.

asalmgren (approved):

PR #5498, opened 2026-06-18
Reduce2Plane with Guards

ax3l:
## Summary

Include potentially deposited values in guards in the reduction, if specified by the user. Needed for BLAST ImpactX 2D charge deposition projection from 3D.

## Additional background

Bug in https://github.com/BLAST-ImpactX/impactx/issues/1512

## Checklist

The proposed changes:


ax3l:
I am still a bit unsatisfied how complicated the reduction user-interface is :D

Being aware of staggering, node double ownership, synchronization, etc. Aren't parallel ND-arrays in co-array Fortran et al the oldest thing in the world? There must be simpler syntax that we can adopt that just does the right thing for MPI-parallel axis projections a la [np.sum(a, axis=...)](https://numpy.org/doc/stable/reference/generated/numpy.sum.html)

WeiqunZhang (approved):

PR #5507, opened 2026-06-23
Add support for RZ + EB with sigma MF in `MLEBNodeFDLaplacian`

roelof-groenewald:
## Summary

Currently the `MLEBNodeFDLaplacian` does not support a spatially varying sigma for RZ simulations. This PR adds functionality for this when compiled with EB support. Specifically, this is needed for an extension to the WarpX effective potential solver (see https://github.com/BLAST-WarpX/warpx/pull/6961).

## Additional background

I tested the new functionality by comparing the electrostatic solution in WarpX from using the "lab frame" electrostatic solver (uses constant sigma values) and the "effective potential" solver (which uses a sigma MF). The sigma MF was set to 1 everywhere to get the same system as with the constant sigma.

Result from this PR:
image

## Checklist

The proposed changes:

### DIFF
--- a/Src/LinearSolvers/MLMG/AMReX_MLEBNodeFDLap_2D_K.H
+++ b/Src/LinearSolvers/MLMG/AMReX_MLEBNodeFDLap_2D_K.H
@@ -467,6 +467,71 @@ void mlebndfdlap_sig_gsrb (int i, int j, int k, Array4<Real> const& x,
     }
 }
 
+AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+void mlebndfdlap_sig_adotx_rz (int i, int j, int k, Array4<Real> const& y,
+                                Array4<Real const> const& x,
+                                Array4<int const> const& dmsk,
+                                Array4<Real const> const& sig,
+                                Real dr, Real dz, Real rlo, Real alpha) noexcept
+{
+    Real const r = rlo + Real(i)*dr;
+    if (dmsk(i,j,k) || (r == Real(0.0) && alpha != Real(0.0))) {
+        y(i,j,k) = Real(0.0);
+    } else {
+        Real const rp = r + Real(0.5)*dr;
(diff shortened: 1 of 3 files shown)


WeiqunZhang:
@roelof-groenewald Let's remove this outdated comment from .AMReX_MLEBNodeFDLaplacian.H.
```
// New feature: for non-RZ, sigma can also be a single-component
// cell-centered multifab.
```

WeiqunZhang:
It appears that the only case that does not support variable sigma is now 2d RZ without EB. @roelof-groenewald Would you be able to add support for that case too?

WeiqunZhang:
If you don't want to add the support for 2D RZ without EB in this PR, that's fine. But we need to add an assert in this PR in maybe Fapply, because a previous assert has been removed in this PR. Also update the comments in the header.

roelof-groenewald:
> It appears that the only case that does not support variable sigma is now 2d RZ without EB. @roelof-groenewald Would you be able to add support for that case too?

@WeiqunZhang, yes I can add the RZ + non-EB support too, but in the meantime I addressed your other comments so that this PR can be merged. I'll open another PR for the RZ + non-EB support when that is ready (I'll aim to get to that towards the end of the week).

WeiqunZhang:
@roelof-groenewald Unfortunately, we have to address the issue in this PR, because we do something like this.
```
#ifdef AMREX_USE_EB
    if (cutfab) { // could be false if this fab does not have a single cut cell.
    } else
#endif
   {
        // EB build could reach this branch.
   }
```

roelof-groenewald:
Yes, good catch @WeiqunZhang! I now added the non-EB functions with sigma MF for RZ as well. To test this I took the same setup I used before and just removed the EB. With the new changes I get the same result when using the constant sigma (value 1) as when using the sigma MF (with all values 1). I also show a case with spatially varying sigma from the effective potential solver. The result is a similar but somewhat smoother `phi` (which is generally what we would expect from that method).
image

I also completely removed the abort now, and modified the comment in the header file.

WeiqunZhang (approved):

PR #5515, opened 2026-06-25
Fix parser node alignment for 32-bit builds

WeiqunZhang:
On 32-bit platforms, pointer alignment can be only 4 bytes while double still requires 8-byte alignment, so using pointer-based parser_node alignment can underalign parser_number. Align parser_node to the max of pointer and double alignment.


ax3l:
Thank you! That should do it, I'll give it a final test just to be sure.

ax3l (approved):

PR #5520, opened 2026-06-26
Make the "local" and "nGrow" arguments IntVects instead of ints for Redistribute

atmyers:
This could be useful for cases where the particle motion is anisotropic. 

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_ParticleCommunication.H
+++ b/Src/Particle/AMReX_ParticleCommunication.H
@@ -508,29 +508,6 @@ public:
         buildMPIStart(pc, pc.BufferMap(), m_superparticle_size);
     }
 
-    template <class PC>
-    requires (IsParticleContainer<PC>::value)
-    void build (const PC& pc,
-                const ParticleCopyOp& op,
-                const Vector<int>& int_comp_mask,
-                const Vector<int>& real_comp_mask,
-                IntVect max_cells_moved)
-    {
-        build(pc, op, int_comp_mask, real_comp_mask,
-              max_cells_moved.allGT(0), max_cells_moved);
-    }
-
-    template <class PC>
-    requires (IsParticleContainer<PC>::value)
-    void build (const PC& pc,
-                const ParticleCopyOp& op,
-                const Vector<int>& int_comp_mask,
-                const Vector<int>& real_comp_mask,
-                int local)
-    {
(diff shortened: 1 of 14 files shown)


atmyers:
We need to update the Python wrappers too: https://github.com/AMReX-Codes/pyamrex/pull/590

WeiqunZhang:
If someone calls Redistribute with local = IntVect(2,2,0), what is the expected behavior from the user's perspective? Do they want to say the particles will move up to two cells in the x- and y- directions and zero cells in the z-direction? Or do they want to say the last 0 means it can move any number of cells in the z-direction?

@RemiLehe What do you expect in WarpX?

atmyers:
> If someone calls Redistribute with local = IntVect(2,2,0), what is the expected behavior from the user's perspective? Do they want to say the particles will move up to two cells in the x- and y- directions and zero cells in the z-direction? Or do they want to say the last 0 means it can move any number of cells in the z-direction?
> 
> @RemiLehe What do you expect in WarpX?

The current version of this adds a bool flag to new overload. Now, the "local" argument is which algorithm to use, while "max_cells_moved" is the number of cells to use in each direction if "local" is true.

Src/Particle/AMReX_ParticleCommunication.H
```diff
@@ -507,6 +508,29 @@ public:
...
+    template <class PC>
+    requires (IsParticleContainer<PC>::value)
+    void build (const PC& pc,
+                const ParticleCopyOp& op,
+                const Vector<int>& int_comp_mask,
+                const Vector<int>& real_comp_mask,
+                IntVect max_cells_moved)
+    {
+        build(pc, op, int_comp_mask, real_comp_mask,
+              max_cells_moved.allGT(0), max_cells_moved);
```
WeiqunZhang: Should this be `allGE`?

WeiqunZhang: Actually should it be `.allGE(0) && != 0`?

WeiqunZhang: Or should this overload be dropped. It seems confusing.

atmyers: ParticleCopyPlan::build is not meant to be a user function, so I have removed these convenience overloads (which I agree are confusing) and just used the full, 6-argument version everywhere.

WeiqunZhang (approved):

PR #5522, opened 2026-06-27
placementNew/placementDelete: simplify with if constexpr and trivially_copyable

WeiqunZhang:
Replace the three requires-overloads of placementNew with a single function using if constexpr.  Use std::is_trivially_copyable instead of std::is_trivially_default_constructible — trivially copyable types (including arithmetic and POD types) begin their lifetime implicitly once storage is allocated and need no placement-new.  Add matching Doxygen comments for both functions.

Breaking change: BaseFab<T> no longer default-constructs trivially copyable element types. Code that relied on BaseFab<T> allocation/resize to initialize values, including BaseFab<GpuComplex<T>> zero-initialization, must explicitly fill or write the data before reading it. This is consistent with the existing behavior for arithmetic types, which are not initialized on allocation; in that sense, the new behavior is arguably the less surprising one for BaseFab storage.

### DIFF
--- a/Src/Base/AMReX_BaseFab.H
+++ b/Src/Base/AMReX_BaseFab.H
@@ -94,48 +94,51 @@ makeArray4 (T* p, Box const& bx, int ncomp) noexcept
     return Array4<T>{p, amrex::begin(bx), amrex::end(bx), ncomp};
 }
 
+/**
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
Oh, there is EBCellFlagFab that depends on the default ctor.

PR #5524, opened 2026-06-27
[CMake] Allow building for Blackwell+ GPUs

BenWibking:
## Summary
Updates the CMake configuation so it allows `AMREX_CUDA_ARCH` to be >= 10 without silently discarding this value. This is required to build AMReX for Blackwell or newer GPUs, which have `sm_100` or higher.

## Checklist

The proposed changes:


BenWibking:
@ax3l If there is a simpler way to do this, let me know. The CUDA architecture helper seems to use a lot of old CMake-isms that might have more modern replacements.

ax3l:
Thanks, this is super important. I am on PTO this week, but can take it on next week.

BenWibking:
I realized this was actually the same issue, and might be a cleaner solution: https://github.com/AMReX-Codes/amrex/pull/4572

BenWibking:
Also related: https://github.com/AMReX-Codes/amrex/issues/3948.

BenWibking:
Closing in favor of https://github.com/AMReX-Codes/amrex/pull/4773.

PR #5525, opened 2026-06-28
Docs: document the TinyProfileParser script

### LINKED ISSUE #4160: document TinyProfileParser
There is a Python script to convert TinyProfiler output into JSON format: https://github.com/AMReX-Codes/amrex/blob/development/Tools/TinyProfileParser/profileparser.py

It is not documented in the AMReX docs.

SuryanshSS1011:
Closes #4160.

The `Tools/TinyProfileParser/profileparser.py` script wasn't mentioned anywhere in the docs, so this adds a short subsection to the Tiny Profiling page covering it.

The new "Converting TinyProfiler Output for Hatchet" subsection goes right after "Hot Spots and Load Balance", since the script consumes TinyProfiler stdout. It covers:
- what the script does (parses TinyProfiler stdout into a JSON file for Hatchet)
- how to run it, including the optional output-directory argument
- the metrics it emits per function (n_calls, the excl_* set, and the incl_* set)
- a note that it expects a single plain run-end summary (relevant if you use BL_PROFILE_TINY_FLUSH)

I went with a subsection rather than a separate page since it's one small utility and this keeps it next to the profiler output it parses. Happy to split it into its own page under the chapter toctree instead if you'd prefer that.

Docs-only change; the wording is checked against the current script.

### DIFF
--- a/Docs/sphinx_documentation/source/AMReX_Profiling_Tools.rst
+++ b/Docs/sphinx_documentation/source/AMReX_Profiling_Tools.rst
@@ -134,6 +134,56 @@ loaded processes have to wait for messages sent by the heavily loaded
 processes. See also :ref:`sec:profopts` for a diagnostic option that may
 provide more insight on the load imbalance.
 
+.. _sec:tiny:profparser:
+
+Converting TinyProfiler Output for Hatchet
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


SuryanshSS1011:
Friendly bump on this one. It's a small docs-only change (documents the existing TinyProfileParser script) and has been open a couple of weeks. Is anyone able to take a look, or is there a better person to tag?

PR #5538, opened 2026-07-05
Refactor amrex::clz using C++20 std::countl_zero

WeiqunZhang:


WeiqunZhang:
One HIP test failed in the gitlab CI. But it had nothing to do with this PR. It was due to the machine running out of memory. For the CLZ tests, the runs were successful on all three GPUs.

asalmgren (approved):

PR #5539, opened 2026-07-05
Remove amrex::clz

WeiqunZhang:


WeiqunZhang:
I don't think `amrex::clz` is used anywhere. Since C++20 has a constexpr function for counting the leading zeros, I think we should remove this function. @atmyers You added this function. If you think we should keep it, let's refactor it using std::countl_zero (#5538).

asalmgren (approved):

atmyers:
I have no memory of what this was for. Let's remove it, we can add back a version based on std::count_zero if needed.

atmyers (approved):

PR #5548, opened 2026-07-08
Add face-centered EB factories

skang67:
## Summary

This PR adds face-centered (FC) Embedded Boundary (EB) factories.

## Additional background

- Currently, the `EB2` class only supports EB factory on cell-centered (CC) grids.
- The FC EB factories are generated by coarsening the refined CC EB factory data onto staggered grids.
- The FC EB factory data can be accessed through the same wrapper as the CC factory using overloaded constructors.
- Currently, the FC factory must be built after the CC factory.

## Key changes

- Added a new data container `struct FCData` to `class Level`.
- Added `EB2::BuildFC`, which generates FC factories and attaches them to the static `IndexSpace` instance `m_instance`.
- Added `GShopLevel<G>::buildFCData`, which constructs FC factories for each `face_dir` using a refined CC `GShopLevel<G>`.
  - For each level, a temporary refined CC level is created. 
  - `coarsen_from_fine` function is extended (with only three more lines) to merge subcells in a staggered fashion.
- Added an overloading `EBDataCollection` constructor to access the FC EB data.
- A new regression test `Tests/EB/FCFactory` is added.

## Examples

To make factories, call `EB2::BuildFC` after `EB2::Build`.
```
EB2::Build(gshop, geom, ngrow);
EB2::BuildFC();
```

Then, instantiate a factory by specifying the face direction, `idim=0,1,2`:
```
u_factory = std::make_unique<EBFArrayBoxFactory>(
            level, geom, ba, dm,
            Vector<int>{nghost_basic(), nghost_volume(), nghost_full()},
            m_support_level, idim);
```

To access the FC EB data,
```
Array4<const EBCellFlag> u_cellflg = u_factory->getMultiEBCellFlagFab()[mfi].const_array();
Array4<const Real> u_volfrac = u_factory->getVolFrac().const_array(mfi);
Array4<const Real> u_afrac_x = u_factory->getAreaFrac()[0]->const_array(mfi);
Array4<const Real> u_bnorm = u_factory->getBndryNormal().const_array(mfi);
```

## Suggestions for Future Work

- I found that `EBCellFlag::isConnected()` returns true for neighboring zero-volume cells, especially when the neighboring cells are outside the valid region. This behavior may be corrected so that all cells with `volfrac = 0` are disconnected.
- Currently, `ApplyRedistribution` only accepts cell-centered `Box` objects. So, for redistribution of face-centered data, we need to convert the box to (0,0,0) type and adjust its bounds. It may be worth adding flexibility of grid type to the redistribution scheme.

## Checklist

The proposed changes:


WeiqunZhang:
1. Blocker — buildFCData mutates the shared cell-centered level and corrupts its level set

  AMReX_EB2_Level.H, inside GShopLevel<G>::buildFCData:

  m_grids = amrex::coarsen(refined_cc_level->m_grids, 2);
  m_dmap  = refined_cc_level->m_dmap;
  m_levelset.define(amrex::convert(m_grids,IntVect::TheNodeVector()), m_dmap, 1, 0);

  These are the members of the live CC level held in IndexSpaceImp::m_gslevel — the one every existing EBFArrayBoxFactory reads from. Measured on a 64³ sphere:

  CC level grids BEFORE: n=1  minimalBox=((-16,-16,-16) (79,79,79))
  CC level grids AFTER : n=8  minimalBox=((-8,-8,-8) (71,71,71))

  Level::boxArray() — public API — now describes a smaller region than the CC EB data it is supposed to describe. And because m_levelset is rebuilt on those grids with zero ghost cells,
  Level::fillLevelSet falls back to its setVal(-1.0) default outside them:

  sum(LEVELSET over valid+12 ghost)
    before BuildFC = -1210425.062
    after  BuildFC = -1374766.578
    delta          = -164341.5156        (~13.6%)

  So any application that calls EB2::BuildFC() and afterwards constructs a cell-centered factory silently gets wrong level-set data. m_volfrac/m_cellflag/m_areafrac survive only because they
  aren't redefined — the metadata describing them is now wrong. The FC build should write into m_fc_data[face_dir] using local variables and never touch m_grids, m_dmap, or m_levelset.

  2. Blocker — FC area fractions and edge centroids have the wrong index type

  CC   areafrac[0]=(N,C,C)  areafrac[1]=(C,N,C)  edgecent[0]=(C,N,N)
  FC0  areafrac[0]=(C,C,C)  areafrac[1]=(C,C,C)  edgecent[0]=(N,N,N)
  FC1  areafrac[0]=(C,C,C)  areafrac[1]=(C,C,C)  edgecent[0]=(C,C,N)
  FC2  areafrac[0]=(C,C,C)  areafrac[1]=(C,C,C)  edgecent[0]=(C,N,C)

  Every CC definition uses amrex::convert(m_grids, IntVect::TheDimensionVector(idim)) (AMReX_EB2_Level.cpp:69,71,175,177; AMReX_EB2_Level.H:482,484), but buildFCData defines

  Every CC definition uses amrex::convert(m_grids, IntVect::TheDimensionVector(idim)) (AMReX_EB2_Level.cpp:69,71,175,177; AMReX_EB2_Level.H:482,484), but buildFCData defines
  fc_areafrac[idim]/fc_facecent[idim] on bare cell-centered m_grids. Result: all three areafrac[idim] are layout-identical cell-centered arrays, which cannot be right for areas of three different
  face families. Since coarsen_from_fine writes them at nodal indices (xbx = surroundingNodes(bx,0), …), the last face plane in each direction lands in ghost storage rather than valid data —
  only the hard-coded ng = 3 keeps it from being an out-of-bounds write.

  The edge_type construction has the same problem, duplicated in buildFCData and the new EBDataCollection ctor. CC sets exactly one direction to cell (an edge); the PR sets two when idim != 
  face_dir (a face) and none when idim == face_dir (a node). The PR's own usage example — u_factory->getAreaFrac()[0] — hands users a mis-typed array.

  3. Blocker — all-regular geometry aborts

  isAllRegular = 1
  hasFCData(0) after BuildFC = 0
  Assertion `a_level.hasFCData(face_dir)' failed ... SIGABRT

  buildFCData returns early on m_allregular without setting m_built, so EB2::BuildFC() is a silent no-op and the next FC factory aborts. The CC path handles all-regular fine. Any application
  using this crashes the moment the body leaves the domain or a coarse level resolves regular.

  Also needs attention

  - buildCellFlag() at the end of buildFCData acts on the cell-centered arrays. Level::buildCellFlag() (AMReX_EB2_Level.cpp:403-425) does m_areafrac[idim].FillBoundary(...) and writes m_cellflag
  — not fc_areafrac/fc_cellflag. So FC flags never get the build_cellflag_from_ap fixup the CC path depends on, while the CC flags are rebuilt 3 × nlevels times for nothing.
  - 3× redundant refined-level construction. IndexSpaceImp<G>::buildAllFCData() loops face_dir outer / ilev inner and buildFCData unconditionally builds a fresh GShopLevel at 2× resolution (8×
  the cells in 3D) each call. One refined level per level would serve all three directions; EB generation is often the dominant setup cost.
  - Failures are swallowed. m_fc_data[face_dir]->m_built = true is set even when the error reduction is non-zero, so hasFCData() reports success on data that failed to coarsen. (// return error;
  is left commented out.)
  - No tests. ~800 lines of new geometry code with zero coverage. At minimum: FC volume/area fractions against an analytic geometry, plus the all-regular and multi-level paths.
  - EB2::BuildFC() silently no-ops for non-GShop index spaces — IndexSpace::buildAllFCData() has an empty default body, so a checkpoint-file index space fails later with an unhelpful assert
  instead of a clear "not supported".
  - The FC EBDataCollection allocates a level set and setVal(0.0) — reads as "exactly on the boundary" everywhere. Better to leave it unallocated and have getLevelSet() abort for FC factories.
  - const int ng = 3 hard-codes ghost width for all FC data, ignoring m_ngrow; m_ngrow[0] > 0 ? m_ngrow[0] : 4 uses a magic 4.
  - The int face_dir = -1 default argument on coarsen_from_fine changes a hot inner-loop function shared with the CC path. Passing it explicitly at the CC call sites would be safer than a silent
  default.

  Cleanup

  // return error;; Level const* refined_cc_level = nullptr; with a "Get or build" comment but no branch; // Consider delete m_levelset construction, ...; the (ERF EB-FC development) comment
  referencing a downstream project; and a no-op block in the new factory ctor:

  { auto const& ma = ebflags.const_arrays(); amrex::ignore_unused(ma); }

Tests/EB/FCFactory/main.cpp
```diff
@@ -127,7 +127,8 @@ void main_main()
     // Numerical accuracy validation - check volume for each FC factory
     Real dx = geom.CellSize(0), dy = geom.CellSize(1), dz = geom.CellSize(2);
     Real cell_volume = dx * dy * dz;
-    Real vol_analytical = (4.0/3.0) * M_PI * std::pow(sphere_radius, 3);
+    const Real pi = 4.0 * std::atan(1.0);
```
WeiqunZhang: There is `std::numbers::pi_v<Real>` for this.

WeiqunZhang (approved):
There are known issues. The experimental feature is for ERF only.

Issue #4592, opened 2025-07-31
SIMD: Align Particle Properties


ax3l:
For loading/writing SIMD packaged data efficiently, we want to use [stdx::vector_aligned](https://en.cppreference.com/w/cpp/experimental/simd/vector_aligned.html)

For this to work, we need to properly align the AMReX SoA attributes.

By default, we can use the "native" size (in elements) for this, but in practice I implemented #4520 in a way that also multiples of that size can be used in user-code, to make use of ILP. That said, I have not yet seen big benefits of the latter, but likely because I did not benchmark a lot with small sims that can fit all their particle data into L1/L2/L3.

Follow-up to #4520

PR #5562, opened 2026-07-14
SIMD: Vectorized Particle Reductions (ParticleReduceSIMD)

ax3l:
## Summary

Add a SIMD-vectorized CPU path for particle reductions, following the single-source SIMD design (`SIMDindex`, `load_1d`, `AMReX_SIMD=ON`):

- `Src/Base/AMReX_ReduceSIMD.H`: generic host reduction engine evalReduceSIMD<WIDTH>` that accumulates in SIMD registers over full lanes, handles the remainder scalar, and folds into `ReduceData` like the scalar `ReduceOps::eval`. Supports `ReduceOpSum/Min/Max` (enforced at compile time), any `GpuTuple` element types (`MakeSIMDTuple`).
- Src/Particle/AMReX_ParticleReduceSIMD.H: ParticleReduceSIMD (generic ReduceOps/ReduceData form) and `ReduceSumSIMD/ReduceMinSIMD/ReduceMaxSIMD` value helpers, mirroring `ParallelForSIMD` conventions:
  explicit `WIDTH` template parameter (default: native width), same user source compiles to a scalar loop with `AMReX_SIMD=OFF` and to the regular device path on GPUs. The kernel invocation is force-inlined into the accumulation loop (`particle_detail::SIMDKernelBinder`), which keeps the SIMD accumulators out of caller-saved spill/reload traffic, lets loads lower to single `vmovupd`s, and fuses multiply+accumulate into FMAs.
- `Src/Base/AMReX_SIMD.H`: scalar fallbacks for `stdx::reduce/hmin/hmax/min/max` so portable kernels compile with `AMReX_SIMD=OFF`.
- `Tests/Particles/ParticleReduceSIMD`: correctness test + benchmark reproducing the ImpactX `ReducedBeamCharacteristics` kernel (fused 34 sums + 6 mins + 6 maxes over a pure-SoA container, 11 real  components). Compares `SuperParticle`-lambda and `(ptd,i)`-lambda `ParticleReduce` against `ParticleReduceSIMD` at native and 2x width. (2×-native width loses in both precisions (W=16 in SP is 55–60% slower than W=8). Register pressure for the 46-operand kernel; native width as the default confirmed.)
- `Tests/SIMD`: `evalReduceSIMD` unit test with a mixed-type tuple.
- Docs: new "Particle Reductions" section in Particle.rst.

Benchmark (i9-12900H P-core, AVX2+FMA, pinned, 1 thread, GCC 13.3 `-O3 -march=native`, best of 5 batches, ~2e8 particle evaluations per measurement), ns/particle:

### Double precision (native SIMD width 4):
```
  N     ParticleReduce  ParticleReduce  ParticleReduceSIMD  speedup
        (SuperParticle) (ptd,i)         (W=4)               vs (ptd,i)
  1k    7.09            9.62            3.58                2.7x
  10k   7.42            9.45            3.57                2.6x
  100k  7.58            9.58            3.51                2.7x
  1M    7.86            10.69           4.52                2.4x
  10M   7.96            10.77           4.55                2.4x
```
### Single precision (native SIMD width 8):
```
  N     ParticleReduce  ParticleReduce  ParticleReduceSIMD  speedup
        (SuperParticle) (ptd,i)         (W=8)               vs (ptd,i)
  1k    7.27            8.96            1.88                4.8x
  10k   7.64            9.35            1.90                4.9x
  100k  7.78            9.35            1.83                5.1x
  1M    8.08            9.75            2.11                4.6x
  10M   8.13            9.90            2.20                4.5x
```

Peak throughput: ~545 Mp/s in SP, ~285 Mp/s in DP. Twice-native `WIDTH` is consistently slower than native width for this 46-operand kernel (register pressure); native is the right default. With `AMReX_SIMD=OFF`, the same single-source kernel is SLP-auto-vectorized by GCC thanks to the force-inlined invocation chain and, at 7.0-7.4 ns/particle (DP), is now the *fastest* scalar form — faster than both classic `ParticleReduce` lambda styles. Single-source kernels win in every build configuration.

## Additional background

ImpactX is in some of our science cases spending 50% of its runtime in those AMReX reduces: https://github.com/BLAST-ImpactX/impactx/issues/1102

Assembly-level analysis of the kernel variants (SLP vectorization of the SuperParticle form, call-boundary accumulator spills before the force-inline fix): https://github.com/AMReX-Codes/amrex/pull/5562#discussion_r3592507831

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

## Checklist

The proposed changes:

### DIFF
--- a/Docs/sphinx_documentation/source/Particle.rst
+++ b/Docs/sphinx_documentation/source/Particle.rst
@@ -507,6 +507,105 @@ In this way, code can be written that is agnostic as to the data layout. For mor
 of pure SoA particles, please see the SOA tests in :cpp:`amrex/Tests/Particles/`, or refer
 to `WarpX <https://github.com/BLAST-WarpX/warpx>`_, `Hipace++ <https://github.com/Hi-PACE/hipace>`_, or `ImpactX <https://github.com/BLAST-ImpactX/impactx>`_, which use this type of particle container.
 
+.. _sec:Particles:Reductions:
+
+Particle Reductions
+===================
+
+AMReX provides functions to compute reductions (sums, minima, maxima, and
+logical and/or) over the particles in a :cpp:`ParticleContainer`, on CPUs and
+GPUs alike. The simple entry points :cpp:`amrex::ReduceSum`,
+:cpp:`amrex::ReduceMin`, :cpp:`amrex::ReduceMax`,
+:cpp:`amrex::ReduceLogicalAnd` and :cpp:`amrex::ReduceLogicalOr` reduce a
+single quantity computed per particle by a user-provided callable. The general
+:cpp:`amrex::ParticleReduce` computes many reductions in a single pass over
+the particles, combining :cpp:`amrex::ReduceOps` (e.g., several sums, minima
+and maxima at once) into one :cpp:`amrex::ReduceData` result tuple.
+
+The per-particle callable can take the particle in several forms; the most
+efficient one takes a :cpp:`ConstParticleTileData` and the particle index,
+which reads the particle components directly from the underlying (SoA) arrays:
+
+.. highlight:: c++
+
+::
+
+    using ConstPTDType = typename PC::ConstPTDType;
+    amrex::ReduceOps<ReduceOpSum, ReduceOpMin, ReduceOpMax> reduce_ops;
+    auto r = amrex::ParticleReduce<ReduceData<amrex::Real, amrex::Real, int>> (
+                 pc, [=] AMREX_GPU_DEVICE (const ConstPTDType& ptd, const int i) noexcept
+                               -> amrex::GpuTuple<amrex::Real, amrex::Real, int>
+             {
+                 const amrex::Real a = ptd.rdata(1)[i];
+                 const amrex::Real b = ptd.rdata(2)[i];
+                 const int c = ptd.idata(1)[i];
+                 return {a, b, c};
+             }, reduce_ops);
+
+Note that no MPI reduction is performed at the end of these operations; call
+the reductions in :cpp:`ParallelDescriptor` manually if you want that behavior.
+
+.. _sec:Particles:SIMDReductions:
+
+SIMD-Vectorized Particle Reductions
+-----------------------------------
+
+When AMReX is compiled with SIMD support (CMake option ``AMReX_SIMD=ON``, see
+:ref:`sec:build:cmake`), the entry points :cpp:`amrex::ParticleReduceSIMD`,
+:cpp:`amrex::ReduceSumSIMD`, :cpp:`amrex::ReduceMinSIMD` and
+:cpp:`amrex::ReduceMaxSIMD` evaluate the reduction vectorized over ``WIDTH``
+particles at a time, using SIMD registers as accumulators. The same user code
+compiles and runs unchanged in every configuration:
+
+* ``AMReX_SIMD=ON`` (CPU): vectorized main loop plus a scalar remainder loop,
+* ``AMReX_SIMD=OFF`` (CPU): a scalar loop, semantically identical to
+  :cpp:`amrex::ParticleReduce`,
+* GPU builds: the regular device reduction.
+
+The callable is invoked as :cpp:`f(ptd, si)` where ``si`` is an
+:cpp:`amrex::SIMDindex`. Write it generically over the index type and load
+particle components with :cpp:`amrex::simd::load_1d`, which returns a SIMD
+register in the vectorized main loop and a scalar value otherwise:
+
+.. highlight:: c++
+
+::
+
+    #include <AMReX_ParticleReduceSIMD.H>
+
+    using ConstPTDType = typename PC::ConstPTDType;
+    amrex::ReduceOps<ReduceOpSum, ReduceOpMin> reduce_ops;
+    auto r = amrex::ParticleReduceSIMD<ReduceData<amrex::ParticleReal, amrex::ParticleReal>> (
+                 pc, [=] AMREX_GPU_DEVICE (const ConstPTDType& ptd, auto si) noexcept
+             {
+                 auto const a = amrex::simd::load_1d(ptd.rdata(1), si);
+                 auto const w = amrex::simd::load_1d(ptd.rdata(2), si);
+                 return amrex::makeTuple(a * w, a);
+             }, reduce_ops);
+
+For CUDA builds, note that extended device lambdas have compiler-version
+dependent restrictions on generic (``auto``) parameters; a small functor with
+a templated :cpp:`operator()` is the most portable form of the kernel.
+
+The SIMD width can be chosen explicitly via the ``WIDTH`` template parameter
+(e.g., :cpp:`ParticleReduceSIMD<RD, 8>`); it defaults to the native SIMD
+register width for :cpp:`amrex::ParticleReal`, which is a good default in
+practice, in particular for kernels with many reduction components.
+
+Two differences from :cpp:`amrex::ParticleReduce` to be aware of:
+
+* Only :cpp:`ReduceOpSum`, :cpp:`ReduceOpMin` and :cpp:`ReduceOpMax` are
+  supported (enforced at compile time).
+* SIMD sum reductions reassociate floating-point additions (per-lane partial
+  sums plus a final horizontal fold), so results can differ from the scalar
+  evaluation order by rounding. In particular,
+  :cpp:`amrex::system::regtest_reduction` is not honored by the SIMD path;
+  use the scalar entry points when bitwise reproducible regression data is
+  required.
+
+For a complete example, including a benchmark against the scalar entry
+points, see ``Tests/Particles/ParticleReduceSIMD``.
+
 .. _sec:Particles:Interacting:
 
 Interacting with Mesh Data
--- a/Docs/sphinx_documentation/source/Particle.rst
+++ b/Docs/sphinx_documentation/source/Particle.rst
@@ -522,7 +522,7 @@ single quantity computed per particle by a user-provided callable. The general
 the particles, combining :cpp:`amrex::ReduceOps` (e.g., several sums, minima
 and maxima at once) into one :cpp:`amrex::ReduceData` result tuple.
 
-The per-particle callable can take the particle in several forms; the most
+The per-particle callable can take the particle in several forms: the most
 efficient one takes a :cpp:`ConstParticleTileData` and the particle index,
 which reads the particle components directly from the underlying (SoA) arrays:
 
@@ -531,8 +531,8 @@ which reads the particle components directly from the underlying (SoA) arrays:
 ::
 
     using ConstPTDType = typename PC::ConstPTDType;
-    amrex::ReduceOps<ReduceOpSum, ReduceOpMin, ReduceOpMax> reduce_ops;
-    auto r = amrex::ParticleReduce<ReduceData<amrex::Real, amrex::Real, int>> (
+    amrex::ReduceOps<amrex::ReduceOpSum, amrex::ReduceOpMin, amrex::ReduceOpMax> reduce_ops;
+    auto r = amrex::ParticleReduce<amrex::ReduceData<amrex::Real, amrex::Real, int>> (
                  pc, [=] AMREX_GPU_DEVICE (const ConstPTDType& ptd, const int i) noexcept
                                -> amrex::GpuTuple<amrex::Real, amrex::Real, int>
              {
@@ -542,7 +542,7 @@ which reads the particle components directly from the underlying (SoA) arrays:
                  return {a, b, c};
              }, reduce_ops);
 
-Note that no MPI reduction is performed at the end of these operations; call
+Note that no MPI reduction is performed at the end of these operations. Call
 the reductions in :cpp:`ParallelDescriptor` manually if you want that behavior.
 
 .. _sec:Particles:SIMDReductions:
@@ -574,8 +574,8 @@ register in the vectorized main loop and a scalar value otherwise:
     #include <AMReX_ParticleReduceSIMD.H>
 
     using ConstPTDType = typename PC::ConstPTDType;
-    amrex::ReduceOps<ReduceOpSum, ReduceOpMin> reduce_ops;
-    auto r = amrex::ParticleReduceSIMD<ReduceData<amrex::ParticleReal, amrex::ParticleReal>> (
+    amrex::ReduceOps<amrex::ReduceOpSum, amrex::ReduceOpMin> reduce_ops;
+    auto r = amrex::ParticleReduceSIMD<amrex::ReduceData<amrex::ParticleReal, amrex::ParticleReal>> (
                  pc, [=] AMREX_GPU_DEVICE (const ConstPTDType& ptd, auto si) noexcept
              {
                  auto const a = amrex::simd::load_1d(ptd.rdata(1), si);
@@ -583,12 +583,12 @@ register in the vectorized main loop and a scalar value otherwise:
                  return amrex::makeTuple(a * w, a);
              }, reduce_ops);
 
-For CUDA builds, note that extended device lambdas have compiler-version
-dependent restrictions on generic (``auto``) parameters; a small functor with
+For CUDA builds, note that extended device lambdas have nvcc-version
+dependent restrictions on generic (``auto``) parameters and a small functor with
 a templated :cpp:`operator()` is the most portable form of the kernel.
 
 The SIMD width can be chosen explicitly via the ``WIDTH`` template parameter
-(e.g., :cpp:`ParticleReduceSIMD<RD, 8>`); it defaults to the native SIMD
+(e.g., :cpp:`ParticleReduceSIMD<RD, 8>`). It defaults to the native SIMD
 register width for :cpp:`amrex::ParticleReal`, which is a good default in
 practice, in particular for kernels with many reduction components.
 
@@ -599,9 +599,17 @@ Two differences from :cpp:`amrex::ParticleReduce` to be aware of:
 * SIMD sum reductions reassociate floating-point additions (per-lane partial
   sums plus a final horizontal fold), so results can differ from the scalar
   evaluation order by rounding. In particular,
-  :cpp:`amrex::system::regtest_reduction` is not honored by the SIMD path;
+  :cpp:`amrex::system::regtest_reduction` is not honored by the SIMD path,
   use the scalar entry points when bitwise reproducible regression data is
   required.
+
+Under the hood, :cpp:`amrex::ParticleReduceSIMD` evaluates each particle tile
+with :cpp:`amrex::evalReduceSIMD<WIDTH>(n, reduce_data, f, reduce_ops)`, the
+SIMD analogue of the element loop in :cpp:`ReduceOps::eval`. This engine is
+not particle-specific: it can be used directly to reduce over any indexed
+range on the host, with the same single-source callable contract (``f(si)``
+with an :cpp:`amrex::SIMDindex`, provided by ``AMReX_ReduceSIMD.H``). See
+``Tests/SIMD`` for a small stand-alone example with a mixed-type tuple.
 
 For a complete example, including a benchmark against the scalar entry
 points, see ``Tests/Particles/ParticleReduceSIMD``.
--- a/Src/Particle/AMReX_ParticleReduceSIMD.H
+++ b/Src/Particle/AMReX_ParticleReduceSIMD.H
@@ -35,7 +35,8 @@ namespace particle_detail {
 
 /**
  * \brief A general SIMD-vectorized reduction method for the particles in a
- * ParticleContainer. This version operates from the specified lev_min to lev_max.
+ * ParticleContainer. This version operates from the specified lev_min to
+ * lev_max.
  *
  * This is the SIMD counterpart of amrex::ParticleReduce. When AMReX is compiled
  * with SIMD support (\c AMReX_SIMD=ON), the reduction runs vectorized over
@@ -53,9 +54,11 @@ namespace particle_detail {
  *    \code{.cpp}
  *       using ConstPTDType = typename PC::ConstPTDType;
  *       amrex::ReduceOps<amrex::ReduceOpSum, amrex::ReduceOpMin> reduce_ops;
- *       auto r = amrex::ParticleReduceSIMD<amrex::ReduceData<amrex::ParticleReal, amrex::ParticleReal>>(
- *                    pc, lev_min, lev_max,
- *                    [=] AMREX_GPU_DEVICE (const ConstPTDType& ptd, auto si) noexcept
+ *       auto r =
+ * amrex::ParticleReduceSIMD<amrex::ReduceData<amrex::ParticleReal,
+ * amrex::ParticleReal>>( pc, lev_min, lev_max,
+ *                    [=] AMREX_GPU_DEVICE (const ConstPTDType& ptd, auto si)
+ * noexcept
  *                {
  *                    auto const a = amrex::simd::load_1d(ptd.rdata(1), si);
  *                    auto const w = amrex::simd::load_1d(ptd.rdata(2), si);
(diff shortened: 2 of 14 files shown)


AlexanderSinn:
Since beam_moments takes everything by reference, the test with (ptd,i) is supposed to access some particle components multiple times instead of reading them all upfront. That might explain the performance difference compared to the SuperParticle version.

Docs/sphinx_documentation/source/Particle.rst
```diff
@@ -507,6 +507,105 @@ In this way, code can be written that is agnostic as to the data layout. For mor
...
+logical and/or) over the particles in a :cpp:`ParticleContainer`, on CPUs and
+GPUs alike. The simple entry points :cpp:`amrex::ReduceSum`,
+:cpp:`amrex::ReduceMin`, :cpp:`amrex::ReduceMax`,
+:cpp:`amrex::ReduceLogicalAnd` and :cpp:`amrex::ReduceLogicalOr` reduce a
+single quantity computed per particle by a user-provided callable. The general
+:cpp:`amrex::ParticleReduce` computes many reductions in a single pass over
+the particles, combining :cpp:`amrex::ReduceOps` (e.g., several sums, minima
+and maxima at once) into one :cpp:`amrex::ReduceData` result tuple.
+
+The per-particle callable can take the particle in several forms; the most
```
ax3l:
```suggestion
The per-particle callable can take the particle in several forms: the most
```

Docs/sphinx_documentation/source/Particle.rst
```diff
@@ -507,6 +507,105 @@ In this way, code can be written that is agnostic as to the data layout. For mor
...
+                 pc, [=] AMREX_GPU_DEVICE (const ConstPTDType& ptd, const int i) noexcept
+                               -> amrex::GpuTuple<amrex::Real, amrex::Real, int>
+             {
+                 const amrex::Real a = ptd.rdata(1)[i];
+                 const amrex::Real b = ptd.rdata(2)[i];
+                 const int c = ptd.idata(1)[i];
+                 return {a, b, c};
+             }, reduce_ops);
+
+Note that no MPI reduction is performed at the end of these operations; call
```
ax3l:
```suggestion
Note that no MPI reduction is performed at the end of these operations. Call
```

Docs/sphinx_documentation/source/Particle.rst
```diff
@@ -507,6 +507,105 @@ In this way, code can be written that is agnostic as to the data layout. For mor
...
+                 pc, [=] AMREX_GPU_DEVICE (const ConstPTDType& ptd, auto si) noexcept
+             {
+                 auto const a = amrex::simd::load_1d(ptd.rdata(1), si);
+                 auto const w = amrex::simd::load_1d(ptd.rdata(2), si);
+                 return amrex::makeTuple(a * w, a);
+             }, reduce_ops);
+
+For CUDA builds, note that extended device lambdas have compiler-version
+dependent restrictions on generic (``auto``) parameters; a small functor with
+a templated :cpp:`operator()` is the most portable form of the kernel.
```
ax3l, line 588:
```suggestion
For CUDA builds, note that extended device lambdas have nvcc-version
dependent restrictions on generic (``auto``) parameters and a small functor with
a templated :cpp:`operator()` is the most portable form of the kernel.
```

Docs/sphinx_documentation/source/Particle.rst
```diff
@@ -507,6 +507,105 @@ In this way, code can be written that is agnostic as to the data layout. For mor
...
+
+The SIMD width can be chosen explicitly via the ``WIDTH`` template parameter
+(e.g., :cpp:`ParticleReduceSIMD<RD, 8>`); it defaults to the native SIMD
```
ax3l:
```suggestion
(e.g., :cpp:`ParticleReduceSIMD<RD, 8>`). It defaults to the native SIMD
```

Docs/sphinx_documentation/source/Particle.rst
```diff
@@ -507,6 +507,105 @@ In this way, code can be written that is agnostic as to the data layout. For mor
...
+practice, in particular for kernels with many reduction components.
+
+Two differences from :cpp:`amrex::ParticleReduce` to be aware of:
+
+* Only :cpp:`ReduceOpSum`, :cpp:`ReduceOpMin` and :cpp:`ReduceOpMax` are
+  supported (enforced at compile time).
+* SIMD sum reductions reassociate floating-point additions (per-lane partial
+  sums plus a final horizontal fold), so results can differ from the scalar
+  evaluation order by rounding. In particular,
+  :cpp:`amrex::system::regtest_reduction` is not honored by the SIMD path;
```
ax3l:
```suggestion
  :cpp:`amrex::system::regtest_reduction` is not honored by the SIMD path,
```

Tests/Particles/ParticleReduceSIMD/main.cpp
```diff
@@ -0,0 +1,424 @@
...
+    auto operator() (PTD const& ptd, SI const si) const noexcept
+    {
+        return simd::load_1d(ptd.rdata(Comp), si);
+    }
+};
+
+//! Deterministic pseudo-random number in [0, 1) from an integer (splitmix64 finalizer)
+AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+ParticleReal hash01 (std::uint64_t z) noexcept
+{
```
ax3l, line 154: Wild choice, but I like it.

Docs/sphinx_documentation/source/Particle.rst
```diff
@@ -507,6 +507,105 @@ In this way, code can be written that is agnostic as to the data layout. For mor
...
+                 return {a, b, c};
+             }, reduce_ops);
+
+Note that no MPI reduction is performed at the end of these operations; call
+the reductions in :cpp:`ParallelDescriptor` manually if you want that behavior.
+
+.. _sec:Particles:SIMDReductions:
+
+SIMD-Vectorized Particle Reductions
+-----------------------------------
```
ax3l, line 551: Interestingly, `ReduceOps::eval` and `evalReduceSIMD` are not advertised to users yet. Maybe they should?

Src/Particle/AMReX_ParticleReduceSIMD.H
```diff
@@ -0,0 +1,411 @@
...
+        // We avoid ParIter here to allow reductions after grids change but prior to calling Redistribute()
+#if !defined(AMREX_USE_GPU) && defined(AMREX_USE_OMP)
+#pragma omp parallel for
+#endif
+        for (int pmap_it = 0; pmap_it < std::ssize(ptile_ptrs); ++pmap_it)
+        {
+            const auto& tile = plev.at(grid_tile_ids[pmap_it]);
+            const auto np = tile.numParticles();
+            const auto& ptd = tile.getConstParticleTileData();
+#ifdef AMREX_USE_GPU
```
ax3l, line 149: Codex:
`evalReduceSIMD` is currently the only place that checks both `WIDTH >= 1` and the supported operation pack. This GPU branch bypasses that function, while the non-SIMD host path replaces the requested width with `EffWidth = 1`. Consequently, invalid widths can silently compile on GPU/non-SIMD builds, and unsupported operations compile on GPU even though the documented contract says they are rejected at compile time. Could we enforce both invariants before this backend split, for example by taking `ReduceOps<Ps...>&` here and asserting on `Ps...`?

Docs/sphinx_documentation/source/Particle.rst
```diff
@@ -507,6 +507,105 @@ In this way, code can be written that is agnostic as to the data layout. For mor
...
+register in the vectorized main loop and a scalar value otherwise:
+
+.. highlight:: c++
+
+::
+
+    #include <AMReX_ParticleReduceSIMD.H>
+
+    using ConstPTDType = typename PC::ConstPTDType;
+    amrex::ReduceOps<ReduceOpSum, ReduceOpMin> reduce_ops;
```
ax3l: Codex:
These examples mix namespace-qualified APIs with unqualified `ReduceOp*` and `ReduceData` types. Since neither snippet includes `using namespace amrex;`, copying them produces lookup errors. Please qualify `amrex::ReduceOpSum`, `amrex::ReduceOpMin`, `amrex::ReduceOpMax`, and `amrex::ReduceData` in both reduction examples so they are self-contained.

Tests/Particles/ParticleReduceSIMD/main.cpp
```diff
@@ -0,0 +1,424 @@
...
+    ReduceOpsT reduce_ops;
+    return ParticleReduce<ReduceDataT>(pc,
+        [=] AMREX_GPU_DEVICE (const ConstPTDType& ptd, const int i) noexcept
+        {
+            return beam_moments(ptd.rdata(RealSoA::w)[i],
+                                ptd.rdata(RealSoA::x)[i], ptd.rdata(RealSoA::y)[i], ptd.rdata(RealSoA::t)[i],
+                                ptd.rdata(RealSoA::px)[i], ptd.rdata(RealSoA::py)[i], ptd.rdata(RealSoA::pt)[i],
+                                ptd.rdata(RealSoA::sx)[i], ptd.rdata(RealSoA::sy)[i], ptd.rdata(RealSoA::sz)[i],
+                                s);
```
ax3l: Codex:
Because `beam_moments` accepts all inputs by `const&`, this baseline binds them directly to SoA storage, while variant C first materializes every component in a local. That difference can expose variant B to repeated loads or weaker alias analysis and makes the benchmark measure more than SIMD vectorization. Hoisting equivalent scalar loads here would make the comparison more like-for-like:

```suggestion
            auto const p_w  = ptd.rdata(RealSoA::w)[i];
            auto const p_x  = ptd.rdata(RealSoA::x)[i];
            auto const p_y  = ptd.rdata(RealSoA::y)[i];
            auto const p_t  = ptd.rdata(RealSoA::t)[i];
            auto const p_px = ptd.rdata(RealSoA::px)[i];
            auto const p_py = ptd.rdata(RealSoA::py)[i];
            auto const p_pt = ptd.rdata(RealSoA::pt)[i];
            auto const p_sx = ptd.rdata(RealSoA::sx)[i];
            auto const p_sy = ptd.rdata(RealSoA::sy)[i];
            auto const p_sz = ptd.rdata(RealSoA::sz)[i];

            return beam_moments(p_w, p_x, p_y, p_t, p_px, p_py, p_pt,
                                p_sx, p_sy, p_sz, s);
```

ax3l: Claude Fabel:
Applied the suggested hoisting: variant B now reads all 10 components into `auto const` locals upfront (like-for-like with variant C's `load_1d` locals). To keep measuring whether the binding style matters, the previous form is retained as an explicitly labeled extra variant `B2(ptd,i by-ref)`.

Updated numbers (i9-12900H P-core #5 pinned, 1 OMP thread, GCC 13.3 `-O3 -march=native`, double precision, `AMReX_SIMD=ON`, best of 5 batches at ~2e8 particle evaluations per measurement), ns/particle:

| N | A(SuperParticle) | B(ptd,i upfront) | B2(ptd,i by-ref) | C(SIMD W=4) | C2(SIMD W=8) |
|---|---|---|---|---|---|
| 1k | 7.74 | 10.42 | 9.78 | 4.41 | 5.00 |
| 100k | 7.73 | 9.67 | 9.59 | 4.36 | 5.02 |
| 10M | 8.05 | 10.94 | 10.25 | 5.18 | 5.70 |

Takeaways:

- Hoisting does not change the picture: the two `(ptd,i)` forms stay within ~6% of each other, and the sign *flips* between builds (in the previous build, where the roles were swapped, the by-ref form was the slower one). GCC 13 CSEs the const loads after inlining, so by-ref binding does not systematically expose repeated loads or weaker alias analysis here.
- The A-vs-B gap therefore does not come from the binding style. The remaining plausible explanation is `getSuperParticle`'s access pattern itself: a compile-time-unrolled gather of *all* components (incl. unused `qm` + `idcpu`) into a contiguous local struct before any arithmetic, which schedules better than interleaving loads with the 46-accumulator update chain.
- The C-vs-B comparison is now like-for-like as requested, and unchanged: ~2x in DP (native W=4), ~4.5-5x in SP (native W=8).

Docs/sphinx_documentation/source/Particle.rst
ax3l: added

Tests/Particles/ParticleReduceSIMD/main.cpp
ax3l: @AlexanderSinn does this also address your comment in https://github.com/AMReX-Codes/amrex/pull/5562#issuecomment-4974375087 ?

AlexanderSinn: Yes, maybe then in this case the perf difference between (ptd,i) and (SuperParticle) comes down to compiler randomness in the layout of the assembly. Maybe you could look at the assembly output to spot the difference if you have time, but it's also not that important.

ax3l: Yeah I understood the difference at some point in the past better, maybe the SuperParticle cannot omit loading some particle memory that ptd access patterns can optimize out. Once in a (homogeneous) struct, SIMD usually works well.  cc @atmyers

ax3l: # Here is what Fabel analyzed

(take with a grain of salt, I have not validated it manually for this analysis):

Took up the suggestion to look at the assembly (`objdump -d`, GCC 13.3 `-O3 -march=native`, DP, AVX2) — the A-vs-B gap turns out to be deterministic, not layout randomness:

**Variant A (SuperParticle) is SLP-auto-vectorized; B/B2 are not.** Per-particle hot loops:

| variant | loop insns | FP arith packed / scalar | dominant ops |
|---|---|---|---|
| A (SuperParticle) | 122 | **32** / 14 | `vsubpd`×15, `vfmadd213pd`×10, `vmulpd`×7 |
| B (ptd,i upfront) | 168 | 8 / **56** | `vfmadd213sd`×25, `vmulsd`×16 |
| B2 (ptd,i by-ref) | 183 | 4 / **66** | `vfmadd213sd`×25, `vmulsd`×20 |

`getSuperParticle` gathers all components into a contiguous local struct, and from there GCC's SLP pass packs the 46 independent tuple products/updates four-at-a-time into ymm registers — A is effectively partially SIMD *across tuple components*. The identical arithmetic fed from 10 separate `ptd.rdata(k)` pointers defeats SLP and compiles to scalar chains with ~2x the stack traffic. Static insn ratio B/A = 1.38 vs measured runtime ratio 1.31 — consistent. So "read everything upfront" was directionally right, but it only helps when it lands in a contiguous struct that triggers SLP; plain scalar locals (variant B) don't.

Bonus findings on the explicit SIMD path (`evalReduceSIMD<4>`): the main loop is the ideal 46 packed updates (`vaddpd`×34 + `vminpd`/`vmaxpd`×6+6), but (1) GCC does not inline the 46-operand kernel into the loop — one `call` per 4 particles, and since ymm registers are caller-saved, all 46 accumulators get saved/restored around it (`vmovapd`×92 = most of the loop); (2) the `element_aligned` `copy_from` loads lower to element-wise `vmovsd`+`vinsertf128` packing instead of `vmovupd` (the known `fixed_size_simd` quirk, cf. #4592). C wins ~2x (DP) / ~4.5-5x (SP) despite both — so there is known headroom (force-inline the kernel invocation; improve the load lowering) for follow-up PRs.

<details>

# Assembly Analysis: `ParticleReduce` Variants (PR #5562)

Why is the SuperParticle lambda form (A) consistently ~30% faster than the
`(ptd,i)` form (B), and what exactly does the explicit SIMD path (C) compile
to? Follow-up to the review thread
https://github.com/AMReX-Codes/amrex/pull/5562#discussion_r3583129599 —
AlexanderSinn suggested the A-vs-B gap might be "compiler randomness in the
layout of the assembly" and to check the assembly output.

**TL;DR: it is not randomness.** Variant A gets **SLP-auto-vectorized** by GCC
because `getSuperParticle` materializes a contiguous local struct; the same
arithmetic fed directly from SoA pointers (B/B2) compiles to scalar code.

## Setup

- Binary: `build-simd/Tests/Particles/ParticleReduceSIMD/3d/Test_Particles_ParticleReduceSIMD_3d`
  (GCC 13.3, `-O3 -march=native -mtune=native`, double precision, `AMReX_SIMD=ON`,
  i9-12900H / AVX2+FMA, native SIMD width 4).
- Method: `objdump -d --demangle`, then isolate each variant's per-particle hot
  loop. The tile loops are outlined into `[clone ._omp_fn.0]` functions
  (`#pragma omp parallel for` → `GOMP_parallel`); the per-particle loops of
  A/B/B2 are inlined there, while variant C's element loop lives in the
  standalone `amrex::evalReduceSIMD<4, ...>` instantiation.
- Kernel: the ImpactX `ReducedBeamCharacteristics` shape — per particle
  10 SoA loads, 9 subtractions, ~33 products (CSE'd/FMA-fused by GCC),
  46 accumulator updates (34 sum + 6 min + 6 max).

## Per-particle hot loops (double precision)

| Variant | loop insns | FP arith (packed / scalar) | dominant ops | stack stores/loads |
|---|---|---|---|---|
| A (SuperParticle) | 122 | 46 (**32 / 14**) | `vsubpd`×15, `vfmadd213pd`×10, `vmulpd`×7, `vminsd`/`vmaxsd`×6+6 | 27 / 3 |
| B (ptd,i upfront locals) | 168 | 64 (8 / **56**) | `vmovsd`×82, `vfmadd213sd`×25, `vmulsd`×16 | 54 / 16 |
| B2 (ptd,i by-ref) | 183 | 70 (4 / **66**) | `vmovsd`×98, `vfmadd213sd`×25, `vmulsd`×20 | 57 / 19 |

### Variant A: SLP vectorization

`getSuperParticle(i)` copies all components (including the unused `qm` and
`idcpu`) into a contiguous `Particle<11,0>` on the stack. From that contiguous
struct, GCC's SLP (superword-level parallelism) pass packs the 46 mutually
independent per-component products and accumulator updates four-at-a-time into
ymm registers: the loop computes the tuple math with packed `vsubpd`/`vmulpd`/
`vfmadd213pd` — i.e., **A is partially SIMD across tuple components within a
single particle**. Only the 6+6 min/max updates remain scalar.

### Variants B/B2: SLP defeated

The identical arithmetic fed from 10 separate `ptd.rdata(k)` array pointers
(B: hoisted into scalar locals, B2: bound by `const&`) does not trigger SLP.
GCC emits scalar `vmulsd`/`vfmadd213sd` chains with roughly twice the stack
traffic. Hoisting loads into plain scalar locals (B, the review suggestion)
does not help — only the contiguous-struct gather unlocks SLP.

Consistency check: static instruction ratio B/A = 168/122 ≈ 1.38, measured
runtime ratio ≈ 10.5/8.0 ≈ 1.31 ns/particle. The B-vs-B2 delta (168 vs 183
insns) is within the ±6% runtime noise band and flips sign between builds, as
reported in the PR thread.

So the earlier review intuition ("read everything upfront") was half right:
upfront reading matters, but only when it lands in a contiguous struct that
enables SLP — plain scalar locals do not.

## Variant C: explicit SIMD path (`evalReduceSIMD<4>`)

The W=4 main loop is exactly the ideal accumulation pattern:

```
loop@0x213d8: 146 insns, FP=46 packed / 0 scalar
  vmovapd×92, vaddpd×34, vminpd×6, vmaxpd×6, call×1
```

34 packed sums + 6 packed mins + 6 packed maxes = the 46 accumulator updates,
all 256-bit. Two codegen quirks are visible, though:

1. **The kernel is not inlined into the accumulation loop.** GCC keeps the
   46-operand kernel (wrapper lambda → `BeamMomentsKernel`) as an out-of-line
   function and `call`s it once per 4 particles. Because ymm registers are
   caller-saved, all 46 SIMD accumulators are stored to and reloaded from the
   stack around the call — that is the `vmovapd`×92, the bulk of the loop.
2. **`element_aligned` loads are not `vmovupd`.** Inside the kernel, the
   `copy_from(ptr, element_aligned)` loads lower to element-wise
   `vmovsd`+`vunpcklpd`+`vinsertf128` vector construction instead of single
   unaligned vector loads — the known libstdc++ `fixed_size_simd` codegen
   quirk (see amrex issue #4592 on alignment guarantees).

Despite both quirks, C wins ~2x over B in DP (4.4–5.2 vs 9.5–10.9 ns/particle)
and ~4.5–5x in SP (native W=8).

## Take-aways / possible follow-ups

- The A-vs-B gap is a deterministic vectorization difference, not noise:
  contiguous gather → SLP. It also explains why ImpactX's current
  SuperParticle-form diagnostic is not as slow as the per-particle copy
  suggests.
- Headroom in the SIMD path: force-inlining the kernel invocation chain in
  `evalReduceSIMD` (so the accumulator traffic folds into the arithmetic
  instead of a bulk ymm save/restore around a `call`) and/or improving the
  `load_1d` lowering to true `vmovupd` could push the SIMD speedup further.
  With 46 live ymm accumulators against 16 architectural registers, some
  spilling is unavoidable, but the call-boundary save/restore is not.

## Reproducing

```bash
objdump -d --demangle build-simd/Tests/Particles/ParticleReduceSIMD/3d/Test_Particles_ParticleReduceSIMD_3d > bench.asm
# A/B/B2 hot loops: the '[clone ._omp_fn.0]' functions whose demangled names
# contain the respective lambda signatures; C: 'evalReduceSIMD<4, int, ...>'.
# Identify B vs B2 by the function pointer each run_variant_* passes to GOMP_parallel.
```

</details>

ax3l: Claude Fabel:
Took up the suggestion to look at the assembly (`objdump -d`, GCC 13.3 `-O3 -march=native`, DP, AVX2) — the A-vs-B gap turns out to be deterministic, not layout randomness:

**Variant A (SuperParticle) is SLP-auto-vectorized; B/B2 are not.** Per-particle hot loops:

| variant | loop insns | FP arith packed / scalar | dominant ops |
|---|---|---|---|
| A (SuperParticle) | 122 | **32** / 14 | `vsubpd`×15, `vfmadd213pd`×10, `vmulpd`×7 |
| B (ptd,i upfront) | 168 | 8 / **56** | `vfmadd213sd`×25, `vmulsd`×16 |
| B2 (ptd,i by-ref) | 183 | 4 / **66** | `vfmadd213sd`×25, `vmulsd`×20 |

`getSuperParticle` gathers all components into a contiguous local struct, and from there GCC's SLP pass packs the 46 independent tuple products/updates four-at-a-time into ymm registers — A is effectively partially SIMD *across tuple components*. The identical arithmetic fed from 10 separate `ptd.rdata(k)` pointers defeats SLP and compiles to scalar chains with ~2x the stack traffic. Static insn ratio B/A = 1.38 vs measured runtime ratio 1.31 — consistent. So "read everything upfront" was directionally right, but it only helps when it lands in a contiguous struct that triggers SLP; plain scalar locals (variant B) don't.

The same analysis exposed headroom in the explicit SIMD path: GCC kept the 46-operand kernel out-of-line, calling it once per SIMD width inside the accumulation loop — and since ymm registers are caller-saved, all 46 SIMD accumulators were spilled/reloaded around every call (`vmovapd`×92 = most of the loop). **Fixed in the new commit**: `ParticleReduceSIMD` now passes a named `AMREX_FORCE_INLINE` binder (`particle_detail::SIMDKernelBinder`) instead of a plain wrapper lambda. With the kernel folded into the loop, the loads become true `vmovupd` (the element-wise `vmovsd`+`vinsertf128` packing was a call-boundary artifact, not the feared `fixed_size_simd` quirk) and the multiply+accumulate pairs fuse into `vfmadd213pd` directly into the accumulators — 79 packed / 0 scalar FP ops at W=4.

Updated numbers (same protocol as above; `ParticleReduceSIMD`, best of 5 batches, ns/particle; speedup vs the scalar `(ptd,i)` form in parentheses):

| N | DP before | DP after | SP before | SP after |
|---|---|---|---|---|
| 1k | 4.41 | **3.58** (2.7x) | 2.25 | **1.88** (4.8x) |
| 10k | 4.37 | **3.57** (2.6x) | 2.24 | **1.90** (4.9x) |
| 100k | 4.36 | **3.51** (2.7x) | 2.21 | **1.83** (5.1x) |
| 1M | 5.14 | **4.52** (2.4x) | 2.51 | **2.11** (4.6x) |
| 10M | 5.15 | **4.55** (2.4x) | 2.58 | **2.20** (4.5x) |

Peak throughput is now ~545 Mp/s in SP (~285 Mp/s DP). Unexpected bonus: with the kernel force-inlined into the width-1 loop, the `AMReX_SIMD=OFF` scalar fallback of the same single-source kernel now gets SLP-vectorized too (like variant A) and runs at 7.0-7.4 ns/particle (DP) — faster than every other scalar form, including SuperParticle. So the single-source kernel now wins in every build configuration.

atmyers (approved):
Looks great!

ax3l:
Thanks!

ImpactX benchmark also looks great:
https://github.com/BLAST-ImpactX/impactx/pull/1564#issuecomment-4999484385

PR #5569, opened 2026-07-17
Add fconvert plotfile tool for native-to-HDF5 conversion

jrood-nrel:
## Summary

Copilot with Opus 4.8 wrote this.

This adds `Tools/Plotfile/fconvert`, a native AMReX plotfile postprocessing tool that rewrites one or more native plotfiles as HDF5 plotfiles using AMReX’s existing HDF5 writer and compression descriptors. It follows the existing `f*` tool conventions and is wired into both CMake and GNUmake builds.

- **New tool**
  - Adds `Tools/Plotfile/fconvert.cpp` with the standard plotfile-tool structure (`main_main()`, command-line parsing via `command_argument_count()` / `get_command_argument()`, `Initialize(argc, argv, false)`).
  - Reads all levels/components from `amrex::PlotFileData`, reconstructs `Geometry`, `level_steps`, and `ref_ratio`, and calls `WriteMultiLevelPlotfileHDF5(...)`.
  - Supports multiple input plotfiles, `--compression` / `-c`, and `--output` / `-o`.

- **HDF5 / ZFP behavior**
  - HDF5-specific codepaths are guarded with `#ifdef AMREX_USE_HDF5`.
  - Non-HDF5 builds abort at runtime with an actionable message instead of failing at compile time.
  - Compression handling is passed through unchanged to AMReX, e.g.
    ```bash
    fconvert --compression ZFP_ACCURACY@0.001 plt00010
    ```

- **Build integration**
  - Adds `fconvert` to `Tools/Plotfile/CMakeLists.txt` so it builds/installs alongside the existing plotfile tools.
  - Adds `fconvert` to `Tools/Plotfile/GNUmakefile`.

- **Documentation**
  - Documents `fconvert` in `Docs/sphinx_documentation/source/Post_Processing.rst`.
  - Adds an HDF5 postprocessing note in `Docs/sphinx_documentation/source/IO.rst`.

## Additional background

`fconvert` is intended for post-simulation conversion/compression of existing native plotfiles without requiring applications to emit HDF5 directly. Output naming defaults to the input plotfile name, with optional override for single-file output or multi-file prefixing.

## Checklist

The proposed changes:

### DIFF
--- a/Docs/sphinx_documentation/source/Post_Processing.rst
+++ b/Docs/sphinx_documentation/source/Post_Processing.rst
@@ -295,6 +295,31 @@ Typing ``./fextrema.gnu.ex`` without inputs will bring up usage and options.
                        0.03                       1          2.349724636         8.277319027e-17          1.157052145         8.277319027e-17          1.156713078           0.03595941273                    1         8.277319027e-17         0.4924203149         8.277319027e-17         0.4922760141           0.01530367099                    1         -0.005172583789       0.005172583789         -0.005172583789       0.005172583789         -0.005287367803       0.005287367803         -0.005287367803       0.005287367803         -0.004924487345        0.05687549245
 
 
+fconvert
+--------
+
+Converts one or more native AMReX plotfiles to HDF5 plotfiles. The tool passes
+through AMReX HDF5 compression descriptors such as ``None@0``, ``ZLIB@5``,
+and ``ZFP_ACCURACY@0.001``.
+
+**How to build and run**
+
+In ``amrex/Tools/Plotfile``, type ``make programs=fconvert`` and then
+``./fconvert.gnu.ex`` to run. Typing ``./fconvert.gnu.ex`` without inputs will
+bring up usage and options.
+
+**Example**
+
+.. code-block:: console
+
+    user@machine:~/AMReX/amrex/Tools/Plotfile$ ./fconvert.gnu.ex \
+    > --compression ZFP_ACCURACY@0.001 \
+    > --output converted \
+    > ~/AMReX/FHDeX/exec/multispec/Reg_DetBubble_2d_Bench/plt0000000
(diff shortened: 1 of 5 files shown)


Docs/sphinx_documentation/source/Post_Processing.rst
```diff
@@ -295,6 +295,31 @@ Typing ``./fextrema.gnu.ex`` without inputs will bring up usage and options.
...
+fconvert
+--------
+
+Converts one or more native AMReX plotfiles to HDF5 plotfiles. The tool passes
+through AMReX HDF5 compression descriptors such as ``None@0``, ``ZLIB@5``,
+and ``ZFP_ACCURACY@0.001``.
+
+**How to build and run**
+
+In ``amrex/Tools/Plotfile``, type ``make programs=fconvert`` and then
```
WeiqunZhang: We should add `USE_HDF5=TRUE` to the make command. We also need to note that if the compiler cannot find `hdf5.h`, one might also need to provide the path to HDF5 (e.g., `make programs=fconvert USE_HDF5=TRUE HDF5_HOME=/usr/lib/x86_64-linux-gnu/hdf5/serial/` on Ubuntu 24.04).

Tools/Plotfile/fconvert.cpp
```diff
@@ -0,0 +1,171 @@
...
+        ++farg;
+    }
+
+    if (farg > narg) {
+        PrintUsage();
+        return;
+    }
+
+#ifndef AMREX_USE_HDF5
+    amrex::Abort("fconvert requires AMReX to be built with AMReX_HDF5=ON. "
```
WeiqunZhang: This is for CMake. For GNU Make, it's USE_HDF5=TRUE and USE_HDF5_ZFP=TRUE.

WeiqunZhang:
When I run `./fconvert.gnu.ex plt00100/`, I end up with `plt00100/.h5`. So I think we need to strip the trailing `/` if there is one.

WeiqunZhang (approved):

PR #5575, opened 2026-07-21
Generalize local mask

atmyers:
The "local" argument to Redistribute really lumps together two distinct things: 1) whether to do point-to-point MPI communication for the particle count exchange, and 2) whether to use a mask for the grid/tile lookup. This separates these into distinct options. 

It also adds a heuristic to fall back to the full BoxArray search in cases where the mask would be too expensive too use.

The proposed changes:

### DIFF
--- a/Docs/sphinx_documentation/source/Particle.rst
+++ b/Docs/sphinx_documentation/source/Particle.rst
@@ -267,6 +267,14 @@ the supplied cell count in each direction since the last :cpp:`Redistribute()`
 call. Use the global algorithm after initialization, regridding, load balancing,
 or any update that may send particles arbitrarily far.
 
+On CPU builds, local :cpp:`Redistribute()` can use a level-0 mask to speed up
+particle-to-grid lookup for short moves. AMReX bypasses this mask for multi-level
+redistribution and when the estimated mask storage is too large; in those cases,
+the local MPI communication algorithm can still be used while particle lookup
+falls back to a box search. The mask policy can be controlled with
+``particles.redistribute_use_mask``, ``particles.redistribute_mask_max_ratio``,
+and ``particles.redistribute_mask_max_bytes``.
+
 Application codes will likely want to create their own derived
 ParticleContainer class that specializes the template parameters and adds
 additional functionality, like setting the initial conditions, moving the
--- a/Docs/sphinx_documentation/source/RuntimeParameters.rst
+++ b/Docs/sphinx_documentation/source/RuntimeParameters.rst
@@ -1182,20 +1182,29 @@ Particles
    :type: bool
    :value: true
 
+   .. versionadded:: 26.08
+      Runtime parameter ``particles.redistribute_use_mask``.
+
    On CPU builds, this controls whether local ``Redistribute()`` may use a
    level-0 mask for particle-to-grid lookup when the call is otherwise eligible.
 
 .. py:data:: particles.redistribute_mask_max_ratio
-   :type: real
+   :type: double
    :value: 8.0
+
+   .. versionadded:: 26.08
+      Runtime parameter ``particles.redistribute_mask_max_ratio``.
 
    This is the maximum ratio of grown mask cells to valid cells for using the
    CPU local ``Redistribute()`` lookup mask. Non-positive values disable this
    ratio limit.
 
 .. py:data:: particles.redistribute_mask_max_bytes
-   :type: int
+   :type: long
    :value: 268435456
+
+   .. versionadded:: 26.08
+      Runtime parameter ``particles.redistribute_mask_max_bytes``.
 
    This is the maximum estimated per-rank storage, in bytes, for using the CPU
    local ``Redistribute()`` lookup mask. Negative values disable this byte limit.
--- a/Docs/sphinx_documentation/source/RuntimeParameters.rst
+++ b/Docs/sphinx_documentation/source/RuntimeParameters.rst
@@ -1178,6 +1178,28 @@ Particles
    container when writing checkpoint and plot files for particles. The
    special value of ``-1`` indicates one file per process.
 
+.. py:data:: particles.redistribute_use_mask
+   :type: bool
+   :value: true
+
+   On CPU builds, this controls whether local ``Redistribute()`` may use a
+   level-0 mask for particle-to-grid lookup when the call is otherwise eligible.
+
+.. py:data:: particles.redistribute_mask_max_ratio
+   :type: real
+   :value: 8.0
+
+   This is the maximum ratio of grown mask cells to valid cells for using the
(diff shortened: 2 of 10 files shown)


atmyers:
Let's hold off on merging this until after the monthly release. However, the WarpX tests all pass with this change: https://github.com/BLAST-WarpX/warpx/pull/7090. After the release, we can update the PR to WarpX and use this feature.

WeiqunZhang:
PR 5575 — Generalize local mask

  Verification: built the branch with MPI (-DAMReX_SPACEDIM=3 -DAMReX_MPI=ON -DAMReX_ENABLE_TESTS=ON, Release) and ran Test_Particles_Redistribute_3d. The two new tests
  pass at 2 ranks. All 73 CI checks pass.

  The refactor itself is right, and it fixes two real bugs: the level-0 mask was previously used for lookups at all levels under local=true (indexing a level-0 iMultiFab
  with fine-level IntVects), and moving the local assertions below the lev_max == -1 resolution turns AMREX_ASSERT(numParticlesOutOfRange(*this, lev_min, lev_max, ...) == 
  0) from vacuous into a real check — with lev_max = -1 the loop in numParticlesOutOfRange ran zero iterations and always returned 0. Worth calling out in the PR body,
  since downstream debug builds may start tripping that assert where they never did.

  Blocker: RedistributeMaskLooksCheap() is a rank-local decision gating a collective

  RedistributeMaskLooksCheap (AMReX_ParticleContainerBase.cpp:332) accumulates valid_cells/grown_cells from an MFIter over this rank's boxes only, so the answer varies per
  rank whenever the DistributionMapping isn't perfectly uniform. But it gates BuildRedistributeMask, whose body calls
  redistribute_mask_ptr->FillBoundary(geom.periodicity()) — an MPI-collective. When ranks disagree, one rank posts FillBoundary sends/recvs that nobody matches, and those
  stale messages collide with the subsequent particle communication.

  Reproduced with the PR's own test binary. Input is inputs.rt.local_nomask with nlevs=1, max_grid_size=8 on a 16³ domain (8 boxes → 3/3/2 over 3 ranks), and one knob
  changed:

  particles.redistribute_mask_max_bytes = 20000

  With nghost=1, a grown box is 10³ = 1000 cells at 2*sizeof(int) = 8 B/cell, so a 3-box rank estimates 24000 B and a 2-box rank 16000 B — they straddle the limit:

  [DIAG] rank 0 grown_cells=3000 bytes=24000 -> use_mask=0
  [DIAG] rank 1 grown_cells=3000 bytes=24000 -> use_mask=0
  [DIAG] rank 0 grown_cells=3000 bytes=24000 -> use_mask=0
  [DIAG] rank 1 grown_cells=3000 bytes=24000 -> use_mask=0
  [DIAG] rank 2 grown_cells=2000 bytes=16000 -> use_mask=1
  [DIAG] rank 2 ENTERING mask FillBoundary (collective)
  Segfault
  *** An error occurred in MPI_Waitall
  *** MPI_ERR_TRUNCATE: message truncated
  *** MPI_ERRORS_ARE_FATAL

  (The [DIAG] lines are a temporary instrumentation I added and reverted; the crash reproduces without them.)

  Controls isolate it to the disagreement, not to 3 ranks or to the no-mask path:

  ┌───────────────────────────────────────────────────┬─────────────────────────────┐
  │                        run                        │           outcome           │
  ├───────────────────────────────────────────────────┼─────────────────────────────┤
  │ 3 ranks, max_bytes=268435456 (all ranks use mask) │ pass                        │
  ├───────────────────────────────────────────────────┼─────────────────────────────┤
  │ 3 ranks, max_bytes=1 (all ranks skip mask)        │ pass                        │
  ├───────────────────────────────────────────────────┼─────────────────────────────┤
  │ 3 ranks, max_bytes=20000 (ranks disagree)         │ MPI_ERR_TRUNCATE + segfault │
  ├───────────────────────────────────────────────────┼─────────────────────────────┤
  │ 2 ranks, max_bytes=20000 (4/4 boxes, ranks agree) │ pass                        │
  └───────────────────────────────────────────────────┴─────────────────────────────┘

  This is not an exotic corner — it fires precisely in the regime the feature targets. Under the shipping defaults, max_bytes = 256 MiB trips when a rank exceeds ~33.5 M
  grown cells, and ranks in a large run sit close enough together to straddle that; max_ratio = 8.0 straddles even more easily once box sizes differ across ranks
  (non-divisible domains, post-regrid box mixes, large max_cells_moved). The new inputs.rt.local_nomask uses max_bytes = 1, which makes every rank agree, so the added test
  can't catch this.

  Also note if (grown_cells == 0) { return true; } — a rank owning no boxes votes "use mask", which is an independent way to diverge.

  The fix needs the decision to be identical on every rank. Cheapest correct version: reduce it (ParallelAllReduce::LogicalAnd on the local verdict, or sum
  valid_cells/grown_cells before deciding), and cache the result next to redistribute_mask_ptr keyed on the same BoxArray/DistributionMapping/nghost triple that
  BuildRedistributeMask already tests, so it costs one small collective per regrid rather than one per Redistribute().

  ParmParse in the Redistribute hot path

  RedistributeMaskLooksCheap constructs a ParmParse and runs three query() calls on every Redistribute() — several times per timestep in a typical code, in a function
  whose whole purpose is to save time. Everything else under the particles. prefix is read once behind static bool initialized in ParticleContainer_impl::Initialize()
  (AMReX_ParticleContainerI.H:75-99). These three should follow that pattern; a function-local static works fine here since the base class isn't templated. Folding this
  into the caching suggested above handles both at once.

  Minor

  - RuntimeParameters.rst gives particles.redistribute_mask_max_bytes as :type: int, but the code reads it into a Long — relevant since the point of the knob is large
  values.
  - The three new parameters have no .. versionadded:: directive; the neighbouring tiny_profiler.* entries do.
  - static_cast<long double> for the byte estimate is out of step with the rest of AMReX (and is 64-bit on ARM/MSVC). Plain Long arithmetic reaches the same answer here —
  grown_cells * 8 can't overflow at any size that fits in memory. Likewise Real max_ratio becomes float under AMReX_PRECISION=SINGLE; double would be more predictable for
  a threshold.
  - The if (AMReX_OMP) … elseif (AMReX_MPI) … chain means the new tests never run under MPI in an OMP+MPI build. That matches the existing block below it in the same file,
  so it's a pre-existing pattern rather than something this PR introduced — but it's why CI wouldn't have surfaced the blocker even with a divergent input.

atmyers:
We don't need to do an MPI reduction to fix this. Each rank has the full BoxArray and DistributionMapping metadata, so each rank can agree on the result of `RedistributeMaskLooksCheap` without needing to do MPI communication.

WeiqunZhang (approved):

PR #5576, opened 2026-07-22
Add developer docstring to getTileArray

atmyers:
The proposed changes:

### DIFF
--- a/Src/Base/AMReX_FabArrayBase.H
+++ b/Src/Base/AMReX_FabArrayBase.H
@@ -392,6 +392,32 @@ public:
     //! parallel copy or add
     enum CpOp { COPY = 0, ADD = 1 };
 
+    /**
+    * \brief Return the cached tile decomposition used by MFIter.
+    *
+    * The cache is keyed by this FabArrayBase BDKey plus the pair of
+    * requested tile size and BoxArray coarsening ratio. If no entry exists,
+    * this builds it lazily with buildTileArray() and records the use before
+    * returning.
+    *
+    * It is not guaranteed that the all the tiles in the decomposition will be
+    * have exactly the input tile_size. Some of the tiles may be slightly larger.
+    * It is guaranteed, however, that the tiles associated with a given box
+    * will be non-overlapping.
+    *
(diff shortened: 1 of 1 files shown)


ax3l:
cc @markabate :)

Src/Base/AMReX_FabArrayBase.H
```diff
@@ -392,6 +392,32 @@ public:
...
+    * of tiles in that FAB, and the cell-centered tile boxes consumed by
+    * MFIter.
+    *
+    * TileArray::tileArray always stores cell-centered boxes; MFIter converts
+    * them to the requested IndexType for tilebox(), growntilebox(), and
+    * related accessors. The returned pointer is owned by the static tile-array
+    * cache; treat it as immutable and do not keep it after the cache entry can
+    * be flushed.
+    */
     const TileArray* getTileArray (const IntVect& tilesize) const;
```
ax3l, line 421: @atmyers what pointer does this return if untiled? A `nullprt` or something else?

Src/Base/AMReX_FabArrayBase.H
```diff
@@ -392,6 +392,32 @@ public:
     //! parallel copy or add
     enum CpOp { COPY = 0, ADD = 1 };
 
+    /**
+    * \brief Return the cached tile decomposition used by MFIter.
+    *
+    * The cache is keyed by this FabArrayBase BDKey plus the pair of
+    * requested tile size and BoxArray coarsening ratio. If no entry exists,
+    * this builds it lazily with buildTileArray() and records the use before
+    * returning.
```
ax3l: maybe clearer for some readers:
```suggestion
    * this builds it on first use with buildTileArray() and records the use
    * before returning.
```

Src/Base/AMReX_FabArrayBase.H
atmyers: If tiling is off, this returns the same thing as in the tiled case, but with only one entry per owned Box.

ax3l: Thanks!

ax3l (approved):

PR #5577, opened 2026-07-23
ParticleHeader: Parse Grid Table

ax3l:
## Summary

Extract the parsing of the per-level grid table (data file index, particle count and byte offset for each grid) from `ParticleContainer::Restart` into the shared `ParticleHeader` (follow-up to #5476), so standalone consumers - e.g. language bindings and format converters - can locate each grid's binary particle data without duplicating the on-disk format logic.

- `ParticleHeader` gains a nested `GridEntry {which, count, where}` and `Vector<Vector<GridEntry>> grids`, filled by the new `parse_grid_table()`.
- `ParticleHeader::read()` now returns a fully parsed header including the table; `parse()` still consumes exactly the metadata prefix, keeping the existing stream contract.
- `Restart` consumes the shared parser instead of inline `HdrFile >>` reads - one source of truth for the Header format.

Motivation: pyAMReX tools (runtime-SoA plotfile reading https://github.com/AMReX-Codes/pyamrex/pull/596, AMReX-Codes/pyamrex#581, and a plotfile-to-openPMD converter https://github.com/AMReX-Codes/pyamrex/pull/597) need the grid table for per-grid particle counts and parallel-read planning.

## Testing

```
cmake -S . -B build -DAMReX_ENABLE_TESTS=ON -DAMReX_TEST_TYPE=All -DAMReX_PARTICLES=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo
cmake --build build -j6
ctest --test-dir build -R 'CheckpointRestart' -E 'HDF5|AsyncIO' --output-on-failure
```

All pass: `Particles_CheckpointRestart_3d`, `Particles_CheckpointRestartDualGrid_3d`, `Particles_CheckpointRestartDualGridSOA_3d`, `Particles_CheckpointRestartSOA_3d` (plus `Particles_Redistribute_3d` from the Small set).

## Checklist

The proposed changes:

### DIFF
--- a/Src/Particle/AMReX_ParticleHeader.cpp
+++ b/Src/Particle/AMReX_ParticleHeader.cpp
@@ -15,6 +15,8 @@ namespace amrex {
 void
 ParticleHeader::parse (std::istream& is)
 {
+    grids.clear();
+
     is >> version;
     AMREX_ASSERT(!version.empty());
 
--- a/Src/Particle/AMReX_ParticleHeader.cpp
+++ b/Src/Particle/AMReX_ParticleHeader.cpp
@@ -76,6 +76,30 @@ ParticleHeader::parse (std::istream& is)
     is >> finest_level;
 }
 
+void
+ParticleHeader::parse_grid_table (std::istream& is)
+{
+    AMREX_ASSERT(finest_level >= 0);
+
+    // The number of grids of every level comes first, ...
+    Vector<int> ngrids(finest_level+1);
+    for (int lev = 0; lev <= finest_level; ++lev) {
+        is >> ngrids[lev];
+        AMREX_ASSERT(ngrids[lev] > 0);
+    }
+
+    // ... followed by each level's table of (data file index, particle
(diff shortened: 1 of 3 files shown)


atmyers (approved):

Src/Particle/AMReX_ParticleHeader.cpp
```diff
@@ -101,6 +125,7 @@ ParticleHeader::read (const std::string& dir, const std::string& file)
 
     ParticleHeader header;
     header.parse(HdrFile);
```
ax3l, line 129: Codex:
Should `ParticleHeader::parse()` clear grids? Reusing an object to parse another metadata prefix can otherwise leave the previous grid table inconsistent with the new `finest_level`. This does not affect `read()` or `Restart()`.

ax3l: Maybe save to clear out... Will add.

PR #5581, opened 2026-07-27
Add For overloads with compile time options

ax3l:
## Summary

`ParallelFor` with `CompileTimeOptions` had no `For` counterpart, so kernels with non-independent iterations (e.g., particle-to-mesh scatter through `Gpu::Atomic` adds, which are plain updates on host) had no way to opt out of `AMREX_PRAGMA_SIMD` while keeping the compile-time-option dispatch. Under the pragma (GCC `ivdep`), GCC 15 vectorizes such accumulation loops and silently drops updates from SIMD lanes that hit the same address, observed in the WarpX charge/current deposition:
https://github.com/BLAST-WarpX/warpx/issues/7097

Add the six `For(TypeList<CompileTimeOptions<...>>, ...)` overloads (1D range, Box, Box + ncomp, each with and without the MT template parameter), dispatching via `AnyCTO` to the pragma-free For. On GPU, `For` and `ParallelFor` are identical.

## Additional background

https://github.com/BLAST-WarpX/warpx/issues/7097

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_CTOParallelForImpl.H
+++ b/Src/Base/AMReX_CTOParallelForImpl.H
@@ -396,6 +396,110 @@ void ParallelFor (TypeList<CTOs...> ctos,
     ParallelFor<AMREX_GPU_MAX_THREADS>(ctos, option, box, ncomp, std::forward<F>(f));
 }
 
+template <int MT, std::integral T, class F, typename... CTOs>
+void For (TypeList<CTOs...> ctos,
+          std::array<int,sizeof...(CTOs)> const& runtime_options,
+          T N, F&& f)
+{
+    AnyCTO(ctos, runtime_options,
+        [&](auto cto_func){
+            For<MT>(N, cto_func);
+        },
+        std::forward<F>(f)
+    );
+}
+
+template <int MT, class F, int dim, typename... CTOs>
+void For (TypeList<CTOs...> ctos,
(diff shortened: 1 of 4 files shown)


Src/Base/AMReX_CTOParallelForImpl.H
```diff
@@ -396,6 +396,110 @@ void ParallelFor (TypeList<CTOs...> ctos,
...
+/**
+ * \brief For with compile time optimization of kernels with run time options.
+ *
+ * Like the ParallelFor overload taking a TypeList of CompileTimeOptions, but
+ * without the implied promise that loop iterations are independent.  On CPU,
+ * ParallelFor marks the loop with AMREX_PRAGMA_SIMD, which licenses the
+ * compiler to vectorize across iterations; that is invalid for kernels whose
+ * iterations may touch the same memory location (e.g., particle-to-mesh
+ * scatter with Gpu::Atomic adds, which are plain updates on host).  Use this
+ * overload for such kernels.  On GPU, For and ParallelFor are identical.
```
ax3l:
```suggestion
 * without the implied promise that loop iterations are independent.  On CPU,
 * ParallelFor marks the loop with AMREX_PRAGMA_SIMD, which licenses the
 * compiler to vectorize across iterations. On GPU, For and ParallelFor are identical.
```

WeiqunZhang (approved):

PR #5585, opened 2026-07-31
TinyProfiler: stdout/stderr keywords for tiny_profiler.output_file

ax3l:
## Summary

Add the special values `stdout` and `stderr` for `tiny_profiler.output_file`, which select the default out stream and the default error stream of AMReX, respectively. An explicit `stdout` value was not expressible before: the empty string served as the default, but `tiny_profiler.output_file=` is rejected by `ParmParse` ("no values for definition"), so users could not spell it out on the command line or in an inputs file.

This came up for applications that redirect the report to a file by default and want a simple way to switch back to the standard output, see the discussion in https://github.com/BLAST-WarpX/warpx/pull/7077.

The default value of the parameter is now the literal `stdout` instead of the empty string, which is self-documenting; the behavior is unchanged and an empty value is still treated the same as `stdout`.

## Additional background

For https://github.com/BLAST-WarpX/warpx/pull/7077 and https://github.com/BLAST-ImpactX/impactx/pull/1566

## Checklist

The proposed changes:

### DIFF
--- a/Docs/sphinx_documentation/source/RuntimeParameters.rst
+++ b/Docs/sphinx_documentation/source/RuntimeParameters.rst
@@ -1605,7 +1605,7 @@ enabled.
    .. versionadded:: 24.09
       Runtime parameter ``tiny_profiler.output_file``.
 
-   .. versionchanged:: 26.08
+   .. versionchanged:: 26.09
       Special names ``stdout`` and ``stderr``; the default value is now
       ``stdout`` (same behavior as the previous default, the empty string).
 
--- a/Docs/sphinx_documentation/source/RuntimeParameters.rst
+++ b/Docs/sphinx_documentation/source/RuntimeParameters.rst
@@ -1600,12 +1600,17 @@ enabled.
 
 .. py:data:: tiny_profiler.output_file
    :type: string
-   :value: [empty]
+   :value: stdout
 
    .. versionadded:: 24.09
       Runtime parameter ``tiny_profiler.output_file``.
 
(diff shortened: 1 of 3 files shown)


dpgrote (approved):
Looks good!

WeiqunZhang:
> An explicit stdout value was not expressible before: the empty string served as the default, but tiny_profiler.output_file= is rejected by ParmParse ("no values for definition"), so users could not spell it out on the command line or in an inputs file.

I am just curious. This might work. But I did not test.
```
tiny_profile.output_file=""
```

Docs/sphinx_documentation/source/RuntimeParameters.rst
```diff
@@ -1600,12 +1600,17 @@ enabled.
...
+   :value: stdout
 
    .. versionadded:: 24.09
       Runtime parameter ``tiny_profiler.output_file``.
 
-   If this parameter is empty, the output of tiny profiling is dumped to the
-   default output stream of AMReX. If it is not empty, it specifies the file
-   name for the output. Note that ``/dev/null`` is a special name that means
-   no output.
+   .. versionchanged:: 26.08
```
WeiqunZhang:
```suggestion
   .. versionchanged:: 26.09
```

WeiqunZhang (approved):

ax3l:
Yep, either way it would not be super portable :)
https://github.com/AMReX-Codes/amrex/pull/5585#issuecomment-5148088630

PR #5587, opened 2026-08-03
Add tools for performing Kahan summation, along with a stochastic heat equation test that uses them.

atmyers:
The stochastic heat equation test is based on a Fortran code from John Bell. It gives the same results to the expected roundoff level when run in single versus double precision, both on CPU and GPU. The needed changes to achieve this were:

1. Use compensated summation when accumulating temperature statistics. 
2. Use a reference temperature when calculating skew and kurtosis to avoid cancellation errors. 
3. Make sure to use the same random number sequence between float and double. 

With these changes, the final statistics in single precision are:
```
aver stats 300.0039062 84.62759399 0.05907286704 2.936065435 45.98757553 21027.66797
cov stats -2.733321905 -1.717116833 -674.5462646 -1.432207465 -679.3804321 7172.049316
```

and in double:
```
aver stats 300 84.62537922 0.05907257817 2.936066093 45.98554629 21026.57194
cov stats -2.733323726 -1.71713382 -674.5293965 -1.432627767 -679.3645906 7171.662404
```

Single precision, without changes 1 and 2 above:
```
aver stats 300.0044556 84.42797852 0.1245079786 2.834512711 59.875 43584
cov stats -2.955393076 80.38709259 -28341.67578 36.77419281 -101210.8359 -20215.74219
```

Note that problem is particularly bad if you are averaging over many time steps. Without the precision changes, and with `nstat=1`, I get in single precision:
```
aver stats 214.7484131 8858.719727 -1.508285284 40.72777176 -1257593.125 3196192512
cov stats 8858.719727 -1520838.25 2904403968 -1520954.375 2907278848 2706444032
```

This also adds a function for doing compensated summation to AMReX. Note that certain compiler optimizations that assume associative arithmetic will break this function by reverting it to a simple sum. We therefore manually disable those optimizations for our supported compilers.

The simulated temperature variance versus theory for single precision:

variance

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_CompensatedSum.H
+++ b/Src/Base/AMReX_CompensatedSum.H
@@ -8,6 +8,34 @@
 namespace amrex
 {
 
+// Kahan compensation relies on (t - sum) - y evaluating to the rounding
+// error lost by sum + y, rather than to the exact-arithmetic value of 0.
+// Optimizations that assume floating-point addition is associative (e.g.
+// -ffast-math, -fassociative-math) or that fuse operations across the
+// subtractions (aggressive FP contraction) are free to simplify that
+// expression to 0, which silently turns compensatedAdd into a plain,
+// uncompensated sum. AMReX enables -ffast-math by default for several
+// compilers (see Flags_FASTMATH in Tools/CMake/AMReXFlagsTargets.cmake and
+// the gnu.mak/cray.mak GNU Make compiler flags), so this function's body is
+// pinned to strict floating-point semantics regardless of the flags used by
+// the translation unit that calls it.
+//
+// This function is deliberately NOT AMREX_FORCE_INLINE. Forcing
+// inlining across an optimization-setting boundary like the one below
+// causes both compilers to inline the -ffast-math-optimized call-site
+// version instead of honoring this function's own pragma-scoped settings,
+// silently defeating the protection below.
+#if defined(__CUDACC__) || defined(__HIPCC__)
+#pragma STDC FP_CONTRACT OFF
+#endif
+
+#if defined(__clang__)
+#pragma clang fp reassociate(off) contract(off)
+#elif defined(__GNUC__)
+#pragma GCC push_options
+#pragma GCC optimize ("no-fast-math")
+#endif
+
 /**
  * \brief Add a value to a running sum with Kahan compensation.
  *
@@ -15,7 +43,7 @@ namespace amrex
  * should be initialized before the first call, usually to zero.
  */
 template <typename T, typename U>
-AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+AMREX_GPU_HOST_DEVICE
 void compensatedAdd (T& sum, T& compensation, U value) noexcept
 {
     T const y = static_cast<T>(value) - compensation;
@@ -23,6 +51,12 @@ void compensatedAdd (T& sum, T& compensation, U value) noexcept
     compensation = (t - sum) - y;
     sum = t;
 }
+
+#if defined(__clang__)
+#pragma clang fp reassociate(on) contract(on)
+#elif defined(__GNUC__)
+#pragma GCC pop_options
+#endif
 
 }
 
--- a/Src/Base/AMReX_CompensatedSum.H
+++ b/Src/Base/AMReX_CompensatedSum.H
@@ -8,12 +8,20 @@
 namespace amrex
 {
 
+/**
+ * \brief Add a value to a running sum with Kahan compensation.
+ *
+ * Callers own both the accumulated sum and the compensation storage. Both
+ * should be initialized before the first call, usually to zero.
+ */
 template <typename T, typename U>
 AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
-void compensatedSet (T& sum, T& compensation, U value) noexcept
+void compensatedAdd (T& sum, T& compensation, U value) noexcept
 {
-    sum = static_cast<T>(value);
-    compensation = T(0);
+    T const y = static_cast<T>(value) - compensation;
+    T const t = sum + y;
+    compensation = (t - sum) - y;
+    sum = t;
 }
 
 }
--- a/Src/Base/AMReX_CompensatedSum.H
+++ b/Src/Base/AMReX_CompensatedSum.H
@@ -1,0 +1,301 @@
+#ifndef AMREX_COMPENSATEDSUM_H_
+#define AMREX_COMPENSATEDSUM_H_
+#include <AMReX_Config.H>
+
+#include <AMReX_BLassert.H>
+#include <AMReX_GpuQualifiers.H>
+#include <AMReX_MultiFab.H>
+#include <AMReX_REAL.H>
+
+#include <type_traits>
+#include <utility>
+
+namespace amrex
+{
+
+template <typename T, typename AccumT = T>
+struct CompensatedSum
+{
+    AccumT sum = AccumT(0);
+    AccumT comp = AccumT(0);
+
+    template <typename U>
+    AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+    void add (U value) noexcept
+    {
+        AccumT const y = static_cast<AccumT>(value) - comp;
+        AccumT const t = sum + y;
+        comp = (t - sum) - y;
+        sum = t;
+    }
+
+    AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+    void add (CompensatedSum const& rhs) noexcept
+    {
+        add(rhs.value());
+    }
+
+    AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+    AccumT value () const noexcept { return sum; }
+
+    AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+    AccumT compensation () const noexcept { return comp; }
+
+    AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+    void reset () noexcept
+    {
+        sum = AccumT(0);
+        comp = AccumT(0);
+    }
+
+    AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+    void set (T value) noexcept
+    {
+        sum = static_cast<AccumT>(value);
+        comp = AccumT(0);
+    }
+};
+
+template <typename T, typename U>
+AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+void compensatedAdd (T& sum, T& comp, U value) noexcept
+{
+    T const y = static_cast<T>(value) - comp;
+    T const t = sum + y;
+    comp = (t - sum) - y;
+    sum = t;
+}
+
+template <typename T>
+struct CompensatedRef
+{
+    T* sum = nullptr;
+    T* comp = nullptr;
+
+    template <typename U>
+    AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+    void add (U value) const noexcept
+    {
+        compensatedAdd(*sum, *comp, value);
+    }
+
+    AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+    T value () const noexcept { return *sum; }
+
+    AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+    T compensation () const noexcept { return *comp; }
+
+    AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+    void set (T value) const noexcept
+    {
+        *sum = value;
+        *comp = T(0);
+    }
+};
+
+template <typename SumView, typename CompView>
+struct CompensatedView
+{
+    SumView sum;
+    CompView comp;
+
+    template <typename... I>
+    AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+    auto operator() (I... idx) const noexcept
+    {
+        using T = std::remove_cv_t<std::remove_reference_t<decltype(sum(idx...))>>;
+        return CompensatedRef<T>{&(sum(idx...)), &(comp(idx...))};
+    }
+
+    template <typename U>
+    AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+    void add (int i, int j, int k, U value) const noexcept
+    {
+        (*this)(i,j,k).add(value);
+    }
+
+    template <typename U>
+    AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+    void add (int i, int j, int k, int n, U value) const noexcept
+    {
+        (*this)(i,j,k,n).add(value);
+    }
+};
+
+template <typename SumView, typename CompView>
+AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+auto makeCompensatedView (SumView const& sum, CompView const& comp) noexcept
+{
+    return CompensatedView<SumView,CompView>{sum, comp};
+}
+
+class CompensatedMultiFab
+{
+public:
+    CompensatedMultiFab () noexcept = default;
+
+    CompensatedMultiFab (CompensatedMultiFab const&) = delete;
+    CompensatedMultiFab& operator= (CompensatedMultiFab const&) = delete;
+
+    CompensatedMultiFab (CompensatedMultiFab&& rhs)
+        : m_owned_value(std::move(rhs.m_owned_value)),
+          m_owned_comp(std::move(rhs.m_owned_comp))
+    {
+        m_value = (rhs.m_value == &rhs.m_owned_value) ? &m_owned_value : rhs.m_value;
+        m_comp = (rhs.m_comp == &rhs.m_owned_comp) ? &m_owned_comp : rhs.m_comp;
+        rhs.m_value = nullptr;
+        rhs.m_comp = nullptr;
+    }
+
+    CompensatedMultiFab& operator= (CompensatedMultiFab&& rhs)
+    {
+        if (this != &rhs) {
+            bool const value_is_owned = (rhs.m_value == &rhs.m_owned_value);
+            bool const comp_is_owned = (rhs.m_comp == &rhs.m_owned_comp);
+            m_owned_value = std::move(rhs.m_owned_value);
+            m_owned_comp = std::move(rhs.m_owned_comp);
+            m_value = value_is_owned ? &m_owned_value : rhs.m_value;
+            m_comp = comp_is_owned ? &m_owned_comp : rhs.m_comp;
+            rhs.m_value = nullptr;
+            rhs.m_comp = nullptr;
+        }
+        return *this;
+    }
+
+    explicit CompensatedMultiFab (MultiFab& value)
+    {
+        define(value);
+    }
+
+    CompensatedMultiFab (MultiFab& value, MultiFab& compensation) noexcept
+    {
+        define(value, compensation);
+    }
+
+    CompensatedMultiFab (BoxArray const& bxs, DistributionMapping const& dm,
+                         int ncomp, int ngrow,
+                         MFInfo const& info = MFInfo(),
+                         FabFactory<FArrayBox> const& factory = FArrayBoxFactory())
+    {
+        define(bxs, dm, ncomp, ngrow, info, factory);
+    }
+
+    CompensatedMultiFab (BoxArray const& bxs, DistributionMapping const& dm,
+                         int ncomp, IntVect const& ngrow,
+                         MFInfo const& info = MFInfo(),
+                         FabFactory<FArrayBox> const& factory = FArrayBoxFactory())
+    {
+        define(bxs, dm, ncomp, ngrow, info, factory);
+    }
+
+    void define (MultiFab& value)
+    {
+        m_owned_value = MultiFab();
+        m_owned_comp.define(value.boxArray(), value.DistributionMap(), value.nComp(),
+                            value.nGrowVect(), MFInfo().SetArena(value.arena()),
+                            value.Factory());
+        m_value = &value;
+        m_comp = &m_owned_comp;
+        resetCompensation();
+    }
+
+    void define (MultiFab& value, MultiFab& compensation) noexcept
+    {
+        BL_ASSERT(value.boxArray() == compensation.boxArray());
+        BL_ASSERT(value.DistributionMap() == compensation.DistributionMap());
+        BL_ASSERT(value.nComp() == compensation.nComp());
+        BL_ASSERT(value.nGrowVect() == compensation.nGrowVect());
+        m_owned_value = MultiFab();
+        m_owned_comp = MultiFab();
+        m_value = &value;
+        m_comp = &compensation;
+    }
+
+    void define (BoxArray const& bxs, DistributionMapping const& dm,
+                 int ncomp, int ngrow,
+                 MFInfo const& info = MFInfo(),
+                 FabFactory<FArrayBox> const& factory = FArrayBoxFactory())
+    {
+        m_owned_value.define(bxs, dm, ncomp, ngrow, info, factory);
+        m_owned_comp.define(bxs, dm, ncomp, ngrow, info, factory);
+        m_value = &m_owned_value;
+        m_comp = &m_owned_comp;
+        setVal(Real(0.0));
+    }
+
+    void define (BoxArray const& bxs, DistributionMapping const& dm,
+                 int ncomp, IntVect const& ngrow,
+                 MFInfo const& info = MFInfo(),
+                 FabFactory<FArrayBox> const& factory = FArrayBoxFactory())
+    {
+        m_owned_value.define(bxs, dm, ncomp, ngrow, info, factory);
+        m_owned_comp.define(bxs, dm, ncomp, ngrow, info, factory);
+        m_value = &m_owned_value;
+        m_comp = &m_owned_comp;
+        setVal(Real(0.0));
+    }
+
+    void resetCompensation ()
+    {
+        BL_ASSERT(m_comp != nullptr);
+        m_comp->setVal(Real(0.0));
+    }
+
+    void setVal (Real value)
+    {
+        BL_ASSERT(m_value != nullptr);
+        m_value->setVal(value);
+        resetCompensation();
+    }
+
+    [[nodiscard]] MultiFab& value () noexcept
+    {
+        BL_ASSERT(m_value != nullptr);
+        return *m_value;
+    }
+
+    [[nodiscard]] MultiFab const& value () const noexcept
+    {
+        BL_ASSERT(m_value != nullptr);
+        return *m_value;
+    }
+
+    [[nodiscard]] MultiFab& compensation () noexcept
+    {
+        BL_ASSERT(m_comp != nullptr);
+        return *m_comp;
+    }
+
+    [[nodiscard]] MultiFab const& compensation () const noexcept
+    {
+        BL_ASSERT(m_comp != nullptr);
+        return *m_comp;
+    }
+
+    [[nodiscard]] auto array (MFIter const& mfi) noexcept
+    {
+        return makeCompensatedView(m_value->array(mfi), m_comp->array(mfi));
+    }
+
+    [[nodiscard]] auto array (MFIter const& mfi, int start_comp) noexcept
+    {
+        return makeCompensatedView(m_value->array(mfi, start_comp),
+                                   m_comp->array(mfi, start_comp));
+    }
+
+    [[nodiscard]] auto array (MFIter const& mfi, int start_comp, int num_comps) noexcept
+    {
+        return makeCompensatedView((*m_value)[mfi].array(start_comp, num_comps),
+                                   (*m_comp)[mfi].array(start_comp, num_comps));
+    }
+
+private:
+    MultiFab m_owned_value;
+    MultiFab m_owned_comp;
+    MultiFab* m_value = nullptr;
+    MultiFab* m_comp = nullptr;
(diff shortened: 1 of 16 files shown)


zingale:
fyi... we have an exact summation routine in Microphysics:
https://github.com/AMReX-Astro/Microphysics/blob/main/util/esum.H

the routines are generated via:
https://github.com/AMReX-Astro/Microphysics/blob/main/util/esum_cxx.py

although I think now we could probably template it

WeiqunZhang:
Blockers

  1. main.cpp:361-362 — the random flux sign is drawn from a normal instead of a uniform, so the noise is biased.

  auto const normal = local_random_normal<double>(0.0, 1.0, engine);
  ra(i,j,k) = (normal >= 0.5) ? Real(1.0) : Real(-1.0);

  Thresholding a standard normal at 0.5 gives P(+1) = 1-Φ(0.5) = 0.3085, hence E[ra] = -0.383 and Var[ra] = 0.853 — not the 0 and 1 that the flux amplitude alpha*uave*ra/sqrt(dx*dt) assumes.
  Measured over 192,320 draws: empirical mean −0.3807 (predicted −0.3829).

  This breaks the fluctuation–dissipation relation. Equilibrium runs (200k steps, 10k samples):

```  ┌──────────────────────────────────┬───────────────────────────────────────────────────────────┐
  │                                  │                    aver stats variance                    │
  ├──────────────────────────────────┼───────────────────────────────────────────────────────────┤
  │ PR as written                    │ 81.42                                                     │
  ├──────────────────────────────────┼───────────────────────────────────────────────────────────┤
  │ thresholding the uniform instead │ 95.09                                                     │
  ├──────────────────────────────────┼───────────────────────────────────────────────────────────┤
  │ theory k_B T²/(ρ c_v A Δx)       │ 87.72 (×~1.09 forward-Euler inflation at cfl=0.1 → ~95.6) │
  └──────────────────────────────────┴───────────────────────────────────────────────────────────┘
```

  The ratio 81.42/95.09 = 0.856 matches the predicted 0.853 variance deficit exactly. The equilibrium variance is ~14% low, plus a spurious mean flux.

  The >= 0.5 strongly suggests the original Fortran thresholded a uniform, so the fix is probably:
  ra(i,j,k) = (precision_independent_uniform(engine) >= 0.5) ? Real(1.0) : Real(-1.0);
  (which also makes local_random_normal/inverse_standard_normal dead code). Using ra = normal directly would also be statistically valid — your call which was intended.
  ra(i,j,k) = (precision_independent_uniform(engine) >= 0.5) ? Real(1.0) : Real(-1.0);
  (which also makes local_random_normal/inverse_standard_normal dead code). Using ra = normal directly would also be statistically valid — your call which was intended.

  2. main.cpp:893-894 — every MPI rank gets the same RNG seed, so the noise is duplicated across ranks.

  amrex::InitRandom(static_cast<ULong>(inputs.seed), ParallelDescriptor::NProcs(),
                    static_cast<ULong>(inputs.seed));

  InitRandom seeds thread tid with cpu_seed + tid*nprocs (AMReX_Random.cpp:115) — no rank dependence. AMReX's own default is InitRandom(MyProc()+1, NProcs()) (AMReX.cpp:700). The GPU path is the
  same: ResizeRandomSeed(gpu_seed) with an identical gpu_seed everywhere.

  Verified with mpiexec -n 2 (npts=32, max_grid_size=16, one box per rank). Dumping ranflux after OverrideSync, step 1:
  faces  0..15 :  1  1 -1 -1 -1 -1  1 -1 -1 -1  1 -1 -1 -1  1 -1
  faces 17..31 :  1 -1 -1 -1 -1  1 -1 -1 -1  1 -1 -1 -1  1 -1      <- == faces 1..15
  Reproduced on step 3 as well. The second half of the domain gets a bit-for-bit copy of the first half's noise — spatially periodic noise with period = box size, which is exactly what a
  fluctuating-hydro test must not have. And setup_test runs this test under mpiexec -n 2 whenever AMReX_MPI=ON, so the CI configuration hits it. Folding ParallelDescriptor::MyProc() into both
  seeds fixes it while keeping reproducibility at fixed rank count.

  ---
  Should address

  3. AMReX_CompensatedSum.H — the compensation is silently optimized away under -ffast-math. Demonstrated (10M adds of 1e-4f):
  -O3            : compensated = 1000.000000   naive = 1087.724243
  -O3 -ffast-math: compensated = 1028.590942   naive = 1028.590942   <- identical
  -fassociative-math proves (t - sum) - y == 0. This matters because Cray clang-based builds add -ffast-math unconditionally (Tools/GNUMake/comps/cray.mak:56), and AMReX_FASTMATH=ON does the same
  for GCC/Clang/Intel. Since this is a new public API whose only purpose is the compensation, it should be robust or say it isn't. Making the two intermediates volatile survives -ffast-math on
  both gcc 14 and clang 21 (verified). nvcc's --use_fast_math doesn't reassociate adds, so CUDA device code is unaffected.

  4. The Doxygen block has no \param entries. Worth adding, along with a note that compensation must be zero-initialized and T should be a floating-point type.

  ---
  Nits

  - inputs:26-27 — jmidl = 26, jmidr = 75 with npts = 32. Inert (midfact = 1.0) but a confusing leftover from a 100-cell setup.
  - inputs:38 — write_ascii = 1 makes every ctest run drop 24 .dat files into the test working directory. Suggest 0 for the checked-in inputs.
  - main.cpp:744-750 and 764-770 — var_mean…var_mom4 and c11_v…c22_v are computed and then discarded via ignore_unused. Either print them (the Fortran presumably did) or delete them.
  - main.cpp:216-217 — jpartl/jpartr are parsed but never used.
  - is_gaussian only sets ncoef=2 and the filename tag; there's no Gaussian solver path. Unlike is_hybrid and irestart, it isn't guarded — consider aborting on nonzero.
  - The test's only pass criterion is exit status, so it won't catch answer changes. Fine if intentional, but worth stating in the PR.

  The central-moment algebra in build_diagnostics (Cor11–Cor22, mom3, mom4) all checks out symbolically, and the reference-state subtraction does correctly avoid the cancellation the PR set out
  to fix. The header is self-contained and correctly registered in both CMakeLists.txt and Make.package.

WeiqunZhang:
We should address (3) and (4).

Src/Base/AMReX_CompensatedSum.H
```diff
@@ -0,0 +1,89 @@
...
+void compensatedAdd (T& sum, T& compensation, U value) noexcept
+{
+#if defined(__clang__)
+#pragma clang fp reassociate(off) contract(off)
+#endif
+    T const y = static_cast<T>(value) - compensation;
+    T const t = sum + y;
+    compensation = (t - sum) - y;
+    sum = t;
+}
```
ax3l: @atmyers in case it's useful, there are also higher orders of this when needed:
- https://github.com/BLAST-ImpactX/impactx/pull/1195
- https://en.wikipedia.org/wiki/Kahan_summation_algorithm#Further_enhancements

Each of those variants though push up your memory/register footprint by another `T`.

WeiqunZhang (approved):

WeiqunZhang:
I did not test it myself. But my AI agent claims that at least for GCC 15.2.0 and Clang 21.1.8 on Ubuntu 26.04, the following code without any pragmas works with fastmath.
```                            
template <typename T, typename U>
void compensatedAdd (T& sum, T& compensation, U value) noexcept
{
    T const y = static_cast<T>(value) - compensation;
    T volatile t = sum + y;
    T volatile c = t - sum;
    compensation = c - y;
    sum = t;
}
```

Would you be able to test it with more compilers using the compiler explorer setup you mentioned?

WeiqunZhang:
I asked AI to test with Compiler Explorer.

# Kahan Summation: `volatile` vs. Pragma Protection — Findings

Investigation: can `volatile` replace the compiler-specific fast-math pragmas in
`AMReX_CompensatedSum.H` (PR #5587)?

## TL;DR

Yes. Marking the two intermediates `T volatile t = sum + y;` and
`T volatile c = t - sum;` as `volatile` makes the function robust to
`-ffast-math` / `/fp:fast` on every compiler where fast-math is dangerous,
**and** it survives `AMREX_FORCE_INLINE` — which the pragma version explicitly
cannot. It eliminates the entire wall of `float_control` / `GCC optimize` /
`clang fp` / `nv_diag_suppress` pragmas.

## Critical detail

`volatile` on `t` **alone is not sufficient** (and can be *worse* than naive).
Under `-fassociative-math`, `(t - sum) - y` can reassociate to `t - (sum + y)`;
since the recomputed `sum + y` is bit-identical to the stored `t`, the result
collapses to 0. Both `t` and the intermediate `(t - sum)` must be `volatile`
to force materialization at two points the optimizer cannot cross.

## The function

```cpp
/**
 * \brief Add a value to a running sum with Kahan compensation.
 *
 * Callers own both the accumulated sum and the compensation storage. Both
 * should be initialized before the first call, usually to zero.
 *
 * Kahan compensation relies on (t - sum) - y evaluating to the rounding
 * error lost by sum + y, rather than to the exact-arithmetic value of 0.
 * Optimizations that assume floating-point addition is associative (e.g.
 * -ffast-math, -fassociative-math) are free to simplify that expression to
 * 0, which silently turns this into a plain, uncompensated sum. The
 * `volatile` qualifiers below force the intermediate values to be
 * materialized, creating barriers the optimizer cannot reassociate across.
 * This is robust to inlining and to the caller's optimization flags, so the
 * function may be AMREX_FORCE_INLINE'd.
 *
 * Note: volatile on `t` alone is NOT sufficient -- the compiler can still
 * reassociate `(t - sum) - y` into `t - (sum + y)`, and the recomputed
 * `sum + y` is bit-identical to the stored `t`, collapsing the result to 0.
 * Both `t` and the intermediate `(t - sum)` must be volatile.
 *
 * \param[in,out] sum Running sum updated by \p value.
 * \param[in,out] compensation Kahan compensation term updated to track the
 *                             rounding error from the addition.
 * \param[in] value Value to add to \p sum.
 */
template <typename T, typename U>
void compensatedAdd (T& sum, T& compensation, U value) noexcept
{
    T const y = static_cast<T>(value) - compensation;
    T volatile t = sum + y;
    T volatile c = t - sum;
    compensation = c - y;
    sum = t;
}
```

## Test methodology

The test program below builds three accumulators in one translation unit and
reads `base` and `n` from stdin (preventing constant-folding):

1. `double` reference (exact),
2. plain Kahan (no protection) — fast-math is expected to break this,
3. volatile Kahan (the candidate).

Verdict logic:
- `plain_err > 100` and `volatile_err <= 16`  → **PASS** (fast-math broke
  plain; volatile held).
- `plain_err > 100` and `volatile_err > 100`  → **FAIL** (both broke).
- `plain_err <= 100`                          → **INCONCLUSIVE** (fast-math
  did not break plain; nothing to protect).

Inputs: `base = 1e8` (float ULP = 8, so adding 1.0f is lost by a naive sum),
`n = 10000`, reference = `100010000`.

## Test source

```cpp
#include <cstdio>
#include <cmath>

#if defined(_MSC_VER) && !defined(__clang__)
#  define FINLINE __forceinline
#else
#  define FINLINE inline __attribute__((always_inline))
#endif

// Unprotected Kahan -- fast-math is expected to collapse this.
template <typename T, typename U>
FINLINE void kahan_plain (T& sum, T& comp, U value)
{
    T const y = static_cast<T>(value) - comp;
    T const t = sum + y;
    comp = (t - sum) - y;
    sum = t;
}

// Volatile-protected Kahan -- the candidate.
template <typename T, typename U>
FINLINE void kahan_volatile (T& sum, T& comp, U value)
{
    T const y = static_cast<T>(value) - comp;
    T volatile t = sum + y;
    T volatile c = t - sum;
    comp = c - y;
    sum = t;
}

int main() {
    float base; int n;
    if (std::scanf("%f %d", &base, &n) != 2) { base = 1e8f; n = 10000; }

    double sref = base;
    for (int i = 0; i < n; ++i) sref += 1.0;

    float sp = base, cp = 0.f;
    for (int i = 0; i < n; ++i) kahan_plain<float,float>(sp, cp, 1.0f);

    float sv = base, cv = 0.f;
    for (int i = 0; i < n; ++i) kahan_volatile<float,float>(sv, cv, 1.0f);

    double ep = std::fabs(double(sp) - sref);
    double ev = std::fabs(double(sv) - sref);

    std::printf("ref=%.1f plain_err=%.0f volatile_err=%.0f\n", sref, ep, ev);
    if (ep > 100.0 && ev <= 16.0)
        std::printf("VERDICT: PASS\n");          // fast-math broke plain, volatile held
    else if (ep > 100.0 && ev > 100.0)
        std::printf("VERDICT: FAIL\n");          // fast-math broke both
    else
        std::printf("VERDICT: INCONCLUSIVE\n");  // fast-math did not break plain
    return 0;
}
```

## Results (Compiler Explorer, execution via API)

Local verification: GCC 15.2.0 and Clang 21.1.8 on the box; the table below is
Compiler Explorer. `plain_err` is the error of the unprotected Kahan sum under
the listed fast-math flag; `volatile_err` is the error of the volatile form.

| Compiler | fast-math flags | plain_err | volatile_err | Verdict |
|---|---|---:|---:|---|
| GCC 15.3 (`g153`) | `-O3 -ffast-math` | 10000 | **0** | **PASS** |
| Clang 21.1.0 (`clang2110`) | `-O3 -ffast-math` | 1256 | **0** | **PASS** |
| clang-cl 22.1.6 (`clang-cl2216`) | `/O2 /fp:fast` | 1256 | **0** | **PASS** |
| MSVC v19.43 (`vcpp_v19_43_VS17_13_x64`) | `/O2 /fp:fast` | 10000 | **0** | **PASS** |
| ROCm clang 6.1.2 (`clang-rocm-60102`) | `-O3 -ffast-math` | 1256 | **0** | **PASS** |
| NVCC 13.3 host (`nvcc133`) | `-O3 -Xcompiler -ffast-math` | 10000 | **0** | **PASS** |
| Intel icx 2026 (`icxlatest`) | `-Ofast` / `-fp-model=fast=2` | 0 | 0 | n/a* |
| NVHPC nvc++ 26.5 (`nvcxx_x86_cxx26_5`) | `-fast` | 0 | 0 | n/a* |

\* icx and NVHPC do not break *plain* Kahan even at their most relaxed settings,
so there is nothing to protect against; `volatile_err = 0` confirms the
volatile form is harmless there.

### icx precision-model sweep (all INCONCLUSIVE, plain_err = 0)

| icx flags | plain_err | volatile_err |
|---|---:|---:|
| `-fp-model=fast` | 0 | 0 |
| `-fp-model=fast=1` | 0 | 0 |
| `-fp-model=fast=2` | 0 | 0 |
| `-fp-model=fast -fimf-precision=low` | 0 | 0 |
| plain `-O3` (no fp-model) | 0 | 0 |

Unlike vanilla clang, icx never reassociates this loop-carried reduction.

## AMREX_FORCE_INLINE survival

The pragma version in the PR is *deliberately not* `AMREX_FORCE_INLINE`,
because inlining pulls the body into the caller's fast-math context and the
per-function pragma no longer applies. The volatile form is intrinsic to the
memory operations, so it survives forced inlining. Verified locally on GCC 15
and Clang 21 (`always_inline` honored: zero `call`s in the assembly) across
`-O3`, `-O3 -ffast-math`, and `-Ofast` — `volatile_err = 0` in every case.

## Why volatile works (assembly evidence)

Under `-O3 -ffast-math`, GCC reduces the *plain* Kahan loop to a vectorized
`addps` with **zero** `subss` — the compensation was folded to 0 and deleted.
The volatile form retains the subtraction steps with forced stack round-trips
as barriers:

```
movss  %xmm1, -8(%rsp)   ; store t (volatile)
movss  -8(%rsp), %xmm1   ; reload t (opaque barrier)
subss  %xmm2, %xmm1      ; c = t - sum
movss  %xmm1, -4(%rsp)   ; store c (volatile)
movss  -4(%rsp), %xmm1   ; reload c (opaque barrier)
```

## Caveat / not exercised

NVCC **device** code under `--use_fast_math` was not executed (Compiler
Explorer has no GPU). The PR already established device does not collapse the
expression, so volatile is neutral-or-better there.

## Cost

Two extra stack store/load round-trips per call (visible above). Negligible for
a statistics accumulator; the `AMREX_FORCE_INLINE` win (no call overhead)
partially offsets it.

## Reproduction

- Local: `g++ -O3 -ffast-math kahan_fi_test.cpp` (and clang++).
- Compiler Explorer: `/tmp/opencode/ce_run.py` posts `kahan_ce.cpp` to
  `https://godbolt.org/api/compiler/<id>/compile` with `execute` enabled and
  stdin `100000000 10000`.

zingale:
`Array1D` currently has a `sum()` function.  What do we think about making another function `csum()` that uses this? or some other clever name?

WeiqunZhang:
```
T const y = static_cast<T>(value) - compensation;
```
should also be `T volatile y = ..`. Then with 3 volatiles, the CI tests in this PR pass.

WeiqunZhang:
This is the updated report.

# Kahan Summation: `volatile` vs. Pragma Protection — Findings

Investigation: can `volatile` replace the compiler-specific fast-math pragmas in
`AMReX_CompensatedSum.H` (PR #5587)?

## TL;DR

Yes. Marking **all three** intermediates `T volatile y`, `T volatile t`, and
`T volatile c` as `volatile` makes the function robust to `-ffast-math` /
`/fp:fast` on every compiler where fast-math is dangerous, **and** it survives
`AMREX_FORCE_INLINE` — which the pragma version explicitly cannot. It
eliminates the entire wall of `float_control` / `GCC optimize` / `clang fp` /
`nv_diag_suppress` pragmas.

> **Correction from an earlier draft:** an initial version of this note
> recommended only `t` and `c` as volatile. That was *incomplete* and was caught
> by AMReX's `CompensatedSum` regression test (see "The GCC `y`-substitution
> bug" below). All three — `y`, `t`, and `c` — must be volatile.

## Critical detail: why all three intermediates must be volatile

Fast-math reassociation attacks the Kahan expression at three distinct points.
Each volatile qualifier closes one of them:

1. **`t` volatile** — without it, `(t - sum) - y` reassociates to
   `t - (sum + y)`, and since the recomputed `sum + y` is bit-identical to the
   stored `t`, the result collapses to **0** (total defeat, worse than naive).

2. **`c = t - sum` volatile** — even with `t` volatile, GCC/Clang can still
   reassociate `(t - sum) - y` across the subtraction. Materializing `c` to
   memory forces `fl(t - sum)` to be computed and stored before `- y`.

3. **`y` volatile** — even with `t` and `c` volatile, GCC substitutes the
   non-volatile `y = value - compensation` into `compensation = c - y` and
   reassociates `c - (value - compensation)` into `(c + compensation) - value`,
   a *different rounding* that loses precision. Clang does not do this, which is
   why an earlier Clang-only / ULP=8 test missed it. Making `y` volatile prevents
   the substitution.

Only `y + t + c` (all three) is robust across GCC and Clang at every magnitude.

## The function

```cpp
/**
 * \brief Add a value to a running sum with Kahan compensation.
 *
 * Callers own both the accumulated sum and the compensation storage. Both
 * should be initialized before the first call, usually to zero.
 *
 * Kahan compensation relies on (t - sum) - y evaluating to the rounding
 * error lost by sum + y, rather than to the exact-arithmetic value of 0.
 * Optimizations that assume floating-point addition is associative (e.g.
 * -ffast-math, -fassociative-math) are free to simplify that expression to
 * 0, which silently turns this into a plain, uncompensated sum. The
 * `volatile` qualifiers below force the intermediate values to be
 * materialized, creating barriers the optimizer cannot reassociate across.
 * This is robust to inlining and to the caller's optimization flags, so the
 * function may be AMREX_FORCE_INLINE'd.
 *
 * Note: ALL THREE of y, t, and c must be volatile. Each closes a distinct
 * reassociation path the optimizer can otherwise take under -ffast-math:
 *   - t volatile  : stops `(t - sum) - y` collapsing to 0 via `t - (sum + y)`;
 *   - c volatile  : stops `(t - sum) - y` reassociating across the subtraction;
 *   - y volatile  : stops GCC substituting y into `c - y` and reassociating to
 *                   `(c + compensation) - value` (a different rounding).
 * Marking only a subset is silently wrong on some compiler/magnitude combos.
 *
 * \param[in,out] sum Running sum updated by \p value.
 * \param[in,out] compensation Kahan compensation term updated to track the
 *                             rounding error from the addition.
 * \param[in] value Value to add to \p sum.
 */
template <typename T, typename U>
void compensatedAdd (T& sum, T& compensation, U value) noexcept
{
    T volatile y = static_cast<T>(value) - compensation;
    T volatile t = sum + y;
    T volatile c = t - sum;
    compensation = c - y;
    sum = t;
}
```

## Test methodology

The test program below builds three accumulators in one translation unit and
reads `base` and `n` from stdin (preventing constant-folding):

1. `double` reference (exact),
2. plain Kahan (no protection) — fast-math is expected to break this,
3. volatile Kahan (the candidate).

Verdict logic:
- `plain_err > 100` and `volatile_err <= 16`  → **PASS** (fast-math broke
  plain; volatile held).
- `plain_err > 100` and `volatile_err > 100`  → **FAIL** (both broke).
- `plain_err <= 100`                          → **INCONCLUSIVE** (fast-math
  did not break plain; nothing to protect).

Inputs: `base = 1e8` (float ULP = 8, so adding 1.0f is lost by a naive sum),
`n = 10000`, reference = `100010000`.

> **Caveat on the CE results below:** this `1e8` (ULP=8) input *masks* the GCC
> `y`-substitution problem — at that magnitude the partially-reassociated
> result happened to round to the correct value. AMReX's own test uses `2^25`
> (ULP=4) with exact equality and exposes the flaw; see the next section.

## Test source

```cpp
#include <cstdio>
#include <cmath>

#if defined(_MSC_VER) && !defined(__clang__)
#  define FINLINE __forceinline
#else
#  define FINLINE inline __attribute__((always_inline))
#endif

// Unprotected Kahan -- fast-math is expected to collapse this.
template <typename T, typename U>
FINLINE void kahan_plain (T& sum, T& comp, U value)
{
    T const y = static_cast<T>(value) - comp;
    T const t = sum + y;
    comp = (t - sum) - y;
    sum = t;
}

// Volatile-protected Kahan -- the candidate (all three intermediates volatile).
template <typename T, typename U>
FINLINE void kahan_volatile (T& sum, T& comp, U value)
{
    T volatile y = static_cast<T>(value) - comp;
    T volatile t = sum + y;
    T volatile c = t - sum;
    comp = c - y;
    sum = t;
}

int main() {
    float base; int n;
    if (std::scanf("%f %d", &base, &n) != 2) { base = 1e8f; n = 10000; }

    double sref = base;
    for (int i = 0; i < n; ++i) sref += 1.0;

    float sp = base, cp = 0.f;
    for (int i = 0; i < n; ++i) kahan_plain<float,float>(sp, cp, 1.0f);

    float sv = base, cv = 0.f;
    for (int i = 0; i < n; ++i) kahan_volatile<float,float>(sv, cv, 1.0f);

    double ep = std::fabs(double(sp) - sref);
    double ev = std::fabs(double(sv) - sref);

    std::printf("ref=%.1f plain_err=%.0f volatile_err=%.0f\n", sref, ep, ev);
    if (ep > 100.0 && ev <= 16.0)
        std::printf("VERDICT: PASS\n");          // fast-math broke plain, volatile held
    else if (ep > 100.0 && ev > 100.0)
        std::printf("VERDICT: FAIL\n");          // fast-math broke both
    else
        std::printf("VERDICT: INCONCLUSIVE\n");  // fast-math did not break plain
    return 0;
}
```

## Results (Compiler Explorer, execution via API)

Local verification: GCC 15.2.0 and Clang 21.1.8 on the box; the table below is
Compiler Explorer. `plain_err` is the error of the unprotected Kahan sum under
the listed fast-math flag; `volatile_err` is the error of the volatile form.

| Compiler | fast-math flags | plain_err | volatile_err | Verdict |
|---|---|---:|---:|---|
| GCC 15.3 (`g153`) | `-O3 -ffast-math` | 10000 | **0** | **PASS** |
| Clang 21.1.0 (`clang2110`) | `-O3 -ffast-math` | 1256 | **0** | **PASS** |
| clang-cl 22.1.6 (`clang-cl2216`) | `/O2 /fp:fast` | 1256 | **0** | **PASS** |
| MSVC v19.43 (`vcpp_v19_43_VS17_13_x64`) | `/O2 /fp:fast` | 10000 | **0** | **PASS** |
| ROCm clang 6.1.2 (`clang-rocm-60102`) | `-O3 -ffast-math` | 1256 | **0** | **PASS** |
| NVCC 13.3 host (`nvcc133`) | `-O3 -Xcompiler -ffast-math` | 10000 | **0** | **PASS** |
| Intel icx 2026 (`icxlatest`) | `-Ofast` / `-fp-model=fast=2` | 0 | 0 | n/a* |
| NVHPC nvc++ 26.5 (`nvcxx_x86_cxx26_5`) | `-fast` | 0 | 0 | n/a* |

\* icx and NVHPC do not break *plain* Kahan even at their most relaxed settings,
so there is nothing to protect against; `volatile_err = 0` confirms the
volatile form is harmless there.

### icx precision-model sweep (all INCONCLUSIVE, plain_err = 0)

| icx flags | plain_err | volatile_err |
|---|---:|---:|
| `-fp-model=fast` | 0 | 0 |
| `-fp-model=fast=1` | 0 | 0 |
| `-fp-model=fast=2` | 0 | 0 |
| `-fp-model=fast -fimf-precision=low` | 0 | 0 |
| plain `-O3` (no fp-model) | 0 | 0 |

Unlike vanilla clang, icx never reassociates this loop-carried reduction.

## The GCC `y`-substitution bug (found via AMReX's `CompensatedSum` test)

After porting the (then) `t`+`c`-only volatile version to the `kahan_volatile`
branch, AMReX's `Tests/Base/CompensatedSum` regression test **failed on GCC but
passed on Clang**, built with `CXXFLAGS += -ffast-math`. The test adds `1.0f` to
`2^25` (float ULP = 4) twelve times and asserts **exact** equality with 12.

Reproduced in isolation (`g++ -O3 -ffast-math`):

| intermediates volatile | GCC result | Clang result | expected |
|---|---:|---:|---:|
| `t` only | 0 (collapsed) | 0 (collapsed) | 12 |
| `t`, `c` | **8 (wrong)** | 12 (ok) | 12 |
| `y`, `t`, `c` | **12 (ok)** | 12 (ok) | 12 |

With only `t`+`c` volatile, GCC returns **8, not 12** — and `AMREX_FORCE_INLINE`
does not change this. GCC substitutes the non-volatile
`y = value - compensation` into `compensation = c - y` and, under
`-ffast-math`, rewrites `c - (value - compensation)` as
`(c + compensation) - value`, which rounds differently. Clang does not perform
this substitution, which is why the earlier ULP=8 / Clang-biased testing missed
it. Making `y` volatile closes the last path.

### Faithfulness verification (volatile reproduces strict Kahan bit-for-bit)

Because Kahan does *not* always recover the exact arithmetic sum (e.g. it
genuinely returns 96, not 100, for 100 increments at ULP>=4), the right test is
"does volatile-under-fast-math match strict Kahan," not "does it hit an ideal."
A magnitude sweep (float & double, ULP 0.25 -> 16) was compiled twice — once
strict (plain Kahan, the reference) and once `-O3 -ffast-math` (all-volatile
Kahan, the candidate) — and the numeric outputs diffed:

```
diff <(grep res strict.out) <(grep res fm.out)
# (no differences)
```

Result: the all-volatile form is **numerically identical** to strict plain
Kahan across every tested magnitude, on **both GCC and Clang**. The AMReX test
now passes on both (`make` and `make COMP=llvm`).

## AMREX_FORCE_INLINE survival

The pragma version in the PR is *deliberately not* `AMREX_FORCE_INLINE`,
because inlining pulls the body into the caller's fast-math context and the
per-function pragma no longer applies. The all-volatile form is intrinsic to
the memory operations, so it survives forced inlining. Verified locally on GCC
15 and Clang 21 (`always_inline` honored: zero `call`s in the assembly) across
`-O3`, `-O3 -ffast-math`, and `-Ofast`, at the discriminating ULP=4 magnitude —
the all-volatile form returns the correct value (12) with and without
`always_inline`.

## Why volatile works (assembly evidence)

Under `-O3 -ffast-math`, GCC reduces the *plain* Kahan loop to a vectorized
`addps` with **zero** `subss` — the compensation was folded to 0 and deleted.
The volatile form retains the subtraction steps with forced stack round-trips
as barriers (one per volatile intermediate):

```
movss  ..., y               ; store y (volatile)
movss  ..., t               ; store t (volatile)
movss  ..., t ; subss       ; reload t (opaque); c = t - sum
movss  ..., c               ; store c (volatile)
movss  ..., c ; subss       ; reload c (opaque); compensation = c - y
```

## Caveat / not exercised

NVCC **device** code under `--use_fast_math` was not executed (Compiler
Explorer has no GPU). The PR already established device does not collapse the
expression, so volatile is neutral-or-better there.

## Cost

Three extra stack store/load round-trips per call (one per volatile
intermediate). Negligible for a statistics accumulator; the
`AMREX_FORCE_INLINE` win (no call overhead, which the pragma version cannot
use) partially offsets it.

## Reproduction

- AMReX test: `cd Tests/Base/CompensatedSum && make -j8 && ./main3d.gnu.ex`
  (and `make -j8 COMP=llvm`).
- Local repro of the GCC `y` bug: `/tmp/opencode/repro2.cpp` —
  `g++ -O3 -ffast-math -std=c++20 repro2.cpp`.
- Faithfulness sweep: `/tmp/opencode/cmp.cpp`, compiled `-DMODE=0` (strict ref)
  and `-DMODE=1` (volatile, with `-ffast-math`), diff the `res=` lines.
- Compiler Explorer: `/tmp/opencode/ce_run.py` posts `kahan_ce.cpp` to
  `https://godbolt.org/api/compiler/<id>/compile` with `execute` enabled and
  stdin `100000000 10000` (note: ULP=8 input masks the `y` bug — use the AMReX
  ULP=4 test as the authoritative correctness check).

PR #5597, opened 2026-08-05
SYCL: use platform default context

### LINKED ISSUE #5593: Adopt External GPU Contexts?
Seen in https://github.com/AMReX-Codes/pyamrex/pull/454

When AMReX owns and shuts down the GPU context, I see that at least in SYCL that can cause process level segfaults when mixing AMReX with Intel GPU libraries like [dpctl/dpnp](https://github.com/IntelPython/dpctl).

Should we, similar to how we handle external MPIs, adopt existing contexts and refrain from shutting them down in `amrex::Finalize()`?

ax3l:
## Summary

Use the KHR platform default context when available, with the legacy oneAPI extension and private context as fallbacks. Keep AMReX's asynchronous error handler on its queues.

The initialization API and rank-to-GPU mapping are unchanged: every queue still targets the GPU selected for that MPI rank. Sharing the default context enables interoperability with other SYCL libraries.

## Additional background

Close #5593

After merge, we need to merge https://github.com/AMReX-Codes/pyamrex/pull/606

## Checklist

The proposed changes:

### DIFF
--- a/Docs/sphinx_documentation/source/GPU.rst
+++ b/Docs/sphinx_documentation/source/GPU.rst
@@ -1686,8 +1686,13 @@ it is reset, at which point AMReX restores the previous external stream.
   :cpp:`The_Async_Arena` are still pending, in which case AMReX forces a
   synchronization to keep the arena safe.  The external stream or queue must
   belong to the active AMReX device.  For SYCL, the queue must also use the
-  same SYCL context as AMReX and must be an in-order queue.  AMReX selects the
-  active GPU during :cpp:`amrex::Initialize`, whose overloads accept an
+  same SYCL context as AMReX and must be an in-order queue.  When supported by
+  the SYCL implementation, AMReX uses the selected device platform's default
+  context.  That context can contain several devices, but each MPI rank's
+  AMReX queues still target only the GPU selected for that rank.  This also
+  makes an in-order queue created for that device with the platform default
+  context compatible with AMReX.  AMReX selects the active GPU during
+  :cpp:`amrex::Initialize`, whose overloads accept an
   optional trailing :cpp:`int device_id` argument; pass the desired GPU there
   if an external runtime needs AMReX to adopt a specific device before the
   stream is created.  Conversely, if AMReX should drive the selection, query
--- a/Docs/sphinx_documentation/source/GPU.rst
+++ b/Docs/sphinx_documentation/source/GPU.rst
@@ -1686,13 +1686,9 @@ it is reset, at which point AMReX restores the previous external stream.
   :cpp:`The_Async_Arena` are still pending, in which case AMReX forces a
   synchronization to keep the arena safe.  The external stream or queue must
   belong to the active AMReX device.  For SYCL, the queue must also use the
-  same SYCL context as AMReX and must be an in-order queue.  When supported by
-  the SYCL implementation, AMReX uses the selected device platform's default
-  context.  That context can contain several devices, but each MPI rank's
-  AMReX queues still target only the GPU selected for that rank.  This also
-  makes an in-order queue created for that device with the platform default
-  context compatible with AMReX.  AMReX selects the active GPU during
-  :cpp:`amrex::Initialize`, whose overloads accept an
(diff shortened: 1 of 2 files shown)


Docs/sphinx_documentation/source/GPU.rst
```diff
@@ -1686,8 +1686,13 @@ it is reset, at which point AMReX restores the previous external stream.
...
   belong to the active AMReX device.  For SYCL, the queue must also use the
-  same SYCL context as AMReX and must be an in-order queue.  AMReX selects the
-  active GPU during :cpp:`amrex::Initialize`, whose overloads accept an
+  same SYCL context as AMReX and must be an in-order queue.  When supported by
+  the SYCL implementation, AMReX uses the selected device platform's default
+  context.  That context can contain several devices, but each MPI rank's
+  AMReX queues still target only the GPU selected for that rank.  This also
+  makes an in-order queue created for that device with the platform default
+  context compatible with AMReX.  AMReX selects the active GPU during
+  :cpp:`amrex::Initialize`, whose overloads accept an
```
ax3l: Not sure if this level of detail is needed.

Old text:
```suggestion
  same SYCL context as AMReX and must be an in-order queue.  AMReX selects the
  active GPU during :cpp:`amrex::Initialize`, whose overloads accept an
```

Or maybe a small edition like:
```suggestion
  same SYCL context as AMReX and must be an in-order queue.  AMReX
  selects an active GPU during :cpp:`amrex::Initialize` from the device
  platform's default context, whose overloads accept an
```

ax3l:
@WeiqunZhang I was only able to test this on a local device and have no multi-GPU intel machine at hand. Can you potentially check on Aurora pls that I do not break anything for MPI-parallel runs?

WeiqunZhang:
Does this fix the python issue? Do we need to include a header like sycl/ext/khr/default_context.hpp?

WeiqunZhang:
The oneapi extension is deprecated. https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/deprecated/sycl_ext_oneapi_default_context.asciidoc We can probably remove it.

WeiqunZhang:
Looks like sycl.hpp includes the extension header.

Src/Base/AMReX_GpuDevice.cpp
```diff
@@ -554,9 +554,16 @@ Device::initialize_gpu (bool minimal)
         sycl::platform platform(sycl::gpu_selector_v);
         auto const& gpu_devices = platform.get_devices();
         sycl_device = std::make_unique<sycl::device>(gpu_devices[device_id]);
+#if defined(SYCL_KHR_DEFAULT_CONTEXT)
+        sycl_context = std::make_unique<sycl::context>(platform.khr_get_default_context());
+#elif defined(SYCL_EXT_ONEAPI_DEFAULT_CONTEXT)
+        sycl_context = std::make_unique<sycl::context>(platform.ext_oneapi_get_default_context());
```
ax3l: On the standard `khr_get_default_context`:
- SYCL specification: Khronos-approved optional extension in SYCL 2020 Rev. 10 and later. Supporting implementations define `SYCL_KHR_DEFAULT_CONTEXT == 1`. It is not                      implied merely by `SYCL_LANGUAGE_VERSION`.

- Intel oneAPI 2025.2 is the first release implementing `sycl_khr_default_context`. Not present in 2025.1.
  - 2026.0 shows first deprecation of the oneAPI extension.

ax3l: AMReX already documents Intel oneAPI DPC++ 2025.2 or newer as supported, so I think we can remove here indeed

```suggestion
```

ax3l:
Updated accordingly! Checked the OneAPI versions and the fallback is one point release older than what we document we support.

Yes, this fixed the teardown segfault when co-loading `dpnp` with pyAMReX in the same process. Now I can run tests cleanly and can rely on their exit code again.

WeiqunZhang:
It seems to work on aurora.

WeiqunZhang (approved):

PR #5606, opened 2026-08-12
Optimize STL BVH traversal

ankithadas:
## Summary
  - Using a slab-based ray-box intersection test with reused inverse ray directions.
  - Simplifying point-to-box distance calculations and visiting nearer BVH children first.
  - Allowing the CPU `getBoxType` path to stop once both covered and regular points are found.

## Additional background

For a representative `(128^3)` Stanford bunny case, these changes reduced `STLtools::getBoxType` from about 1.87 s to 0.003 s and reduced build time from about 5.50 s to 3.81s. Results remained unchanged.

## Checklist

The proposed changes:


WeiqunZhang:
Can we split them into three PRs so that it's easier to review?

ankithadas:
Sounds good.

Issue #5605, opened 2026-08-11
Add LLVM Flang CI


WeiqunZhang:
We need to add a CI that can catch #5603.

muk2:
Hi @WeiqunZhang, I have opened a PR to address this, here it is: #5614. Thanks!

Issue #5603, opened 2026-08-11
Fortran Interface Bug - Duplicate `public` attribute


ktras:
I tried to build AMReX and its Fortran Interface with `v22.1.3` of LLVM Flang and encountered a compile time error because of a duplicate `public` attribute in `Src/F_Interfaces/AmrCore/AMReX_fillpatch_mod`. This requires only a very minor change to fix and then LLVM Flang can build AMReX and its Fortran interface.

PR #5604 has been opened to address this.

PR #5614, opened 2026-08-16
Add LLVM Flang CI

muk2:
## Summary

Adds a Linux CI workflow that builds libamrex **and its Fortran interfaces** with LLVM Flang, addressing #5605.

The existing Clang CI compiles C++ with Clang but still uses `gfortran` for Fortran, so it cannot catch LLVM-Flang-specific Fortran errors — e.g. the duplicate `public` attribute in `AMReX_fillpatch_mod.F90` that was fixed in #5604 (#5603). This new job builds with `flang` (and matching `clang`/`clang++`) with `-DAMReX_FORTRAN_INTERFACES=ON`, so such regressions are caught going forward.

### Changes
- `.github/workflows/flang.yml` — new `LinuxFlang` workflow. Builds/installs libamrex + Fortran interfaces with LLVM Flang 20 (NOMPI, Debug, EB + Fortran interfaces on). Uses the shared `check_changes` gate to skip docs-only PRs.
- `.github/workflows/dependencies/dependencies_flang.sh` — installs `flang`/`clang` from apt.llvm.org.

## Checklist

The proposed changes:
- [x] add new capabilities to AMReX (CI coverage for LLVM Flang)


WeiqunZhang:
Just for testing, can we revert the change in #5604 to see if the CI covers it?

WeiqunZhang:
I tested it locally. Without the fix in #5604, the CI would fail.

WeiqunZhang (approved):

PR #5615, opened 2026-08-17
[WIP] Host Thread Safety: Make Global State Race-Free

ax3l:
I wanted to see what it takes to make AMReX thread-safe (enough) to support a [GIL-free Python](https://github.com/AMReX-Codes/pyamrex/issues/612) ([PR 614](https://github.com/AMReX-Codes/pyamrex/pull/614), [docs preview](https://pyamrex--614.org.readthedocs.build/en/614/usage/threading.html)) implementation of pyAMReX. This is what it took and the overhead seems basically negligently (+0.2%) low so far.

I do not expect we need to merge this desperately, but it's an idea I would like to explore and discuss :)

---

Claude artifacts summary: https://claude.ai/code/artifact/81c8f43e-4cdd-483f-8060-2bcf6c24f6f5?via=auto_preview

---

## Summary

AMReX can be driven from host threads it did not create — most commonly Python threads on a free-threaded ([PEP 703](https://peps.python.org/pep-0703/)) interpreter through [pyAMReX](https://github.com/AMReX-Codes/pyamrex), but equally `std::thread` in a C++ application. This makes the process-global state such a program reaches race-free.

One line explains why the surface is as large as it is — `Src/Base/AMReX_OpenMP.H:38`:

```cpp
constexpr int get_thread_num () { return 0; }   // when !AMREX_USE_OMP
```

Without OpenMP every host thread reports itself as OpenMP thread 0. So per-thread state indexed by that number collapsed onto one slot, and `omp critical` / `omp atomic` — AMReX's only mutual exclusion for several global caches — excluded nothing at all.

## What changed

| State | Where | Fix |
|---|---|---|
| `MFIter::depth`, the counter behind `AMREX_ALWAYS_ASSERT(depth == 1)` | `AMReX_MFIter.cpp` | `thread_local` — nesting is a per-thread property, so **each thread may now drive its own MFIter**, including over the same `FabArray` |
| Tile-array, FillBoundary, ParallelCopy, rotate, FillPatch, CrseFine, ParFor caches; `m_BD_count`; all flushers; the region-tag stack | `AMReX_FabArrayBase.cpp` | one `std::recursive_mutex`; the one reader, `FabArray::AllocFabs()`, goes through a new locked `regionTags()` snapshot |
| `FabArrayStats` counters, `comm_meta_data_id` | `AMReX_FabArrayBase.{H,cpp}` | atomics |
| `DistributionMapping::getIndexArray()` — lazily fills two vectors in the *shared* `DMRef` | `AMReX_DistributionMapping.cpp` | mutex, with no unlocked fast path |
| `BoxArray::getHashMap()` / `clear_hash_bin()`, `EBCellFlagFab::m_typemap` (including `resetType()`), ParticleTile runtime-pointer caches | 3 files | `omp critical` → `std::mutex` |
| The **default `Geometry`**. `Setup()` is a check-then-act on it and *every* `Geometry::define()` calls it, so two threads building the first `Geometry` of the process both ran the body and either could read a half-written `RealBox`. The window reopens on every `Initialize()`, since the default hangs off `AMReX::top()` | `AMReX_Geometry.cpp` | one mutex over `Setup()`, `define()`'s reads, all three `ResetDefault*`, the default-copying constructor and `operator>>` |
| `amrex::Random()` generators, indexed by `get_thread_num()` | `AMReX_Random.cpp` | per-thread generator; `InitRandom()` serialised, and it claims `generators[0]` so that slot stays tied to the OpenMP master |
| ParmParse table — a *query* writes the entry's use count, type hint and last value | `AMReX_ParmParse.cpp` | one `std::recursive_mutex`; `queryAdd()`/`queryAddWithParser()` hold it across the query-then-add (a new `tableMutex()` accessor lets the header templates do that); a new `tableCopy()` for callers that want to walk the table; the `omp atomic` / `omp single nowait` directives are gone, the latter having silently dropped every team member's write but the first |
| **GPU execution-scope flags** — `in_launch_region`, `in_graph_region`, `in_single_stream_region`, `in_nosync_region` | `AMReX_GpuControl.{H,cpp}` | `thread_local` behind out-of-line accessors, with the setters broadcasting so an OpenMP team still observes a guard opened outside it |
| `Gpu::Device::gpu_stream_index` | `AMReX_GpuDevice.{H,cpp}` | `thread_local` — each host thread gets its own stream |
| **`CArena`** — every mutator already took `carena_mutex`, but `PrintUsage`, `operator<<`, `heap_space_used()`, `heap_space_actually_used()`, `sizeOf()` and `freeableMemory()` read that state unlocked, including `m_busylist.size()` on a container another thread is rehashing. `Arena::print_usage`/`print_usage_to_files` are bound in pyAMReX | `AMReX_CArena.{H,cpp}` | readers locked; mutex made recursive because the out-of-memory path re-enters on the same thread; the OOM diagnostic itself uses `try_to_lock` (see below) |
| `Particle::the_next_id` (three mirrored copies) | `AMReX_Particle.H` et al. | `std::atomic<Long>` |
| `Tokenize()` scratch + `strtok`; `VisMF::persistentIFStreams`; `VisMF`'s `s_layout_cache`, reached by `VisMF::Read()` whenever it is handed an undefined MultiFab — which is what pyAMReX's `amr.VisMF.Read(name)` does every call | `AMReX_Utility.cpp`, `AMReX_VisMF.*` | `thread_local`, `strtok_r`, and a mutex for the layout cache |
| `VisMF::Read`'s `static double totalTime` accumulator (only written when `vismf.verbose`) | `AMReX_VisMF.cpp` | `std::atomic<double>` |
| `ParallelContext::Frame::get_inc_mpi_tag()` | `AMReX_ParallelContext.cpp` | mutex — locally unique tags (see the caveat below) |
| Lazy `static X; static bool first;` blocks | `AMReX_ParticleContainerBase.cpp`, `AMReX_ParticleContainerI.H` | function-local statics; the shared `particles.*` settings move to `ParticleContainerBase::ReadStaticParameters()`, guarded once per `Initialize()` rather than once per template instantiation |

## Three worth calling out

**`MFIter::currentDepth()` changes meaning.** It now reports the *calling thread's* nesting, not a process-wide count. It has no in-tree callers, but note this is also an OpenMP behaviour change: a non-master team member used to read the master's incremented value and now reads zero.

**`InitRandom()` reseeds global state.** `ParticleContainer::InitRandom()` forwards to it, and on a GPU build it frees and reallocates the device RNG array — four threads calling it concurrently was a host-heap double free. It is now serialised, but no locking makes reseed-while-others-draw meaningful, so it is documented as belonging on one thread before workers start. That is an API-shape question worth a separate look.

**The GPU execution-scope flags could not be left to caller discipline.** They were process-global with save/restore RAII guards, and the first version of this PR documented them as the caller's problem. That is not achievable: AMReX opens those guards *itself* inside ordinary API calls — `ParticleContainer::addParticles()` opens a `Gpu::NoSyncRegion`, `Gpu::AnyOf()` behind `MultiFab::contains_nan()`/`contains_inf()` opens a `Gpu::LaunchSafeGuard`, and both are bound in pyAMReX. Two threads calling `add_particles()` interleave as save(false) / save(true) / restore(false) / restore(true), leaving `in_nosync_region` stuck `true` for the rest of the run, after which AMReX silently stops synchronizing streams. Hence the per-thread rework. AMReX's own `omp parallel if (Gpu::notInLaunchRegion())` clauses are evaluated by the encountering thread, so they read the thread-local; only queries made *inside* a team read the broadcast value.

Relatedly, the out-of-memory diagnostic must never block. `Arena::out_of_memory_abort()` walks *every* arena taking each `carena_mutex`, and `the_pinned_arena` is a separate `CArena` even in a CPU-only build — so a thread OOMing inside `The_Arena` and one OOMing inside `The_Pinned_Arena` would deadlock against each other, with no diagnostic, in exactly the situation the diagnostic exists for. Two threads exhausting memory at once is the ordinary case under free threading, so `CArena::PrintUsage(std::ostream&)` and `operator<<` take the lock with `try_to_lock` and print unlocked (marked as such) when it is busy.

## Docs

New `Host Thread Safety` section in `Basics.rst` stating what is guaranteed and what stays the caller's responsibility: concurrent writes to the same data, `Initialize`/`Finalize`, MPI without `AMReX_MPI_THREAD_MULTIPLE`, `TinyProfiler`, `Gpu::setExternalStream()`, dynamic `MFIter` scheduling, the unlocked `AMReX::empty()`/`size()`/`top()` accessors, and `EB2::Build()`'s unguarded index-space stack. The GPU execution-scope guards have *moved off* that list, since they are now per-thread. The `Random()` FAQ entry is updated, including the caveat that inside an OpenMP parallel region the generator is still selected by thread number.

## Testing

Exercised through pyAMReX's free-threading suite (AMReX-Codes/pyamrex#614) on CPython 3.14 free-threaded, built against this branch:

- **CPU** — full pyAMReX suite 249 passed / 95 skipped with `PYTHON_GIL=0` and `=1` alike; 450/450 over 30 repeats of the concurrency tests. pyAMReX CI now has a job on a genuinely free-threaded interpreter (`3.14t`) which asserts `Py_GIL_DISABLED`, asserts the GIL is still off *after* importing pyAMReX, and runs the suite both ways — previously nothing in CI checked the headline claim.
- **AMReX's own test suite** — 76/76 with `AMReX_OMP=ON` + MPI + EB in Debug, and 76/76 on CUDA (nvcc 13.3, RTX A2000), including `GPU_ExternalStream_3d`.
- **ThreadSanitizer** — no remaining races in AMReX code. TSan found two of the fixes above (`FabArrayStats`, the `ParticleContainer` lazy init) that value-asserting tests could not. Run against an earlier revision of this branch.
- **SYCL** — run on hardware (Intel Iris Xe, icpx 2026.1) against an earlier revision. At the current head the backends are compile-verified only: AMReX's own CI plus pyAMReX#614's matrix, pinned at this commit, build it green under SYCL, ROCm/HIP 6.3.2, ICPX, MSVC and clang-cl. The Windows jobs matter here — MSVC forbids `dllexport` on a `thread_local`, which is why the stream index and the execution-scope flags sit behind out-of-line accessors.
- **Cost** — an `MFIter` + `FillBoundary` loop, six alternating runs against an unmodified `development` build: paired mean **+2.0 %, standard deviation 4.4 %**, best-of-six **−0.3 %**. Each build's own run-to-run spread is 10–13 % on this machine, so the added locking is below what this setup can resolve; it is not a measurement of zero, just of "smaller than the noise". A quiet machine would be needed to put a real bound on it.

Several of the fixes came from tests failing (the `DistributionMapping` segfault, the CUDA `InitRandom` double free), two from TSan, and a large majority from six rounds of adversarial review — including three cases where a fix in one round introduced the bug found in the next (the `CArena` OOM deadlock and its cross-arena successor, and the RNG slot-0 claim).

## Not addressed here

- **`Gpu::Device::gpuStream()` no longer inlines** (GPU builds). It calls the out-of-line `Device::getStreamIndex()`, so every kernel launch, `streamSynchronize()` and `freeAsync()` pays a call plus a TLS access plus a bounds check, where `gpu_stream_pool[gpu_stream_index[tid]]` used to inline; the scope-flag queries above are out-of-line for the same reason. Deliberate: the alternative is an `extern thread_local` in the header, and MSVC forbids `dllexport` on a `thread_local` — on a Windows shared build an unexported one would give the DLL and the module *separate* instances, i.e. silently wrong streams. Against a ~5–10 µs launch the added ~10–20 ns is well under 1 %, but anyone measuring many-small-kernel launch overhead should check it and can add a platform-specific fast path.
- An `AMReX_OMP=ON` build driven from **several host threads at once** remains out of scope: the RNG selects by OpenMP thread number inside a parallel region, and the GPU scope flags share their broadcast slots. Pick one thread pool.
- `noexcept` functions that now take a lock (`CArena::heap_space_used`/`heap_space_actually_used`/`sizeOf`, `Geometry::ResetDefault*`, the four `EBCellFlagFab` queries): `std::scoped_lock`'s constructor can in principle throw `std::system_error`, which inside `noexcept` is `std::terminate`. Dropping `noexcept` would change public signatures, and AMReX aborts on errors anyway.
- `EBCellFlagFab::getType(Box)` takes a process-global mutex per call. That matches the `omp critical` it replaced when `AMReX_OMP=ON`, but in an OMP=OFF build the old code compiled to nothing — the serial build is the one that pays for the correctness. Same shape as the `fabarray_cache_mutex` and `g_table_mutex` additions.
- `FabArray::build_arrays()` (`AMReX_FabArray.H:2092`) builds a `mutable` cache from a `const` method with no lock. Not reachable from pyAMReX, but it contradicts "concurrent reads of one object are safe".
- `MFIter::nextDynamicIndex` is correct within an OpenMP team but shared across teams.
- The region-tag stack is now locked, but a *stack* shared by threads is only meaningful when one thread uses it — `RegionTag` is RAII, so two overlapping guards on different threads still pop each other's tag. `thread_local` would fix that and would also stop an OpenMP team inheriting a tag opened outside it; only `Src/Amr` uses it today.
- MPI: `get_inc_mpi_tag()` is now locally unique, but two ranks can still pair up mismatched tags if they admit threads in different orders; the same applies to the `ReduceBoolAnd` inside `VisMF::Read`. AMReX also never stores the provided MPI thread level. Worth its own PR.
- `NFilesIter::currentDeciderIndex` (`AMReX_NFiles.cpp:10`) and `StreamRetry::nStreamErrors` (`AMReX_Utility.cpp:558`) are process-global and written unlocked on the plotfile-write path. Covered by the existing "plotfile writes are MPI-collective, one thread" rule, except in a no-MPI build where two concurrent writes would otherwise be reasonable.
- The lazy MPI datatype/op handles in `AMReX_ParallelDescriptor` are check-then-act on process-global handles. In-tree the only instantiation pyAMReX can reach needs multi-GiB-per-rank MultiFabs; the real exposure is downstream `Bcast<IntVect>`/`<Box>`.
- `AMREX_MEM_PROFILING`-only counters — `num_multifabs*` (`AMReX_MultiFab.cpp:31`) and `BARef::numboxarrays`/`total_*_bytes*` (`AMReX_BoxArray.cpp:26`) — are plain `++`/`--`/max in constructors and destructors. Left alone deliberately: no CI job sets `AMReX_MEM_PROFILE=ON`, so converting them would be an untested change to an untested configuration.
- `EB2::IndexSpace::m_instance` is an unguarded push/erase stack; already documented as the caller's responsibility, as is `Lazy::reduction_queue` (which the CMake build never enables).

## Build note for SYCL

`AMReX_SYCL_SPLIT_KERNEL` defaults `ON`, emitting one device image per kernel. Linking pyAMReX that way drove `sycl-post-link` past 31 GB and it was still climbing after an hour on a 62 GB machine; with `AMReX_SYCL_SPLIT_KERNEL=OFF` the whole build takes about twenty minutes.

### DIFF
--- a/Docs/sphinx_documentation/source/Basics.rst
+++ b/Docs/sphinx_documentation/source/Basics.rst
@@ -2738,10 +2738,16 @@ AMReX's process-global state is safe under that:
   generator, seeded from the same sequence as the OpenMP ones. Only the OpenMP
   generators take part in :cpp:`SaveRandomState`/:cpp:`RestoreRandomState`.
 - Particle IDs (:cpp:`Particle::NextID()`) are handed out atomically.
-- The GPU stream index is per thread, so each host thread submits to a stream of
-  its own from :cpp:`Gpu::Device::gpuStream()`.
+- The current-stream index is per thread, so one host thread can no longer
+  change the stream another is submitting to. The stream *pool* is still
+  shared, and an :cpp:`MFIter` walks it from index 0 in every thread, so two
+  threads can still land on the same stream.
 - :cpp:`amrex::Tokenize()` uses per-thread scratch and returns a per-thread
   result.
+- :cpp:`VisMF`'s persistent input streams are per thread, since an
+  :cpp:`std::ifstream` carries a single read position. Note that
+  :cpp:`VisMF::CloseAllStreams()` therefore closes only the calling thread's;
+  the rest close when their thread exits.
 
 What remains the caller's responsibility:
 
@@ -2753,7 +2759,7 @@ What remains the caller's responsibility:
 - :cpp:`amrex::InitRandom()` -- and therefore
   :cpp:`ParticleContainer::InitRandom()`, which calls it -- *reseeds* the
   global generators, and on a GPU build frees and reallocates the device RNG
-  state. Concurrent calls are serialised, but reseeding while another thread
+  state. Concurrent calls are serialized, but reseeding while another thread
   draws numbers is still a logical race (a use-after-free on the device). Seed
   before starting worker threads.
 - **MPI**, unless AMReX is built with ``AMReX_MPI_THREAD_MULTIPLE=ON``.
--- a/Docs/sphinx_documentation/source/Basics.rst
+++ b/Docs/sphinx_documentation/source/Basics.rst
@@ -2708,6 +2708,65 @@ assertion, one could call
     logical :: old_flag
     old_flag = amrex_mfiter_allow_multiple(.true.)
 
+.. _sec:basics:hostthreads:
+
+Host Thread Safety
+==================
+
+.. highlight:: c++
+
+Besides OpenMP, AMReX can also be driven from host threads it did not create
+itself -- most commonly Python threads on a free-threaded (`PEP 703
+<https://peps.python.org/pep-0703/>`__) interpreter through `pyAMReX
+<https://pyamrex.readthedocs.io>`__, but equally :cpp:`std::thread` in a C++
+application. Such a thread is not part of any OpenMP team, so
+:cpp:`amrex::OpenMP::get_thread_num()` reports ``0`` for it -- and in a build
+without OpenMP it reports ``0`` for every thread.
+
+AMReX's process-global state is safe under that:
+
+- The caches behind :cpp:`MFIter`, :cpp:`FillBoundary` and :cpp:`ParallelCopy`,
+  and the reference counts that decide when they are flushed, are mutex
+  protected.
+- The nesting counter behind the "Nested or multiple active MFIters" assertion
+  is per thread, so **each thread may drive its own MFIter**, including over the
+  same :cpp:`FabArray`. Consequently :cpp:`MFIter::currentDepth()` reports the
+  calling thread's nesting only.
+- The :cpp:`ParmParse` table is mutex protected, including on *query*, which
+  updates the entry's use count and type hint.
+- :cpp:`amrex::Random()` and friends give a thread AMReX did not create its own
+  generator, seeded from the same sequence as the OpenMP ones. Only the OpenMP
+  generators take part in :cpp:`SaveRandomState`/:cpp:`RestoreRandomState`.
+- Particle IDs (:cpp:`Particle::NextID()`) are handed out atomically.
+- The GPU stream index is per thread, so each host thread submits to a stream of
+  its own from :cpp:`Gpu::Device::gpuStream()`.
+- :cpp:`amrex::Tokenize()` uses per-thread scratch and returns a per-thread
+  result.
+
+What remains the caller's responsibility:
+
+- **Concurrent writes to the same data.** Two threads writing the same box, or
+  overlapping regions of one :cpp:`Array4`, is a data race, exactly as it would
+  be inside an OpenMP region.
+- :cpp:`amrex::Initialize()` and :cpp:`amrex::Finalize()`, and process-wide
+  policy such as :cpp:`SetVerbose()` or the error handler: one thread only.
+- :cpp:`amrex::InitRandom()` -- and therefore
+  :cpp:`ParticleContainer::InitRandom()`, which calls it -- *reseeds* the
+  global generators, and on a GPU build frees and reallocates the device RNG
+  state. Concurrent calls are serialised, but reseeding while another thread
+  draws numbers is still a logical race (a use-after-free on the device). Seed
+  before starting worker threads.
+- **MPI**, unless AMReX is built with ``AMReX_MPI_THREAD_MULTIPLE=ON``.
+  Otherwise AMReX calls :cpp:`MPI_Init` rather than :cpp:`MPI_Init_thread`, and
+  collectives must not be entered from two threads at once.
+- :cpp:`TinyProfiler`: its section stack is global and asserts on nesting.
+- :cpp:`Gpu::setExternalStream()`, and dynamic :cpp:`MFIter` scheduling
+  (:cpp:`MFItInfo::SetDynamic`), which shares one counter across an OpenMP team.
+
+Note that host threads and OpenMP are two thread pools over the same cores.
+Using both oversubscribes the node; pick one, or cap ``OMP_NUM_THREADS`` so that
+the product matches the core count.
+
 .. _sec:basics:fortran:
 
 Fortran and C++ Kernels
--- a/Docs/sphinx_documentation/source/Basics.rst
+++ b/Docs/sphinx_documentation/source/Basics.rst
@@ -2768,6 +2768,16 @@ What remains the caller's responsibility:
 - :cpp:`TinyProfiler`: its section stack is global and asserts on nesting.
 - :cpp:`Gpu::setExternalStream()`, and dynamic :cpp:`MFIter` scheduling
   (:cpp:`MFItInfo::SetDynamic`), which shares one counter across an OpenMP team.
+- The GPU execution-scope guards -- :cpp:`Gpu::LaunchSafeGuard`,
+  :cpp:`Gpu::GraphSafeGuard`, :cpp:`Gpu::SingleStreamRegion`,
+  :cpp:`Gpu::NoSyncRegion` -- and :cpp:`Gpu::setExternalStream()`. These set
+  process-wide mode flags and restore them on scope exit, so one thread's guard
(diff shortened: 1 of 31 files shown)


ax3l:
Closing for now, because so far just a draft and can be revived later.

I want to explore one more approach as in https://github.com/AMReX-Codes/pyamrex/pull/614#issuecomment-5373398259 to try something minimal around the compute regions, where we nearly all AMReX functions in a single-threaded region.

PR #5643, opened 2026-08-24
Fix `Random()` returning 1.0f & add `RandomPositive`

### LINKED ISSUE #5638: amrex::Random() can return exactly 1.0f in single precision (GPU), violating its documented [0,1) range; breaks 1-Random() 0-guards
### Summary

`amrex::Random()` is documented to return a uniform real in **`[0, 1)` — "0.0 included, 1.0 excluded"** (`Src/Base/AMReX_Random.H:19-20`). In **single precision on the GPU this contract is violated: `amrex::Random()` can return exactly `1.0f`.** As a direct consequence, the common `1 - amrex::Random()` idiom used to obtain a sqrt- and log-safe `(0, 1]` value **does not work in single precision** — it can still produce exactly `0`, so  `std::sqrt(1.0 - amrex::Random())` or `std::log(amrex::Random())` / `-std::log(...)` sampling can hit `log(0) = -inf`.

This is a refinement of the old #727 / #726: back then the GPU path was changed from `curand_uniform()` to `1.0f - curand_uniform()` to align it with the CPU's `[0, 1)`. That transform is correct in exact arithmetic, but float rounding makes both endpoints reachable in single precision.

### Where

```cpp
// Src/Base/AMReX_Random.H, Real Random (RandomEngine const&)
#ifdef BL_USE_FLOAT
    // GPU, single precision:
    return 1.0f - curand_uniform(random_engine.rand_state);   // L37 (HIP: L36)
#else
    // GPU, double precision:
    return 1.0  - curand_uniform_double(random_engine.rand_state);  // L44 (HIP: L43)
#endif
```
Host path (`Src/Base/AMReX_Random.cpp:133-137`): `std::uniform_real_distribution<Real>(0.0, 1.0)`.

### The bug (single precision, GPU)

`curand_uniform()` returns `(0, 1]` (0 excluded, 1 included). The intended mapping `1.0f - curand_uniform()` sends `(0, 1] → [0, 1)`. But in `float`:

1. **Top endpoint — contract violation.** The representable float just below `1.0f` is `1 - 2⁻²⁴`, so the round-to-nearest midpoint is `1 - 2⁻²⁵`. Whenever `curand_uniform() ≤ 2⁻²⁵`, `1.0f - curand_uniform()` **rounds to exactly `1.0f`** — i.e. `Random()` returns `1.0`, violating the documented "1.0 excluded". Probability ≈ `2⁻²⁵ ≈ 3×10⁻⁸` per draw. (Double precision: threshold `2⁻⁵⁴`, prob ≈ `5×10⁻¹⁷` — rare, but nonzero.)

2. **Bottom endpoint — log(0) footgun.** `curand_uniform()` deliberately **excludes 0** so that `-log(curand_uniform())` is always finite. The `1 - curand_uniform()` transform **reintroduces `0.0`** at the bottom (when `curand_uniform() == 1.0f`, prob ≈ `2⁻²⁴`), so `std::log(amrex::Random())` becomes `log(0) = -inf`. This bites AMReX's **own** `RandomGamma`, which does `std::log(u)` with `u = amrex::Random()` at `Src/Base/AMReX_Random.H:154`.

3. **`1 - amrex::Random()` is not a sufficient guard in single precision.** Downstream codes (e.g. WarpX `SampleGaussianFluxDistribution.H`) guard log-sampling with `1._rt - amrex::Random(engine); // ensures urand > 0`. Because of (1), when `curand_uniform()` is tiny, `amrex::Random()` is exactly `1.0f`, so `1 - amrex::Random() == 1.0f - 1.0f == 0.0f` — back to `log(0)`. This is **not** a fast-math artifact: `amrex::Random()` is a materialized value (the `curand` result is opaque to the compiler), so `1 - (1 - c)` cannot be reassociated back to `c`; it reproduces under both IEEE and `--use_fast_math`.

### Minimal reproduction (the float arithmetic)

```c
#include <stdio.h>
float amrex_random(float curand){ volatile float r = 1.0f - curand; return r; } // materialized, like the real path
int main(){
    for (float c : (float[]){0x1p-24f, 0x1p-25f, 0x1p-26f, 0x1p-32f}) {
        float R = amrex_random(c);              // = amrex::Random()  when curand_uniform()==c
        printf("curand=%.3e  Random=%.9g  1-Random=%.9g\n", c, R, 1.0f - R);
    }
}
// curand=5.96e-08 (2^-24)  Random=1            1-Random=0   <-- Random() == 1.0f (contract violation) AND 1-Random == 0
// curand=2.98e-08 (2^-25)  Random=1            1-Random=0
// ...  identical under `-O2` and `-O2 -ffast-math`
```

### End-to-end symptom (real-world hit)

Found via cross-code benchmarking. In ImpactX single precision on GPU, Box-Muller beam sampling does `std::sqrt(-2 * std::log(u1))` with `u1 = amrex::Random()`. A single `log(0) = -inf` yields a `±inf` particle coordinate; ImpactX fits its dynamically-sized space-charge mesh to the particle min/max, so one `inf` particle collapses the domain to `[-inf, +inf]` and aborts inside `Geometry::computeRoundoffDomain()` **before any tracking**. It is single-precision-only and N-dependent (near-certain by `N ≳ 10⁶`, matching the ~`10⁻⁸` per-draw probabilities above); double precision never triggers in practice. Notably, switching to the `1 - amrex::Random()` guard did **not** fix it (see point 3); an explicit clamp did.

### Possible remedies (for discussion)

Any of, in rough order of preference:
- **Guarantee the documented `[0, 1)` in single precision**, e.g. `amrex::min(1.0f - curand_uniform(...), 1.0f - 2⁻²⁴f)` (or a `nextafter`-style clamp of the top endpoint), so `Random()` is never `1.0f`.
- **Offer/keep a log-safe `(0, 1]` generator** for the `-log(Random())` use case (curand's native interval already provides this — the `1 -` flip is what discards it), and use it inside `RandomGamma`.
- **At minimum, document** that `amrex::Random()` can return exactly `0.0`, and (in single precision) exactly `1.0f`, so consumers of `std::log(Random())` must clamp — and that `1 - amrex::Random()` is *not* a sufficient guard in single precision.

### References

- Related: #727 / #726 (origin of the current `1 - curand_uniform` form; CPU/GPU interval alignment).
- AMReX self-exposure: `Src/Base/AMReX_Random.H:154` (`RandomGamma` → `std::log(amrex::Random())`).
- Downstream mitigation (clamp): ImpactX PR BLAST-ImpactX/impactx#1627.

ax3l:
Fixes #5638.

**What was achieved.** `amrex::Random()` is documented as `[0,1)` but could return exactly `1.0` — on CUDA/HIP because `1 - curand_uniform()` rounds back up for a tiny draw (~3e-8 per call in single precision), and on the host because `std::uniform_real_distribution` is permitted to reach its upper bound (LWG 2524). Both endpoints are now obtained by **relocating a single endpoint of the underlying generator rather than by arithmetic**, which is exact by construction and cheaper. Adds **`amrex::RandomPositive()`**, a uniform real in `(0,1]`, for samplers singular at zero (`-log(u)`, `pow(u,-a)`, Box-Muller) — user code previously wrote `1 - amrex::Random()`, which is *not* a valid guard since it evaluates to exactly `0.0` in single precision. `RandomGamma` now uses it; both are documented in a new "Random Numbers" section of the user's guide.

**What was tested.** New `Tests/Base/Random` in CTest: a deterministic maximal-URBG case (the endpoints are far too rare to sample), the endpoint behaviour of the interval conversions, interval assertions through the `RandomEngine` overloads, and `RandomGamma` finiteness — validated by breaking a guard on purpose to confirm it fails. Passes on CPU (9/9, Small) and single-precision CUDA. On an A2000 the original bug reproduces at **109 violations in 3.9e9 draws** (predicted ~118) and goes to **0**; chi-square over 1024 bins is clean for both generators (z = +0.28 and −0.00). Both precision branches compile clean under IEEE and `--use_fast_math -ftz=true`.

A third commit pins the **MSVC minimum to 19.44 (VS 2022 17.14)**, up from an untested 19.34 — see the details block.

**Does it break user interfaces?** The API is purely additive — one new function, no signature or default changed. **`Random()` does return different values on GPU for a given engine state**, because the interval conversion is no longer `1 - c`, so GPU results and any checksums taken from them will shift. This is not a distribution change: `P(== 0)` is unchanged and chi-square is indistinguishable. CPU streams are unaffected.

<details>
<summary>Details, corner cases, and rejected ideas</summary>

### The clamp constant

Used on the host and SYCL paths, where the native interval is already `[0,1)` and only the vendor defect needs guarding. It is `1 - eps/2`, exactly `std::nextafter(Real(1), Real(0))` in both precisions (verified for `float` and `double`, with and without `-ffast-math`). Being `constexpr`, it works in device code where `std::nextafter` is unavailable — so neither workaround discussed in #5638 is needed: not `1 - epsilon` (2 ULPs below one, needlessly discarding a representable value), and not a host-computed value stashed in `RandomEngine`. Incidentally, libstdc++ already clamps a maximal URBG to exactly this value.

### Safety under all math modes

Both guarantees hold however AMReX is built, because no path performs arithmetic on the draw: every interval is established by a **comparison**, so there is nothing to reassociate, contract or flush. `almost_one` is exactly representable, so constant folding is stable, and the smallest value `RandomPositive()` can return is a **normal** number in both precisions. Verified: all GPU results are byte-identical under `-O3` and `--use_fast_math -ftz=true`.

### Does any of this perturb uniformity?

No. In every case only a single endpoint value is affected and every other value passes through bit-for-bit unchanged, so the distribution restricted to the documented interval is untouched.

- **The GPU remap** relocates `2^-25` of mass from `1.0`, which was never in contract, to `0.0` — which is exactly where the old `1 - c` form put it too, so `P(Random() == 0)` is unchanged at ~`2^-25` (measured: 3.75e-08 over 4e8 draws).
- **The host/SYCL clamp** moves `2^-25` of mass by one ULP, bounding any moment shift by `2^-49` (~2e-15).
- **`RandomPositive`** performs no relocation at all on CUDA/HIP, and on host/SYCL moves `P(u == 0)` onto `1.0` (measured `P(RandomPositive() == 1)` = 2.5e-08).

Measured chi-square over 1024 bins at 4e8 draws in single precision on GPU: `Random()` z = +0.28, `RandomPositive()` z = −0.00. The engine stream is untouched in all cases — no extra draws are consumed and engine state is not altered.

One corner case on the record: under the GPU remap the atom at exactly `0.0` sits next to the generator's fine `~2^-32` grid, so zero is locally overweighted relative to its immediate neighbours, where under `1 - c` the grid near zero was a uniform `2^-24`. `P(== 0)` is identical either way and this is invisible at any binning coarser than ULP scale, but it is the one respect in which the remap's distribution is less locally regular.

### Why the interval conversion is an endpoint remap, not `1 - c`

`(c == 1) ? 0 : c` converts `(0,1]` to `[0,1)` exactly, because the two intervals differ in precisely one endpoint. Subtracting from one also converts the interval, but not exactly, and costs more:

- for a draw below half an ULP of one, `1 - draw` rounds back up to `1.0` — the bug;
- `1 - draw` is exact by Sterbenz above `0.5`, which collapses every returned value below `0.5` onto a `2^-24` grid. So the old form could not return anything between `0` and `2^-24`, discarding roughly nine bits for small values. The remap keeps the generator's own `~2^-33` floor (measured: smallest positive `5.8e-10` vs `6.0e-08`).
- it is one comparison instead of a subtract plus a guard.

Measured over 3.93e9 draws, the two are statistically indistinguishable (chi-square 1089.6 vs 1089.2) and have identical `P(== 0)`, so the resolution and cost are gained for nothing. The cost is that every GPU draw changes value, which is called out above.

An endpoint-only variant that keeps the flip — `v = 1 - c; return (v == 1) ? 0 : v;` — would have preserved GPU streams (differing in only 111 draws in 3.9e9), but it **doubles** `P(Random() == 0)` (5.59e-08 vs 2.77e-08), and zero is the endpoint that makes `std::log(u)` infinite. That is also why the host and SYCL paths, which need only a defect guard rather than an interval conversion, *clamp* instead of relocating: clamping does not raise the probability of returning zero. Full analysis with measurements in [this comment](https://github.com/AMReX-Codes/amrex/pull/5643#issuecomment-5414475212).

### Why not just `amrex::RandomNormal` for the Box-Muller case?

For a plain normal deviate, yes — the guide's distribution table lists it directly below, and the doxygen carries a `\see`. Not a coincidence that it is unaffected: `curand_uniform` excludes zero *precisely because* curand's own normal generator needs `log(u)`; the `1 - c` flip discarded exactly the property the vendor put there on purpose. Hand-written Box-Muller still earns its place when the **radial variable itself** is needed — a radially truncated Gaussian (rejection on `RandomNormal` would be an unbounded loop with divergent warps, variable draw counts, and a *square* rather than *circular* cut), or an isotropic direction from a normalized vector of deviates.

### Pinning the MSVC minimum (third commit)

The declared floor was MSVC 19.34 (VS 2022 17.4, Nov 2022), but nothing exercises it. Since AMReX requires C++20, the Windows toolchains it is actually built with are much newer:

- **CI**: the `windows-2022` and `windows-latest` images currently provide **MSVC 19.44.35228** and **19.51.36252**. So 19.44 is the oldest MSVC any test has ever run against.
- **conda-forge**: the `amrex` feedstock, and the downstream `warpx` and `impactx` feedstocks, all build Windows with the `vs2022` compiler package (`c_stdlib: vs`, no lower pin), so distributed binaries are not built with anything older either.

Raised to 19.44 in `CMakeLists.txt` and `BuildingAMReX_Chapter.rst` so the claim matches what is verified. Note this is a **policy change**, not a consequence of the RNG work — happy to split it out or pick a different number. The natural alternative is **19.42 (VS 17.12)**, the first release containing the `generate_canonical` fix — less aggressive, though still untested. Note the tested floor is a moving target: `windows-2022` is on its final servicing baseline and `windows-latest` has already moved to VS 2026, so this number will need revisiting when that image retires.

This also corrected the framing of the host clamp. An earlier revision justified it by "MSVC 19.34 is defective and is supported"; with the floor at 19.44 every supported standard library clamps already. The clamp stays, but for the accurate reason: **the C++20 wording itself does not guarantee the bound** (LWG 2524; MSVC only conformed once it implemented P0952R2 in VS 2022 17.12 = MSVC 19.42, so the whole of the old 19.34–19.41 window is affected and none of the new one is). The interval should be a property of AMReX, not of whichever standard library is in use.

### Known limitations

- The deterministic test pins the *guard* rather than `Random()` itself, because there is no URBG injection point; it catches a regression in the guard or its constant, but the call-site coverage remains statistical. A seam for injecting an engine would fix that — out of scope here, happy to add.
- `Random()`'s interval is a guarantee about the returned value, not about expressions computed from it: `problo + Random()*dx` can still round onto the upper cell face (verified in both precisions). Noted in the guide.
- `FillRandom` remains backend-dependent. Its CPU path is now clamped (free — already looping), but the GPU paths forward to vendor *bulk* generators where a guarantee needs a full extra pass over the array. Documented accurately rather than papered over.
- Naming is bikesheddable: `RandomPositive` fits the `Random*` prefix, `RandomNonZero` reads more literally.
- A `(0,1)` variant excluding *both* endpoints would additionally serve logistic/Cauchy inversion, at the cost of a clamp on every backend. Left out.

### Downstream

ImpactX BLAST-ImpactX/impactx#1627 can drop its `amrex::max(u1, numeric_limits::min())` clamps in favour of `RandomPositive`; WarpX has four sites that want it (`SampleGaussianFluxDistribution.H:49`, `InjectorMomentum.H:320,322`, `DefaultInitialization.H:71`).

</details>

## To Do

- [x] draft 🤖 Generated with [Claude Code](https://claude.com/claude-code)
- [x] self-review
- [x] LLM-based review
- [x] final self-review

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -131,7 +131,7 @@ endif ()
 set_mininum_compiler_version(CXX GNU 11.0)
 set_mininum_compiler_version(CXX Clang 14.0)
 set_mininum_compiler_version(CXX AppleClang 14.0)
-set_mininum_compiler_version(CXX MSVC 19.34)
+set_mininum_compiler_version(CXX MSVC 19.44)
 
 #
 # Set CMAKE_<LANG>_FLAGS_<CONFIG> if not already defined
--- a/Docs/sphinx_documentation/source/Basics.rst
+++ b/Docs/sphinx_documentation/source/Basics.rst
@@ -3412,6 +3412,126 @@ The downside of this is we have to use :fortran:`pointer` instead of
 pass the Fortran pointer to a procedure with explicit array argument
 to get rid of the pointerness completely.
 
+.. _sec:basics:random:
+
+Random Numbers
+==============
+
+AMReX provides a set of pseudo-random number generators.  Most come in two
+overloads: one that takes no arguments, for use on the host, and one that takes
+a :cpp:`RandomEngine` and can be called from inside a GPU kernel.  The engine is
+supplied by :cpp:`ParallelForRNG`, the random-number variant of
+:cpp:`ParallelFor`:
+
+.. highlight:: c++
+
+::
+
+    amrex::ParallelForRNG(N,
+    [=] AMREX_GPU_DEVICE (int i, amrex::RandomEngine const& engine) noexcept
+    {
+        p[i] = amrex::Random(engine);
+    });
+
+The generators are thread safe.  With OpenMP, each thread draws from its own
+independent generator.  The seed can be set with
+:cpp:`amrex::ResetRandomSeed(cpu_seed, gpu_seed)`.
+
+The uniform generators and their intervals
+------------------------------------------
+
+There are two uniform generators on the unit interval, and which one you want
+depends on what you do with the result:
+
+.. table::
+   :align: center
+
+   +----------------------------------+------------+-----------------------------------------+
+   | Function                         | Interval   | Use for                                 |
+   +==================================+============+=========================================+
+   | :cpp:`amrex::Random()`           | ``[0,1)``  | indices, positions within a cell,       |
+   |                                  |            | rejection tests                         |
+   +----------------------------------+------------+-----------------------------------------+
+   | :cpp:`amrex::RandomPositive()`   | ``(0,1]``  | anything singular at zero:              |
+   |                                  |            | ``log(u)``, ``1/u``, ``pow(u,-a)``      |
+   +----------------------------------+------------+-----------------------------------------+
+
+Both endpoints matter in practice, and each generator excludes the one that is
+dangerous for its intended use.
+
+:cpp:`amrex::Random` excludes the upper endpoint, so that
+:cpp:`int(N*amrex::Random())` is always a valid index into an array of length
+``N`` -- the product cannot round up to ``N``.  It *includes* zero.
+
+Note that the guarantee is on the value :cpp:`amrex::Random` returns, not on
+whatever is computed from it.  Arithmetic downstream can still round onto a
+boundary: :cpp:`problo + amrex::Random()*dx` may evaluate to exactly the upper
+face of the cell, since the rounded sum can land there even though the draw is
+strictly below one.  Code that needs a coordinate strictly inside the cell must
+guard the final value, not just the draw.
+
+:cpp:`amrex::RandomPositive` excludes the lower endpoint, so that the result can
+be passed straight to a function that is singular at zero:
+
+.. highlight:: c++
+
+::
+
+    // exponential distribution with mean tau
+    amrex::Real const t = -tau * std::log(amrex::RandomPositive(engine));
+
+    // power law with exponent -a
+    amrex::Real const x = std::pow(amrex::RandomPositive(engine), -a);
+
+There is no accuracy or performance penalty for picking the interval that fits
+the use case: whichever one you ask for is obtained from the underlying
+generator by relocating at most a single endpoint, never by arithmetic.  Both
+guarantees therefore hold in single and double precision, and under every
+floating-point mode AMReX may be built with, including ``-ffast-math`` /
+``--use_fast_math`` and flush-to-zero.
+
+.. warning::
+
+   Do **not** write :cpp:`1 - amrex::Random()` to obtain a non-zero value; draw
+   from :cpp:`amrex::RandomPositive` instead.  The two are not equivalent: the
+   subtraction costs an arithmetic operation and discards resolution, since it
+   is exact only above ``0.5`` and collapses smaller values onto a coarse
+   grid.
+
+Other distributions
+-------------------
+
+.. table::
+   :align: center
+
+   +--------------------------------------------------+------------------------------------------+
+   | Function                                         | Distribution                             |
+   +==================================================+==========================================+
+   | :cpp:`amrex::RandomNormal(mean, stddev)`         | normal                                   |
+   +--------------------------------------------------+------------------------------------------+
+   | :cpp:`amrex::RandomPoisson(lambda)`              | Poisson                                  |
+   +--------------------------------------------------+------------------------------------------+
+   | :cpp:`amrex::RandomGamma(alpha, beta)`           | Gamma                                    |
+   +--------------------------------------------------+------------------------------------------+
+   | :cpp:`amrex::Random_int(n)`                      | uniform integer on ``[0,n-1]``           |
+   +--------------------------------------------------+------------------------------------------+
+   | :cpp:`amrex::Random_long(n)`                     | uniform long on ``[0,n-1]``, host only   |
+   +--------------------------------------------------+------------------------------------------+
+
+To fill an array rather than draw one value at a time, use
+:cpp:`amrex::FillRandom(p, N)` for the uniform distribution and
+:cpp:`amrex::FillRandomNormal(p, N, mean, stddev)` for the normal distribution.
+
+.. note::
+
+   :cpp:`amrex::FillRandom` does not give the same interval on every backend
+   the way :cpp:`amrex::Random` and :cpp:`amrex::RandomPositive` do.  It is
+   ``[0,1)`` on the CPU, but the GPU paths hand back whatever the vendor bulk
+   generator produces -- nominally ``[0,1)`` with SYCL and ``(0,1]`` with CUDA
+   and HIP -- so code that depends on either endpoint should guard the values
+   it reads back, or draw with :cpp:`amrex::Random` /
+   :cpp:`amrex::RandomPositive` instead.
+
 Abort, Assertion and Backtrace
 ==============================
 
--- a/Docs/sphinx_documentation/source/Basics.rst
+++ b/Docs/sphinx_documentation/source/Basics.rst
@@ -3449,8 +3449,8 @@ depends on what you do with the result:
    +----------------------------------+------------+-----------------------------------------+
    | Function                         | Interval   | Use for                                 |
    +==================================+============+=========================================+
-   | :cpp:`amrex::Random()`           | ``[0,1)``  | indices, positions within a cell,       |
-   |                                  |            | rejection tests                         |
+   | :cpp:`amrex::Random()`           | ``[0,1)``  | positions within a cell, rejection      |
+   |                                  |            | tests, general uniform sampling         |
    +----------------------------------+------------+-----------------------------------------+
    | :cpp:`amrex::RandomPositive()`   | ``(0,1]``  | anything singular at zero:              |
    |                                  |            | ``log(u)``, ``1/u``, ``pow(u,-a)``      |
@@ -3459,16 +3459,14 @@ depends on what you do with the result:
 Both endpoints matter in practice, and each generator excludes the one that is
 dangerous for its intended use.
 
-:cpp:`amrex::Random` excludes the upper endpoint, so that
-:cpp:`int(N*amrex::Random())` is always a valid index into an array of length
-``N`` -- the product cannot round up to ``N``.  It *includes* zero.
+:cpp:`amrex::Random` excludes the upper endpoint and *includes* zero.
 
 Note that the guarantee is on the value :cpp:`amrex::Random` returns, not on
 whatever is computed from it.  Arithmetic downstream can still round onto a
 boundary: :cpp:`problo + amrex::Random()*dx` may evaluate to exactly the upper
 face of the cell, since the rounded sum can land there even though the draw is
-strictly below one.  Code that needs a coordinate strictly inside the cell must
-guard the final value, not just the draw.
+strictly below one.  Code that needs a value strictly inside a range must
+therefore guard the result of the computation, not rely on the draw alone.
 
 :cpp:`amrex::RandomPositive` excludes the lower endpoint, so that the result can
 be passed straight to a function that is singular at zero:
--- a/Docs/sphinx_documentation/source/BuildingAMReX_Chapter.rst
+++ b/Docs/sphinx_documentation/source/BuildingAMReX_Chapter.rst
@@ -29,7 +29,7 @@ The minimum toolchain versions we target for C++20 builds are:
 
 * GCC 11 or newer (for both host builds and CUDA host compilers).
 * LLVM Clang 14 or newer, including AppleClang 14 on macOS.
-* Microsoft Visual Studio 2022 (MSVC 19.34 / 17.4) or newer.
+* Microsoft Visual Studio 2022 (MSVC 19.44 / 17.14) or newer.
 * NVIDIA CUDA Toolkit 12.2 or newer.
 * AMD ROCm/HIP 6.0 or newer.
 * Intel oneAPI DPC++ 2025.2 or newer.
--- a/Src/Base/AMReX_Random.H
+++ b/Src/Base/AMReX_Random.H
@@ -38,9 +38,15 @@ namespace amrex
     Real clamp_below_one (Real v) noexcept
     {
         // The constexpr equivalent of std::nextafter(Real(1), Real(0)), which
-        // is not available in device code.
+        // is not available in device code.  That this really is the immediate
+        // predecessor of one follows from the two assertions below: it is less
+        // than one, and the midpoint between it and one already rounds to one,
+        // so no Real lies in between.  Tests/Base/Random additionally checks it
+        // against std::nextafter itself on the host.
         constexpr Real almost_one
             = Real(1) - Real(0.5)*std::numeric_limits<Real>::epsilon();
+        static_assert(almost_one < Real(1));
+        static_assert(Real(1) - Real(0.25)*std::numeric_limits<Real>::epsilon() == Real(1));
         // No generator produces NaN, but note this maps it to almost_one
         // rather than propagating it, since the comparison is false for NaN.
         return (v < almost_one) ? v : almost_one;
--- a/Src/Base/AMReX_Random.H
+++ b/Src/Base/AMReX_Random.H
@@ -22,6 +22,12 @@ namespace amrex
     AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
     Real one_to_zero (Real v) noexcept { return (v == Real(1)) ? Real(0) : v; }
 
+    //! Convert a [0,1) draw to the left-open (0,1] the same way.  Note that
+    //! this is also correct if the generator reaches 1.0 in violation of its
+    //! own contract, since 1.0 is in range here.
+    AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+    Real zero_to_one (Real v) noexcept { return (v == Real(0)) ? Real(1) : v; }
+
     //! Clamp a value into the half-open unit interval [0,1) by pulling an
     //! exact 1.0 down to the largest representable Real below one.  Used
     //! where the generator is natively [0,1) and only needs guarding against
@@ -52,12 +58,11 @@ namespace amrex
     *
     *  \note Since 0.0 is included, the result must not be passed unguarded
     *  to a function that is singular at zero, such as std::log or std::pow
-    *  with a negative exponent.  Note that `1 - amrex::Random()` is \b not
-    *  a way around that: it costs an arithmetic operation and discards
+    *  with a negative exponent.  Draw from amrex::RandomPositive for those
+    *  cases rather than deriving a value with `1 - amrex::Random()`, which is
+    *  \b not equivalent: it costs an arithmetic operation and discards
     *  resolution, because the subtraction is exact only above 0.5 and
-    *  collapses smaller values onto a coarse grid.  Before the upper bound
-    *  above was guaranteed it was unsafe as well, evaluating to exactly 0.0
-    *  whenever Random() returned exactly 1.0.
+    *  collapses smaller values onto a coarse grid.
     *
     */
     Real Random ();
@@ -96,6 +101,98 @@ namespace amrex
         AMREX_IF_ON_HOST((
                 amrex::ignore_unused(random_engine);
                 return Random();
+        ))
+#endif
+    }
+
+    /**
+    * \ingroup amrex_utilities
+    * \brief Generate a pseudo-random real from uniform distribution,
+    *  excluding zero
+    *
+    *  Generates one pseudorandom real number from a uniform
+    *  distribution between 0.0 and 1.0 (0.0 excluded, 1.0 included).
+    *
+    *  This is the counterpart to amrex::Random for samplers that are
+    *  singular at zero, for example
+    *
+    *  \code{.cpp}
+    *      // exponential distribution with mean tau
+    *      Real const t = -tau * std::log(amrex::RandomPositive(engine));
+    *
+    *      // power law with exponent -a
+    *      Real const x = std::pow(amrex::RandomPositive(engine), -a);
+    *  \endcode
+    *
+    *  Prefer this over deriving a non-zero value from amrex::Random by hand
+    *  with `1 - amrex::Random()`, which is lossier for no benefit.  See the
+    *  note on amrex::Random.
+    *
+    *  \note If all you need is a normally distributed value, use
+    *  amrex::RandomNormal rather than a hand-rolled Box-Muller transform on
+    *  top of this function.  It is faster (on GPU it is the vendor's own
+    *  normal generator, and it does not consume two uniforms per deviate) and
+    *  it never touches std::log, so it is unaffected by the endpoint issue
+    *  described above.  Box-Muller written out by hand is only worthwhile
+    *  when the radial variable itself is needed, for instance to truncate a
+    *  Gaussian at a given number of sigma by inverting the radial CDF, or to
+    *  draw an isotropic direction by normalizing a vector of deviates.
+    *
+    *
+    *  There is no accuracy or performance penalty for choosing this interval
+    *  over the half-open [0,1) of amrex::Random.  On CUDA/HIP it is the
+    *  native interval of the underlying generator, so the value is returned
+    *  as drawn; on the host and with SYCL the generator is natively [0,1) and
+    *  only its zero endpoint is relocated to one.  No arithmetic is performed
+    *  on any path, so no resolution is lost and there is nothing that a
+    *  floating-point mode could change: the exclusion of zero holds under
+    *  -ffast-math / --use_fast_math and flush-to-zero alike, because it is
+    *  established by a comparison rather than by a computed bound.
+    *
+    *  The smallest value that can be returned is in every case the smallest
+    *  positive value of the underlying generator, and is therefore
+    *  backend-dependent: roughly 2^-33 in single precision with curand,
+    *  2^-32 with rocrand, and the granularity of the standard library's
+    *  distribution on the host, which for a 32-bit engine is 2^-32 in single
+    *  precision and 2^-64 in double.  All of these are normal numbers, so
+    *  flush-to-zero cannot reach them.
+    *
+    *  \see amrex::Random, amrex::RandomNormal
+    */
+    Real RandomPositive ();
+
+    /// \ingroup amrex_utilities
+    AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+    Real RandomPositive (RandomEngine const& random_engine)
+    {
+#if defined (__SYCL_DEVICE_ONLY__)
+        // oneMKL draws from [0,1); relocating the zero endpoint converts that
+        // to (0,1] with no arithmetic, and stays correct even if the
+        // generator reaches its own upper bound, since 1.0 is in range here
+        mkl::rng::device::uniform<Real> distr;
+        return random_util::zero_to_one(mkl::rng::device::generate(distr, *random_engine.engine));
+#else
+        // (0,1] is the native interval of hiprand/curand, so unlike Random()
+        // this needs no endpoint relocated and no clamp: the draw is
+        // returned exactly as the generator produced it.
+#ifdef BL_USE_FLOAT
+        AMREX_IF_ON_DEVICE((
+                AMREX_HIP_OR_CUDA(
+                        return hiprand_uniform(random_engine.rand_state); ,
+                        return curand_uniform(random_engine.rand_state);
+                )
+        ))
+#else
+        AMREX_IF_ON_DEVICE((
+                AMREX_HIP_OR_CUDA(
+                        return hiprand_uniform_double(random_engine.rand_state); ,
+                        return curand_uniform_double(random_engine.rand_state);
+                )
+        ))
+#endif
+        AMREX_IF_ON_HOST((
+                amrex::ignore_unused(random_engine);
+                return RandomPositive();
         ))
 #endif
     }
@@ -193,7 +290,7 @@ namespace amrex
             } while (v <= 0.0_rt);
 
             v = v * v * v;
-            u = amrex::Random(random_engine);
+            u = amrex::RandomPositive(random_engine); // std::log(u) below
 
             if (u < 1.0_rt - 0.0331_rt * x * x * x * x) {
                 break;
@@ -215,9 +312,9 @@ namespace amrex
     *  Generates one real number (single or double)
     *  extracted from a Gamma distribution, given the Real parameters alpha and beta.
     *  alpha and beta must both be > 0.
-    *  The CPU version of this function relies on the Standard Template Library
-    *  The GPU version of this function relies is implemented in terms of Random
-    *  and RandomNormal.
+    *  The CPU version of this function relies on the Standard Template Library.
+    *  The GPU version is implemented in terms of amrex::RandomPositive and
+    *  amrex::RandomNormal.
     *
     */
     Real RandomGamma (Real alpha, Real beta);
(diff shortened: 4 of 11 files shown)


Docs/sphinx_documentation/source/Basics.rst
```diff
@@ -3412,6 +3412,122 @@ The downside of this is we have to use :fortran:`pointer` instead of
...
+
+::
+
+    // exponential distribution with mean tau
+    amrex::Real const t = -tau * std::log(amrex::RandomPositive(engine));
+
+    // Box-Muller
+    amrex::Real const u1 = amrex::RandomPositive(engine);
+    amrex::Real const u2 = amrex::RandomPositive(engine);
+    amrex::Real const g  = std::sqrt(-2.0*std::log(u1)) * std::cos(2.0*pi*u2);
```
ax3l: I wonder if this is a good example since we also have `amrex::RandomNormal`. Was there a reason we did not end up using this in ImpactX and instead did a Box-Muller @cemitch99 ?

Src/Base/AMReX_Random.H
```diff
@@ -52,6 +88,81 @@ namespace amrex
...
+    *  singular at zero, for example
+    *
+    *  \code{.cpp}
+    *      // exponential distribution with mean tau
+    *      Real const t = -tau * std::log(amrex::RandomPositive(engine));
+    *
+    *      // Box-Muller transform
+    *      Real const u1 = amrex::RandomPositive(engine);
+    *      Real const u2 = amrex::RandomPositive(engine);
+    *      Real const g  = std::sqrt(-2.0*std::log(u1)) * std::cos(2.0*pi*u2);
```
ax3l: Example: to check, as above

Docs/sphinx_documentation/source/Basics.rst
ax3l: Ah, ImpactX was (Claude):

   Writing Box-Muller out by hand pays off only when the
   radial variable is needed in its own right, for example to truncate a
   Gaussian at a given number of sigma by inverting the radial CDF, or to draw
   an isotropic direction by normalizing a vector of deviates.

and

  Gaussian.H genuinely can't use RandomNormal. The cutx/cuty/cutt truncation inverts the
  radial CDF directly — r² = -2 log(1 - u₁(1-e^{-f})) with f = cut²/2, plus the r2scale
  variance renormalization. That needs the radius as an explicit variable. Rejection
  sampling on RandomNormal would be an unbounded loop with divergent warps on GPU, would
  consume a variable number of draws (so reproducibility goes), and — the decisive part
  — would give a square cut in (x, px) rather than the intended circular one. That's a
  different distribution, not a slower route to the same one.

  Waterbag/Thermal use Gaussians as a means to an end — normalizing a 6-vector onto the
  unit sphere for an isotropic direction. RandomNormal would work fine here; the
  normalization doesn't care where the deviates came from.

  Semigaussian, and the pt draws in KVdist/Kurth4D/Triangle, are just plain normals.
  RandomNormal is a drop-in. Those sites are also wasteful: they burn two uniforms and
  throw away the sin half to get one deviate.


Will update example here.

Src/Base/AMReX_Random.H
ax3l: Will update example here.

ax3l:
Codex:

I would request changes on this PR at head `1c683ff3`.

## Findings

1. **High — Host `RandomPositive()` can still return zero on a supported MSVC toolchain.**

   [`AMReX_Random.cpp:140`](https://github.com/AMReX-Codes/amrex/blob/1c683ff3f2d20c89b8b472ea9abe262c21b5bb76/Src/Base/AMReX_Random.cpp#L140)

   The implementation assumes `std::uniform_real_distribution<float>(0,1)` never returns `1.0f`, then subtracts its result from one. However, AMReX supports MSVC 19.34 / VS 2022 17.4 ([compiler floor](https://github.com/AMReX-Codes/amrex/blob/1c683ff3f2d20c89b8b472ea9abe262c21b5bb76/CMakeLists.txt#L134)), whose STL implementation returned the rounded quotient without an endpoint clamp ([source](https://github.com/microsoft/STL/blob/vs-2022-17.4/stl/inc/random#L271-L292)). Microsoft tracked this exact `uniform_real_distribution<float>` defect in [STL #1074](https://github.com/microsoft/STL/issues/1074); it was only fixed in 2024 by [STL #4740](https://github.com/microsoft/STL/pull/4740).

   For an `mt19937` output of `UINT32_MAX`, the supported implementation produces:

   ```text
   distribution=0x1p+0 RandomPositive=0x0p+0
   ```

   Thus `RandomPositive()` can still feed zero to `log`, while host `Random()` can still return one. The same issue remains in CPU [`FillRandom`](https://github.com/AMReX-Codes/amrex/blob/1c683ff3f2d20c89b8b472ea9abe262c21b5bb76/Src/Base/AMReX_Random.cpp#L277), which independently constructs `std::uniform_real_distribution`. All three host paths should clamp the distribution result with `random_util::clamp_below_one` before returning, complementing, or storing it.

2. **Medium — The documented cell-position guarantee is false after subsequent floating-point operations.**

   [`Basics.rst:3462`](https://github.com/AMReX-Codes/amrex/blob/1c683ff3f2d20c89b8b472ea9abe262c21b5bb76/Docs/sphinx_documentation/source/Basics.rst#L3462)

   Although `Random()` itself is strictly below one, the expression `problo + Random()*dx` can round onto the cell's upper boundary. For example, in double precision with `problo = dx = 1` and `Random() = nextafter(1,0)`, the expression evaluates to exactly `2.0`. The text should not say this expression “always stays inside the cell”; callers requiring that guarantee must guard the final coordinate.

3. **Medium — The endpoint fix and new public API have no durable regression coverage.**

   The PR changes no tests. The existing [`RandomNumberGeneration` test](https://github.com/AMReX-Codes/amrex/blob/1c683ff3f2d20c89b8b472ea9abe262c21b5bb76/Tests/GPU/RandomNumberGeneration/main.cpp#L40) only prints sample moments: it does not assert either interval, call `RandomPositive` or `RandomGamma`, or participate in the CMake/CTest suite. Since the original failure is rare, statistical sampling is insufficient; please add deterministic endpoint tests (including a maximal URBG result) and register them in CTest.

4. **Low — The documented minimum return value is wrong for CUDA/HIP.**

   [`AMReX_Random.H:133`](https://github.com/AMReX-Codes/amrex/blob/1c683ff3f2d20c89b8b472ea9abe262c21b5bb76/Src/Base/AMReX_Random.H#L133)

   The comment says the smallest `RandomPositive()` value is `eps/2` (`2^-24` for float). That applies to the complemented host/SYCL path, but CUDA/HIP return their native generator result. CUDA's conversion uses a `2^-32` scale with a `2^-33` offset, so values substantially below `eps/2` are possible. The zero-exclusion/FTZ argument remains valid because these values are still normal, but the lower bound should be described as backend-dependent.

5. **Low — The guide incorrectly says every listed generator has a GPU overload.**

   [`Basics.rst:3420`](https://github.com/AMReX-Codes/amrex/blob/1c683ff3f2d20c89b8b472ea9abe262c21b5bb76/Docs/sphinx_documentation/source/Basics.rst#L3420)

   `Random_long` is included in the later distribution table, but its declaration explicitly states that [there is no GPU version](https://github.com/AMReX-Codes/amrex/blob/1c683ff3f2d20c89b8b472ea9abe262c21b5bb76/Src/Base/AMReX_Random.H#L357). Please qualify the opening paragraph and mark `Random_long` as host-only in the table.

6. **Low — `RandomGamma` Doxygen is stale after the implementation change.**

   [`AMReX_Random.H:291`](https://github.com/AMReX-Codes/amrex/blob/1c683ff3f2d20c89b8b472ea9abe262c21b5bb76/Src/Base/AMReX_Random.H#L291)

   It says the GPU implementation uses `Random` and `RandomNormal`; after this PR it uses `RandomPositive` and `RandomNormal`. This is also a good opportunity to fix the existing “relies is implemented” wording.

## Validation

- `git diff --check 28ba7777ca50c76287aa119252e3f8275e4638f7..1c683ff3f2d20c89b8b472ea9abe262c21b5bb76` — passed.
- `gh pr checks 5643 --repo AMReX-Codes/amrex` — all current checks passed, including CUDA SP, HIP, SYCL, fast-math, Doxygen, and codespell.
- A deterministic reproduction of the VS 17.4 float formula produced `distribution=1.0f` and `RandomPositive=0.0f`.
- A deterministic floating-point check confirmed `1.0 + nextafter(1.0,0.0) == 2.0`, demonstrating that the documented coordinate expression can hit its upper boundary.
- No local AMReX runtime test was run; the current checkout is a different topic branch and its pre-existing untracked files were left untouched.

WeiqunZhang:
What if we do this in Random()
```
auto r = curand_....;
return (r == 1.0f) ? 0.0f : r; 
```
Can you ask your agents to comment or test on it?

ax3l:
Claude:
Thanks — all six addressed at `0cb471fc`. I verified each finding locally rather than taking it on trust; both of the arithmetic claims reproduce exactly as described.

**1. High — host paths can return the endpoint.** Confirmed, and it is not really MSVC-specific: `uniform_real_distribution` is built on `generate_canonical`, which **LWG 2524** allows to return `1.0`. So no implementation is obliged to avoid it — libstdc++ and libc++ happen to clamp, MSVC (in our supported range) did not. I therefore clamped **every** path rather than just the known-defective one — both GPU paths, SYCL, and the host — so the interval becomes a property of AMReX instead of a property of whichever vendor generator or standard library is in use. `FillRandom`'s CPU path is clamped too, as you asked; the GPU bulk paths are not, since a guarantee there needs a full extra pass over the array, so those stay documented as vendor-native.

Incidental confirmation of the constant: on libstdc++ a maximal URBG returns `0.99999994`, which is exactly the `1 - eps/2` this PR clamps to. An implementation that already clamps independently picked the same value.

**2. Medium — cell-position guarantee.** Correct, and my wording was wrong. Reproduced in both precisions, not just double:

```
float   Random=0.99999994039535522461  problo+Random*dx = 2  == problo+dx? YES
double  Random=0.99999999999999988898  problo+Random*dx = 2  == problo+dx? YES
```

The guide now says the guarantee is on the returned value, not on expressions computed from it, and that code needing a strictly-interior coordinate must guard the final value. I dropped the `problo + Random()*dx` claim from the indexing sentence entirely.

**3. Medium — regression coverage.** Added `Tests/Base/Random`, registered in CTest. The deterministic core is the maximal-URBG case you describe. It also asserts both intervals through the `RandomEngine` overloads, which exercises the device paths on a GPU build, and that `RandomGamma` stays finite on both branches. I validated that it actually bites by breaking the guard on purpose:

```
0::Assertion `random_util::clamp_below_one(Real(1)) == below_one' failed
```

Passes on CPU (9/9 in the Small suite) and on a single-precision CUDA build on an A2000.

One limitation worth stating plainly: because `amrex::Random()` draws from an internal generator with no URBG injection point, the deterministic test pins the *guard* (`clamp_below_one`) rather than `Random()` itself. It will catch a regression in the guard or its constant, but not removal of a call site; the interval assertions covering the call sites remain statistical. Making that fully deterministic would need a seam for injecting an engine, which felt out of scope here — happy to add one if you'd prefer.

**4. Low — lower bound.** Fixed; documented as backend-dependent, with the CUDA `2^-32` scale / `2^-33` offset named explicitly. The FTZ argument is restated in terms of "that minimum is a normal number in both precisions" so it stays correct on either path.

**5. Low — `Random_long`.** Fixed: the opening paragraph now says "most come in two overloads" and the table marks it host-only.

**6. Low — stale `RandomGamma` doxygen.** Fixed, including the "relies is implemented" wording.

I also restructured the two commits so the host-side fix sits in the fix commit rather than the feature commit, and retitled it from "on GPU" to "can return exactly 1.0" now that it is not GPU-specific. The tree was verified byte-identical to the state I tested before and after that rewrite.

Local validation: full SP CUDA build and CPU build both 0 warnings / 0 errors; 9/9 CTest; both precision branches compile clean under IEEE and `--use_fast_math -ftz=true`; doxygen 1.9.8 (matching the CI apt version, which is how I reproduced the earlier `log(u)` warning) and sphinx both clean.

ax3l:
> What if we do this in Random()
> ```C++
> auto r = curand_....;
> return (r == 1.0f) ? 0.0f : r; 
> ```
>
> Can you ask your agents to comment or test on it?

@WeiqunZhang I am checking, but our current clamp is already as simple as

```C++
return (v < almost_one) ? v : almost_one;
```
with only a `constexpr` for `almost_one`.

ax3l:
Claude:

## Evaluating `return (c == 1.0f) ? 0.0f : c;` for the GPU path

An alternative to the clamp was proposed for the CUDA/HIP path of `Random()`: rather than flipping the interval and clamping, take the raw generator value and map only its top endpoint down to zero.

```cpp
auto c = curand_uniform(random_engine.rand_state);   // native (0,1]
return (c == 1.0f) ? 0.0f : c;                       // -> [0,1)
```

It is correct: `c` is `(0,1]`, so non-`1` values give `(0,1)` and the `1` case gives `0`, and the union is exactly the documented `[0,1)`. It is also attractive on its face — a pure compare-and-select with **no arithmetic at all**, so it is cheaper than `1 - c` plus a clamp, and there is nothing for reassociation, contraction or flush-to-zero to act on.

I measured it against the other two candidates on real `curand_uniform` rather than reasoning about it alone.

### Measurements

RTX A2000 (`sm_86`), CUDA 13.2, single precision, 3,932,160,000 draws, histogram over 1024 bins.

| variant | `>= 1` | `P(== 0)` | min positive | max value | chi2 / 1023 dof | z |
|---|---|---|---|---|---|---|
| **A** `1 - c` (pre-fix) | **111** | 2.772e-08 | 5.960464e-08 | `1` | 1089.2 | +1.46 |
| **B** `clamp_below_one(1 - c)` (this PR) | 0 | 2.772e-08 | 5.960464e-08 | 0.99999994 | 1089.2 | +1.46 |
| **C** `(c == 1) ? 0 : c` (proposed) | 0 | 2.772e-08 | **5.820766e-10** | 0.99999994 | 1089.6 | +1.47 |

Means and variances agree with theory for all three (`0.5` and `0.0833333`). Every number above is byte-identical when rebuilt with `--use_fast_math -ftz=true`.

So C is fully correct, has exactly the same probability of returning zero as B, and is statistically indistinguishable from it.

### Where C is actually better

C preserves the generator's resolution near zero. Under B the smallest non-zero value is `2^-24` (`5.96e-08`); under C it is roughly two decades finer (`5.82e-10`, approaching curand's `2^-33` floor).

The reason is that for `c` in `[0.5, 1]` the subtraction `1 - c` is exact by Sterbenz, and `c` in that range already lives on the coarse float grid near one, whose spacing is `2^-24`. So `1 - c` lands on a `2^-24` grid near zero: **B cannot represent any value between `0` and `2^-24`**, having discarded about nine bits of resolution for every draw below `0.5`. C never performs the subtraction, so it keeps them.

### Why it is not in this PR

Switching B to C changes the value of essentially **every** draw, not just the pathological ones:

```
draws=655360000  differ=655359954  (100.0000% of draws change value)
```

(The 46 that agree are `c == 0.5`, where `1 - c == c`, plus `c == 1`, where both yield zero.)

That is expected — `1 - c` and `c` are simply different numbers — but it means every downstream GPU checksum and regression baseline would need rebaselining, across AMReX's own tests and dependent codes. B, by contrast, changes only the ~3e-8 of draws that were out of contract to begin with, and only by one ULP, so existing streams are preserved.

A bug fix should not carry a stream break in its luggage. My suggestion is to keep the clamp here and, if the cleaner form and the better near-zero resolution are wanted, take C as its own PR with a release note, where the rebaselining is the visible, intended subject rather than a side effect.

### One corner case worth recording

Under C the atom at exactly `0.0` carries mass `2^-25` while its immediate neighbours sit on curand's fine `~2^-32` grid, so zero is roughly 128x overweighted relative to the values next to it. Under B the grid near zero is `2^-24` throughout, so the same atom is locally consistent.

This is invisible at any binning coarser than ULP scale — the chi-square figures above are indistinguishable — and `P(== 0)` is identical between the two, so it is not a reason to prefer B. It is simply the one respect in which C's distribution is less locally regular, and it should be on the record if C is revisited later.


---

## Addendum: the endpoint-only variant, `1 - c` with `1.0 -> 0`

The 100% figure above deserves a clarification, because it is easy to read variant C as an endpoint tweak when it is not. C returns the raw generator value `c`; the current code returns `1 - c`. Dropping the flip is what changes every draw — the `== 1.0f` test has nothing to do with it. On identical RNG state:

```
   curand c       B: clamp(1-c)  C: (c==1)?0:c  D: 1-c, 1->0
   0.46224481     0.53775519     0.46224481     0.53775519
   0.71655625     0.28344375     0.71655625     0.28344375
   0.30665722     0.69334280     0.30665722     0.69334280
   0.47598666     0.52401334     0.47598666     0.52401334
```

C returns the mirror value on every ordinary draw. B and D agree.

That fourth column is the variant one might actually have in mind when saying "map the 1.0 to 0": keep the interval flip, and send the rounded-up endpoint to zero rather than clamping it down.

```cpp
float v = 1.0f - c;
return (v == 1.0f) ? 0.0f : v;      // D
```

D **is** stream-preserving, and the objection to C does not apply to it. Over the same 3,932,160,000 draws it differs from the clamp in only **111 draws** (2.8e-08 of the total), and it is equally correct — nothing lands outside `[0,1)`. So it is a legitimate candidate, and worth recording why it is still not the one to take.

| | `P(result == 0)` | `P(result == 1-2^-24)` |
|---|---|---|
| **B** clamp (this PR) | 2.7720e-08 | 8.9264e-08 |
| **D** map `1 -> 0` | **5.5949e-08** | 6.1035e-08 |
| *ideal on a `2^-24` grid* | 2.9802e-08 (`2^-25`) | 5.9605e-08 (`2^-24`) |

Both misplace exactly the same `2^-25` of probability mass; they differ only in where they put it. B overweights the top value by a factor 1.5. D instead **doubles the probability of returning exactly zero**.

That is the deciding difference. Zero is the endpoint that carries the hazard — it is what turns `std::log(u)` into `-inf`, which is the failure that opened #5638 in the first place. D would make that outcome twice as likely, while B leaves `P(== 0)` at its natural rate and moves the surplus onto `1-2^-24`, where nothing in common use is singular.

Given that some `2^-25` of mass has to be displaced either way, it should land on the benign endpoint rather than the dangerous one. That is the clamp.

WeiqunZhang:
Variant C is readable to everyone including mortals. Unless one is an expert in both the IEEE-754 standard and how compilers implement the standard, it's hard to be convinced that `1.0f - 0.5f*std::numeric_limits<float>::epsilon()` will always work.

So I think variant C is the way to go.

ax3l:
@WeiqunZhang If you like for the GPU path, I think the flipping 1 to zero instead of 1-c is ok if we are ok that people that have GPU checksums in place will see some noise.

ax3l:
I think so, too. The design tried to avoid churn for drawn values but we can do a simple flip since it seems safe and efficient. (Drawing numbers is anyway more expensive than a compare-and-swap op.)

Docs/sphinx_documentation/source/BuildingAMReX_Chapter.rst
```diff
@@ -23,13 +23,13 @@ Fortran compiler that supports the Fortran 2003 standard, and a C
...
 available in any Unix-like environment (e.g., Perl and sed).  For
-building with CMake, the minimal requirement is version 3.25.
+building with CMake, the minimal requirement is version 3.18.
 
 The minimum toolchain versions we target for C++20 builds are:
 
 * GCC 11 or newer (for both host builds and CUDA host compilers).
 * LLVM Clang 14 or newer, including AppleClang 14 on macOS.
-* Microsoft Visual Studio 2022 (MSVC 19.34 / 17.4) or newer.
+* Microsoft Visual Studio 2022 (MSVC 19.44 / 17.14) or newer.
```
ax3l, line 32: Clarified because this is what we can test in GH CI.
Part of this PR, because versions just before this had a bug in their stdlib returning 1 as a value of the uniform distribution (as a consequence of an ISO C++ lingo defect).

Src/Base/AMReX_Random.cpp
```diff
@@ -130,11 +130,31 @@ Real RandomNormal (Real mean, Real stddev)
...
 }
 
+// std::uniform_real_distribution is specified as [0,1), but that bound is not
+// something to depend on: it is built on std::generate_canonical, whose C++20
+// specification is defective -- LWG 2524 -- in that the quotient it forms can
+// round up to exactly 1.0.  In single precision with a 32-bit engine,
+// (2^32-1)/2^32 rounds to 1.0f.  Implementations are free to clamp and the
+// ones AMReX currently supports do (MSVC only since it implemented P0952R2 in
+// 2024, microsoft/STL#1074), but nothing in the standard AMReX targets
+// requires it.  Clamp rather than trust it: that makes the interval of
```
ax3l:
```suggestion
// round up to exactly 1.0f in single precision. Implementations are free to clamp
// and the ones AMReX currently supports do, including MSVC (fixed
// P0952R2 in 2024 via microsoft/STL#1074).  Since this is a C++20 defect, we
// clamp rather than trust it: that makes the interval of
```

CMakeLists.txt
```diff
@@ -122,7 +131,7 @@ endif ()
 set_mininum_compiler_version(CXX GNU 11.0)
 set_mininum_compiler_version(CXX Clang 14.0)
 set_mininum_compiler_version(CXX AppleClang 14.0)
-set_mininum_compiler_version(CXX MSVC 19.34)
+set_mininum_compiler_version(CXX MSVC 19.44)
```
ax3l, line 134: Intentional, see [inline comment above](https://github.com/AMReX-Codes/amrex/pull/5643#discussion_r3856551420)

Docs/sphinx_documentation/source/Basics.rst
```diff
@@ -3412,6 +3412,126 @@ The downside of this is we have to use :fortran:`pointer` instead of
...
+   +----------------------------------+------------+-----------------------------------------+
+   | :cpp:`amrex::RandomPositive()`   | ``(0,1]``  | anything singular at zero:              |
+   |                                  |            | ``log(u)``, ``1/u``, ``pow(u,-a)``      |
+   +----------------------------------+------------+-----------------------------------------+
+
+Both endpoints matter in practice, and each generator excludes the one that is
+dangerous for its intended use.
+
+:cpp:`amrex::Random` excludes the upper endpoint, so that
+:cpp:`int(N*amrex::Random())` is always a valid index into an array of length
```
WeiqunZhang: N might be so big that it cannot be represented exactly by float. We can avoid discussing this.

lucafedeli88 (commented):
Ah! Good catch...

I think that the proposed solution should work well. It's also a good thing to offer both [0,1) and (0,1] with a clear API.

Potentially, an alternative compact implementation could be to rely on a random unsigned int64 `r64` and, for all the architectures, compute : 

```
(r64 >> 11)*0x1.0p-53
```
to get a uniformly distributed number in [0,1) 

and 

```
(r64 >> 11)*0x1.0p-53 + 0x1.0p-54
```
to get a uniformly distributed number in (0,1].

Similar expressions can be written starting from a 32 bit unsigned integer using 8 and 24 instead of 11 and 53.

From the point of view of portability, the main advantage would be that the behavior of the generator would be much more similar across different architectures. The only difference would be the specific generator used to generate random integers.

Docs/sphinx_documentation/source/Basics.rst
ax3l: Agreed, sub-ideal example here.

ax3l: Updated! :)

ax3l:
> I think that the proposed solution should work well. It's also a good thing to offer both [0,1) and (0,1] with a clear API.

@lucafedeli88 thanks a lot! Glad the API design is all agreed :tada:

> Potentially, an alternative compact implementation ...

That is an interesting idea, but I would avoid that change here for the following reason: what you propose here is a change in sampling algorithm pulling the responsibility for the quality of numbers drawn onto our table instead of std/vendor interfaces. I assume this is based on the papers/work linked in https://www.reddit.com/r/programming/comments/jijhwa/generating_random_floatingpoint_numbers_by/ but really would not open that can of worms here.

WeiqunZhang (approved):

PR #5645, opened 2026-08-25
Use CMake language support for HIP builds

BenWibking:
## Summary
Migrates CMake HIP builds to use CMake's build in support for HIP that is available in CMake 3.21 and newer. (Note that CMake 3.25+ is now already required for CUDA.)

**User-visible changes:**
* Instead of `AMREX_AMD_ARCH` and `AMReX_AMD_ARCH` variables, users should migrate to using `CMAKE_HIP_ARCHITECTURES` to specify the GPU architecture for HIP builds. This unifies the variable names for specifying the GPU arch for both CUDA and HIP. The old spellings are deprecated.
* Since CMake now treats HIP as a language similar to CUDA, `setup_target_for_hip_compilation(target)` is required whenever `setup_target_for_cuda_compilation(target)` would also be used.
* HIP Clang ≥ 17.0 is required (ROCm 6.x or newer)

## Checklist

The proposed changes:


BenWibking:
While theoretically appealing, making this switch will unfortunately break all downstream codes.

PR #5648, opened 2026-08-26
Initialize/Finalize: Pop Hooks Before Calling Them

ax3l:
## Summary

`amrex::Initialize` and `amrex::Finalize` drained their hook stacks with `top()()` followed by `pop()`. A hook that registers another one with `ExecOnInitialize`/`ExecOnFinalize` therefore left the *calling* hook on top of the stack: it ran a second time, and the following `pop()` silently discarded the newly registered hook without ever running it.

Move the hook off the stack before calling it, so that a hook which registers another one runs exactly once and the newly registered hook runs as well.

## Additional background

Found while making the ABLASTR warn manager state per AMReX cycle in https://github.com/BLAST-WarpX/warpx/pull/7186: a warning recorded from a finalize hook resurrects the warn manager singleton and loses the teardown hook that would have released it again.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX.cpp
+++ b/Src/Base/AMReX.cpp
@@ -87,6 +87,7 @@
 #include <stack>
 #include <string>
 #include <thread>
+#include <utility>
 #include <limits>
 #include <vector>
 #include <algorithm>
@@ -441,13 +442,13 @@ amrex::Initialize (int& argc, char**& argv, bool build_parm_parse,
     while ( ! The_Initialize_Function_Stack.empty())
     {
         //
-        // Call the registered function.
+        // Remove the function from the stack before calling it, so that a function
+        // that calls ExecOnInitialize itself neither runs twice nor drops the newly
+        // registered function.
         //
-        The_Initialize_Function_Stack.top()();
-        //
-        // And then remove it from the stack.
-        //
+        auto f = std::move(The_Initialize_Function_Stack.top());
         The_Initialize_Function_Stack.pop();
+        f();
     }
 
     BL_PROFILE_INITIALIZE();
@@ -854,13 +855,13 @@ amrex::Finalize (amrex::AMReX* pamrex)
     while (!The_Finalize_Function_Stack.empty())
     {
         //
-        // Call the registered function.
+        // Remove the function from the stack before calling it, so that a function
+        // that calls ExecOnFinalize itself neither runs twice nor drops the newly
+        // registered function.
         //
-        The_Finalize_Function_Stack.top()();
-        //
-        // And then remove it from the stack.
-        //
+        auto f = std::move(The_Finalize_Function_Stack.top());
         The_Finalize_Function_Stack.pop();
+        f();
     }
 
     // The MemPool stuff is not using The_Finalize_Function_Stack so that


ax3l:
`StochasticHeatEquation_1d` seems to generally fail @WeiqunZhang ? Unrelated timeout, I think?

ax3l:
HPSF issue analyzed by Claude:

Unrelated to this PR — confirmed by bisecting the GitLab pipeline history.

## What actually failed

Pipeline [1779961](https://gitlab.spack.io/amrex/amrex/-/pipelines/1779961): all three GPU jobs (Nvidia-H100 SP, AMD-MI300A, Intel-PVC) died at exactly `execution took longer than 1h0m0s`. In each, the sole failing test is `StochasticHeatEquation_1d`, which never completes:

| job | test wall time when killed |
|---|---|
| Nvidia-H100-single-precision | `***Failed 2383.62 sec` |
| AMD-MI300A | `***Failed 2668.30 sec` |
| Intel-PVC | `Subprocess terminated***Exception: 2520.69 sec` |

`ctest -j 2` starts test #80 early, it occupies one slot for the rest of the hour while the other slot finishes everything else, then the runner kills the job.

## Why it is not this PR

- The nightly schedule on plain `development` from the same day ([job 24106057](https://gitlab.spack.io/amrex/amrex/-/jobs/24106057), no PR code) fails identically: `227/227 Test #80: StochasticHeatEquation_1d ***Failed 2361.67 sec`.
- Every HPSF pipeline has failed since the test landed. Last green: pipeline 1749499, `2026-08-11T03:13Z`. 52ff27ef590429cb4845759ed6b3b1333619c248 (#5587, which adds the test) merged `2026-08-11T06:14Z`. First red: pipeline 1749700, `2026-08-11T06:25Z` — same test, same hang. Unbroken red ever since.
- Mechanically it cannot be this PR anyway: all in-tree hook registrations in `Src/` are `ExecOnFinalize` (teardown), and the hang is mid-time-loop, thousands of seconds before `amrex::Finalize` is reached.

## Root cause of the test itself

`Tests/StochasticHeatEquation/inputs` runs `nstep = 20000000` + `ntherm = 2000000`, times `ensemble = 2` → **44M steps on a 32-cell 1D domain**. Per step, `advance()` with `scheme = 1` does ~10 kernel launches plus `FillBoundary`/`OverrideSync`, and the loop additionally calls `u.sum(icor_box, 0)` (a blocking device reduction) and `accumulate_stats` every single step (`nstat = 1`).

That is pure launch-latency work with no arithmetic to hide it. On CPU it is fine — GitHub's `GNU@13 1D Debug Fortran [tests]` passes it in **354.69 s** (~8 µs/step). On GPU each step costs an order of magnitude more, so it blows the 1 h budget. The Intel-PVC job builds with `AMReX_MPI=OFF` and hangs too, so this is kernel-launch overhead, not MPI.

It only bites HPSF because `.gitlab/hpsf-gitlab-ci.yml` does not set `AMReX_TEST_TYPE`, so it defaults to `All`; the GitHub GPU workflows (`cuda.yml`, `clang.yml`, `gcc.yml`) pass `-DAMReX_TEST_TYPE=Small`, which excludes it.

## Suggested fix

The fix likely belongs in the test rather than in CI timeouts: a CI-sized inputs file (e.g. `nstep`/`ntherm` cut ~100x, `nstat` raised) registered as the ctest command, keeping the full-length inputs for manual science runs. Alternatively, gate the test off when `AMReX_GPU_BACKEND` is set.

WeiqunZhang (approved):

PR #5653, opened 2026-08-28
Warnings: CXX23 Features

ax3l:
## Summary

Both Clang and GCC warn when we use CXX23 features already. Supress them and/or use intrinsics for now to silence them.

## Additional background

Seen downstream in WarpX with Clang 22 (counter) and a recent GCC (for assume).

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_Extension.H
+++ b/Src/Base/AMReX_Extension.H
@@ -168,6 +168,28 @@
 #define AMREX_UNROLL_LOOP(n)
 #endif
 
+// __COUNTER__
+//
+// __COUNTER__ has been a de-facto extension in all compilers AMReX supports for
+// a very long time, but it was only recently adopted into C2y.  Clang 22 and
+// later therefore diagnose it as a C2y extension under -Wpedantic, even in C++
+// (https://github.com/llvm/llvm-project/issues/196557).  Bracket a use of
+// __COUNTER__ with these macros to silence that diagnostic.  The nested
+// -Wunknown-warning-option suppression keeps older Clang, which does not know
+// -Wc2y-extensions, quiet as well.
+
+#if defined(__clang__)
+#define AMREX_COUNTER_WARNING_PUSH                                        \
+    _Pragma("clang diagnostic push")                                      \
+    _Pragma("clang diagnostic ignored \"-Wunknown-warning-option\"")      \
+    _Pragma("clang diagnostic ignored \"-Wc2y-extensions\"")
+#define AMREX_COUNTER_WARNING_POP                                         \
+    _Pragma("clang diagnostic pop")
+#else
+#define AMREX_COUNTER_WARNING_PUSH
+#define AMREX_COUNTER_WARNING_POP
+#endif
+
 // __attribute__((weak))
 
 #if defined(AMREX_TYPECHECK)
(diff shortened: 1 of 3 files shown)


dpgrote (approved):
Thanks! This fixed both the `__COUNTER__` and `ASSUME` warnings.

Src/Base/AMReX_Extension.H
```diff
@@ -168,6 +168,28 @@
...
+// -Wunknown-warning-option suppression keeps older Clang, which does not know
+// -Wc2y-extensions, quiet as well.
+
+#if defined(__clang__)
+#define AMREX_COUNTER_WARNING_PUSH                                        \
+    _Pragma("clang diagnostic push")                                      \
+    _Pragma("clang diagnostic ignored \"-Wunknown-warning-option\"")      \
+    _Pragma("clang diagnostic ignored \"-Wc2y-extensions\"")
+#define AMREX_COUNTER_WARNING_POP                                         \
+    _Pragma("clang diagnostic pop")
```
lucafedeli88:
```suggestion
#define AMREX_COUNTER_WARNING_PUSH                                        \
    _Pragma("clang diagnostic AMReXCounterWarning.push")                                      \
    _Pragma("clang diagnostic ignored \"-Wunknown-warning-option\"")      \
    _Pragma("clang diagnostic ignored \"-Wc2y-extensions\"")
#define AMREX_COUNTER_WARNING_POP                                         \
    _Pragma("clang diagnostic AMReXCounterWarning.pop")
```

Can we use namespaces for these, in case at some point in the future we want to nest multiple PUSH/POP declarations? See here https://clang.llvm.org/docs/LanguageExtensions.html#specifying-an-attribute-for-multiple-declarations-pragma-clang-attribute . 
Warning : I haven't tested the pragmas with namespaces yet.

Src/Base/AMReX_Extension.H
```diff
@@ -233,7 +255,15 @@
...
 //   - nvcc >= 11.1.0
 //   - icx >= 2021.1.2
-#if defined(__has_cpp_attribute) && __has_cpp_attribute(assume)
+//
+// [[assume]] itself is C++23.  Clang >= 19 and GCC >= 13 advertise it through
+// __has_cpp_attribute in older language modes as well, and Clang then diagnoses
+// every use as a C++23 extension under -Wpedantic
+// (-Wc++23-attribute-extensions).  So only take the attribute when the language
+// mode really is C++23 and rely on the compiler-specific spellings below
+// otherwise; they carry the same information to the optimizer.
```
lucafedeli88, line 269: Shouldn't we use `__builtin_assume` for clang in case the feature is supported by the compiler but the standard is not `C++23` ?

lucafedeli88 (changes requested):
Thanks for this PR, @ax3l . These warnings were indeed very annoying ;-) .
I've left a couple of comments.

Src/Base/AMReX_Extension.H
WeiqunZhang: Is that already covered in the else part?

Src/Base/AMReX_Extension.H
WeiqunZhang: Claude says

> The namespaced-pragma review suggestion does not work. That syntax belongs to #pragma clang attribute, not #pragma clang diagnostic; clang 21 rejects it and the diagnostic is not suppressed. Plain push and pop already nest.

WeiqunZhang (approved):

Src/Base/AMReX_Extension.H
lucafedeli88: Oh, yes, it was not expanded on my screen... I didn't see it, sorry.

Src/Base/AMReX_Extension.H
lucafedeli88: True, namespaces can't be used for diagnostics. 
But then `AMREX_COUNTER_WARNING_POP` is maybe a bit misleading. Maybe it could be changed into something like `AMREX_DIAGNOSTIC_POP`, as it is more descriptive of what it actually does.
For the moment it's not a problem, but if in the future other similar macros `AMREX_FOO_WARNING_PUSH/POP` are defined, the following code would not work as expected:
```
AMREX_COUNTER_WARNING_PUSH
// stuff...
AMREX_FOO_WARNING_PUSH
// stuff...
AMREX_COUNTER_WARNING_POP
// stuff...
AMREX_FOO_WARNING_POP
```

ax3l: I see what you mean, the macro suggests a false sense of namespacing.

How about we define only:
```C++
#define AMREX_WARNING_PUSH_COUNTER ...
#define AMREX_WARNING_PUSH_FOO ...
#define AMREX_WARNING_POP ...
```
to ensure the asymmetry is clearer?

ax3l: Changed accordingly & comment added :)

lucafedeli88 (approved):
Perfect! Thanks, @ax3l

PR #5657, opened 2026-08-31
MLMG: sync duplicated nodal DOFs in MLNodeLinOp::applyBC

WeiqunZhang:
MLEBNodeFDLaplacian::Fapply picks its kernel per box: cut boxes use mlebndfdlap_adotx_eb, regular boxes mlebndfdlap_adotx.  Same operator, different floating-point expression, so a node shared by a cut box and a regular box ended up with two copies differing by one ulp.  Nothing reconciled them, and the bottom BiCGStab amplified the difference by more than an order of magnitude per iteration until it saturated at the bottom tolerance and the solve stopped converging.  An EB nodal solve that converged with a single box failed once the domain was split into three or more.

applyBC now calls FillBoundaryAndSync instead of FillBoundary, so every operator application starts from data whose duplicated nodes agree.  This fixes all the nodal operators at once and is robust to any other source of asymmetry (FMA contraction, vectorization, GPU).

That makes the trailing nodalSync in the smoothers redundant, so drop it. The one in MLNodeABecLaplacian's GPU sweep loop stays, since no applyBC separates those sweeps.  Iteration counts are unchanged, but nodal results are no longer bitwise identical across box decompositions and rank counts.

Also document that the mask is a weight rather than a boolean.

### DIFF
--- a/Src/Base/AMReX_FabArrayUtility.H
+++ b/Src/Base/AMReX_FabArrayUtility.H
@@ -2000,9 +2000,16 @@ Dot (FabArray<FAB> const& x, int xcomp, int ncomp, IntVect const& nghost, bool l
 }
 
 /**
- * \brief Compute dot product of two FabArrays in region that mask is true
+ * \brief Compute weighted dot product of two FabArrays
  *
- * \param mask   mask
+ * The mask is a weight, not a boolean.  The result is the sum of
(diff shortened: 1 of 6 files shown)


n01r (approved):
Confirmed, this PR fixes https://github.com/BLAST-WarpX/warpx/issues/7210! Thanks a lot!

atmyers (approved):

PR #5658, opened 2026-09-01
Replace deprecated `#pragma omp master` with `#pragma omp masked`

lucafedeli88:
## Summary

This PR replaces occurrences of  ` #pragma omp master` with `#pragma omp masked` , as the former is deprecated is deprecated in OpenMP 5.1 . `masked` is more general than `master` (it's possible to specify a `filter` argument), but without additional arguments the two constructs behave in the same way:  https://www.openmp.org/spec-html/5.1/openmpse16.html .

## Additional background

Today I compiled WarpX with `g++ 16.2` and I got several warnings about the deprecation of ` #pragma omp master` . I've checked that with the replacement introduced by this PR in AMReX the warnings disappear and WarpX compiles successfully. I haven't performed other tests.

## Checklist

The proposed changes:
- [X] fix a bug or incorrect behavior in AMReX [[**it was not really a bug, just a deprecated construct.**]]

### DIFF
--- a/Src/Base/AMReX_BLProfiler.cpp
+++ b/Src/Base/AMReX_BLProfiler.cpp
@@ -306,7 +306,7 @@ void BLProfiler::PStop() {
 
 void BLProfiler::start() {
 #ifdef AMREX_USE_OMP
-#pragma omp master
+#pragma omp masked
 #endif
 {
   bltelapsed = 0.0;
@@ -341,7 +341,7 @@ void BLProfiler::start() {
 
 void BLProfiler::stop() {
 #ifdef AMREX_USE_OMP
-#pragma omp master
+#pragma omp masked
 #endif
 {
   double tDiff(amrex::second() - bltstart);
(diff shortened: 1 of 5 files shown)


WeiqunZhang:
We still need to support GCC 11, which does not appear to support `omp masked`.

https://godbolt.org/z/MYaqa9hPe

lucafedeli88:
> We still need to support GCC 11, which does not appear to support `omp masked`.
> 
> https://godbolt.org/z/MYaqa9hPe

ah, because GCC11 only partially supports OpenMP 5.0 ! Sorry, I thought `#pragma omp masked` was sufficiently old .
Then this is not a viable solution.

Maybe adding `-Wno-deprecated-openmp` as an option could be considered to silence this warning  for newer compilers (until the support for GCC11 is dropped).

WeiqunZhang:
Yes, I agree that `-Wno-deprecated-openmp` seems to be a better solution for now.

lucafedeli88:
> Yes, I agree that `-Wno-deprecated-openmp` seems to be a better solution for now.

ok. I will open a new PR and close this one, then.

PR #5661, opened 2026-09-02
BLProfiler: give each BL_PROFILE_REGION a unique variable name

WeiqunZhang:
Both definitions of BL_PROFILE_REGION take a parameter named rname but pasted the token vname, which is not a parameter, so every expansion produced the same identifier: bl_profile_region_vname for the full profiler and tiny_profile_region_vname for the tiny one. Two regions in the same scope therefore failed to compile with a redefinition error. It looks like a copy-paste leftover from BL_PROFILE_VAR(fname, vname).

Pasting rname would not work either, since the argument is a string literal, so use `__COUNTER__` through BL_PROFILE_PASTE the way BL_PROFILE already does.


WeiqunZhang:
Wait for #5653.

asalmgren (approved):

PR #5663, opened 2026-09-02
Fix ParmParse Leaks Tables Between Init/Finalize Cycles

ax3l:
## Summary

When used in pyAMReX bindings, `amrex::ParmParse` leaks state between sequential simulations in the same process. This fixes it.

## Additional background

Seen in ImpactX with a simulation that turns a non-default option on, like CSR modeling, followed by a simulation that relies on the default (off).

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX.cpp
+++ b/Src/Base/AMReX.cpp
@@ -460,10 +460,6 @@ amrex::Initialize (int& argc, char**& argv, bool build_parm_parse,
         {
             // No command line to parse. argc == 0 is the library/embedded entry
             // point (e.g. language bindings), argc == 1 the executable name only.
-            // Both must still initialize ParmParse, because that is what registers
-            // ParmParse::Finalize and thereby clears the parameter table in
-            // amrex::Finalize(). Skipping it would leak every runtime-added
-            // parameter into the next Initialize/Finalize cycle.
             ParmParse::Initialize(0,nullptr,nullptr);
         }
         else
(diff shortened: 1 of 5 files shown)


Src/Base/AMReX.cpp
```diff
@@ -456,11 +456,17 @@ amrex::Initialize (int& argc, char**& argv, bool build_parm_parse,
...
     {
-        if (argc == 1)
+        if (argc <= 1)
         {
+            // No command line to parse. argc == 0 is the library/embedded entry
+            // point (e.g. language bindings), argc == 1 the executable name only.
+            // Both must still initialize ParmParse, because that is what registers
+            // ParmParse::Finalize and thereby clears the parameter table in
+            // amrex::Finalize(). Skipping it would leak every runtime-added
+            // parameter into the next Initialize/Finalize cycle.
```
ax3l: Optional if less verbose desired:
```suggestion
```

Src/Base/AMReX_ParmParse.cpp
```diff
@@ -1770,6 +1770,13 @@ ParmParse::Finalize ()
     g_parser_recursive_symbols.clear();
     g_parser_recursive_symbols.resize(1);
 
+    // Reset the remaining file-scope state, so that the next Initialize starts
+    // from a clean slate. Applications that call SetParserPrefix do so after
+    // Initialize (e.g. from the func_parm_parse hook), so they re-establish
+    // their prefix on every cycle.
```
ax3l: Optional, if less verbose desired:
```suggestion
```

WeiqunZhang (approved):

PR #5666, opened 2026-09-02
Fix MLEBNodeFDLap for the case where STL triangle vertex exactly lands on nodal point

atmyers:
In that case, both "h" and "scale" are 0.0. The current code divides by h first, then multiplies by scale, resulting in a floating point exception, even though the limit is finite.

This also adds new tests for NodalPoissonEB, using both implicit and STL geometries. The STL one fails without this fix.

## Checklist

The proposed changes:


WeiqunZhang:
Could you add mlmg.getGradSolution() to the test and also update compGrad?

WeiqunZhang (approved):

PR #5670, opened 2026-09-03
Add asynchronous GPU TinyProfiler markers

WeiqunZhang:
Add opt-in async timer and region macros with marked output and runtime filtering. Include coverage for CPU, MPI, disabled, legacy, and CUDA builds.


WeiqunZhang:
See also #5669 for an alternative approach.

asalmgren (approved):

PR #5831, opened 2026-09-06
Fix VisMF reads that silently miss data

### LINKED ISSUE #5711: VisMF::Read returns success without covering the on-disk data: files beyond nProcs skipped, iMultiFab layout unchecked
**Severity:** high/medium · **Category:** correctness · **Subsystem:** Base-io (blast radius 3/5)

**Locations:** `Src/Base/AMReX_VisMF.cpp:1675`, `Src/Base/AMReX_VisMF.H:602`

Two VisMF read entry points can finish "successfully" while the FabArray does not actually hold what is on disk: the synchronous file-order path drops whole files when there are more data files than reader ranks, and the iMultiFab Read template never validates ncomp/ngrow (and validates the BoxArray only in debug builds) before reading straight into the caller's fabs.

## The defect

**`Src/Base/AMReX_VisMF.cpp:1675`** — In the vismf.usesynchronousreads file-order path, NItemsPerBin(nProcs, nRanksPerFile) gives zero reader ranks to every file beyond the first nProcs; those files' boxes are never assigned (nBoxesPerRank empty), never read, and mf.ParallelCopy(fafabFileOrder) silently leaves them untouched. _(also reported as F145)_

**`Src/Base/AMReX_VisMF.H:602`** — The iMultiFab Read template only checks BoxArray match with a debug-only AMREX_ASSERT and never checks nComp/nGrow; in release builds a pre-defined FabArray whose BoxArray, ncomp or ngrow differ from the on-disk header is silently read into, and IArrayBox::readFrom resizes each local fab, leaving the FabArray metadata inconsistent with its fabs.

## Why it matters

*F012:* MPI build. Write with vismf.headerversion=2 (any NoFabHeader version) on 8 ranks (-> 8 mf_D_ files); read the same MultiFab on 2 ranks with vismf.usesynchronousreads=1 and nGrow=0. Reproduced: VisMF::Read returns without error, prints 'not inFileOrder', and 12 of 16 boxes keep their pre-read contents (max|read-orig| = 27472 vs 0 for the default path or for 8 reader ranks).

*F143:* Release build (no AMREX_DEBUG), MPI on or off. iMultiFab imf(ba_new, dm, 1, 0) defined after regridding; amrex::Read(imf, chkdir+"/mask") where mask was written on ba_old with ngrow=1 or 2 components. read_fab -> IArrayBox::readFrom (AMReX_IArrayBox.cpp) does this->resize(tmp_box,tmp_ncomp); imf.nGrowVect()/nComp()/fabbox(i) no longer match imf[i]; later FillBoundary/ParallelCopy/setVal operate on the wrong region; no error. VisMF::Read for MultiFab instead uses AMREX_ALWAYS_ASSERT(match && ncomp==) and handles ngrow via a temporary fab.

## How to reach it

- *F012:* config: AMReX_MPI=ON (path is inside #ifdef BL_USE_MPI), any SPACEDIM/precision; MPI builds are in clang.yml, hypre.yml, petsc.yml, apps.yml. Needs runtime vismf.usesynchronousreads=1 and vismf.headerversion=2/3/4 (both non-default).
- *F012:* API path: ParmParse: vismf.headerversion=2, vismf.usesynchronousreads=1 On 8 ranks: VisMF::Write(mf,"mf") // 8 mf_D_ files On 2 ranks: MultiFab mf2(ba,dm,nc,0); VisMF::Read(mf2,"mf") -> files 3..8 get zero reader ranks; their FABs keep prior contents, no error.
- *F012:* test coverage: none; no Tests/ program sets vismf.usesynchronousreads. Src callers: Amr::restart / AmrLevel::restart via VisMF::Read (StateData) hit it whenever an app uses those two vismf keys.
- *F143:* config: Release build (no AMReX_DEBUG/AMReX_ASSERTIONS), AMReX_SPACEDIM=3, MPI on or off; standard CI release jobs (gcc.yml/clang.yml) build this template but no test instantiates it.
- *F143:* API path: amrex::Write(imf_old, "chk/mask") with iMultiFab(ba_old, dm, 2, 1); later iMultiFab imf(ba_new, dm, 1, 0); amrex::Read(imf, "chk/mask"); -> fabs resized to on-disk boxes/ncomp, metadata unchanged; a subsequent imf.setVal/FillBoundary on comp>=on-disk ncomp or on ghost cells absent on disk writes out of bounds. Public template; no in-tree caller.
- *F143:* test coverage: none — no Tests/ program calls amrex::Read/Write(FabArray<IArrayBox>).

## Suggested fix

**F012 (file-order path, `AMReX_VisMF.cpp` ~1674-1703).** The bin assignment assumes `nFiles <= nProcs`, but `NFilesIter` writes `min(NProcs, nOutFiles)` files, so restarting on fewer ranks than the writer used produces exactly `nFiles > nProcs`. The diff keeps the existing scheme and makes it total: every file gets at least one reader (`std::max(n, 1)` after `NItemsPerBin`) and `currentRank` wraps with `% nProcs` instead of clamping at `nProcs-1`. When `nFiles <= nProcs` the bins already sum to `nProcs`, so both edits are no-ops on the path that works today; extra files become single-rank `NFilesIter` chains, which need no cross-rank messaging, so one rank reading several files is safe. **Maintainer decision:** the alternative is to bail to the non-synchronous branch when `FileReadChains.size() > nProcs`; that is simpler but loses the file-order optimisation for the restart-on-fewer-ranks case. Either way a hard check that `indexFileOrder == nBoxes` after the assignment loop would turn any future gap into an abort rather than a silent skip. **F143 (`AMReX_VisMF.H` ~599-603).** Mirror the MultiFab twin `VisMF::Read` (cpp ~1636), which uses `AMREX_ALWAYS_ASSERT` on BoxArray match and ncomp. The iMultiFab path needs the ngrow check as well: `detail::read_fab` calls `IArrayBox::readFrom` directly on `fa[i]`, which resizes the fab to the on-disk box, and there is no temporary-fab copy (as in `VisMF::readFAB`) to absorb an ngrow difference. The precondition is already documented in the Doxygen block, so this only converts silent metadata corruption into an abort in release builds. Landing order does not matter; the two diffs touch different files and are independent.

For `Src/Base/AMReX_VisMF.cpp:1675` (F012):

```diff
--- a/Src/Base/AMReX_VisMF.cpp
+++ b/Src/Base/AMReX_VisMF.cpp
@@ -1673,6 +1673,9 @@
 
     Vector<int> nRanksPerFile(FileReadChains.size());
     NItemsPerBin(nProcs, nRanksPerFile);
+    for(int & nrpf : nRanksPerFile) {  // ---- every file needs a reader
+      nrpf = std::max(nrpf, 1);
+    }
     int currentFileIndex(0);
 
     for(frcIter = FileReadChains.begin(); frcIter != FileReadChains.end(); ++frcIter) {
@@ -1699,7 +1702,7 @@
           ++indexFileOrder;
         }
         ++currentRank;
-        currentRank = std::min(currentRank, nProcs - 1);
+        currentRank = currentRank % nProcs;
       }
       ++currentFileIndex;
     }
```

<sub>Gives every file at least one reader and wraps currentRank so files beyond nProcs go to rank f % nProcs instead of being dropped. When nFiles <= nProcs the bins sum to nProcs, so both changes are no-ops there. Extra files form single-rank NFilesIter chains (no inter-rank messaging), so a rank reading several files is safe. Maintainer may prefer falling back to the non-synchronous path instead. Covers F145.</sub>

For `Src/Base/AMReX_VisMF.H:602` (F143):

```diff
--- a/Src/Base/AMReX_VisMF.H
+++ b/Src/Base/AMReX_VisMF.H
@@ -599,7 +599,9 @@ Read (FabArray<FAB>& fa, const std::string& name)
     if (fa.empty()) {
         fa.define(ba, DistributionMapping{ba}, ncomp, ngrow);
     } else {
-        AMREX_ASSERT(amrex::match(ba, fa.boxArray()));
+        AMREX_ALWAYS_ASSERT(amrex::match(ba, fa.boxArray()) &&
+                            ncomp == fa.nComp() &&
+                            ngrow == fa.nGrowVect());
     }
 
 #ifdef AMREX_USE_MPI
```

<sub>Mirrors the MultiFab twin VisMF::Read (AMReX_VisMF.cpp:1636-1637), which uses AMREX_ALWAYS_ASSERT on BoxArray match and ncomp. Adds the ngrow check too because detail::read_fab reads straight into fa[i] via IArrayBox::readFrom, which resizes the fab to the on-disk box; unlike the MultiFab path there is no temporary-fab copy to absorb an ngrow difference. Turns silent metadata corruption into an abort in release builds; the API precondition is already documented in the Doxygen block above.</sub>

## Verification evidence

<details><summary><code>F012</code> — confirmed (one verifier lens); duplicates folded in: F145</summary>

Lens 1 (refutation attempt): Src/Base/AMReX_VisMF.cpp:1674-1675 `Vector<int> nRanksPerFile(FileReadChains.size()); NItemsPerBin(nProcs, nRanksPerFile);` with NItemsPerBin (lines 53-65) giving countForAll = nProcs/nFiles = 0 and only the first nProcs bins +1. For files with 0 ranks `Vector<int> nBoxesPerRank(0)` -> NItemsPerBin returns early -> inner loop never runs, so baFileOrder.set/ranksFileOrder/readFileRanks are never touched for those boxes; nothing reads them and `mf.ParallelCopy(fafabFileOrder)` (line 1886) copies only the empty default boxes. NFilesIter::ActualNFiles clamps writer file count to min(NProcs, nOutFiles), so nFiles > reader NProcs is exactly the restart-on-fewer-ranks case. Unchanged since 3c0d2fcde3; no dev-branch fix, no open issue/PR.

</details>

<details><summary><code>F143</code> — confirmed (one verifier lens)</summary>

Lens 1 (refutation attempt): Src/Base/AMReX_VisMF.H:599-603: `if (fa.empty()) { fa.define(...); } else { AMREX_ASSERT(amrex::match(ba, fa.boxArray())); }` — no check of ncomp or ngrow against the header at all, and the BoxArray check is debug-only (unchanged since 458679dbff 'iMultiFab I/O (#2495)'). detail::read_fab (H:535-548) calls IArrayBox::readFrom, which at Src/Base/AMReX_IArrayBox.cpp:133-135 does `if (this->box() != tmp_box || this->nComp() != tmp_ncomp) this->resize(tmp_box, tmp_ncomp);`, so each local fab silently takes the on-disk box/ncomp while fa.nComp()/nGrowVect()/fabbox() keep the caller's values. The MultiFab twin VisMF::Read (cpp:1636-1637) uses AMREX_ALWAYS_ASSERT(match && ncomp==) and handles ngrow via a temporary (cpp:1511-1514). No fix on development; not filed. Severity lowered: requires caller-side metadata mismatch and the BoxArray precondition is documented.

</details>

---

<sub>Based on commit `248fcfb7a5`, the tree the audit verified against. From an automated audit of AMReX `Src/`. Audit finding ids: F012, F143, F145. Reviewer unit(s): Base/IO-VisMF-2, Base/IO-VisMF-3. Nothing here was compiled or run except where the evidence says so — the failure scenarios are code reasoning, so the reaching configuration above is the cheapest way to confirm or refute it.</sub>

WeiqunZhang:
In the synchronous file-order path, NItemsPerBin gives no reader to files beyond the first nProcs, so restarting on fewer ranks than the writer used left those files' boxes untouched and Read returned without error. Give every file a reader and wrap the rank counter, and assert that every box is covered.

The FabArray Read template checked the BoxArray only in debug builds and never checked ncomp or ngrow, but it reads straight into the caller's fabs, which resize. Check all three unconditionally.

Fixes #5711.

### DIFF
--- a/Src/Base/AMReX_VisMF.H
+++ b/Src/Base/AMReX_VisMF.H
@@ -599,7 +599,11 @@ Read (FabArray<FAB>& fa, const std::string& name)
     if (fa.empty()) {
         fa.define(ba, DistributionMapping{ba}, ncomp, ngrow);
     } else {
-        AMREX_ASSERT(amrex::match(ba, fa.boxArray()));
+        // read_fab reads straight into fa[i], which resizes the fab, so unlike
+        // VisMF::Read there is nothing here to absorb a mismatch.
+        AMREX_ALWAYS_ASSERT(amrex::match(ba, fa.boxArray()) &&
+                            ncomp == fa.nComp() &&
+                            ngrow == fa.nGrowVect());
     }
 
 #ifdef AMREX_USE_MPI
--- a/Src/Base/AMReX_VisMF.cpp
+++ b/Src/Base/AMReX_VisMF.cpp
@@ -1673,6 +1673,11 @@ VisMF::Read (FabArray<FArrayBox> &mf,
 
     Vector<int> nRanksPerFile(FileReadChains.size());
     NItemsPerBin(nProcs, nRanksPerFile);
+    // There may be more files than ranks (e.g., restarting on fewer ranks than
+    // the run that wrote the data used).  Every file still needs a reader.
+    for(int & nrpf : nRanksPerFile) {
+      nrpf = std::max(nrpf, 1);
+    }
     int currentFileIndex(0);
 
     for(frcIter = FileReadChains.begin(); frcIter != FileReadChains.end(); ++frcIter) {
@@ -1699,10 +1704,13 @@ VisMF::Read (FabArray<FArrayBox> &mf,
           ++indexFileOrder;
         }
         ++currentRank;
-        currentRank = std::min(currentRank, nProcs - 1);
+        currentRank = currentRank % nProcs;
       }
       ++currentFileIndex;
     }
+
+    AMREX_ALWAYS_ASSERT_WITH_MESSAGE(indexFileOrder == nBoxes,
+                                     "VisMF::Read: not all boxes are assigned a reader");
 
     DistributionMapping dmFileOrder(std::move(ranksFileOrder));
 


WeiqunZhang:
# PR #5831: Fix VisMF reads that silently miss data

PR: https://github.com/AMReX-Codes/amrex/pull/5831
Reviewed commit: `e0d9d0261f80fea3dda67b629a29a3edddfcf4fe`
Reviewed at: 2026-09-07 20:37 UTC
Status: Complete

## Findings

No actionable findings. Every file now gets at least one reader, with rank assignment wrapping when files outnumber ranks. The coverage assertion detects incomplete scheduling. The generic FabArray reader checks the full allocation contract before reading into existing FABs.

## Validation

Inspected PR description, conversation, inline comments, and the complete diff with `git diff ae079342654b60f95effddd4377cd9ffaf573f5f e0d9d0261f80fea3dda67b629a29a3edddfcf4fe`. Traced file-chain construction, per-file item distribution, reader assignment and the file-order temporary path at the reviewed head using git show. No local multi-file restart or malformed-destination runtime test was run. Restarts with fewer ranks than files and release-build contract failures remain local coverage limits.

Reviewed by gpt-6-astra.

asalmgren (approved):

PR #5836, opened 2026-09-06
Simplify AMREX_DEVICE_PRINTF for SYCL

WeiqunZhang:
The compiler moves a printf format string to the constant address space itself, so the local opencl_constant copy is no longer needed, and the extension calls std::printf on the host, so the __SYCL_DEVICE_ONLY__ branch is not needed either. Intel's own tests stopped using the attribute by default some time ago.

The macro now takes the same arguments on every backend. The old form needed at least one argument after the format, and could not take a format string that was not a literal, which is what AMReX_Array4.H does for an out-of-bounds message.

### DIFF
--- a/Src/Base/AMReX_GpuPrint.H
+++ b/Src/Base/AMReX_GpuPrint.H
@@ -9,21 +9,14 @@
 #endif
 
 #if defined(AMREX_USE_SYCL)
-#if defined(__SYCL_DEVICE_ONLY__)
-#  define AMREX_DEVICE_PRINTF(format,...) { \
-      static const __attribute__((opencl_constant)) char amrex_i_format[] = format ; \
-      sycl::ext::oneapi::experimental::printf(amrex_i_format, __VA_ARGS__); }
-#else
-#  define AMREX_DEVICE_PRINTF(format,...) { \
-      std::printf(format, __VA_ARGS__); }
-#endif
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
# PR #5836: Simplify AMREX_DEVICE_PRINTF for SYCL

PR: https://github.com/AMReX-Codes/amrex/pull/5836
Reviewed commit: `2296544450bbdcc37a203837dfba8b8f350f6a36`
Reviewed at: 2026-09-07 20:40 UTC
Status: Complete

## Findings

No actionable findings from source inspection. The macro now forwards the entire argument list consistently, including format-only calls and Array4's computed format string. CUDA/HIP expansions are unchanged.

## Validation

Inspected PR description, conversation, inline comments, and the complete diff with `git diff ae079342654b60f95effddd4377cd9ffaf573f5f 2296544450bbdcc37a203837dfba8b8f350f6a36`. Inspected GpuPrint.H and Array4 call sites at the reviewed head. No local SYCL compiler was found through command -v icpx or the standard oneAPI compiler path, so host/device SYCL compilation and compatibility with older supported toolchains were not verified locally. The live PR checks were successful; this does not establish every compiler-version combination.

Reviewed by gpt-6-astra.

asalmgren (approved):

PR #5850, opened 2026-09-07
Fix data race in BaseFab::lockAdd for mixed destbox shapes

### LINKED ISSUE #5736: BaseFab::lockAdd picks lock planes from destbox shape, so mixed flat/thick calls into one fab race
**Severity:** medium · **Category:** gpu-parallel-mpi · **Subsystem:** Base-core (blast radius 5/5)

**Location:** `Src/Base/AMReX_BaseFab.H:3067`

## The defect

**What is wrong** — lockAdd chooses its locking plane dimension from the destbox shape (len.z==1 -> lock y-planes, else lock z-planes), so two concurrent lockAdd calls into the same fab with overlapping destboxes of different z-thickness lock different plane sets and race on the same cells, losing '+=' updates.

## Why it matters

3D CPU build with OpenMP (AMREX_USE_OMP, AMREX_SPACEDIM=3): thread A calls fab.lockAdd(srcA, bA, dA={0:15,0:15,5:5},...) (flat, locks y-plane j) while thread B calls fab.lockAdd(srcB, bB, dB={0:15,0:15,0:7},...) (thick, locks z-plane k). A holds OpenMP::get_lock(j), B holds get_lock(k) — different locks — and both execute unsynchronized 'pdst[ii] += psrc[ii]' on cells (i,j,5): silent lost updates, unlike the atomicAdd it documents itself as replacing.

## How to reach it

- config: AMReX_SPACEDIM=3, AMReX_OMP=ON, AMReX_GPU_BACKEND=NONE (any precision). Built by .github/workflows/gcc.yml job 'GNU@14 OMP SIMD [tests]' (-DAMReX_OMP=ON, default 3D) and windows.yml OMP job.
- API path: amrex::Initialize(argc,argv); FArrayBox dst(Box({0,0,0},{15,15,7}),1); dst.setVal(0.); #pragma omp parallel num_threads(2) { FArrayBox src(dst.box(),1); src.setVal(1.); if (omp_get_thread_num()==0) dst.lockAdd<RunOn::Host>(src, Box({0,0,5},{15,15,5}), Box({0,0,5},{15,15,5}),0,0,1); else dst.lockAdd<RunOn::Host>(src, dst.box(), dst.box(),0,0,1); } Public API; NO in-tree Src/ caller. Downstream WarpX (tiles grown by ng>=1, always thick) and HiPACE++ (fab itself flat) never mix shapes, so no known user hits it.
- test coverage: Tests/OpenMP/atomicAdd/main.cpp calls lockAdd with uniform growntilebox tiles (never mixed z-thickness) and prints timings only; it does NOT assert on values.

## Suggested fix

Make the locking dimension a property of the destination fab rather than of each call. Replace the `len.z == 1` test (where `len` is the destbox length) with a test on `amrex::length(this->domain).z`, keeping `nplanes`/`plo` derived from the destbox as now. Every `lockAdd` into a given fab then serializes on the same lock set, which is the invariant the OpenMP-lock scheme needs and the one `atomicAdd` provides for free. This preserves the intent of #3700: a fab that is itself one cell thick in z (HiPACE++ slices, and every 2D build, where the fab box always has z-length 1) still spreads work over y-plane locks. The only behavioral change is a flat destbox into a thick fab, which now takes its single z-plane lock instead of y-plane locks: correct, with somewhat more contention for that unusual pattern. WarpX tiles are always thick, so it is unaffected. Maintainer decisions: confirm HiPACE++ allocates its slice fabs with a flat box (not a thick fab written one plane at a time), otherwise this trades the race for a performance regression there; and add a sentence to the `lockAdd` Doxygen note stating that the lock set is chosen per fab so concurrent calls of any shape are safe. Single-site change, lands as one commit; `Tests/OpenMP/atomicAdd` could gain a mixed-shape check comparing against `atomicAdd`.

```diff
--- a/Src/Base/AMReX_BaseFab.H
+++ b/Src/Base/AMReX_BaseFab.H
@@ -3064,7 +3064,7 @@
         int planedim;
         int nplanes;
         int plo;
-        if (len.z == 1) {
+        if (amrex::length(this->domain).z == 1) {
             planedim = 1;
             nplanes = len.y;
             plo = dlo.y;
```

<sub>Chooses the locking plane from the destination fab's box rather than the per-call destbox, so all concurrent lockAdd calls into one fab lock along the same dimension. Flat fabs (HiPACE++ slices, and all 2D builds) still lock y-planes; a flat destbox into a thick fab now locks its single z-plane (correct, slightly more contention). Maintainer should confirm HiPACE++'s flat fabs are flat in the fab box, not only in destbox.</sub>

## Verification evidence

<details><summary><code>F102</code> — confirmed (one verifier lens)</summary>

Lens 1 (refutation attempt): Src/Base/AMReX_BaseFab.H:3067-3075 `if (len.z == 1) { planedim = 1; nplanes = len.y; plo = dlo.y; } else { planedim = 2; ...}` then `auto* lock = OpenMP::get_lock(m);` (3088) — lock set chosen per-call from destbox shape. A flat destbox holds get_lock(j), a thick one get_lock(k); OpenMP::get_lock (AMReX_OpenMP.cpp:236-241) returns distinct omp_lock_t for j!=k mod 128, so both threads run `pdst[ii] += psrc[ii]` (3105) on (i,j,5) unsynchronized. Doc (line 924-926) promises 'Atomically add', no shape restriction or assert. Deliberately introduced by 0c59bad946 (#3700) for HiPACE++ flat fabs; PR text addresses only single-lock contention, not mixed-shape concurrency. Not in KNOWN_STATE; no development commits touch the file.

</details>

---

<sub>Based on commit `248fcfb7a5`, the tree the audit verified against. From an automated audit of AMReX `Src/`. Audit finding id: F102. Reviewer unit(s): Base/BaseFab-2. Nothing here was compiled or run except where the evidence says so — the failure scenarios are code reasoning, so the reaching configuration above is the cheapest way to confirm or refute it.</sub>

WeiqunZhang:
lockAdd picked its OpenMP lock planes from the destbox shape, so a flat and a thick destbox into the same fab locked different plane sets and raced. Choose the plane dimension from the fab's own box instead, so all concurrent calls into one fab serialize on the same locks.

Fixes #5736.

### DIFF
--- a/Src/Base/AMReX_BaseFab.H
+++ b/Src/Base/AMReX_BaseFab.H
@@ -3061,10 +3061,12 @@ BaseFab<T>::lockAdd (const BaseFab<T>& src, const Box& srcbox, const Box& destbo
         auto const& slo = amrex::lbound(srcbox);
         Dim3 const offset{.x = slo.x-dlo.x, .y = slo.y-dlo.y, .z = slo.z-dlo.z};
 
+        // The lock dimension must be a property of the destination fab, not of
+        // destbox, so that concurrent calls into the same fab use the same locks.
         int planedim;
         int nplanes;
         int plo;
-        if (len.z == 1) {
+        if (amrex::length(this->domain).z == 1) {
             planedim = 1;
             nplanes = len.y;
             plo = dlo.y;


WeiqunZhang:
# PR #5850: Fix data race in BaseFab::lockAdd for mixed destbox shapes

PR: https://github.com/AMReX-Codes/amrex/pull/5850
Reviewed commit: `994e0e269d10320b0e572178ed5d5d568247e307`
Reviewed at: 2026-09-07 20:47 UTC
Status: Complete

## Findings

No actionable correctness issues found. Selecting the plane direction from the destination FAB makes every lockAdd call on that FAB choose the same lock for any shared destination cell, even when one destination subbox is flat and the other is thick. Plane bounds still come from the requested destination subbox, so the update does not extend outside the requested region.

## Validation

Inspected PR description, conversation, inline comments, and the complete diff with `git diff ae079342654b60f95effddd4377cd9ffaf573f5f 994e0e269d10320b0e572178ed5d5d568247e307`. Inspected the complete lockAdd loop, plane-mask traversal, source offsets, and OpenMP lock lookup, including 2D (z length always one), single-z-plane 3D FABs, and thick 3D FABs receiving flat subboxes. The GPU fallback and non-OpenMP paths are unchanged. No local threaded build, stress test, or race-detector run was performed for this PR; the finding assessment is based on lock-identity and loop-bound analysis.

Reviewed by gpt-6-astra.

asalmgren (approved):

PR #5856, opened 2026-09-07
Fix bisect losing the root when the sign-test product underflows

### LINKED ISSUE #5758: amrex::bisect loses the root when flo*fmi underflows to zero; bracketing assert also passes
**Severity:** medium · **Category:** numerics · **Subsystem:** Base-io (blast radius 4/5)

**Location:** `Src/Base/AMReX_Algorithm.H:192`

## The defect

**What is wrong** — amrex::bisect decides which half to keep with the product test `fmi*flo < T(0)`; when both function values are small enough that the product underflows to (-)0, the test is false and `lo = mi` is taken even when signs differ, so the bracket loses the root and the routine converges to `hi`.

## Why it matters

Any build; worst with AMREX_USE_FLOAT. Call `amrex::bisect(0.f, 1.f, [](float x){ return 1e-25f*(x-0.3f); })` (or double with scale 1e-200): fhi*flo underflows to -0, `fmi*flo < 0` is false, lo is moved past the root, result is 0.99999976 instead of 0.3 (verified with a verbatim copy of lines 136-202). No assert fires; the same underflow also makes the bracketing assert on line 180 pass.

## How to reach it

- config: Any configuration (header-only template, independent of AMReX_PRECISION; user instantiates with float or double). Default CMake or clang.yml -DAMReX_PRECISION=SINGLE job builds it, but no CI test calls the floating-point overload.
- API path: #include `<AMReX_Algorithm.H>` float r = amrex::bisect(0.f, 1.f, [](float x){ return 1e-25f*(x-0.3f); }); // or double: amrex::bisect(0.0, 1.0, [](double x){ return 1e-200*(x-0.3); }); // r ~ 1.0 instead of 0.3; no assert fires. Public Doxygen-documented function with NO in-tree caller (Src callers at AMReX_TagParallelFor.H:367 and AMReX_OpenBC.cpp:429 use the integer-index overload).
- test coverage: none - no Tests/ program or Src/ caller invokes the floating-point bisect overload; nothing asserts on its return value.

## Suggested fix

Replace both product-sign tests in the floating-point `amrex::bisect` overload with direct sign comparisons: the half-selection on line 192 becomes `(fmi < T(0)) != (flo < T(0)) ? hi = mi : lo = mi;` and the bracketing assert on line 180 becomes `(flo < T(0)) != (fhi < T(0))`. A product of two finite values can underflow to +/-0 while the operands keep their signs, so comparing signs is the only test that faithfully encodes "opposite sign"; it is also cheaper than the multiply. This is a pure equivalence rewrite in the non-underflow case: the `flo == T(0)`, `fhi == T(0)` and `fmi == T(0)` early-outs immediately above already handle exact zeros, so at the two test sites both operands are nonzero and `a*b < 0` is exactly `sign(a) != sign(b)`. The sketch avoids `std::signbit` deliberately because `-0` would then read as negative, and because plain `<` is trivially host/device safe under `AMREX_GPU_HOST_DEVICE`. No API or documentation change; the integer-index overload used by `AMReX_TagParallelFor.H` and `AMReX_OpenBC.cpp` is untouched. Land as a single commit. Maintainer decision: whether to add a small test (e.g. `f(x)=1e-25f*(x-0.3f)` on `[0,1]` in float) since no in-tree test currently exercises the floating-point overload.

```diff
--- a/Src/Base/AMReX_Algorithm.H
+++ b/Src/Base/AMReX_Algorithm.H
@@ -177,7 +177,7 @@
         if (flo == T(0)) { return lo; }
         if (fhi == T(0)) { return hi; }
 
-        AMREX_ASSERT_WITH_MESSAGE(flo * fhi <= T(0),
+        AMREX_ASSERT_WITH_MESSAGE((flo < T(0)) != (fhi < T(0)),
             "Error - calling bisect but lo and hi don't bracket a root.");
 
         T mi = (lo + hi) / T(2);
@@ -189,7 +189,7 @@
             mi = (lo + hi) / T(2);
             fmi = f(mi);
             if (fmi == T(0)) { break; }
-            fmi*flo < T(0) ? hi = mi : lo = mi;
+            ((fmi < T(0)) != (flo < T(0))) ? hi = mi : lo = mi;
             flo = f(lo);
             fhi = f(hi);
             ++n;
```

<sub>Replaces the product-sign tests in amrex::bisect with direct sign comparisons, so underflow of flo*fhi or fmi*flo to +/-0 can no longer misdirect the bracket or let the bracketing assert pass. Exact-zero cases are already handled by the preceding `== T(0)` early-outs, so the sign comparison is equivalent to the intended `< 0` product test whenever no underflow occurs. Host/device safe, no API change.</sub>

## Verification evidence

<details><summary><code>F098</code> — confirmed (one verifier lens)</summary>

Lens 1 (refutation attempt): Src/Base/AMReX_Algorithm.H:192 `fmi*flo < T(0) ? hi = mi : lo = mi;` and :180 `AMREX_ASSERT_WITH_MESSAGE(flo * fhi <= T(0), ...)`. Unchanged since e915f9e20b (#1209); `git log ai_audit..development -- Src/Base/AMReX_Algorithm.H` empty, development blob identical. numpy float32 walk-through of f(x)=1e-25f*(x-0.3f) on [0,1]: flo*fhi = -0.0 (assert passes), product test returns 0.99999976, sign test returns 0.30000007; float64 with 1e-200 scale returns 0.99999999999909 vs 0.3. No KNOWN_STATE issue/PR matches.

</details>

---

<sub>Based on commit `248fcfb7a5`, the tree the audit verified against. From an automated audit of AMReX `Src/`. Audit finding id: F098. Reviewer unit(s): Base/Memory-Containers-Math-1. Nothing here was compiled or run except where the evidence says so — the failure scenarios are code reasoning, so the reaching configuration above is the cheapest way to confirm or refute it.</sub>

WeiqunZhang:
The floating-point amrex::bisect picked the half to keep with fmi*flo < 0.  For tiny function values that product underflows to +/-0, so the test fails and the bracket drops the root.  Compare signs directly instead; exact zeros already return early, so the tests are otherwise equivalent.  The bracketing assert had the same flaw.

Fixes #5758.

### DIFF
--- a/Src/Base/AMReX_Algorithm.H
+++ b/Src/Base/AMReX_Algorithm.H
@@ -177,7 +177,7 @@ namespace amrex
         if (flo == T(0)) { return lo; }
         if (fhi == T(0)) { return hi; }
 
-        AMREX_ASSERT_WITH_MESSAGE(flo * fhi <= T(0),
+        AMREX_ASSERT_WITH_MESSAGE((flo < T(0)) != (fhi < T(0)),
             "Error - calling bisect but lo and hi don't bracket a root.");
 
         T mi = (lo + hi) / T(2);
@@ -189,7 +189,7 @@ namespace amrex
             mi = (lo + hi) / T(2);
             fmi = f(mi);
             if (fmi == T(0)) { break; }
-            fmi*flo < T(0) ? hi = mi : lo = mi;
+            ((fmi < T(0)) != (flo < T(0))) ? hi = mi : lo = mi;
             flo = f(lo);
             fhi = f(hi);
             ++n;


WeiqunZhang:
# PR #5856: Fix bisect losing the root when the sign-test product underflows

PR: https://github.com/AMReX-Codes/amrex/pull/5856
Reviewed commit: `1dffa8bc5298ede772b0e432222d8162908b2428`
Status: Complete

## Findings

No actionable correctness issues found. Direct sign comparisons avoid multiplication underflow in both the bracket assertion and interval selection. Endpoint and midpoint zero handling remains before the relevant sign decision.

## Validation

Read the PR description, conversation, reviews and inline comments, and inspected the complete diff with `git diff ae079342654b60f95effddd4377cd9ffaf573f5f 1dffa8bc5298ede772b0e432222d8162908b2428 --`. Inspected the entire floating-point bisect overload, endpoint handling, iteration limits and bracket updates.

### Focused local probe

Passed roots at 0.3 for function scales 1, 1e-200 and -1e-200. The comparison-base header returns 1 instead of 0.3 for both tiny scales (exit 2). This is a CPU scalar test.

Probe source: `agent-notes/codex-reviews-2/validation/pr-5856.cpp`. The local `agent-notes/codex-reviews-2/validation/README.md` records the common Debug build and extraction procedure. Exact changed headers/translation units were compiled and linked against that common library; these are not whole builds of every PR.

```sh
python3 agent-notes/codex-reviews-2/validation/build-probe.py 5856 head
timeout 30 /tmp/codex-reviews-2/5856/head/probe
python3 agent-notes/codex-reviews-2/validation/build-probe.py 5856 base
timeout 30 /tmp/codex-reviews-2/5856/control/probe
```

Compiler commands actually run are saved in `validation/pr-5856-commands.txt`.

Reviewed by gpt-6-astra.

asalmgren (approved):

PR #5862, opened 2026-09-07
Fix use-after-free when PODVector value argument aliases the vector

### LINKED ISSUE #5761: PODVector::push_back/insert/resize(const T&) read a_value after freeing the old buffer (self-alias UAF)
**Severity:** medium · **Category:** memory-ub · **Subsystem:** Base-io (blast radius 4/5)

**Location:** `Src/Base/AMReX_PODVector.H:635`

## The defect

**What is wrong** — PODVector::push_back(const T&) (and insert/resize taking const T&) reads a_value AFTER AllocateBufferForPush has deallocated the old buffer, so an argument aliasing an element of the same vector (v.push_back(v.front())) is a use-after-free that silently stores garbage; std::vector guarantees this case.

## Why it matters

CPU build (Release, MPI on, OMP off, double), Gpu::HostVector<double> (= PODVector<double,std::allocator>): v.push_back(3.5); fill until size()==capacity(); v.push_back(v.front()). Verified on this tree linked against libamrex_3d.a: printed back()=1.25579e-313 (glibc tcache pointer written into the freed block) instead of 3.5. Same pattern in insert(pos,count,const T&) after memMoveImpl and in resize(n,const T&).

## How to reach it

- config: Any configuration; default CPU CMake (AMReX_SPACEDIM=3, AMReX_PRECISION=DOUBLE, AMReX_MPI=ON, AMReX_OMP=OFF, no GPU backend) suffices. Built by .github/workflows/gcc.yml jobs 'GNU@11 Release [lib]' and 'GNU@13 3D Debug Fortran [tests]'. Also affects HostVector/PinnedVector in GPU builds when arena in-place growth fails.
- API path: amrex::Gpu::HostVector<double> v; // PODVector<double,std::allocator<double>> v.push_back(3.5); while (v.size() < v.capacity()) v.push_back(1.0); v.push_back(v.front()); // AllocateBufferForPush frees old block, then a_value read from freed memory double x = v.back(); // garbage (glibc tcache pointer bits), not 3.5 No in-tree caller self-aliases a PODVector (ErrorList.H:518 does it on amrex::Vector, which is std::vector); public API reachable by applications.
- test coverage: none. Tests/GPU/Vector/main.cpp exercises push_back/insert/assign on all vector flavors but never with a self-referencing argument and asserts nothing on this path.

## Suggested fix

Take a local copy of the by-reference argument before anything can reallocate or shift the buffer, and fill from that copy. In `push_back(const T&)` the copy must precede `AllocateBufferForPush`; in `insert(pos, count, const T&)` it must precede both the `AllocateBufferForInsert` branch and the in-place `memMoveImpl` branch (the move alone can clobber an aliased element even when no reallocation happens); in `resize(n, const T&, GrowthStrategy)` it must precede `resize_without_init_snan`. `insert(pos, const T&)` and `insert(pos, T&&)` delegate to the three-argument overload, so they are covered for free. This is the right fix rather than an assert or a doc caveat because `PODVector` is used as a drop-in for `std::vector` (`Gpu::HostVector`, `PinnedVector`, particle containers), and the standard guarantees `v.push_back(v[i])`, `v.insert(pos, v[i])` and `v.resize(n, v[i])` work. `T` is required to be trivially copyable, so the extra stack copy costs nothing; for device allocators `uninitializedFillNImpl` already captures the value by copy into the fill kernel, so behaviour on GPU vectors is unchanged. One call for the maintainer: `assign(count, const T&)` has the same clear/reserve-then-fill order, but the standard makes a self-referencing `t` a precondition violation for `assign`, so it is defensible to leave it alone. Adding the same one-line copy there is cheap and would make the whole class alias-safe; either way, document the decision. A small case in `Tests/GPU/Vector/main.cpp` doing a push_back of `front()` at full capacity would pin the behaviour. All three sites can land in one commit.

```diff
--- a/Src/Base/AMReX_PODVector.H
+++ b/Src/Base/AMReX_PODVector.H
@@ -505,6 +505,7 @@
         {
             auto* pos = const_cast<iterator>(a_pos);
             if (a_count > 0) {
+                T const value = a_value; // a_value may alias an element of *this
                 if (m_capacity < m_size + a_count)
                 {
                     std::size_t insert_index = std::distance(m_data, pos);
@@ -517,7 +518,7 @@
                                         (Allocator const&)(*this));
                     m_size += a_count;
                 }
-                detail::uninitializedFillNImpl(pos, a_count, a_value,
+                detail::uninitializedFillNImpl(pos, a_count, value,
                                                (Allocator const&)(*this));
             }
             return pos;
@@ -628,11 +629,12 @@
 
         void push_back (const T& a_value)
         {
+            T const value = a_value; // a_value may alias an element of *this
             if (m_size == m_capacity) {
                 auto new_capacity = GetNewCapacityForPush();
                 AllocateBufferForPush(new_capacity);
             }
-            detail::uninitializedFillNImpl(m_data+m_size, 1, a_value,
+            detail::uninitializedFillNImpl(m_data+m_size, 1, value,
                                            (Allocator const&)(*this));
             ++m_size;
         }
@@ -769,12 +771,13 @@
         void resize (size_type a_new_size, const T& a_val,
                      GrowthStrategy strategy = GrowthStrategy::Poisson)
         {
+            T const value = a_val; // a_val may alias an element of *this
             size_type old_size = m_size;
             resize_without_init_snan(a_new_size, strategy);
             if (old_size < a_new_size)
             {
                 detail::uninitializedFillNImpl(m_data + old_size,
-                                               m_size - old_size, a_val,
+                                               m_size - old_size, value,
                                                (Allocator const&)(*this));
             }
         }
```

<sub>Copies the by-reference argument into a local `T const value` before any reallocation (AllocateBufferForPush/AllocateBufferForInsert/resize_without_init_snan) or memMove, then fills from the copy, in push_back(const T&), insert(pos,count,const T&) and resize(n,const T&). T is trivially copyable so the copy is free. insert(pos,const T&) is covered via delegation. Verified with git apply --check on a scratch copy.</sub>

## Verification evidence

<details><summary><code>F130</code> — confirmed (one verifier lens)</summary>

Lens 1 (refutation attempt): Src/Base/AMReX_PODVector.H push_back: `AllocateBufferForPush(new_capacity);` -> UpdateDataPtr does `deallocate(m_data, capacity());` (L899) then L635 `detail::uninitializedFillNImpl(m_data+m_size, 1, a_value, ...)` reads a_value; for std::allocator `allocate_in_place` always returns a new block (L33-34) so the old one is unconditionally freed. Same order in insert (L516/L948 then L520) and resize(n,val) (reserve_doit->UpdateDataPtr then L776). No assert/doc excludes aliasing; `git log -S alias`, `git log -L` on push_back show no deliberate workaround; `git log ai_audit..development -- AMReX_PODVector.H` empty; not in KNOWN_STATE.

</details>

---

<sub>Based on commit `248fcfb7a5`, the tree the audit verified against. From an automated audit of AMReX `Src/`. Audit finding id: F130. Reviewer unit(s): Base/Memory-Containers-Math-3. Nothing here was compiled or run except where the evidence says so — the failure scenarios are code reasoning, so the reaching configuration above is the cheapest way to confirm or refute it.</sub>

WeiqunZhang:
push_back, insert, resize and assign taking a const T& read the value after the buffer has been reallocated or shifted, so v.push_back(v[0]) stored garbage. Copy the value into a local first.

Fixes #5761.

### DIFF
--- a/Src/Base/AMReX_PODVector.H
+++ b/Src/Base/AMReX_PODVector.H
@@ -505,6 +505,8 @@ namespace amrex
         {
             auto* pos = const_cast<iterator>(a_pos);
             if (a_count > 0) {
+                // Copy first because a_value could be a reference to an element of this vector.
+                T const value = a_value;
                 if (m_capacity < m_size + a_count)
                 {
                     std::size_t insert_index = std::distance(m_data, pos);
@@ -517,7 +519,7 @@ namespace amrex
                                         (Allocator const&)(*this));
                     m_size += a_count;
                 }
-                detail::uninitializedFillNImpl(pos, a_count, a_value,
+                detail::uninitializedFillNImpl(pos, a_count, value,
                                                (Allocator const&)(*this));
             }
             return pos;
@@ -582,12 +584,14 @@ namespace amrex
 
         void assign (size_type a_count, const T& a_value)
         {
+            // Copy first because a_value could be a reference to an element of this vector.
+            T const value = a_value;
             if ( a_count > m_capacity ) {
                 clear();
                 reserve(a_count);
             }
             m_size = a_count;
-            detail::uninitializedFillNImpl(m_data, a_count, a_value,
+            detail::uninitializedFillNImpl(m_data, a_count, value,
                                            (Allocator const&)(*this));
         }
 
@@ -628,11 +632,13 @@ namespace amrex
 
         void push_back (const T& a_value)
         {
+            // Copy first because a_value could be a reference to an element of this vector.
+            T const value = a_value;
             if (m_size == m_capacity) {
                 auto new_capacity = GetNewCapacityForPush();
                 AllocateBufferForPush(new_capacity);
             }
-            detail::uninitializedFillNImpl(m_data+m_size, 1, a_value,
+            detail::uninitializedFillNImpl(m_data+m_size, 1, value,
                                            (Allocator const&)(*this));
             ++m_size;
         }
@@ -769,12 +775,14 @@ namespace amrex
         void resize (size_type a_new_size, const T& a_val,
                      GrowthStrategy strategy = GrowthStrategy::Poisson)
         {
+            // Copy first because a_val could be a reference to an element of this vector.
+            T const value = a_val;
             size_type old_size = m_size;
             resize_without_init_snan(a_new_size, strategy);
             if (old_size < a_new_size)
             {
                 detail::uninitializedFillNImpl(m_data + old_size,
-                                               m_size - old_size, a_val,
+                                               m_size - old_size, value,
                                                (Allocator const&)(*this));
             }
         }


WeiqunZhang:
# PR #5862: Fix use-after-free when PODVector value argument aliases the vector

PR: https://github.com/AMReX-Codes/amrex/pull/5862
Reviewed commit: `1da552a4fef4a917fcba01ae201446033d8e4b4c`
Status: Complete

## Findings

No actionable correctness issues found. Saving the scalar value before growth, deallocation or suffix shifting protects aliased arguments in push_back, insert, resize and assign.

## Validation

Read the PR description, conversation, reviews and inline comments, and inspected the complete diff with `git diff ae079342654b60f95effddd4377cd9ffaf573f5f 1da552a4fef4a917fcba01ae201446033d8e4b4c --`. Rechecked the current full diff, insertion wrappers, allocation/update helpers and uninitializedFillNImpl. The earlier report in codex-reviews has the same reviewed SHA; its 384-case probe was copied and rerun for this batch. The earlier ASan/UBSan results are historical evidence, not a new sanitizer run.

### Focused local probe

Passed all 384 copied self-aliasing cases at the reviewed header: integer and two-integer struct values, exact/spare capacity, all original source elements, insert positions and resize/assign sizes. The base-header control fails the first reallocation push_back value check. This new run was CPU-only and not sanitizer-instrumented; GPU/arena allocator behavior was inspected only.

Probe source: `agent-notes/codex-reviews-2/validation/pr-5862.cpp`. The local `agent-notes/codex-reviews-2/validation/README.md` records the common Debug build and extraction procedure. Exact changed headers/translation units were compiled and linked against that common library; these are not whole builds of every PR.

```sh
python3 agent-notes/codex-reviews-2/validation/build-probe.py 5862 head
timeout 30 /tmp/codex-reviews-2/5862/head/probe
python3 agent-notes/codex-reviews-2/validation/build-probe.py 5862 base
timeout 30 /tmp/codex-reviews-2/5862/control/probe
```

Compiler commands actually run are saved in `validation/pr-5862-commands.txt`.

Reviewed by gpt-6-astra.

asalmgren (approved):

PR #5870, opened 2026-09-08
STL: make EB intersection tolerances scale with the geometry

### LINKED ISSUE #5789: STL: lineseg_tri_intersect uses absolute eps on length^3 Plücker terms, misclassifies small-scale STLs
**Severity:** medium · **Category:** numerics · **Subsystem:** EB (blast radius 3/5)

**Location:** `Src/EB/AMReX_EB_triGeomOps_K.H:218`

## The defect

**What is wrong** — lineseg_tri_intersect classifies the ray/triangle configuration with an ABSOLUTE tolerance eps=numeric_limits<Real>::epsilon() applied to Plücker side operators that scale as length^3 (and to areas that scale as length^2), so for small-coordinate geometries every triangle falls into the 'coplanar' branch and is reported as intersected.

## Why it matters

AMREX_USE_FLOAT, eb2.geom_type=stl with an STL whose (scaled) coordinates are O(1e-3) or less, e.g. a 5 mm part in SI units: |S1|,|S2|,|S3|~X^3~1e-9 < FLT_EPSILON for all triangles, the coplanar branch's area test then compares O(1e-6) differences against 1.2e-7 and returns 'intersects' for every bbox-overlapping triangle; num_intersects parity is random, so fill()/fillFab()/getBoxType() produce a salt-and-pepper level set. Double precision fails the same way below ~1e-5..1e-6 (micron-scale parts in SI).

## How to reach it

- config: AMReX_SPACEDIM=3, AMReX_EB=ON, AMReX_PRECISION=SINGLE (built by .github/workflows/clang.yml first job: -DAMReX_EB=ON -DAMReX_PRECISION=SINGLE). Double + micron-scale STL reaches it in every EB=ON job (gcc.yml, cuda.yml, hip.yml).
- API path: ParmParse: eb2.geom_type=stl, eb2.stl_file=part.stl (vertex coords O(1e-3), stl_scale=1), geometry prob_lo/hi ~ 1e-3. EB2::Build(geom, 0, 0) -> EB2::IndexSpaceSTL -> Level_STL::Level_STL -> STLtools::read_stl_file -> prepare() (m_boundry_is_outside parity) -> STLtools::fill / fillFab / getBoxType -> line_tri_intersects -> tri_geom_ops::lineseg_tri_intersect. STLtools::read_stl_file/fill are also public and callable directly.
- test coverage: Tests/EB/MarchingCubes/main.cpp uses cube.stl with vertices at +/-1 and stl_scale=1 (unit scale, no failure) and only prints build time; nothing asserts on the level set or parity. No test exercises small-coordinate STLs.

## Suggested fix

Make every "is this zero" test in the STL intersection path relative to the geometry scale instead of comparing against a bare `numeric_limits<Real>::epsilon()`. The Plücker side operators `S1..S3` and `ls_s1/ls_s2` are products of three coordinate differences (scale^3), the areas in the coplanar branch are scale^2, and the cross product in `STLtools::sign()` (`AMReX_EB_STL_utils.cpp`) is scale^2; a tolerance that is not scaled the same way collapses the classification to "coplanar/intersects" for any part whose coordinates are well below unity, which is exactly the SI-units case. The attached diff computes `L = max |coordinate|` over the five points entering `lineseg_tri_intersect` and uses `eps*L^3` for the side operators and `eps*L^2` (or `eps*tri_area`) for the area comparison; `sign()` gets the same treatment with `eps*L^2`. This keeps behaviour identical to today's unit-scale STLs (which is what `Tests/EB/MarchingCubes` exercises) and makes it scale-invariant. Land the `triGeomOps_K.H` and `STL_utils.cpp` changes together, since `getIntercept` and `fill`/`getBoxType` otherwise disagree on which triangles are hit. Maintainer decisions: (1) whether to use a small multiplier (e.g. `4*eps`) rather than bare `eps` now that the tolerance follows coordinate magnitude, since large-coordinate STLs currently get an effectively tighter test; (2) whether to instead translate `v1` to the origin and normalize by the segment/triangle bbox diagonal before forming Plücker coordinates, which also removes cancellation for far-from-origin parts; (3) adding a small-scale STL (e.g. `cube.stl` with `stl_scale=1e-3` in single precision) to a test that asserts on the level set, since nothing currently checks the parity result.

```diff
--- a/Src/EB/AMReX_EB_triGeomOps_K.H
+++ b/Src/EB/AMReX_EB_triGeomOps_K.H
@@ -1,6 +1,7 @@
 #ifndef AMREX_EB_TRIGEOMOPS_K_H_
 #define AMREX_EB_TRIGEOMOPS_K_H_
 #include <AMReX_Config.H>
+#include <AMReX_Algorithm.H>
 #include <AMReX_Math.H>
 #include <AMReX_REAL.H>
 
@@ -212,10 +213,17 @@
             //we are assuming there are no intersections initially
             int no_intersections=1;
 
-            Real eps = std::numeric_limits<Real>::epsilon();
+            //tolerances follow the coordinate scale: areas ~ scale^2, side_op ~ scale^3
+            Real scale = amrex::max(std::abs(v1[0]),std::abs(v1[1]),std::abs(v1[2]),
+                                    std::abs(v2[0]),std::abs(v2[1]),std::abs(v2[2]),
+                                    std::abs(t1[0]),std::abs(t1[1]),std::abs(t1[2]),
+                                    std::abs(t2[0]),std::abs(t2[1]),std::abs(t2[2]),
+                                    std::abs(t3[0]),std::abs(t3[1]),std::abs(t3[2]));
+            Real eps_area = std::numeric_limits<Real>::epsilon()*scale*scale;
+            Real eps = eps_area*scale;
 
             //coplanar (S1,S2,S3 = 0)
             if(std::abs(S1) < eps && std::abs(S2) < eps && std::abs(S3) < eps)
             {
                 //Print()<<"line segment and triangle are in the same plane\t"<<S1<<"\t"<<S2<<"\t"<<S3<<"\n";
                 tri_area=triangle_area(t1,t2,t3);
@@ -227,7 +235,7 @@
                 area1=(triangle_area(t1,t2,v1)+triangle_area(t2,t3,v1)+triangle_area(t3,t1,v1));
                 area2=(triangle_area(t1,t2,v2)+triangle_area(t2,t3,v2)+triangle_area(t3,t1,v2));
 
-                if( std::abs(area1-tri_area)>eps || std::abs(area2-tri_area)>eps )
+                if( std::abs(area1-tri_area)>eps_area || std::abs(area2-tri_area)>eps_area )
                 {
                     no_intersections = 0;
                 }
--- a/Src/EB/AMReX_EB_STL_utils.cpp
+++ b/Src/EB/AMReX_EB_STL_utils.cpp
@@ -56,7 +56,9 @@
     Real sign (Real x1, Real y1, Real x2, Real y2, Real x3, Real y3)
     {
         Real cp = (x2-x1)*(y3-y2) - (x3-x2)*(y2-y1);
-        if (std::abs(cp) < std::numeric_limits<Real>::epsilon()) {
+        Real scale = amrex::max(std::abs(x1),std::abs(y1),std::abs(x2),
+                                std::abs(y2),std::abs(x3),std::abs(y3));
+        if (std::abs(cp) < std::numeric_limits<Real>::epsilon()*scale*scale) {
             return 0._rt;
         } else {
             return std::copysign(1.0_rt, cp);
```

<sub>Scales the zero tolerances by the largest |coordinate| entering the Plücker/area terms (side_op ~ scale^3, areas ~ scale^2), so behaviour matches today's unit-scale case at any STL size. Same pattern applied to sign() in STL_utils.cpp (cp ~ scale^2), which drives getIntercept. Adds AMReX_Algorithm.H for variadic amrex::max. Maintainer: for large coordinates the tolerance loosens from absolute eps; a multiplier (e.g. 4*eps) may be preferred.</sub>

## Verification evidence

<details><summary><code>F174</code> — confirmed (one verifier lens)</summary>

Lens 1 (refutation attempt): Src/EB/AMReX_EB_triGeomOps_K.H:215-218 `Real eps = std::numeric_limits<Real>::epsilon(); if(std::abs(S1) < eps && std::abs(S2) < eps && std::abs(S3) < eps)` and :230 `std::abs(area1-tri_area)>eps`; side_op terms are (X^2)*(X) = X^3, areas X^2, eps absolute. Unchanged since ad7974de5b (#1592); no ai_audit..development commit; PR #5618 only reorders STL_utils.cpp (ref point/nearest crossing), not the tolerance; KNOWN_STATE has no matching issue (#5389 is large triangles). numpy replica of the exact algorithm on a 960-tri sphere: float32 R=2.5e-3 -> 62686/62686 tests hit coplanar branch, 437/928 cells >2% from surface misclassified (47%); R=0.1 already 10% wrong; R=1 0%. float64 correct at 2.5e-3, 50% wrong at 2.5e-5, 45% at 2.5e-6.

</details>

---

<sub>Based on commit `248fcfb7a5`, the tree the audit verified against. From an automated audit of AMReX `Src/`. Audit finding id: F174. Reviewer unit(s): EB/STL. Nothing here was compiled or run except where the evidence says so — the failure scenarios are code reasoning, so the reaching configuration above is the cheapest way to confirm or refute it.</sub>

WeiqunZhang:
The Plucker side operators scale as length^3 and the areas as length^2,
so an absolute epsilon is meaningless once the geometry is not order
unity. Shift the triangle to the origin and scale the tolerances by the
resulting size, one factor being the triangle's own size so that a long
segment cannot inflate them. Also make sign() relative to its two
products, and re-enable the NodalPoissonEB_STL ctest that this fixes in
single precision.

Fixes #5789.

### DIFF
--- a/Src/EB/AMReX_EB_STL_utils.cpp
+++ b/Src/EB/AMReX_EB_STL_utils.cpp
@@ -55,8 +55,11 @@ namespace {
     AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
     Real sign (Real x1, Real y1, Real x2, Real y2, Real x3, Real y3)
     {
-        Real cp = (x2-x1)*(y3-y2) - (x3-x2)*(y2-y1);
-        if (std::abs(cp) < std::numeric_limits<Real>::epsilon()) {
+        Real a = (x2-x1)*(y3-y2);
+        Real b = (x3-x2)*(y2-y1);
+        Real cp = a - b;
+        // the tolerance must follow the magnitude of the two products
+        if (std::abs(cp) <= std::numeric_limits<Real>::epsilon()*amrex::max(std::abs(a),std::abs(b))) {
             return 0._rt;
         } else {
             return std::copysign(1.0_rt, cp);
--- a/Src/EB/AMReX_EB_triGeomOps_K.H
+++ b/Src/EB/AMReX_EB_triGeomOps_K.H
@@ -222,14 +222,18 @@ namespace amrex::tri_geom_ops
             int no_intersections=1;
 
             //the tolerances must follow the size of the objects, because the
-            //side operators scale as length^3 and the areas as length^2
-            Real len=Real(0.0);
+            //side operators scale as length^3 and the areas as length^2.  one
+            //factor is the triangle's own size, so that a long line segment
+            //does not inflate the tolerance past the triangle's scale.
+            Real len_t=Real(0.0), len_v=Real(0.0);
             for(int i=0;i<3;i++)
             {
-                len=amrex::max(len,std::abs(v1[i]),std::abs(v2[i]),
-                               std::abs(t2[i]),std::abs(t3[i]));
+                len_t=amrex::max(len_t,std::abs(t2[i]),std::abs(t3[i]));
+                len_v=amrex::max(len_v,std::abs(v1[i]),std::abs(v2[i]));
             }
-            Real eps_area = std::numeric_limits<Real>::epsilon()*len*len;
+            Real len=amrex::max(len_t,len_v);
+            constexpr Real eps_r = std::numeric_limits<Real>::epsilon();
+            Real eps_area = eps_r*len*len_t;
             Real eps = eps_area*len;
 
             //coplanar (S1,S2,S3 = 0)
--- a/Src/EB/AMReX_EB_triGeomOps_K.H
+++ b/Src/EB/AMReX_EB_triGeomOps_K.H
@@ -1,6 +1,7 @@
 #ifndef AMREX_EB_TRIGEOMOPS_K_H_
 #define AMREX_EB_TRIGEOMOPS_K_H_
 #include <AMReX_Config.H>
+#include <AMReX_Algorithm.H>
 #include <AMReX_Math.H>
 #include <AMReX_REAL.H>
 
@@ -197,11 +198,19 @@ namespace amrex::tri_geom_ops
             return(all_ok);
         }
         //================================================================================
-        AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE int lineseg_tri_intersect(const Real v1[3], const Real v2[3],
-                const Real t1[3], const Real t2[3], const Real t3[3])
+        AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE int lineseg_tri_intersect(const Real p1[3], const Real p2[3],
+                const Real q1[3], const Real q2[3], const Real q3[3])
         {
             //see plucker coordinates based method
             //https://members.loria.fr/SLazard/ARC-Visi3D/Pant-project/files/Line_Triangle.html
+
+            //the operators below are translation invariant.  putting q1 at the
+            //origin avoids cancellation for geometries far away from the origin.
+            const Real v1[3]={p1[0]-q1[0],p1[1]-q1[1],p1[2]-q1[2]};
+            const Real v2[3]={p2[0]-q1[0],p2[1]-q1[1],p2[2]-q1[2]};
+            const Real t1[3]={Real(0.0),Real(0.0),Real(0.0)};
+            const Real t2[3]={q2[0]-q1[0],q2[1]-q1[1],q2[2]-q1[2]};
+            const Real t3[3]={q3[0]-q1[0],q3[1]-q1[1],q3[2]-q1[2]};
 
             Real S1,S2,S3;
             Real tri_area,area1,area2;
@@ -212,7 +221,16 @@ namespace amrex::tri_geom_ops
             //we are assuming there are no intersections initially
             int no_intersections=1;
 
-            Real eps = std::numeric_limits<Real>::epsilon();
+            //the tolerances must follow the size of the objects, because the
+            //side operators scale as length^3 and the areas as length^2
+            Real len=Real(0.0);
+            for(int i=0;i<3;i++)
+            {
+                len=amrex::max(len,std::abs(v1[i]),std::abs(v2[i]),
+                               std::abs(t2[i]),std::abs(t3[i]));
+            }
+            Real eps_area = std::numeric_limits<Real>::epsilon()*len*len;
+            Real eps = eps_area*len;
 
             //coplanar (S1,S2,S3 = 0)
             if(std::abs(S1) < eps && std::abs(S2) < eps && std::abs(S3) < eps)
@@ -227,7 +245,7 @@ namespace amrex::tri_geom_ops
                 area1=(triangle_area(t1,t2,v1)+triangle_area(t2,t3,v1)+triangle_area(t3,t1,v1));
                 area2=(triangle_area(t1,t2,v2)+triangle_area(t2,t3,v2)+triangle_area(t3,t1,v2));
 
-                if( std::abs(area1-tri_area)>eps || std::abs(area2-tri_area)>eps )
+                if( std::abs(area1-tri_area)>eps_area || std::abs(area2-tri_area)>eps_area )
                 {
                     no_intersections = 0;
                 }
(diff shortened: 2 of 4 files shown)


WeiqunZhang:
# PR #5870: STL: make EB intersection tolerances scale with the geometry

PR: https://github.com/AMReX-Codes/amrex/pull/5870
Reviewed commit: `31a8f1f0ddb56e794b5d298e6c9a85e7d56a79f4`
Status: Complete
Result: Actionable findings

## Findings

### [P2] Scale the tolerance by the contributing edge lengths, not the longest segment coordinate

The new `epsilon * len^3` bound can classify a non-coplanar miss as coplanar when a segment is much longer than the triangle edges. For triangle `(0,0,0), (1,0,0), (0,1,0)` and segment `(.75,.75,-1e9)` to `(.75,.75,1e9)`, the segment passes the STL wrapper's AABB test but misses the triangle (`x+y=1.5`). At this head, `lineseg_tri_intersect` returns `0` (intersection); the base returns `1` (no intersection). The same regression occurs after scaling every coordinate by `1e-6` or `1e6`. In single precision, segment endpoints at z=+/-1e4 already reproduce it.

The overlarge bound sends the mixed-sign side operators into the coplanar branch, whose area test reports an intersection. This can corrupt STL intersection counts for meshes with small facets relative to the sampling segment. Use an error bound based on the actual determinant products/edge lengths or normalize the calculation without losing the segment-to-triangle scale ratio, and retain this miss as a regression case.

Location: [AMReX_EB_triGeomOps_K.H:226-236](https://github.com/AMReX-Codes/amrex/blob/31a8f1f0ddb56e794b5d298e6c9a85e7d56a79f4/Src/EB/AMReX_EB_triGeomOps_K.H#L226-L236).

## Validation

Read the PR description, conversation, reviews and inline comments, and inspected the complete diff with `git diff ae079342654b60f95effddd4377cd9ffaf573f5f 31a8f1f0ddb56e794b5d298e6c9a85e7d56a79f4 --`. Inspected the complete diff, Plucker side operators, area tests, the STL line-triangle wrapper and its AABB rejection, and BVH intersection consumers.

### Focused local probe

The saved probe prints classifications for six cases in each precision. In double precision, all three long-segment misses are incorrectly classified as intersections at the head and correctly rejected by the base. Short-segment behavior confirms the intended small-scale improvement as well. A separate single-precision header-only compile reproduces all three long-segment false positives with extent 1e4, while the base rejects them. The probe prints observations and exits zero; its output, not the exit status, establishes this finding. No complete STL mesh sampling or GPU execution was run.

Probe source: `agent-notes/codex-reviews-2/validation/pr-5870.cpp`. The local `agent-notes/codex-reviews-2/validation/README.md` records the common Debug build and extraction procedure. Exact changed headers/translation units were compiled and linked against that common library; these are not whole builds of every PR.

```sh
python3 agent-notes/codex-reviews-2/validation/build-probe.py 5870 head
timeout 30 /tmp/codex-reviews-2/5870/head/probe
python3 agent-notes/codex-reviews-2/validation/build-probe.py 5870 base
timeout 30 /tmp/codex-reviews-2/5870/control/probe
```

Compiler commands actually run are saved in `validation/pr-5870-commands.txt`.

Reviewed by gpt-6-astra.

WeiqunZhang:
> segment (.75,.75,-1e9) to (.75,.75,1e9)

Not sure we need to worry about that kind of long triangles.

WeiqunZhang:
We disabled single precision test for NodalPoissonEB_STL in #5672 because it failed. With this branch, it passes now. So this PR will also re-enable it.

WeiqunZhang:
codex:

• No actionable findings at 89f6a039f96d. The previous P2 issue is fixed.

  - All 12 double/single-precision probe cases passed; the old head reproduces the failures.
  - Re-enabled 3D single-precision STL CTest passed.
  - CI remains in progress.

asalmgren (approved):

PR #5874, opened 2026-09-08
Assert AmrLevel FillPatcher preconditions

### LINKED ISSUE #5770: AmrLevel m_fillpatcher cache: single-interp assumption in FillPatcherFill and null deref in FillRKPatch after post-step regrid
**Severity:** medium · **Category:** correctness, memory-ub · **Subsystem:** Amr (blast radius 3/5)

**Locations:** `Src/Amr/AMReX_AmrLevel.cpp:2188`, `Src/Amr/AMReX_AmrLevel.cpp:2283`

AmrLevel keeps one cached FillPatcher per state index in m_fillpatcher, and two callers rely on invariants the cache does not enforce: that one interpolater covers every component of the state, and that the entry exists whenever an RK stage asks for it.

## The defect

**`Src/Amr/AMReX_AmrLevel.cpp:2188`** — AmrLevel::FillPatcherFill builds one cached FillPatcher with desc.interp(scomp) and fills the whole [scomp,scomp+ncomp) range with it, ignoring StateDescriptor::sameInterps, so components registered with a different per-component interpolater are interpolated with the wrong mapper at the coarse/fine boundary.

**`Src/Amr/AMReX_AmrLevel.cpp:2283`** — AmrLevel::FillRKPatch dereferences m_fillpatcher[state_index] without a null check; a post-step regrid replaces the fine AmrLevel with a fresh object whose FillPatchers are all nullptr, so the next RK3/RK4 stage fill is a null-pointer dereference.

## Why it matters

*F078:* Any build, level>=1. App registers State_Type with N components, calling desc_lst.setComponent(State_Type,N-1,"mask",bc,func,&pc_interp) while the rest default to cell_cons_interp, then calls AmrLevel::FillPatcherFill(mf,0,N,ng,t,State_Type,0) (also reached via AmrLevel::RK). The mask ghost cells at the C/F boundary get conservative-linear values instead of piecewise-constant, producing non-integral mask values; AmrLevel::FillPatch on the same data is correct.

*F079:* Any build/dimension with finest_level>=1. A derived AmrLevel calls setPostStepRegrid(1) during advance() at level L and time-steps with AmrLevel::RK(3,...) or RK(4,...). Amr::timeStep runs regrid(L,time) right after advance(L) (Src/Amr/AMReX_Amr.cpp:2090), rebuilding amr_level[L+1] with null m_fillpatcher entries; the immediately following timeStep(L+1) -> advance -> RungeKutta::RK3 fillbndry -> FillRKPatch dereferences the null unique_ptr and segfaults.

## How to reach it

- *F078:* config: Any AMReX_SPACEDIM/backend with AmrLevel (default AMReX_AMRLEVEL=ON), finest_level>=1. Built by gcc.yml CPU jobs.
- *F078:* API path: desc_lst.addDescriptor(S, TheCellType(), Point, ng, N, &cell_cons_interp); desc_lst.setComponent(S, N-1, "mask", bc, func, &pc_interp); // per-component interp // on a level>=1 during advance(): FillPatcherFill(mf, 0, N, ng, time, S, 0); // or AmrLevel::RK(2, S, ...) which calls it for all comps // even FillPatcherFill(mf, N-1, 1, ng, time, S, N-1) afterwards reuses the cached cell_cons FillPatcher
- *F078:* test coverage: Tests/Amr/Advection_AmrLevel/Source/AmrLevelAdv.cpp:283 calls FillPatcherFill with a single-interp descriptor; no assertion, not a ctest target.
- *F079:* config: Any AMReX_SPACEDIM/backend with AmrLevel, amr.max_level>=1. gcc.yml CPU builds; no CI test exercises RK3/4 with post-step regrid.
- *F079:* API path: struct MyLevel : AmrLevel { Real advance(Real t, Real dt, int it, int nc) override { RK(3, State_Type, t, dt, it, nc, rhs); if (level==0) setPostStepRegrid(1); return dt; } }; Amr::coarseTimeStep -> timeStep(0): advance(0) stores level-1 FillPatcher -> regrid(0,time) [Amr.cpp:2094] rebuilds level 1 (m_fillpatcher null) -> timeStep(1) -> advance(1) -> RK(3) -> RungeKutta::RK3 fillbndry(1) -> FillRKPatch -> segfault
- *F079:* test coverage: none. Advection_AmrLevel uses FillPatcherFill (RK2-style) and never sets post_step_regrid; not a ctest target.

## Suggested fix

Both diffs are diagnostic guards, landable independently and in either order; neither changes behavior for the cases that work today. For FillPatcherFill, the guard asserts `desc.identicalInterps(scomp, ncomp)` before building or reusing the cached FillPatcher. This mirrors the existing `BL_ASSERT(identicalInterps)` in FillPatchIteratorHelper and matches what storeRKCoarseData already assumes with `desc.interp(0)`. The alternative is a real fix: loop over `desc.sameInterps(scomp, ncomp)` the way FillCoarsePatch does and hold one cached FillPatcher per range, keyed by (state_index, range start) instead of state_index alone. FillPatcher stores a single InterpBase pointer with no accessor, so the cache structure itself has to change. Maintainer decision: accept the single-interp restriction (and state it in the FillPatcherFill and RK Doxygen) or extend the cache. For FillRKPatch, the guard converts the null unique_ptr dereference into an abort with a message. The root cause sits in Amr::timeStep: the post-step regrid rebuilds the finer AmrLevel through levelbld, discarding the FillPatcher (and the stored coarse RK data) the coarser level just created. Re-creating it from FillRKPatch is not possible because the data died with the old object, so supporting RK3/RK4 with setPostStepRegrid would need Amr::regrid to carry the coarse RK data across the rebuild, or the coarser level to re-store it after regrid. Maintainer decision: support that combination or document it as unsupported (FillPatcherFill's lazy path already handles the RK2 case).

For `Src/Amr/AMReX_AmrLevel.cpp:2188` (F078):

```diff
--- a/Src/Amr/AMReX_AmrLevel.cpp
+++ b/Src/Amr/AMReX_AmrLevel.cpp
@@ -2180,6 +2180,9 @@
             amrex::Abort("FillPatcherFill: Grids are not properly nested.  Must increase blocking factor.");
         }
 
+        AMREX_ALWAYS_ASSERT_WITH_MESSAGE(desc.identicalInterps(scomp,ncomp),
+            "FillPatcherFill: all components must use the same interpolater");
+
         auto& fillpatcher = m_fillpatcher[state_index];
         if (fillpatcher == nullptr) {
             fillpatcher = std::make_unique<FillPatcher<MultiFab>>
```

<sub>Aborts instead of silently interpolating with the wrong mapper, mirroring FillPatchIteratorHelper's BL_ASSERT(identicalInterps) (line ~838). FillPatcher holds a single InterpBase* with no accessor, so a per-range split would require multiple cached FillPatchers per state (cache is keyed only by state_index). Maintainer must decide whether to accept this restriction (also implicit in storeRKCoarseData's desc.interp(0)) or extend the cache to sameInterps ranges.</sub>

For `Src/Amr/AMReX_AmrLevel.cpp:2283` (F079):

```diff
--- a/Src/Amr/AMReX_AmrLevel.cpp
+++ b/Src/Amr/AMReX_AmrLevel.cpp
@@ -2280,6 +2280,8 @@
         StateDataPhysBCFunct physbcf_crse(crse_level.state[state_index], 0,
                                           crse_level.geom);
         auto& fillpatcher = m_fillpatcher[state_index];
+        AMREX_ALWAYS_ASSERT_WITH_MESSAGE(fillpatcher != nullptr,
+            "FillRKPatch: coarse RK data missing; the level was regridded mid-step");
         fillpatcher->fillRK(stage, iteration, ncycle, S, time, physbcf_crse,
                             physbcf, AmrLevel::desc_lst[state_index].getBCs());
     }
```

<sub>Turns the null unique_ptr dereference into a diagnostic abort. The root cause is Amr::timeStep's post-step regrid (AMReX_Amr.cpp ~2090) rebuilding the finer AmrLevel with empty m_fillpatcher after the coarse RK data was stored; the stored data is lost with the old object, so re-creating it is not possible from here. Maintainer must decide whether RK3/RK4 with setPostStepRegrid should be supported (regenerate coarse RK data after regrid) or documented as unsupported.</sub>

## Verification evidence

<details><summary><code>F078</code> — confirmed (one verifier lens)</summary>

Lens 1 (refutation attempt): AmrLevel.cpp:2183-2194: FillPatcher built once per state_index with `desc.interp(scomp)` and `fill(... scomp, dcomp, ncomp ...)` over the whole range; no `sameInterps` split and no `identicalInterps` assert, unlike FillPatchIterator::Initialize (m_range = desc.sameInterps, 1019), FillCoarsePatch (1582) and the helper's BL_ASSERT(identicalInterps) (838). AmrLevel::RK(2) calls FillPatcherFill(mf,0,nComp,...) (AmrLevel.H:925); storeRKCoarseData uses desc.interp(0) (AmrLevel.H:981). Cache keyed only by state_index, so a later per-component call reuses the wrong interp. Docs (AmrLevel.H:590-601, FillPatcher.H:16-74) state no same-interp requirement.

</details>

<details><summary><code>F079</code> — confirmed (one verifier lens)</summary>

Lens 1 (refutation attempt): AmrLevel.cpp:2282-2283 `auto& fillpatcher = m_fillpatcher[state_index]; fillpatcher->fillRK(...)` with no null check; entries are created only by the coarser level's storeRKCoarseData (AmrLevel.H:974-982) or lazily by FillPatcherFill (RK2). Amr::timeStep (Amr.cpp:2090-2094) calls regrid(level,time) after advance(level) when postStepRegrid(); Amr::regrid (2726-2735) always builds a new AmrLevel via levelbld and `amr_level[lev].reset(a)`, whose constructor leaves m_fillpatcher null (AmrLevel.cpp:96). RungeKutta::RK3/RK4 call fillbndry(1,Uold,time) first (RungeKutta.H:222/273) -> FillRKPatch -> null deref. Stored RK data also dies with the old object; combination is unsupported but unasserted/undocumented.

</details>

---

<sub>Based on commit `248fcfb7a5`, the tree the audit verified against. From an automated audit of AMReX `Src/`. Audit finding ids: F078, F079. Reviewer unit(s): Amr/AmrLevel. Nothing here was compiled or run except where the evidence says so — the failure scenarios are code reasoning, so the reaching configuration above is the cheapest way to confirm or refute it.</sub>

WeiqunZhang:
The FillPatcher cached in AmrLevel holds a single interpolater and is keyed by state index only, so all components must share one interpolater.  FillRKPatch also assumed the cached FillPatcher exists, but a post-step regrid rebuilds the level with an empty cache.  Assert both instead of interpolating with the wrong mapper or dereferencing a null pointer, and document the restrictions.

Fixes #5770.

### DIFF
--- a/Src/Amr/AMReX_AmrLevel.H
+++ b/Src/Amr/AMReX_AmrLevel.H
@@ -591,6 +591,8 @@ public:
      * \brief Fill ghost cells using cached FillPatcher (level>0) or FillPatch (level 0).
      *
      * All components in [scomp,scomp+ncomp) must have the same interpolater.
+     * Because the FillPatcher is cached per state, that interpolater must also
+     * be the one used by earlier calls for the same state_index.
      *
      * \param mf          Destination MultiFab.
      * \param dcomp       Destination component offset.
--- a/Src/Amr/AMReX_AmrLevel.H
+++ b/Src/Amr/AMReX_AmrLevel.H
@@ -590,6 +590,8 @@ public:
     /**
      * \brief Fill ghost cells using cached FillPatcher (level>0) or FillPatch (level 0).
      *
+     * All components in [scomp,scomp+ncomp) must have the same interpolater.
+     *
      * \param mf          Destination MultiFab.
      * \param dcomp       Destination component offset.
      * \param ncomp       Number of components to fill.
@@ -648,7 +650,9 @@ public:
      *
      * To use RK, the StateData must have all the ghost cells needed.  See
      * namespace RungeKutta for expected function signatures of the callable
-     * parameters.
+     * parameters.  All components of the StateData must have the same
+     * interpolater.  Note that RK3 and RK4 do not support post-step regrid,
+     * because the coarse level RK data are lost when a level is rebuilt.
      *
      * \param order      order of RK
      * \param state_type index of StateData
@@ -970,6 +974,10 @@ void AmrLevel::storeRKCoarseData (int state_type, Real time, Real dt,
     if (level == parent->finestLevel()) { return; }
 
     const StateDescriptor& desc = AmrLevel::desc_lst[state_type];
+
+    // The FillPatcher holds a single interpolater.
+    AMREX_ALWAYS_ASSERT_WITH_MESSAGE(desc.identicalInterps(0,desc.nComp()),
+        "AmrLevel::RK: all components must have the same interpolater");
 
     auto& fillpatcher = parent->getLevel(level+1).m_fillpatcher[state_type];
     fillpatcher = std::make_unique<FillPatcher<MultiFab>>
--- a/Src/Amr/AMReX_AmrLevel.cpp
+++ b/Src/Amr/AMReX_AmrLevel.cpp
@@ -2166,6 +2166,10 @@ AmrLevel::FillPatcherFill (MultiFab& mf, int dcomp, int ncomp, int nghost,
 
         const StateDescriptor& desc = AmrLevel::desc_lst[state_index];
 
+        // The cached FillPatcher holds a single interpolater.
+        AMREX_ALWAYS_ASSERT_WITH_MESSAGE(desc.identicalInterps(scomp,ncomp),
+            "FillPatcherFill: all components must have the same interpolater");
+
         if (level > 1 &&!amrex::ProperlyNested(fine_level.crse_ratio,
                                                parent->blockingFactor(fine_level.level),
                                                nghost, mf.ixType(),
@@ -2280,6 +2284,8 @@ AmrLevel::FillRKPatch (int state_index, MultiFab& S, Real time,
         StateDataPhysBCFunct physbcf_crse(crse_level.state[state_index], 0,
                                           crse_level.geom);
         auto& fillpatcher = m_fillpatcher[state_index];
+        AMREX_ALWAYS_ASSERT_WITH_MESSAGE(fillpatcher != nullptr,
+            "FillRKPatch: coarse level RK data missing.  Was this level regridded mid-step?");
         fillpatcher->fillRK(stage, iteration, ncycle, S, time, physbcf_crse,
                             physbcf, AmrLevel::desc_lst[state_index].getBCs());
     }
--- a/Src/Amr/AMReX_AmrLevel.cpp
+++ b/Src/Amr/AMReX_AmrLevel.cpp
@@ -2190,6 +2190,11 @@ AmrLevel::FillPatcherFill (MultiFab& mf, int dcomp, int ncomp, int nghost,
                 (parent->boxArray(level), parent->DistributionMap(level), geom_fine,
                  parent->boxArray(level-1), parent->DistributionMap(level-1), geom_crse,
                  IntVect(nghost), desc.nComp(), desc.interp(scomp));
+        } else {
+            // The cache is keyed by state index, so a previous call may have
+            // built it with another component's interpolater.
+            AMREX_ALWAYS_ASSERT_WITH_MESSAGE(fillpatcher->interp() == desc.interp(scomp),
+                "FillPatcherFill: all components must have the same interpolater");
         }
 
         fillpatcher->fill(mf, IntVect(nghost), time,
--- a/Src/AmrCore/AMReX_FillPatcher.H
+++ b/Src/AmrCore/AMReX_FillPatcher.H
@@ -203,6 +203,9 @@ public:
     void fillRK (int stage, int iteration, int ncycle, MF& mf, Real time,
                  BC& cbc, BC& fbc, Vector<BCRec> const& bcs);
 
+    //! Interpolater this FillPatcher was built with.
+    [[nodiscard]] InterpBase* interp () const noexcept { return m_interp; }
+
 private:
 
     BoxArray m_fba;


WeiqunZhang:
# PR #5874: Assert AmrLevel FillPatcher preconditions

PR: https://github.com/AMReX-Codes/amrex/pull/5874
Reviewed commit: `d1977403ac4894bb201e54480292da7538906397`
Status: Complete
Result: Actionable findings

## Findings

### [P2] Check mapper consistency across the lifetime of the state-indexed cache

The new `desc.identicalInterps(scomp,ncomp)` assertion checks only the current component subset. With two state components using different interpolaters, successive one-component calls for `scomp=0` and `scomp=1` both pass: `identicalInterps(...,1)` always returns true. The first call creates `m_fillpatcher[state_index]` with `desc.interp(0)`; the second reuses it without checking `desc.interp(1)`, so the second component is silently interpolated with the first component's mapper.

This is a remaining hole in the precondition this PR intends to enforce, rather than a newly introduced interpolation algorithm defect. Require identical interpolaters for the entire cached state (`0, desc.nComp()`), as `storeRKCoarseData` already does, and update the FillPatcherFill documentation accordingly; alternatively, validate the cached mapper before every reuse. Cover successive calls on different one-component subsets.

Location: [AMReX_AmrLevel.cpp:2169-2171](https://github.com/AMReX-Codes/amrex/blob/d1977403ac4894bb201e54480292da7538906397/Src/Amr/AMReX_AmrLevel.cpp#L2169-L2171); [cache creation/reuse](https://github.com/AMReX-Codes/amrex/blob/d1977403ac4894bb201e54480292da7538906397/Src/Amr/AMReX_AmrLevel.cpp#L2188-L2203).

## Validation

Read the PR description, conversation, reviews and inline comments, and inspected the complete diff with `git diff ae079342654b60f95effddd4377cd9ffaf573f5f d1977403ac4894bb201e54480292da7538906397 --`. Traced the complete FillPatcherFill routine, StateDescriptor::identicalInterps, per-state cache construction/reset, storeRKCoarseData and FillRKPatch. The finding follows directly from the one-component checks and cache key; no full AmrLevel application reproducer was built.

Reviewed by gpt-6-astra.

WeiqunZhang:
codex:

• No actionable findings at 2304157678b2. The previous P2 issue is fixed.

  - Four cache regression cases passed.
  - Old-code control reproduced the wrong-mapper result.
  - Advection CTest passed; CI remains in progress.

asalmgren (approved):

PR #5880, opened 2026-09-08
Reset StateData's preread FabArray header map pointer on restart exit

### LINKED ISSUE #5781: Amr::restart leaves StateData::faHeaderMap pointing at a destroyed function-local map
**Severity:** medium · **Category:** memory-ub · **Subsystem:** Amr (blast radius 3/5)

**Location:** `Src/Amr/AMReX_Amr.cpp:1490`

## The defect

**What is wrong** — `StateData::SetFAHeaderMapPtr(&faHeaderMap)` stores the address of a function-local map in a static pointer that is never reset, so `StateData::faHeaderMap` dangles as soon as `Amr::restart()` returns.

## Why it matters

Any build with MPI/CPU. `faHeaderMap` is declared at line 1464 inside `Amr::restart` and destroyed on return, but `StateData::faHeaderMap` keeps pointing at it. A process that restarts twice (e.g. `{Amr a(bld); a.init(...);}` then `{Amr b(bld); b.init(...);}` with `amr.restart=<chk>`), where the second checkpoint has no `FabArrayHeaders.txt` or is read with `amr.prereadFAHeaders=0`, skips the re-`Set`, so `StateData::restart` at AMReX_StateData.cpp:271-275 dereferences freed memory and hands a garbage `faHeader` pointer to `VisMF::Read`.

## How to reach it

- config: Any CPU build with AMReX_AMRLEVEL=ON (default), e.g. AMReX_SPACEDIM=3, MPI on/off. gcc.yml builds Tests/Amr/Advection_AmrLevel with AMReX_ENABLE_TESTS=ON; no CI job restarts twice.
- API path: amrex::Initialize(); ParmParse pp("amr"); pp.add("restart","chkA"); { Amr a(bld); a.init(0,0); } // chkA has FabArrayHeaders.txt -> static set pp.add("restart","chkB"); pp.add("prereadFAHeaders",0); // or chkB is a legacy checkpoint lacking FabArrayHeaders.txt { Amr b(bld); b.init(0,0); } // Amr::restart (protected, reached via public init) -> AmrLevel::restart -> StateData::restart derefs dangling faHeaderMap. No in-tree caller restarts twice; the sequence uses only public API.
- test coverage: Tests/Amr/Advection_AmrLevel (main.cpp:56 constructs Amr, supports amr.restart) exercises a single restart; no test restarts twice or asserts on the header pointer.

## Suggested fix

Clear the static pointer before `Amr::restart()` returns: call `StateData::SetFAHeaderMapPtr(nullptr)` at the end of the function, next to `BL_PROFILE_REGION_STOP("Amr::restart()")`. The preread map exists only to serve the `StateData::restart` calls made from `AmrLevel::restart` inside this function; nothing consumes it afterwards, and `StateData::restart` (AMReX_StateData.cpp:271) already treats a null pointer as "read the header from disk", so resetting it restores the pre-preread behaviour for any later restart that does not re-populate the map. The only exits between the `Set` and the end of the function are `amrex::Abort` calls, so a single reset at the tail is sufficient. This mirrors what `Amr::Finalize` does for the other Amr statics (`regrid_ba`, `initial_ba`, plot-var lists); adding the same reset there would also be defensible, but it would not cover the two-restarts-in-one-Initialize case, so the in-function reset is the one that matters. Alternative the maintainer may prefer: make the map static (or a `StateData` member) so its lifetime matches the pointer. That costs memory for the whole run, which is why the reset-on-exit approach is the recommended one. Severity was revised to low: it needs two restarts in one process with the second lacking `FabArrayHeaders.txt` or run with `amr.prereadFAHeaders=0`.

```diff
--- a/Src/Amr/AMReX_Amr.cpp
+++ b/Src/Amr/AMReX_Amr.cpp
@@ -1744,5 +1744,6 @@
 
         amrex::Print() << "Restart time = " << dRestartTime << " seconds." << '\n';
     }
+    StateData::SetFAHeaderMapPtr(nullptr);
     BL_PROFILE_REGION_STOP("Amr::restart()");
 }
```

<sub>Resets the static StateData::faHeaderMap pointer before the function-local map goes out of scope, so a later restart that skips the re-Set (no FabArrayHeaders.txt or prereadFAHeaders=0) sees nullptr and falls back to reading headers, instead of dereferencing a destroyed stack map. All StateData::restart/post_restart consumers run inside Amr::restart, so nothing reads the map after this point; the only exits between Set and here are aborts.</sub>

## Verification evidence

<details><summary><code>F075</code> — confirmed (one verifier lens)</summary>

Lens 1 (refutation attempt): Amr.cpp:1464 `std::map<...> faHeaderMap;` (function-local), :1490 `StateData::SetFAHeaderMapPtr(&faHeaderMap);` only inside `if(!faHeaderFileChars.empty())`. grep of Src/ shows the static `StateData::faHeaderMap` (StateData.cpp:28) is written nowhere else and never reset to nullptr; Amr::Finalize (Amr.cpp:141-153) clears other statics but not this one. StateData.cpp:271-275 dereferences `faHeaderMap->find(...)` whenever non-null. Amr::Finalize sets `initialized=false` and ~Amr calls variableCleanUp, so sequential Amr instances are an intended pattern (Tests/Reinit repeats Initialize/Finalize). Second restart that skips the re-Set (legacy checkpoint without FabArrayHeaders.txt, or prereadFAHeaders=0 added via ParmParse) reads a destroyed stack map. Narrow but genuine use-after-scope.

</details>

---

<sub>Based on commit `248fcfb7a5`, the tree the audit verified against. From an automated audit of AMReX `Src/`. Audit finding id: F075. Reviewer unit(s): Amr/Amr-2. Nothing here was compiled or run except where the evidence says so — the failure scenarios are code reasoning, so the reaching configuration above is the cheapest way to confirm or refute it.</sub>

WeiqunZhang:
Amr::restart hands StateData a pointer to a function-local map of preread FabArray headers, but never cleared it, so the static pointer dangled after the function returned.  A second restart in the same run that does not repopulate the map would then read destroyed memory.

Fixes #5781.

### DIFF
--- a/Src/Amr/AMReX_Amr.cpp
+++ b/Src/Amr/AMReX_Amr.cpp
@@ -1744,6 +1744,10 @@ Amr::restart (const std::string& filename)
 
         amrex::Print() << "Restart time = " << dRestartTime << " seconds." << '\n';
     }
+
+    // ---- faHeaderMap is local to this function
+    StateData::SetFAHeaderMapPtr(nullptr);
+
     BL_PROFILE_REGION_STOP("Amr::restart()");
 }
 


WeiqunZhang:
# PR #5880: Reset StateData's preread FabArray header map pointer on restart exit

PR: https://github.com/AMReX-Codes/amrex/pull/5880
Reviewed commit: `b65cf5d0d630d5e34fba012ae66f7d5c3fbec7bb`
Status: Complete

## Findings

No actionable correctness issues found. Clearing the static header-map pointer before the local map is destroyed prevents later restarts from reusing its address. The reset occurs after the restart consumers and post-restart work.

## Validation

Read the PR description, conversation, reviews and inline comments, and inspected the complete diff with `git diff ae079342654b60f95effddd4377cd9ffaf573f5f b65cf5d0d630d5e34fba012ae66f7d5c3fbec7bb --`. Traced faHeaderMap lifetime, SetFAHeaderMapPtr assignment, StateData restart lookup and the end of Amr::restart. No repeated-checkpoint restart was run; exception unwinding was not exercised.

Reviewed by gpt-6-astra.

asalmgren (approved):

PR #5883, opened 2026-09-08
Guard null message and out-of-range argument index

### LINKED ISSUE #5790: AMReX.cpp: Error_host throws runtime_error(nullptr); command_argument_count/get_command_argument miss empty and negative-index guards
**Severity:** medium/low · **Category:** correctness, memory-ub · **Subsystem:** Base-io (blast radius 2/5)

**Locations:** `Src/Base/AMReX.cpp:262`, `Src/Base/AMReX.cpp:984`, `Src/Base/AMReX.cpp:990`

Three small entry points in Src/Base/AMReX.cpp trust their inputs even though the header contracts (AMReX.H:275, 449, 455) explicitly allow a null message, an empty command line, and an out-of-range index. Each fix is a one-line guard that brings the function in line with what the header already promises.

## The defect

**`Src/Base/AMReX.cpp:262`** — With amrex.throw_exception=1, Error_host throws RuntimeError(msg) even when msg is nullptr (the no-argument amrex::Abort()/amrex::Error() overloads), which constructs std::runtime_error from a null char*: libstdc++ throws std::logic_error instead, libc++ crashes. _(also reported as F136)_

**`Src/Base/AMReX.cpp:984`** — command_argument_count() returns size()-1 = -1 when no command line was recorded (argc==0), instead of 0 arguments.

**`Src/Base/AMReX.cpp:990`** — get_command_argument(number) only checks the upper bound; a negative number passes `number < ssize()` and indexes command_arguments[negative], undefined behaviour, although the header promises an empty string for an out-of-range index.

## Why it matters

*F097:* Any build (e.g. CPU, 3D, MPI off), runtime amrex.throw_exception=1 amrex.signal_handling=0 as documented in Debugging.rst. Application catches std::runtime_error and triggers a no-message abort path, e.g. ParmParse::getToken on a malformed inputs string (AMReX_ParmParse.cpp:547), RealBox stream read of a bad token (AMReX_RealBox.cpp:59), FArrayBox::setFormat with a bad format (AMReX_FArrayBox.cpp:269). Instead of a catchable std::runtime_error, libstdc++ throws std::logic_error('basic_string::_S_construct null not valid') (verified with g++ -std=c++20); libc++ dereferences nullptr.

*F297:* Any build. Application initializes via amrex::Initialize(MPI_Comm) (or FFT::Init_FFT / Init_minimal), which passes argc=0 so command_arguments stays empty; amrex::command_argument_count() then returns -1 rather than 0, contradicting the header contract 'number of command line arguments excluding argv[0]' (Fortran's command_argument_count returns 0 in this case). Code sizing a container or looping `for (int i = 1; i <= n; ...)` with a negative count misbehaves.

*F298:* Any build. amrex::Initialize(argc,argv) then amrex::get_command_argument(-1) (e.g. an index derived from command_argument_count() which itself can be -1 per the previous finding, or from user arithmetic): std::vector::operator[] with a negative index reads before the buffer -> garbage std::string copy or segfault instead of the documented std::string().

## How to reach it

- *F097:* config: Any CPU build (e.g. AMReX_SPACEDIM=3, AMReX_MPI=OFF or ON, no GPU); runtime amrex.throw_exception=1 amrex.signal_handling=0. Built by every gcc.yml/clang.yml job; macos.yml (libc++) hits the strlen(nullptr) crash variant.
- *F097:* API path: amrex::Initialize(argc,argv) with inputs `amrex.throw_exception=1 amrex.signal_handling=0 code.a=abc`; then `ParmParse pp("code"); int a; try { pp.query("a",a); } catch (std::runtime_error const&) {}` -> squeryval type-mismatch (ParmParse.cpp:1159) calls amrex::Abort() -> Error_host("Abort", nullptr) -> throw RuntimeError(nullptr). Also RealBox operator>> bad token (RealBox.cpp:59), sgetval not-found (ParmParse.cpp:1188).
- *F097:* test coverage: Tests/ParmParse (ctest, gcc.yml) sets throw_exception=1 and catches std::runtime_error, but its exercised path is amrex::Error("...recursive reference...") with a message; no in-tree test exercises a no-message Abort under throw_exception.
- *F297:* config: Any build (e.g. AMReX_SPACEDIM=3, AMReX_MPI=ON, CPU). Default configuration in every gcc.yml/clang.yml job.
- *F297:* API path: MPI_Init(&argc,&argv); amrex::Initialize(MPI_COMM_WORLD); int n = amrex::command_argument_count(); // n == -1. Same via amrex::Init_minimal(comm) / amrex::Init_FFT(comm). Public API; no in-tree caller uses it after a comm-only Initialize (Tools/Plotfile always go through Initialize(argc,argv) with argc>=1 and their `farg<=narg` loops tolerate -1).
- *F297:* test coverage: none (Tests/MultiBlock/Advection and Tests/FFT/RawPtr use comm-only Initialize/Init_FFT but never call command_argument_count; Tools/Plotfile/f*.cpp use it only after Initialize(argc,argv))
- *F298:* config: Any build (e.g. AMReX_SPACEDIM=3, AMReX_MPI=OFF, CPU). Default configuration in every gcc.yml/clang.yml job.
- *F298:* API path: amrex::Initialize(MPI_COMM_WORLD); std::string last = amrex::get_command_argument(amrex::command_argument_count()); // count is -1 (F297) -> command_arguments[-1]. Or directly amrex::get_command_argument(-1) after any Initialize. Public API; no in-tree caller passes a negative index (Tools/Plotfile use farg>=1).
- *F298:* test coverage: none (no Tests/ or Src/ caller passes a negative index; Tools/Plotfile callers loop from 1)

## Suggested fix

Three independent one-line guards; land them in one PR, in any order. **Error_host (F097).** The non-throw sibling path already tests the pointer (`write_lib_id` and `write_to_stderr_without_buffering` both do `if (msg)`), so the `throw_exception` branch should too. The attached diff falls back to `type` ("Abort"/"Error", always non-null from the inline callers in AMReX.H) so the exception carries some text. Maintainer decision: whether to prefer a combined `type + ": " + msg` string for the exception `what()`; that would change the text seen by existing `catch (std::runtime_error const&)` users (e.g. Tests/ParmParse), so the minimal fallback is the safer default. The `system::error_handler` branch is left as is: the header documents that the handler may receive nullptr, so that is its contract, not a bug. **command_argument_count (F297).** `Initialize(MPI_Comm)`, `Init_minimal` and `Init_FFT` pass `argc = 0`, leaving `command_arguments` empty, so `size()-1` yields -1. Clamp to 0, matching the header wording and the Fortran 2003 `command_argument_count` the function was modelled on (commit da798a3b10). `<algorithm>` is already included. **get_command_argument (F298).** Add the lower-bound check alongside the `std::ssize` comparison introduced in #5368, so a negative index returns the documented empty string instead of indexing before the vector's buffer. No doc change needed: the fixes make the code match Doxygen text that is already correct.

For `Src/Base/AMReX.cpp:262` (F097):

```diff
--- a/Src/Base/AMReX.cpp
+++ b/Src/Base/AMReX.cpp
@@ -259,7 +259,7 @@ amrex::Error_host (const char* type, const char * msg)
     if (system::error_handler) {
         system::error_handler(msg);
     } else if (system::throw_exception) {
-        throw RuntimeError(msg);
+        throw RuntimeError(msg ? msg : type);
     } else {
         write_lib_id(type);
         write_to_stderr_without_buffering(msg);
```

<sub>Guards the nullptr msg before constructing std::runtime_error, matching the non-throw sibling path (write_lib_id/write_to_stderr_without_buffering both test the pointer). Falls back to `type` ("Abort"/"Error", always non-null from the callers) so the exception carries some text. Maintainer may prefer a combined `type: msg` string; the user-installed error_handler branch still receives a raw nullptr, which is its documented contract.</sub>

For `Src/Base/AMReX.cpp:984` (F297):

```diff
--- a/Src/Base/AMReX.cpp
+++ b/Src/Base/AMReX.cpp
@@ -981,7 +981,7 @@ amrex::get_command ()
 int
 amrex::command_argument_count ()
 {
-    return static_cast<int>(command_arguments.size())-1;
+    return std::max(0, static_cast<int>(command_arguments.size())-1);
 }
 
 std::string
```

<sub>Clamps the count to 0 when command_arguments is empty (Initialize(MPI_Comm) / Init_minimal pass argc=0), matching the header promise and Fortran's command_argument_count. `<algorithm>` is already included in this file.</sub>

For `Src/Base/AMReX.cpp:990` (F298):

```diff
--- a/Src/Base/AMReX.cpp
+++ b/Src/Base/AMReX.cpp
@@ -987,7 +987,7 @@ amrex::command_argument_count ()
 std::string
 amrex::get_command_argument (int number)
 {
-    if (number < std::ssize(command_arguments)) {
+    if (number >= 0 && number < std::ssize(command_arguments)) {
         return command_arguments[number];
     } else {
         return std::string();
```

<sub>Adds the missing lower-bound check so a negative index returns the documented empty string instead of indexing std::vector before its buffer. Keeps the existing ssize comparison from #5368.</sub>

## Verification evidence

<details><summary><code>F097</code> — confirmed (one verifier lens); duplicates folded in: F136</summary>

Lens 1 (refutation attempt): Src/Base/AMReX.cpp:262 `throw RuntimeError(msg);` with no null check; AMReX.H:275 documents msg 'may be nullptr' and Abort/Error default `msg = nullptr` (AMReX.H:281,324). RuntimeError = std::runtime_error (AMReX_Exception.H). The non-throw sibling guards `if (str)` (line 207). Compiled scratchpad snippet with g++ 15.2/libstdc++: runtime_error(nullptr) raises std::logic_error 'basic_string::_S_construct null not valid'; libc++ does strlen(nullptr). development blob line 263 identical (only dev commit d12ac9c70d touches hook stacks). gh issue/PR search empty.

Note: `development` has touched this file since the audited commit; the verifier consulted it read-only and did not find this defect fixed there.

</details>

<details><summary><code>F297</code> — confirmed (one verifier lens)</summary>

Lens 1 (refutation attempt): Src/Base/AMReX.cpp:984 `return static_cast<int>(command_arguments.size())-1;`. command_arguments cleared at line 417 and only populated inside `if (argc > 0)` (line 419); Initialize(MPI_Comm,...) at line 347 sets `int argc = 0` so the vector stays empty and the function returns -1. Header AMReX.H:449 promises 'number of command line arguments excluding argv[0]'; originating commit da798a3b10 'add Fortran 2003 style command argument functions' (Fortran returns 0). development blob line 985 identical; no issue/PR.

Note: `development` has touched this file since the audited commit; the verifier consulted it read-only and did not find this defect fixed there.

</details>

<details><summary><code>F298</code> — confirmed (one verifier lens)</summary>

Lens 1 (refutation attempt): Src/Base/AMReX.cpp:990-991 `if (number < std::ssize(command_arguments)) { return command_arguments[number]; }` - no lower-bound check, so number<0 indexes std::vector::operator[] with a negative subscript (UB). Header AMReX.H:455: 'Requested argument or an empty string when number is out of range.' Original da798a3b10 had no check at all; fa795322b4 (#5368) only changed the comparison to ssize. development blob lines 991-992 identical; no issue/PR.

Note: `development` has touched this file since the audited commit; the verifier consulted it read-only and did not find this defect fixed there.

</details>

---

<sub>Based on commit `248fcfb7a5`, the tree the audit verified against. From an automated audit of AMReX `Src/`. Audit finding ids: F097, F297, F298, F136. Reviewer unit(s): Base/Init-Profilers-1. Nothing here was compiled or run except where the evidence says so — the failure scenarios are code reasoning, so the reaching configuration above is the cheapest way to confirm or refute it.</sub>

WeiqunZhang:
Error_host no longer constructs std::runtime_error from a null message; it falls back to the "Error"/"Abort" label. command_argument_count now returns 0 instead of -1 when no command line was recorded, and get_command_argument returns an empty string for a negative index.

Fixes #5790.

### DIFF
--- a/Src/Base/AMReX.cpp
+++ b/Src/Base/AMReX.cpp
@@ -260,7 +260,7 @@ amrex::Error_host (const char* type, const char * msg)
     if (system::error_handler) {
         system::error_handler(msg);
     } else if (system::throw_exception) {
-        throw RuntimeError(msg);
+        throw RuntimeError(msg ? msg : type);
     } else {
         write_lib_id(type);
         write_to_stderr_without_buffering(msg);
@@ -984,13 +984,13 @@ amrex::get_command ()
 int
 amrex::command_argument_count ()
 {
-    return static_cast<int>(command_arguments.size())-1;
+    return std::max(0, static_cast<int>(command_arguments.size())-1);
 }
 
 std::string
 amrex::get_command_argument (int number)
 {
-    if (number < std::ssize(command_arguments)) {
+    if (number >= 0 && number < std::ssize(command_arguments)) {
         return command_arguments[number];
     } else {
         return std::string();


WeiqunZhang:
# PR #5883: Guard null message and out-of-range argument index

PR: https://github.com/AMReX-Codes/amrex/pull/5883
Reviewed commit: `fc08bf735ea0e01548708b2fa6ef9194a3b5a202`
Status: Complete

## Findings

No actionable correctness issues found. The standard Error/Abort labels provide valid exception text for null messages. Command-line queries now handle an empty recorded argument list and negative indices without invalid indexing.

## Validation

Read the PR description, conversation, reviews and inline comments, and inspected the complete diff with `git diff ae079342654b60f95effddd4377cd9ffaf573f5f fc08bf735ea0e01548708b2fa6ef9194a3b5a202 --`. Traced Error/Abort wrappers, Error_host handler/exception branches and command argument storage/query boundaries.

### Focused local probe

Compiled the exact changed AMReX.cpp against the common 2D Debug library. Before Initialize, verified zero argument count and empty strings for indices -1 and 0. With exception mode enabled, Error() and Abort() with default null messages threw RuntimeError with the corresponding labels. All assertions passed; no base control was run.

```sh
python3 agent-notes/codex-reviews-2/validation/build-probe.py 5883 head
/tmp/codex-reviews-2/5883/head/probe
```

Reviewed by gpt-6-astra.

asalmgren (approved):

PR #5887, opened 2026-09-08
Reset Parser/IParser state when compilation fails

### LINKED ISSUE #5797: Parser/IParser::compileHost keeps partial bytecode buffer after parser_compile throws, later compile returns it
**Severity:** medium · **Category:** correctness · **Subsystem:** Base-parser (blast radius 1/5)

**Location:** `Src/Base/Parser/AMReX_Parser.H:311`

## The defect

**What is wrong** — When parser_compile throws (e.g. "Unknown variable"), m_data->m_host_executor stays non-null pointing at a partially written, unterminated bytecode buffer; every later compile()/compileHost() on this Parser (or any copy sharing m_data) skips compilation and returns an executor over that garbage. _(also reported as F146)_

## Why it matters

Any build. Parser p("x+y"); p.registerVariables({"x"}); try { p.compile<1>(); } catch (std::runtime_error const&) {} auto f = p.compile<1>(); f(1.0); -> second compile returns executor whose buffer holds an ADD_PP node with uninitialized i2 and no PARSER_EXE_NULL terminator; parser_exe_eval reads x[garbage]/pstack[garbage] and runs past the buffer end (OOB read / crash / silent garbage instead of an error). setConstant/registerVariables also refuse with a misleading "after compile()" message.

## How to reach it

- config: Any config (default CMake: SPACEDIM=3, DOUBLE, no GPU). Built by every job in gcc.yml/clang.yml/macos.yml/windows.yml; the failure path itself is not exercised by CI.
- API path: Parser p("x+y"); p.registerVariables({"x"}); try { p.compile<1>(); } catch (std::runtime_error const&) {} auto f = p.compile<1>(); // returns executor over unterminated buffer f(1.0); // reads uninitialized i2, walks past buffer Public API; no in-tree caller catches and recompiles, but #2233 introduced the runtime_error specifically for callers to catch.
- test coverage: none. Tests/Parser (ctest target, CPU CI) asserts numerics but never catches a compile failure and reuses the Parser; Src callers (ParmParse::pp_make_parser, EB2.cpp:223) compile once with no catch.

## Suggested fix

Make a failed compile leave the `Data` object in the same state as never having compiled. In the `catch` block of `Parser::compileHost` (Src/Base/Parser/AMReX_Parser.H, before the rethrow) release `m_host_executor` with the same arena-vs-`std::free` choice that `Parser::Data::~Data` uses (AMReX_Parser.cpp), set it back to `nullptr`, and restore `m_use_arena = true` (its declared default) so the next attempt retries the pinned arena. With `m_host_executor` null again, the `if (!(m_data->m_host_executor))` guard re-runs compilation on the next `compile()`/`compileHost()`, and `setConstant`/`registerVariables` stop reporting "after compile()" for a compile that never succeeded. `compile()` needs no change: it calls `compileHost` first and only allocates `m_device_executor` afterwards, so the device path is never reached on failure. `m_exe_size` and `m_locals` may be left stale; both are rewritten on the next successful compile. Apply the identical change to the twin in `IParser::compileHost` (AMReX_IParser.H), mirroring `IParser::Data::~Data`; the two parsers are kept line-for-line parallel and the bug is present in both. Land both in one PR. One judgement call for the maintainer: whether a failed compile should also be made re-triable at all, or whether the parser should simply mark itself invalid. Freeing and nulling is the smaller change and matches what the destructor already does, so it is the recommended option.

```diff
--- a/Src/Base/Parser/AMReX_Parser.H
+++ b/Src/Base/Parser/AMReX_Parser.H
@@ -309,6 +309,13 @@
                 m_data->m_locals = parser_compile(m_data->m_parser, m_uf_ptrs,
                                                   m_data->m_host_executor);
             } catch (const std::runtime_error& e) {
+                if (m_data->m_use_arena) {
+                    The_Pinned_Arena()->free(m_data->m_host_executor);
+                } else {
+                    std::free(m_data->m_host_executor);
+                }
+                m_data->m_host_executor = nullptr;
+                m_data->m_use_arena = true;
                 throw std::runtime_error(std::string(e.what()) + " in Parser expression \""
                                          + m_data->m_expression + "\"");
             }
--- a/Src/Base/Parser/AMReX_IParser.H
+++ b/Src/Base/Parser/AMReX_IParser.H
@@ -218,6 +218,13 @@
             try {
                 iparser_compile(m_data->m_iparser, m_data->m_host_executor);
             } catch (const std::runtime_error& e) {
+                if (m_data->m_use_arena) {
+                    The_Pinned_Arena()->free(m_data->m_host_executor);
+                } else {
+                    std::free(m_data->m_host_executor);
+                }
+                m_data->m_host_executor = nullptr;
+                m_data->m_use_arena = true;
                 throw std::runtime_error(std::string(e.what()) + " in IParser expression \""
                                          + m_data->m_expression + "\"");
             }
```

<sub>On parser_compile failure, free the partially written bytecode buffer (arena vs std::free, mirroring Data::~Data in AMReX_Parser.cpp:54-58 / AMReX_IParser.cpp:53-57), null m_host_executor, and reset m_use_arena to its default so a later compile() re-runs compilation and setConstant/registerVariables work again. Same fix applied to the IParser twin. m_exe_size/m_locals staleness is harmless since they are rewritten on the next compile.</sub>

## Verification evidence

<details><summary><code>F148</code> — confirmed (one verifier lens); duplicates folded in: F146</summary>

Lens 1 (refutation attempt): AMReX_Parser_Exe.H:596 `char* p = nullptr;` in parser_exe_size; every parser_symbol_idx call in AMReX_Parser_Exe.cpp (only source of `throw std::runtime_error("Unknown variable "...)`, :23-24) is under `if (p)` (e.g. :155-159), so the size pass never throws. AMReX_Parser.H:302 allocates m_host_executor before the try at :308; the catch at :311 rethrows without freeing/nulling. `new(p) ParserExeNull` is written only after success (Exe.H:617); ParserExeADD_PP i1/i2 uninitialized (Exe.H:144-148). Next compileHost: `if (!(m_data->m_host_executor))` (:286) is false, returns executor over the partial buffer. No fix on development or in KNOWN_STATE.

</details>

---

<sub>Based on commit `248fcfb7a5`, the tree the audit verified against. From an automated audit of AMReX `Src/`. Audit finding ids: F148, F146. Reviewer unit(s): Base/Parser-2. Nothing here was compiled or run except where the evidence says so — the failure scenarios are code reasoning, so the reaching configuration above is the cheapest way to confirm or refute it.</sub>

WeiqunZhang:
When parser_compile throws, free the partially written bytecode buffer and reset m_host_executor and m_use_arena. Otherwise a later compile() returns an executor over that garbage buffer, and setConstant and registerVariables wrongly refuse with a "after compile()" error.

Fixes #5797.

### DIFF
--- a/Src/Base/Parser/AMReX_IParser.H
+++ b/Src/Base/Parser/AMReX_IParser.H
@@ -176,6 +176,7 @@ private:
         int m_exe_size = 0;
         Data () = default;
         ~Data ();
+        void clear_host_executor ();
         Data (Data const&) = delete;
         Data (Data &&) = delete;
         Data& operator= (Data const&) = delete;
@@ -218,6 +219,8 @@ IParser::compileHost () const
             try {
                 iparser_compile(m_data->m_iparser, m_data->m_host_executor);
             } catch (const std::runtime_error& e) {
+                // Do not leave a partially compiled executor behind.
+                m_data->clear_host_executor();
                 throw std::runtime_error(std::string(e.what()) + " in IParser expression \""
                                          + m_data->m_expression + "\"");
             }
--- a/Src/Base/Parser/AMReX_IParser.cpp
+++ b/Src/Base/Parser/AMReX_IParser.cpp
@@ -50,16 +50,24 @@ IParser::Data::~Data ()
 {
     m_expression.clear();
     if (m_iparser) { amrex_iparser_delete(m_iparser); }
+    clear_host_executor();
+#ifdef AMREX_USE_GPU
+    if (m_device_executor) { The_Arena()->free(m_device_executor); }
+#endif
+}
+
+void
+IParser::Data::clear_host_executor ()
+{
     if (m_host_executor) {
         if (m_use_arena) {
             The_Pinned_Arena()->free(m_host_executor);
         } else {
             std::free(m_host_executor);
         }
+        m_host_executor = nullptr;
+        m_use_arena = true;
     }
-#ifdef AMREX_USE_GPU
-    if (m_device_executor) { The_Arena()->free(m_device_executor); }
-#endif
 }
 /// \endcond
 
--- a/Src/Base/Parser/AMReX_Parser.H
+++ b/Src/Base/Parser/AMReX_Parser.H
@@ -254,6 +254,7 @@ private:
         Vector<char const*> m_locals;
         Data () = default;
         ~Data ();
+        void clear_host_executor ();
         Data (Data const&) = delete;
         Data (Data &&) = delete;
         Data& operator= (Data const&) = delete;
@@ -309,6 +310,8 @@ Parser::compileHost () const
                 m_data->m_locals = parser_compile(m_data->m_parser, m_uf_ptrs,
                                                   m_data->m_host_executor);
             } catch (const std::runtime_error& e) {
+                // Do not leave a partially compiled executor behind.
+                m_data->clear_host_executor();
                 throw std::runtime_error(std::string(e.what()) + " in Parser expression \""
                                          + m_data->m_expression + "\"");
             }
--- a/Src/Base/Parser/AMReX_Parser.cpp
+++ b/Src/Base/Parser/AMReX_Parser.cpp
@@ -51,16 +51,24 @@ Parser::Data::~Data ()
 {
     m_expression.clear();
     if (m_parser) { amrex_parser_delete(m_parser); }
+    clear_host_executor();
+#ifdef AMREX_USE_GPU
+    if (m_device_executor) { The_Arena()->free(m_device_executor); }
+#endif
+}
+
+void
+Parser::Data::clear_host_executor ()
+{
     if (m_host_executor) {
         if (m_use_arena) {
             The_Pinned_Arena()->free(m_host_executor);
         } else {
             std::free(m_host_executor);
         }
+        m_host_executor = nullptr;
+        m_use_arena = true;
     }
-#ifdef AMREX_USE_GPU
-    if (m_device_executor) { The_Arena()->free(m_device_executor); }
-#endif
 }
 /// \endcond
 


WeiqunZhang:
# PR #5887: Reset Parser/IParser state when compilation fails

PR: https://github.com/AMReX-Codes/amrex/pull/5887
Reviewed commit: `cd55ad6b9581512204977d03f72a4a9a568c397e`
Status: Complete

## Findings

No actionable correctness issues found. Compilation failure clears the partially written host executor with its matching allocator and resets the allocation mode. Subsequent compilation or symbol repair can proceed through the normal initialization path.

## Validation

Read the PR description, conversation, reviews and inline comments, and inspected the complete diff with `git diff ae079342654b60f95effddd4377cd9ffaf573f5f cd55ad6b9581512204977d03f72a4a9a568c397e --`. Compared Parser/IParser compileHost, allocation fallback, cleanup/destructors, constant/variable setters and device-executor creation ordering.

### Focused local probe

Passed repeated compile failure, setConstant/registerVariables repair and successful evaluation for both Parser and IParser, before and after AMReX initialization (covering malloc fallback and arena allocation). The expression is x + abs(missing); the base control incorrectly returns an executor on its second Parser compile attempt and fails the assertion. An initial x + missing probe did not trigger the intended unknown-symbol exception in IParser, so that expression was replaced before drawing this conclusion. Device executor allocation was not executed.

Probe source: `agent-notes/codex-reviews-2/validation/pr-5887.cpp`. The local `agent-notes/codex-reviews-2/validation/README.md` records the common Debug build and extraction procedure. Exact changed headers/translation units were compiled and linked against that common library; these are not whole builds of every PR.

```sh
python3 agent-notes/codex-reviews-2/validation/build-probe.py 5887 head
timeout 30 /tmp/codex-reviews-2/5887/head/probe
python3 agent-notes/codex-reviews-2/validation/build-probe.py 5887 base
timeout 30 /tmp/codex-reviews-2/5887/control/probe
```

Compiler commands actually run are saved in `validation/pr-5887-commands.txt`.

Reviewed by gpt-6-astra.

asalmgren (approved):

PR #5889, opened 2026-09-08
Guard the 32-bit item count in CSR::sort

### LINKED ISSUE #5806: Low-severity findings in LinearSolvers: CSR::sort 32-bit nnz, MLNodeTensorLaplacian cfine freed without sync
**Severity:** low · **Category:** memory-ub · **Subsystem:** LinearSolvers (blast radius 4/5)

**Locations:** `Src/LinearSolvers/AMReX_CSR.H:256`, `Src/LinearSolvers/MLMG/AMReX_MLNodeTensorLaplacian.cpp:142`

Two unrelated minor items from the LinearSolvers audit, collected here so they are not lost: an unguarded Long-to-int narrowing in the GPU fallback of CSR::sort(), and a temporary MultiFab in MLNodeTensorLaplacian::interpolation() that is destroyed without a stream sync when MLMG runs in a no-sync region.

## The defect

**`Src/LinearSolvers/AMReX_CSR.H:256`** — CSR::sort() passes the Long member nnz into cub::DeviceSegmentedRadixSort::SortPairs / rocprim::segmented_radix_sort_pairs, whose item-count parameters are 32-bit (int / unsigned int), so a local CSR with more than 2^31 nonzeros silently truncates the count while nrows is explicitly guarded by an assert.

**`Src/LinearSolvers/MLMG/AMReX_MLNodeTensorLaplacian.cpp:142`** — interpolation() destroys the temporary cfine right after launching kernels that read it, with no stream sync when Gpu::inNoSyncRegion() is true (MFIter's destructor sync is disabled there, AMReX_MFIter.cpp:39), unlike the guarded siblings.

## Why it matters

*F402:* AMREX_USE_HIP (or CUDA) build, 1 rank, large-memory GPU (e.g. MI300A 128 GB): SpMatrix<float>::define(partition, csr, CsrSorted{false}) with ~80M local rows x 27-point stencil (nnz > 2^31) whose columns are unsorted and at least one row longer than warp_size*ITEMS_PER_THREAD, so the fallback branch runs. `size`/`num_items` wraps to a small value, the temp-storage query at line 256 under-sizes d_temp, and the sort writes past the allocation -> heap corruption / unsorted column indices and wrong SpMV results.

*F407:* GPU build, MLMG solve run inside a Gpu::noSyncRegion (the use case PR #4351 supports), with agglomerated/consolidated coarse levels so need_parallel_copy is true: interpolation returns and cfine's device memory is released to the Arena while the per-box mlndtslap_interpadd kernels may still be reading it; the block can be re-used by a non-stream-ordered writer -> corrupted interpolation.

## How to reach it

- *F402:* config: AMReX_GPU_BACKEND=CUDA with CUDA <=12.x (CCCL <=2.8) or AMReX_GPU_BACKEND=HIP; AMReX_LINEAR_SOLVERS=ON (default); any SPACEDIM/precision. cuda.yml (12.2/12.5/12.6) and hip.yml (ROCm 6.3.2/7.0) build this, but no GPU CI job runs anything.
- *F402:* API path: AlgPartition part(nlocal_rows); SpMatrix<float> A; A.define(part, csr, CsrSorted{false}); // SpMatrix.H:748 -> CSR::sort() // or A.setVal(f, CsrSorted{false}) (line 950), A.sortCSR() (823), define(ptrs,...,CsrSorted{false}) (810) // fallback branch needs some row with >128 entries and local nnz > 2^31 (CUDA 12) / > 2^32 (HIP); ~80+ GB device memory
- *F402:* test coverage: Tests/Algebra/SpMV/main.cpp:123,250 and Tests/Algebra/GMRES/main.cpp:103 call sort() via CsrSorted{false}, but with tiny matrices and rows <=128, so neither the fallback branch nor the overflow is exercised; no test asserts on it.
- *F407:* config: AMReX_GPU_BACKEND=CUDA (or HIP/SYCL); LPInfo agglomeration or consolidation so need_parallel_copy is true; cuda.yml/hip.yml build this config but do not run tests.
- *F407:* API path: MLNodeTensorLaplacian lop(geom,grids,dmap, LPInfo().setAgglomeration(true)); MLMG mlmg(lop); mlmg.setNoGpuSync(true); // default false on ai_audit mlmg.solve({&sol},{&rhs},tol,0.); // mgVcycle -> interpolation with cfine
- *F407:* test coverage: none: Tests/LinearSolvers/NodeTensorLap does not call setNoGpuSync; no GPU CI runs ctest.

## Suggested fix

Per-finding checklist; the two items are independent and can land in either order or as one small PR. - [ ] **F402, `Src/LinearSolvers/AMReX_CSR.H` (CSR::sort fallback).** Guard `nnz` the same way this function already guards `nrows` a few lines above (`AMREX_ALWAYS_ASSERT(... < Long(std::numeric_limits<int>::max()))`) and the way the SpMatrix transpose path asserts `nnz < INT_MAX` before narrowing (`AMReX_SpMatrix.H` ~471). The assert turns silent truncation and out-of-bounds writes into a hard abort, which is the convention used elsewhere in this file. Maintainer decision: whether that is sufficient, or whether the segmented sort should instead be chunked over row ranges so local nnz >= 2^31 keeps working on CUDA 12 / HIP. Note CUDA 13's CUB takes a 64-bit count, so a chunked path would only matter for older toolkits and ROCm. - [ ] **F407, `Src/LinearSolvers/MLMG/AMReX_MLNodeTensorLaplacian.cpp` (interpolation).** Add the post-loop sync the PR #4351 siblings have (`AMReX_MLNodeABecLaplacian.cpp` ~234, `AMReX_MLNodeLaplacian.cpp` ~689) so `cfine` is not returned to The_Arena while kernels may still read it. The attached diff uses `cfine.local_size() > 0 && Gpu::inNoSyncRegion()` rather than the siblings' `||`, because MFIter::Finalize already syncs outside a no-sync region, and `streamSynchronizeActive()` because MFIter rotates boxes across streams. Maintainer decision: keep the sibling form verbatim for consistency, or switch `cfine` to The_Async_Arena() (as the unmerged `mlmg_nosync` branch does) and drop the sync entirely. `restriction()` needs no change; ParallelCopy syncs before its temporary is destroyed.

For `Src/LinearSolvers/AMReX_CSR.H:256` (F402):

```diff
--- a/Src/LinearSolvers/AMReX_CSR.H
+++ b/Src/LinearSolvers/AMReX_CSR.H
@@ -243,5 +243,7 @@
     if (*h_needs_fallback)
     {
+        AMREX_ALWAYS_ASSERT(nnz < Long(std::numeric_limits<int>::max()));
+
         V<Long> col_index_out(col_index.size());
         V<T> mat_out(mat.size());
         auto* d_col_out = col_index_out.data();
```

<sub>Guards nnz before it is narrowed to the 32-bit item count of cub::DeviceSegmentedRadixSort::SortPairs / rocprim::segmented_radix_sort_pairs, matching the nrows guard at line 155 in this function and the nnz < INT_MAX assert in the SpMatrix.H transpose path (lines 471-473). Turns silent truncation/OOB into a hard abort; the maintainer must decide whether a chunked segmented sort (per-row-range calls) is wanted instead so nnz >= 2^31 keeps working.</sub>

For `Src/LinearSolvers/MLMG/AMReX_MLNodeTensorLaplacian.cpp:142` (F407):

```diff
--- a/Src/LinearSolvers/MLMG/AMReX_MLNodeTensorLaplacian.cpp
+++ b/Src/LinearSolvers/MLMG/AMReX_MLNodeTensorLaplacian.cpp
@@ -167,6 +167,10 @@ MLNodeTensorLaplacian::interpolation (int amrlev, int fmglev, MultiFab& fine,
             });
         }
     }
+
+    if (cfine.local_size() > 0 && Gpu::inNoSyncRegion()) {
+        Gpu::streamSynchronizeActive();
+    }
 }
 
 void
```

<sub>Same guard idea as PR #4351 (MLNodeABecLaplacian.cpp:234), but uses && because MFIter::Finalize already syncs when not in a NoSyncRegion, and streamSynchronizeActive() rather than streamSynchronize() because MFIter rotates boxes across streams (MFIter.cpp:381); this mirrors Gpu::SyncAtExitOnly. Restriction needs no change: ParallelCopy syncs internally (TagParallelFor.H:290, MPI pack path) before cfine is destroyed.</sub>

## Verification evidence

<details><summary><code>F402</code> — confirmed (one verifier lens)</summary>

Lens 1 (refutation attempt): AMReX_CSR.H:256/265 pass `nnz` (Long) as item count. CCCL v2.3.0 (CUDA 12.x, what cuda.yml builds: 12.2/12.5/12.6) declares `int num_items, int num_segments` with `using OffsetT = int`; dispatch sizes temp as `num_items * sizeof(KeyT)` (ping-pong buffer written at true segment offsets). rocPRIM: `unsigned int size`, tmp keys = `size` elements. Sibling transpose path SpMatrix.H:471-473/535-537 asserts `nnz < INT_MAX` before narrowing; sort() has no such guard. Caveats: CUDA 13.3 CUB 3.0.3 takes `int64_t num_items` (unaffected); HIP wraps at 2^32 not 2^31; for 2^31<=nnz<2^32 on CUDA 12 the int is negative so temp_bytes ~1.8e19 and Arena aborts 'Out of memory' (hard fail); silent OOB needs nnz>=2^32.

</details>

<details><summary><code>F407</code> — confirmed (one verifier lens)</summary>

Lens 1 (refutation attempt): MLNodeTensorLaplacian.cpp:138-169: cfine (The_Arena, CArena::free at AMReX_CArena.cpp:296 is an immediate free-list insert, not stream-ordered) is read by per-box AMREX_HOST_DEVICE_PARALLEL_FOR_3D kernels in an MFIter whose device_sync=!Gpu::inNoSyncRegion() (MFIter.cpp:39), then destroyed at function exit with no sync. Sibling MLNodeLaplacian.cpp (from #4351) guards `if (cfine.local_size() > 0 || !Gpu::inNoSyncRegion()) Gpu::streamSynchronize();`. MLMG.H:793-795 sets NoSyncRegion when setNoGpuSync(true). The user's unmerged local branch mlmg_nosync (cac213c4af) switches this exact cfine to The_Async_Arena(), but that is not on development (git log ai_audit..development on the file is empty).

</details>

---

<sub>Based on commit `248fcfb7a5`, the tree the audit verified against. From an automated audit of AMReX `Src/`. Audit finding ids: F402, F407. Reviewer unit(s): LS/Algebra-GMRES-1, LS/NodeABec-NodeTensor. Nothing here was compiled or run except where the evidence says so — the failure scenarios are code reasoning, so the reaching configuration above is the cheapest way to confirm or refute it.</sub>

WeiqunZhang:
The GPU fallback passes nnz to a segmented radix sort whose item count is 32-bit, so assert on it as this function already does for nrows.  The other item in the issue, the cfine lifetime in MLNodeTensorLaplacian, is handled by #4895 with The_Async_Arena.

Fixes #5806.

### DIFF
--- a/Src/LinearSolvers/AMReX_CSR.H
+++ b/Src/LinearSolvers/AMReX_CSR.H
@@ -242,6 +242,9 @@ void CSR<T,V>::sort ()
 
     if (*h_needs_fallback)
     {
+        // The item count of the segmented sort below is 32-bit.
+        AMREX_ALWAYS_ASSERT(nnz < Long(std::numeric_limits<int>::max()));
+
         V<Long> col_index_out(col_index.size());
         V<T> mat_out(mat.size());
         auto* d_col_out = col_index_out.data();


WeiqunZhang:
# PR #5889: Guard the 32-bit item count in CSR::sort

PR: https://github.com/AMReX-Codes/amrex/pull/5889
Reviewed commit: `ab56474eaa839ad865ea173b511a59e6eaec55fc`
Status: Complete

## Findings

No actionable correctness issues found. The fallback now rejects an nnz count outside the supported signed-int range before invoking segmented radix sort or allocating its output buffers. The existing short-row path is unchanged.

## Validation

Read the PR description, conversation, reviews and inline comments, and inspected the complete diff with `git diff ae079342654b60f95effddd4377cd9ffaf573f5f ab56474eaa839ad865ea173b511a59e6eaec55fc --`. Traced fallback detection, nnz type, row-count check and both temporary-size/execution sort calls. No allocation near the 32-bit limit or GPU sort test was run.

CI at final refresh (2026-09-08 04:45 UTC): IN_PROGRESS: 7, SUCCESS: 68. The refresh confirmed the same reviewed head and OPEN state. CI status is separate from the source assessment.

Reviewed by gpt-6-astra.

asalmgren (approved):

PR #5893, opened 2026-09-08
Support disabling FP exception traps on macOS

WeiqunZhang:
get/set/enable/disableFPExcept were implemented for Linux/glibc only, and silently did nothing elsewhere. But we do enable trapping on macOS, on both x86_64 and arm64. So a caller that disables trapping to survive a failing computation was still killed by a signal there. Read and write the trap bits of the FP control register directly.

### DIFF
--- a/Docs/sphinx_documentation/source/Debugging.rst
+++ b/Docs/sphinx_documentation/source/Debugging.rst
@@ -28,8 +28,7 @@ use of uninitialized values, AMReX also initializes ``FArrayBox``\ s in
 with ``TEST=TRUE`` or ``DEBUG=TRUE`` in GNU make, or with ``-DCMAKE_BUILD_TYPE=Debug`` in CMake.
 One can also control this setting for ``FArrayBox`` using the runtime parameter ``fab.init_snan``.
 Note for Macs: Apple silicon chips using the Arm64 architecture are not able to trap
-division by zero.  The trap can still be requested and will be reported as enabled by
-``amrex::getFPExcept()``, but no exception will be raised.
+division by zero.
 
 One can get more information than the backtrace of the call stack by
 instrumenting the code.  Here is an example.
--- a/Docs/sphinx_documentation/source/Debugging.rst
+++ b/Docs/sphinx_documentation/source/Debugging.rst
@@ -27,7 +27,9 @@ use of uninitialized values, AMReX also initializes ``FArrayBox``\ s in
 ``MultiFab``\ s and arrays allocated by ``bl_allocate`` to signaling NaNs when it is compiled
 with ``TEST=TRUE`` or ``DEBUG=TRUE`` in GNU make, or with ``-DCMAKE_BUILD_TYPE=Debug`` in CMake.
 One can also control this setting for ``FArrayBox`` using the runtime parameter ``fab.init_snan``.
-Note for Macs: M1 and M2 chips using the Arm64 architecture are not able to trap division by zero.
+Note for Macs: Apple silicon chips using the Arm64 architecture are not able to trap
+division by zero.  The trap can still be requested and will be reported as enabled by
+``amrex::getFPExcept()``, but no exception will be raised.
 
 One can get more information than the backtrace of the call stack by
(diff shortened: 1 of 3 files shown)


WeiqunZhang:
Needed by https://github.com/BLAST-WarpX/warpx/pull/7213

WeiqunZhang:
# Review notes: PR #5893 — macOS support for FP exception trap control

Files reviewed: `Src/Base/AMReX.H`, `Src/Base/AMReX.cpp` (full diff, read
against the surrounding `Initialize`/`Finalize` FPE code).

## Finding

**`Src/Base/AMReX.cpp:1091` — `~off_bits` truncates the 64-bit FPCR.** Low
severity; unintended write, no observable misbehavior today.

```cpp
env.__fpcr &= ~off_bits;   // off_bits is unsigned int
```

On Apple arm64, `fenv_t` is `struct { unsigned long long __fpsr; unsigned long
long __fpcr; }`. `~off_bits` is a 32-bit value that is zero-extended before the
`&=`, so FPCR bits 63:32 are cleared on *every* call — including
`enableFPExcept`, where `off_bits == 0` and `~0u` becomes `0x00000000FFFFFFFF`.
Those bits are RES0 in AArch64 today, so nothing breaks in practice, but
`<fenv.h>` states that modifying bits outside the documented masks is
undefined. Fix:

```cpp
env.__fpcr &= ~static_cast<unsigned long long>(off_bits);
```

## Checked and correct

- **Overlap semantics** of `apple_set_fpexcept(on, off)` when a bit appears in
  both sets (the `setFPExcept(excepts, FPExcept::all)` path): both branches
  apply `off` first and `on` second, so `on` wins — matching the glibc
  `fedisableexcept(all)` + `feenableexcept(flags)` sequence.
- **x86 mask polarity** is inverted correctly in both directions
  (`_MM_GET_EXCEPTION_MASK` set bit = masked = *not* trapped), and
  `_MM_SET_EXCEPTION_MASK` is fed only mask bits read back from
  `_MM_GET_EXCEPTION_MASK`, so the underflow/denormal/inexact masks and the
  rounding mode in MXCSR are preserved.
- **`prev` is captured via `getFPExcept()` before mutation** in all three
  setters, so the documented `setFPExcept(prev)` restore pattern works on
  macOS.
- **Preprocessor guards line up**: `<xmmintrin.h>` is included under
  `__APPLE__ && __x86_64__`, `<cfenv>` is unconditionally available on Apple,
  and the new anonymous namespace sits inside `namespace amrex` so `FPExcept`
  and `any()` resolve. Universal builds (`CMAKE_OSX_ARCHITECTURES:
  "arm64;x86_64"`) compile each arch in a separate pass, so the
  `#if defined(__x86_64__)` split is safe.
- **The new get/set matches what `amrex::Initialize` installs** on each arch
  (MXCSR mask on x86_64, FPCR trap bits on arm64), so `disableFPExcept` really
  does undo the traps AMReX itself sets — the bug the PR targets.
- No other in-tree caller of `FPExcept` exists, so nothing else changes
  behavior silently.

## Observations, not findings

- The header docs dropped "Linux only", but
  `Docs/sphinx_documentation/source/Debugging.rst:30` still carries the
  standalone note that M1/M2 Arm64 chips cannot trap divide-by-zero. That is
  still true in hardware, but a reader of the new header docs may now expect
  `enableFPExcept(FPExcept::zero)` to work there — and `getFPExcept()` will
  report `zero` as enabled, since the FPCR bit reads back even where the
  hardware will not trap. A sentence in the docs would keep the two consistent.
- At review time the macOS CI jobs (`AppleClang Universal w/o MPI`,
  `AppleClang@11.0 GFortran@9.3`) and `CUDA on Windows` were still pending;
  everything else was green. The two AppleClang jobs are the only ones that
  compile the new code paths, so they are the ones to confirm before merge.

asalmgren (approved):

ax3l:
Thank you for fixing this!

ax3l (approved):
Looks good! Have not tested on real hardware yet.

PR #5894, opened 2026-09-08
MLMG: restore the GPU no-sync region when solve throws

WeiqunZhang:
setNoGpuSync makes solve enter a single-stream no-sync region, but it was left entered if the solve threw, because the flags were reset only on the normal return path. Both convergence failures throw when setThrowException is used. So use the RAII guards instead. This matters for a caller that catches the failure and keeps running.


WeiqunZhang:
needed by https://github.com/BLAST-WarpX/warpx/pull/7213

WeiqunZhang:
Close this because what's in this PR is already in #4895, which also has more changes needed by https://github.com/BLAST-WarpX/warpx/pull/7213.

PR #5876, opened 2026-09-08
Use finite covered-face sentinels in EB face interpolation

WeiqunZhang:
The 2D face-interpolation kernels and the MultiFab initialization in EB_interp_CellCentroid_to_FaceCentroid stored 1e40 into a Real, which becomes +inf in single precision, so a consumer forming apx*phi_face on a covered face got NaN instead of 0.  Follow the 3D kernels and EB_interp_CC_to_FaceCentroid, which already use finite sentinels.

Fixes #5773.

PR #5895, opened 2026-09-09
CI: add a single precision GCC job that runs ctest

WeiqunZhang:
Build in single precision with EB, linear solvers, AmrLevel, particles and FFT, then run ctest and clang-tidy.  Single precision was only exercised by the GitLab H100 GPU runs, so solver breakage kept surfacing late: #4785, #5319 and #5672.  The job found a float-conversion bug in the 2D nodal EB kernel and a NodeTensorLap tolerance that cannot be met in float.  263 -Wfloat-conversion lines remain under Tests/, so that warning is not an error in this job yet; cleanup follows.

### DIFF
--- a/.github/workflows/gcc.yml
+++ b/.github/workflows/gcc.yml
@@ -758,6 +758,69 @@ jobs:
         ccache -s
         du -hs ~/.cache/ccache
 
+  # Single precision build that runs the test suite.  Single precision solvers
+  # used to be exercised only by the GitLab GPU runs.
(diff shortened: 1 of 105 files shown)


WeiqunZhang:
This needs changes from https://github.com/AMReX-Codes/amrex/pull/5876

asalmgren (approved):

atmyers (approved):

atmyers (approved):

PR #5901, opened 2026-09-09
CMake: support LLVMFlang in the compiler flag targets

WeiqunZhang:
Flang got no flags from Flags_Fortran, Flags_FASTMATH or Flags_FPE because LLVMFlang was missing from the compiler id list, so Debug builds were silently left without -fimplicit-none. The two existing Fortran/Clang entries could never match and are now spelled correctly. Of the GNU Fortran debug options, Flang 21 accepts only -fimplicit-none; bounds checking and FPE trapping do not exist.

Also drop a duplicated clause in the typecheck compiler check.

### DIFF
--- a/Tools/CMake/AMReXFlagsTargets.cmake
+++ b/Tools/CMake/AMReXFlagsTargets.cmake
@@ -23,12 +23,12 @@ include_guard(GLOBAL)
 # for every combination of
 #
 #     <lang> = cxx,fortran,cuda
-#     <id>   = gnu,intel,pgi,cray,clang,appleclang,crayclang,ibmclang,intelllvm,msvc,nvidia,nvhpc,xlclang
+#     <id>   = gnu,intel,pgi,cray,clang,appleclang,crayclang,ibmclang,intelllvm,msvc,nvidia,nvhpc,xlclang,llvmflang
 #
 foreach (_language CXX Fortran CUDA )
    string(TOLOWER "${_language}" _lang)
 
-   foreach (_comp GNU Intel PGI Cray Clang AppleClang CrayClang IBMClang IntelLLVM MSVC NVIDIA NVHPC XLClang )
+   foreach (_comp GNU Intel PGI Cray Clang AppleClang CrayClang IBMClang IntelLLVM MSVC NVIDIA NVHPC XLClang LLVMFlang )
       string(TOLOWER "${_comp}" _id)
       # Define variables
       set(_${_lang}_${_id}      "$<COMPILE_LANG_AND_ID:${_language},${_comp}>")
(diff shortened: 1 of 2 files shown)


Tools/CMake/AMReXFlagsTargets.cmake
```diff
@@ -23,12 +23,12 @@ include_guard(GLOBAL)
...
 #
 #     <lang> = cxx,fortran,cuda
-#     <id>   = gnu,intel,pgi,cray,clang,appleclang,crayclang,ibmclang,intelllvm,msvc,nvidia,nvhpc,xlclang
+#     <id>   = gnu,intel,pgi,cray,clang,appleclang,crayclang,ibmclang,intelllvm,msvc,nvidia,nvhpc,xlclang,llvmflang
 #
 foreach (_language CXX Fortran CUDA )
    string(TOLOWER "${_language}" _lang)
 
-   foreach (_comp GNU Intel PGI Cray Clang AppleClang CrayClang IBMClang IntelLLVM MSVC NVIDIA NVHPC XLClang )
+   foreach (_comp GNU Intel PGI Cray Clang AppleClang CrayClang IBMClang IntelLLVM MSVC NVIDIA NVHPC XLClang LLVMFlang )
```
ax3l, line 31: Looks good. It seems `LLVMFlang` is the new ID for the new [LLVM code stack](https://github.com/llvm/llvm-project/tree/main/flang), `Flang` was the [old stack](https://github.com/flang-compiler/flang). CMake 3.28+ adds it, all good here.
https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_ID.html

Tools/CMake/AMReXFlagsTargets.cmake
```diff
@@ -104,6 +104,9 @@ target_compile_options( Flags_Fortran
    $<${_fortran_pgi_rel}:-gopt -fast>
    $<${_fortran_cray_dbg}:-O0 -e i>
    $<${_fortran_cray_rel}:>
+   # Flang has no bounds/uninitialized checking options.
+   $<${_fortran_llvmflang_dbg}:-fimplicit-none>
+   $<${_fortran_llvmflang_rel}:-fimplicit-none>
```
ax3l, line 109: I think this works, too:
```suggestion
   $<${_fortran_llvmflang}:-fimplicit-none>
```

Tools/CMake/AMReXFlagsTargets.cmake
ax3l, line 102: Separate PR, but now also transitioned to the new `intelllvm`

ax3l (approved):
LGTM!

PR #4374, opened 2025-03-13
Guard the use of GNU extensions

WeiqunZhang:
Alpine Linux uses the musl C library, which does not have fegetexcept, feenableexcept, and fedisableexcept.

Close #4373

Issue #4373, opened 2025-03-12
Compling with musl libc (Alpine Linux)


ax3l:
Compiling with alpine/[musl libc](https://musl.libc.org/), which is a common libc used in Python & Docker environments, does not support exception management `fegetexcept`, `feenableexcept`, `fedisableexcept`. They seem to be GNU-only.
- https://man7.org/linux/man-pages/man3/feenableexcept.3.html
- https://linux.die.net/man/3/fedisableexcept

We need to guard the respective functions in `AMReX.cpp` with the GNU libc macro `HAVE_FEDISABLEEXCEPT` (double check name, we can also use other checks to make sure GNU libc is used/not used).

Error message:
```

2025-03-12T18:45:57.8097333Z /project/amrex/Src/Base/AMReX.cpp: In function 'amrex::AMReX* amrex::Initialize(int&, char**&, bool, MPI_Comm, const std::function<void()>&, std::ostream&, std::ostream&, ErrorHandler)':
2025-03-12T18:45:57.8099222Z /project/amrex/Src/Base/AMReX.cpp:592:36: error: 'fegetexcept' was not declared in this scope; did you mean 'fetestexcept'?
2025-03-12T18:45:57.8101686Z   592 |                 prev_fpe_excepts = fegetexcept();
2025-03-12T18:45:57.8102198Z       |                                    ^~~~~~~~~~~
2025-03-12T18:45:57.8102649Z       |                                    fetestexcept
2025-03-12T18:45:57.8144630Z /project/amrex/Src/Base/AMReX.cpp:594:21: error: 'feenableexcept' was not declared in this scope; did you mean 'feraiseexcept'?
2025-03-12T18:45:57.8146351Z   594 |                     feenableexcept(curr_fpe_excepts);  // trap floating point exceptions
2025-03-12T18:45:57.8147659Z       |                     ^~~~~~~~~~~~~~
2025-03-12T18:45:57.8148211Z       |                     feraiseexcept
2025-03-12T18:45:57.8241552Z /project/amrex/Src/Base/AMReX.cpp: In function 'void amrex::Finalize(AMReX*)':
2025-03-12T18:45:57.8242876Z /project/amrex/Src/Base/AMReX.cpp:830:13: error: 'fedisableexcept' was not declared in this scope; did you mean 'feraiseexcept'?
2025-03-12T18:45:57.8244687Z   830 |             fedisableexcept(curr_fpe_excepts);
[...]

PR #5903, opened 2026-09-10
CI: Musl Linux libc / Alpine

ax3l:
## Summary

musl libc is getting more and more popular with cloud users. notably, because it replaces glibc, some GNU-extensions in AMReX need guarding to avoid breakage.

This adds three quick new native runners:
- x86-64
- x86-32
- arm64

on musl libc to ensure we cover any breakage early. The additional architecture (especially the 32bit one) is also helpful to find precision and platform differences quickly to make our APIs, data layouts and numerics more robust.

## Additional background

Related to #4373 and already merged fixes #4385 #4374

## Checklist

The proposed changes:

### DIFF
--- a/.github/workflows/alpine.yml
+++ b/.github/workflows/alpine.yml
@@ -1,0 +1,118 @@
+name: AlpineLinux
+
+on:
+  push:
+    branches-ignore:
+      - 'dependabot/**'
+  pull_request:
+
+concurrency:
+  group: ${{ github.ref }}-${{ github.head_ref }}-alpine-linux
+  cancel-in-progress: true
+
+jobs:
+  check_changes:
+    uses: ./.github/workflows/check_changes.yml
+
+  # Alpine Linux uses musl libc instead of glibc.  Everything we guard with
+  # __GLIBC__ (trapping of floating-point exceptions, backtraces, ...) is
+  # compiled out here, so this catches accidental use of GNU extensions.  The
+  # jobs also cover two CPU architectures that no other Linux job covers,
+  # 32-bit x86 and Arm64, and they are the only place where the SIMD code is
+  # built for Arm NEON.
+  #
+  # We enter Alpine with `docker run` instead of a job-level `container:`
+  # because the Actions runner ships no musl build of Node for Arm64, see
+  # https://github.com/actions/runner/issues/3665.  This way the actions
+  # (checkout, cache) run on the Ubuntu host and only the build runs in Alpine.
+  # Every container runs natively, none is emulated.
+  musl:
+    name: musl ${{ matrix.arch }} [tests]
+    runs-on: ${{ matrix.runner }}
+    needs: check_changes
+    if: needs.check_changes.outputs.has_non_docs_changes == 'true'
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - id: x86_64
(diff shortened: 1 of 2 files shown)


.github/workflows/alpine.yml
```diff
@@ -0,0 +1,118 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - id: x86_64
+            arch: x86-64
+            runner: ubuntu-24.04
+            platform: linux/amd64
+          - id: x86
+            arch: 32bit x86
```
ax3l, line 43:
```suggestion
            arch: x86-32 
```

WeiqunZhang:
● Review of PR #5903 (CI: Musl Linux libc / Alpine) — effort level medium. New .github/workflows/alpine.yml plus a one-line addition to cleanup-cache.yml; no source changes.

  Verified correct: alpine:3.24 publishes 386/arm64/amd64 manifests; AMReX_FFTW_OMP_SUFFIX is real (Tools/CMake/FindAMReXFFTW.cmake:46) and both fftw3_threads/fftw3f_threads are needed; cache
  keys ccache-AlpineLinux-musl-<id>-git-<sha> match what cleanup-cache.yml splits on, and ...-x86-git- does not collide with ...-x86_64-git-; vir-simd v0.4.4 matches gcc.yml; the container's
  missing git is handled by the GIT_FOUND guard in Tools/CMake/AMReXUtils.cmake:16.

  Findings

  1. -DAMReX_TEST_TYPE=Small defeats the stated SIMD purpose — .github/workflows/alpine.yml:107

  I confirmed this directly: SIMD appears only in AMREX_TESTS_SUBDIRS in the else() (All) branch of Tests/CMakeLists.txt:145, and no Small test references SIMD. Everything under AMREX_USE_SIMD is
  templates (Src/Base/AMReX_SIMD.H, AMReX_ReduceSIMD.H, AMReX_GpuLaunchFunctsSIMD.H), so with -DAMReX_SIMD=ON the build only compiles #include <vir/simd.h> — no NEON codegen, no SIMD test run.
  The comment at lines 21-22 ("the only place where the SIMD code is built for Arm NEON") isn't true as configured. Either add SIMD to the Small list, or drop -DAMReX_SIMD=ON and the claim.

  2. Step-level CXXFLAGS leaks into the vir-simd build — .github/workflows/alpine.yml:87

  -Werror -Wall -Wextra -Wpedantic … is exported into the container, so CMake picks it up as the initial CMAKE_CXX_FLAGS for the vir-simd configure/build on lines 89-90. In gcc.yml vir-simd is
  installed in a separate step with no CXXFLAGS, so third-party code is never touched by the warning set. A warning from vir-simd's own compile/try_compile checks — or a new warning in a future
  GCC — fails the job for a reason unrelated to AMReX. Clear it with -e CXXFLAGS= for that portion, or install vir-simd in a separate docker run.

  3. No assertions in the only 32-bit / Linux-Arm64 job — .github/workflows/alpine.yml:96

  Release without -DAMReX_ASSERTIONS=ON. This is exactly where size/alignment assumptions break, and the comparable gcc.yml test jobs enable assertions. A 32-bit
  Long/size_t mismatch would silently give wrong results rather than trip an assert. Low severity, cheap coverage.

PR #5905, opened 2026-09-10
Allow VisMF to read denormalized numbers

dpgrote:
## Summary

There is a bug (I call it a bug) in libc++ when reading in floating point numbers. If the value is a denormalized number that is otherwise valid and representable, the `failbit` is set on the `std::istream` object, even though the number is read in correctly. This PR submits a fix/hack that looks for this case and when it happens, it resets the `failbit` so that the reading of the data can continue. In my case, this problem was happening when reading in the min/max values in a Cell_H file when running amrvis.

## Additional background

In the routine `__num_get_float `, libc++ code uses `strtod` to convert the text string read in to a double precision number. If the number is denormalized the value is interpreted correctly but `strtod` sets the errno to `ERANGE`. Then `__num_get_float` only checks if `errno == ERANGE` and sets the `failbit` when true. It does not distinguish between truly bad numbers and denormalized numbers. The ideal solution is to have the libc++ code fixed.

## Checklist

The proposed changes:

### DIFF
--- a/Src/Base/AMReX_VisMF.cpp
+++ b/Src/Base/AMReX_VisMF.cpp
@@ -7,8 +7,10 @@
 #include <AMReX_VisMF.H>
 
 #include <cerrno>
+#include <cmath>
 #include <cstdio>
 #include <iterator>
+#include <ios>
 #include <limits>
 #include <vector>
 #include <utility>
@@ -297,11 +299,15 @@ operator>> (std::istream&         is,
 
         for(Long j = 0; j < M; ++j) {
 #ifdef BL_USE_FLOAT
-            is >> dtemp >> ch;
+            is >> dtemp;
             ar[i][j] = static_cast<Real>(dtemp);
 #else
-            is >> ar[i][j] >> ch;
+            is >> ar[i][j];
 #endif
+            if (is.fail() and ar[i][j] != 0.0 and std::fpclassify(ar[i][j]) == FP_SUBNORMAL) {
+                is.clear(is.rdstate() & ~std::ios_base::failbit);
+            }
+            is >> ch;
             if( ch != ',' ) {
               amrex::Error("Expected a ',' got something else");
             }
--- a/Src/Base/AMReX_VisMF.cpp
+++ b/Src/Base/AMReX_VisMF.cpp
@@ -304,7 +304,7 @@ operator>> (std::istream&         is,
 #else
             is >> ar[i][j];
 #endif
-            if (is.fail() and ar[i][j] != 0.0 and std::fpclassify(ar[i][j]) == FP_SUBNORMAL) {
+            if (is.fail() && ar[i][j] != 0.0 && std::fpclassify(ar[i][j]) == FP_SUBNORMAL) {
                 is.clear(is.rdstate() & ~std::ios_base::failbit);
             }
             is >> ch;


WeiqunZhang:
@dpgrote You could give https://github.com/AMReX-Codes/amrexplorer a try.

● No, amrexplorer doesn't have this bug. The PR fixes a libc++ quirk: operator>> sets failbit when it reads a denormal number (a valid value so small it's stored at reduced precision), because
  the underlying strtod reports it as out of range (ERANGE) even though the value came back correctly. amrexplorer doesn't use operator>> to read doubles from plotfile files.

  - The code the PR changes (min/max values in the Cell_H headers): readStatisticValue in src/io/plotfile/PlotfileMetadataReader.cpp:94 reads a text token and calls strtod. It never looks at
    errno, so a denormal goes through. It also accepts inf/nan, which AMReX can write there.
  - Other doubles in the Header, like time: readRequired<double> (PlotfileMetadataReader.cpp:52) also uses strtod. It rejects only non-finite results, and a comment there explains that ERANGE was
    deliberately not used as the test so denormals still parse. A unit test covers a 4.9e-324 time (tests/unit/test_plotfile_header.cpp:432).
  - Numbers in expressions: src/expression/Expression.cpp:297 switched from a string stream to strtod for exactly this libc++ reason, and test_expression.cpp:325 tests denormal literals.

WeiqunZhang:
# Review notes — PR #5905 "Allow VisMF to read denormalized numbers"

- Author: dpgrote (David Grote)
- Head: `b4fd532b69301c1fa0b3b32f11e63a36541b00a6` → base `development`
- Touches: `Src/Base/AMReX_VisMF.cpp` (+8 −2)
- Reviewed: 2026-09-14

## Verdict

Diagnosis is correct and the idiom (`is.clear(is.rdstate() & ~std::ios_base::failbit)`,
clearing only `failbit` and preserving `eofbit`/`badbit`) is the right one. Two fixes
needed before merge.

## 1. `BL_USE_FLOAT` branch tests the wrong variable — fix is inert there

`Src/Base/AMReX_VisMF.cpp:301-306`. In the float build `is >> dtemp` reads into a
`double`, then `ar[i][j] = static_cast<Real>(dtemp)` narrows to `float`, and the new
guard classifies `ar[i][j]` instead of `dtemp`.

The only input that makes `is >> dtemp` set `failbit` is a *double*-subnormal
(~1e-320 or smaller), i.e. a header written by a double-precision build. Narrowing
that to `float` yields exactly `0.0f`, so `fpclassify` returns `FP_ZERO`, the
`ar[i][j] != 0.0` term is false, and the guard cannot fire.

Measured (standalone probe, `g++ -O2`):

    BL_USE_FLOAT path: dtemp=9.999889e-321 (SUBNORMAL)
                       ar[i][j]=0.000000e+00  fpclassify(ar)=FP_ZERO  => guard fires? NO

Consequence: a single-precision build reading a double-precision-written `Cell_H`
still aborts with `amrex::Error("Read of Vector<Vector<Real>> failed")`.

Note a *float*-subnormal (e.g. `1e-40`) is a perfectly normal `double`, so `strtod`
never sets `ERANGE` and that read never failed — that case was not broken.

Suggested:

```cpp
#ifdef BL_USE_FLOAT
            is >> dtemp;
            if (is.fail() && std::fpclassify(dtemp) == FP_SUBNORMAL) {
                is.clear(is.rdstate() & ~std::ios_base::failbit);
            }
            ar[i][j] = static_cast<Real>(dtemp);
#else
```

## 2. Sibling FA-level min/max reader has the same bug, untouched

`Src/Base/AMReX_VisMF.cpp:441,447` read `hd.m_famin` / `hd.m_famax`
(`Vector<Real>`, declared `Src/Base/AMReX_VisMF.H:112-113`) with the identical
`is >> famin >> ch` pattern, for `NoFabHeaderFAMinMax_v1` headers. A subnormal there
fails the same way, aborting with
`amrex::Error("Expected a ',' when reading hd.m_famin")`.

Since the same three lines would then live in three places, a file-local helper keeps
the paths from drifting:

```cpp
namespace { void clearSubnormalFailbit (std::istream& is, double v) {
    if (is.fail() && std::fpclassify(v) == FP_SUBNORMAL) {
        is.clear(is.rdstate() & ~std::ios_base::failbit);
    }
} }
```

## Minor

- `ar[i][j] != 0.0` is redundant: `FP_SUBNORMAL` already excludes zero (and normals,
  infinities, NaN). Dropping it also removes a float-equality comparison.
- `#include <ios>` (line 13) is inserted after `<iterator>`; alphabetically it belongs
  before it.

## What was verified

- Compiled PR head `AMReX_VisMF.cpp` with
  `g++ -c -O2 -std=c++20 -fopenmp -DAMREX_SPACEDIM=3 -Wall -Wextra` plus the include
  flags from `build-check/Src/CMakeFiles/amrex_3d.dir/flags.make`, both default and
  with `-DBL_USE_FLOAT`: clean, no warnings.
- Standalone probe of `operator>>` behavior on subnormals. **libstdc++ does not set
  `failbit`** (`fail=0` reading `1e-320` into a `double`), so this change is a no-op
  there — low regression risk.
- Could NOT test libc++ directly (libc++ headers not installed on this machine). The
  libc++ side of the premise rests on the author's report, not on measurement here.

## Housekeeping

- PR was fetched into local branch `pr5905-tmp` to extract the file; still present.
  Remove with `git branch -D pr5905-tmp`.

PR #5910, opened 2026-09-14
Add "--allow_diff_dx" command line argument to fcompare

atmyers:
Previously, the fcompare utility required dx in both plotfiles to be exactly equal, which in general it won't be for mixed precision comparisons.

EDIT - changed to an approach that just adds another command line argument letting you skip the dx comparison.

The proposed changes:


asalmgren (approved):

WeiqunZhang:
Claude is nit-picking, but it has its point. We don't actually have the information about the precision of Geometry.

# Review notes — PR #5910 "Allow fcompare to compare plotfiles with different precisions, even if dx differs"

- Author: atmyers (Andrew Myers)
- Head: `0dc7f2f97d` → base `development` (merge-base `07ebbee51f`)
- Touches: `Src/Base/AMReX_VisMF.{H,cpp}`, `Src/Base/AMReX_PlotFileDataImpl.{H,cpp}`,
  `Src/Base/AMReX_PlotFileUtil.H`, `Tools/Plotfile/fcompare.cpp` (+101 −3)
- Reviewed: 2026-09-14

## Verdict

The happy path works, but the precision heuristic keys on the wrong quantity: it
measures the precision of the *fab data*, while `dx` comes from the ASCII plotfile
`Header`. The two are independent, and the mismatch is reachable with a supported
runtime option (`fab.format = NATIVE_32`).

## Verification

```bash
git worktree add $SD/pr5910 pr-5910
make -C $SD/pr5910/Tools/Plotfile -j12 AMREX_HOME=$SD/pr5910 programs=fcompare DIM=3   # OK
```

Plus a 3^3 single-level writer built against the branch at `PRECISION=DOUBLE` and
`PRECISION=FLOAT` (prob domain `[0,1]^3`, 3 cells/dir, so `dx = 1/3`).

| case | dx in `Header` | data bytes | result |
|---|---|---|---|
| float build vs double build | `0.3333333432674408` / `0.33333333333333331` | 4 / 8 | **AGREE** (error before the PR) |
| same, dx perturbed 1e-3 rel | — | 4 / 8 | ERROR (good) |
| same, dx perturbed 1e-7 rel | — | 4 / 8 | AGREE (good) |
| `fab.format=NATIVE_32` from double build vs float build | `0.33333333333333331` / `0.3333333432674408` | 4 / 4 | **ERROR** ← miss |

The plotfiles used `Cell_H` version `1` (`Version_v1`), so the per-fab-header parsing
branch of the new `VisMF::realDescriptorNumBytes()` is the one exercised.

## 1. The heuristic measures the wrong precision

`dx` is written into the ASCII plotfile `Header` at `HeaderFile.precision(17)` from the
writing build's `Real` (`Src/Base/AMReX_PlotFileUtil.cpp:95,130`). The new check instead
measures the width of the *fab data* on disk. `fab.format = NATIVE_32` / `IEEE_32`
(`Src/Base/AMReX_FArrayBox.cpp:352-396`, `getDataDescriptor` at `:204-219`) decouples
them: a double build then writes 4-byte data with a full-precision double `Header`.

Both directions are wrong:

- **False negative** (row 4 above, reproduced): `NATIVE_32`-from-double vs. a genuine
  single-precision plotfile have matching data widths, so `fuzzy_dx` is false
  (`Tools/Plotfile/fcompare.cpp:193-194`) and fcompare still prints
  `ERROR: grid dx does not match at level 0` — the exact failure this PR exists to fix.
- **False positive**: `NATIVE_32`-from-double vs. plain double enables the fuzzy
  comparison for two `dx` values that are both exact doubles and should be compared
  exactly.

Nothing in the plotfile records the precision the `Header` was written at, so this
cannot be fixed by reading more of the file. Given that, ~100 lines of new `VisMF` +
`PlotFileData` public API buy a signal that is wrong in the mixed case.

Suggested alternatives, both smaller:

- make the near-equality comparison unconditional, or
- expose it as an explicit opt-in flag (`--rtol_dx`), alongside the existing
  `-r/--rel_tol` / `--abs_tol` options.

## 2. Dead branches in `VisMF::realDescriptorNumBytes()`

`Src/Base/AMReX_VisMF.cpp:545-572`. The `c == ':'` ("old" FAB) format is only ever
emitted by `FABio_ascii` and `FABio_8bit` (`Src/Base/AMReX_FArrayBox.cpp:771`, `:891`),
which write `typ = FAB_ASCII|FAB_8BIT` and `wrd = 0`. The `FAB_NATIVE` / `FAB_IEEE` /
`FAB_NATIVE_32` / `FAB_IEEE_32` cases are unreachable for anything AMReX writes, and
AMReX's own reader cannot load such files anyway: `RealDescriptor::newRealDescriptor`
aborts on `FAB_NATIVE`/`FAB_NATIVE_32` (`Src/Base/AMReX_FabConv.cpp:240-243`) and
`FABio::read_header` has no `FAB_IEEE_32` case (`AMReX_FArrayBox.cpp:485-497`).

If the function stays, that whole `switch` collapses to `return 0;`.

## 3. The `0` sentinel is undocumented

`Src/Base/AMReX_VisMF.H:165` says "Number of bytes in the floating-point values stored
on disk" but not that `0` means "unknown". `0` is returned for an empty `BoxArray`, for
ASCII/8BIT fabs, and — via `Src/Base/AMReX_PlotFileDataImpl.cpp:126` — when
`m_ncomp == 0` (null `m_vismf`). The caller reads `0` as "compare exactly", so those
cases silently opt out of the new behavior. The `PlotFileDataImpl` and `PlotFileData`
forwarders have no Doxygen at all.

## 4. Per-call file I/O (minor)

Each call re-opens the data file and allocates a full `ioBufferSize` (256 KiB default,
`Src/Base/AMReX_VisMF.cpp:40`) buffer to read ~40 bytes, on every rank. Irrelevant for
fcompare's `2*nlevels` calls, but this is public API now — worth caching in the `VisMF`
object or computing once in the constructor.

## 5. Tolerance comment (minor)

`Tools/Plotfile/fcompare.cpp:29`: `Real(4.0) * FLT_EPSILON` behaves well in the tests
above; a one-line comment on why 4x would help.

## 6. Docs (minor)

The PR's documentation checkbox is unchecked. Per `AGENTS.md`, the fcompare section of
`Docs/sphinx_documentation/source/Post_Processing.rst:70` should gain a sentence noting
that plotfiles written at different precisions are accepted.

WeiqunZhang (approved):

PR #5940, opened 2026-09-20
Publish build and test results to CDash

WeiqunZhang:
Adds CTestConfig.cmake and submits four existing CI configurations to https://my.cdash.org/index.php?project=AMReX, so results land in one place instead of only in per-run logs: GCC 3D Debug, Clang fastmath, macOS and MSVC.  Reuses the builds CI already does rather than repeating them.

Submission runs only on pushes -- pull requests from forks get no secrets, and per-PR submissions would swamp the dashboard -- and never fails the job.  enable_testing() becomes include(CTest), which enables testing the same way but also reads CTestConfig.cmake so `ctest -D` can submit at all.

Also corrects Testing.rst, which said compilation testing runs on Travis-CI and that the test runner needs Python 2.7.


asalmgren (approved):

ax3l:
Awesome! Good to go!

If you _additionally_ like to record configure and build times, then you need to wrap these two steps also in ctest calls (yeah, it's weird).

We have a PR in WarpX for that open here: https://github.com/BLAST-WarpX/warpx/pull/7246

WeiqunZhang:
Let's do a follow-up on gitlab in the future.

● The comment is accurate, but I would not fold it into this PR. My take:

  What the comment means. CDash only records configure and build timings, and compiler warnings and errors, when those steps run through ctest_configure() and ctest_build(). Our jobs call cmake
  directly and then submit with the branch's submit script, which produces a Test.xml only. So every AMReX build on the dashboard will show "hasconfigure=false" and no build column, exactly as
  WarpX's did.

  Why I would leave it out of PR 5940.

  - Low payoff for AMReX. WarpX's configure fetches AMReX, pyAMReX and PICSAR, so configure time is minutes and worth tracking. Ours is seconds. Our GitHub Actions builds run behind ccache, so
    build times mostly measure the cache hit rate, not the code. And every submitting job already uses -Werror, so the warnings column would be empty by construction.
  - Real restructuring cost. Each of our submitting jobs has one combined run step doing configure, build, clang-tidy on ccache misses, then tests. Routing configure and build through a CTest
    script means splitting every job and replacing cmake calls in four workflows plus three GitLab jobs. The WarpX script shows how much fiddliness that brings: shell-quoting the CMake options
    across the boundary, a generator that must go through an environment variable because -G is silently overridden, a build parallelism knob, and a build that splits into two dashboard entries
    across midnight UTC.
  - The WarpX PR is still open. Better to let it land and settle, then copy what proved out. Their driver script is nearly project-agnostic, so a later AMReX port is small once the design is
    fixed.

  Where it would be worth it. The GitLab GPU jobs on the HPSF runners are the one place the timings would carry signal: no ccache, long CUDA, HIP and SYCL compiles, and nightly cadence. If we
  want anything from this, I would do it there first in a follow-up, using WarpX's script as the template.

  Suggested reply on the PR: thank Axel, agree the dashboard will show tests only, and say configure and build capture is a follow-up once WarpX PR 7246 has landed, starting with the GitLab
  nightly GPU jobs where the timings mean something.

ax3l:
> The GitLab GPU jobs on the HPSF runners are the one place the timings would carry signal

Exactly that is what we do, for the reasons listed.
Yes to follow up, not urgent here.

PR #5941, opened 2026-09-20
CMake: do not require the C MPI component

WeiqunZhang:
AMReX has no C sources, and MPI::MPI_CXX already provides the C API its C++ code calls, so the C component was never needed -- as the `Do we need MPI_C ?` comment next to it suspected.

It was linked PUBLIC, so it reached consumers through the exported target: any downstream project using an MPI-enabled AMReX had to enable the C language, even a pure C++ one, or configuration failed with

    Could NOT find MPI (missing: MPI_C_FOUND C)

Note for downstreams that are themselves C: they now have to request MPI::MPI_C themselves rather than inheriting it from AMReX.

`Tests/CMakeTestInstall` is now a `LANGUAGES CXX` project, so `make test_install` covers this: it imports an MPI-enabled AMReX from a consumer that never enables C. Putting the `C` component back makes it fail with the error above.


WeiqunZhang:
- git ls-files '*.c' is empty — AMReX has no C sources, so nothing needs the C wrapper's flags.
  - No remaining MPI_C / MPI::MPI_C / MPI_C_* reference in any tracked CMake file.
  - The two component lists stay in sync (AMReXParallelBackends.cmake:19 build side, AMReXConfig.cmake.in:195 exported side), both still appending Fortran under the Fortran-interfaces toggle, so
    AMReXTargets.cmake's INTERFACE_LINK_LIBRARIES and the config's find_dependency(MPI ...) agree.
  - MPI::MPI_CXX pulls in the same underlying libmpi for OpenMPI/MPICH/Intel MPI, so C-API consumers and C libraries linked PUBLIC (HYPRE, PETSc) still resolve MPI symbols.
  - The other find_dependency calls in AMReXConfig.cmake.in don't reintroduce a C-language requirement — FindHDF5 explicitly falls back to a C++ test when C isn't enabled;
    FindHYPRE/FindPETSc/FindAMReXFFTW are path/pkg-config based.

  Two notes, neither a defect in this diff:

  - CMakeLists.txt:61 still declares LANGUAGES C CXX. That C looks unused now, but dropping it is a separate, riskier change.
  - The behavior note in the PR body is accurate and worth carrying into the release notes: downstream C projects must now request MPI::MPI_C themselves instead of inheriting it from AMReX.

asalmgren (approved):

ax3l (approved):
This is generally legit.

There are a few corner cases in implementations where I sometimes need to also link the C Language targets as a work-around, but besides these buggy implementations requiring CXX is exactly right to require the MPI (C) API in a CXX project.

In openPMD-api I have a convenient option for the few times I need the work around:
https://github.com/openPMD/openPMD-api/blob/ff916100a6668cf9f6bb102b6d0425bc414c7c23/CMakeLists.txt#L226

PR #5945, opened 2026-09-20
CI: tighten token permissions and pin upload-sarif

WeiqunZhang:
Scorecard's Token-Permissions check scored 0 because the cache cleanup workflows granted actions: write at the workflow level and codeql.yml granted security-events: write there. Keep the top level at contents: read and put the write grants on the jobs that need them. No job's effective permissions change.

Pin github/codeql-action/upload-sarif to the same commit as the other codeql-action steps for the Pinned-Dependencies check.


ax3l (approved):
Cool, nice strengthening

